PR #108, opened 2019-01-23
Work on new payment channels construction

whyrusleeping:
This gets us closer to a payment channel setup that we can actually use for paying storage miners. It also has a few other features that will be important for efficiency.

### DIFF
--- a/payments.md
+++ b/payments.md
@@ -96,11 +96,8 @@ Lane state can be easily tracked on-chain with a compact bitfield.
 
 ```go
 type SpendVoucher struct {
-    // ID is the channel ID for this payment channel
-    ID ChannelID
-    
     // Amount is the amount of FIL that this voucher can be redeemed for
-    Amount *TokenAmount
+    Amount TokenAmount
     
     // Nonce is a number that sets the ordering of vouchers. If you try to redeem
     // a voucher with an equal or lower nonce, the operation will fail. Nonces are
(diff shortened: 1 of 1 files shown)


pooja:
Hey @whyrusleeping lots of great conversation and updates here. Did you want review on this from someone else too (before merging)?

whyrusleeping:
@pooja review from nicola would be good. I also need to add in another payment conditional. We need to be able to pay contingent on a particular file inclusion proof linking user data to a sector.

whyrusleeping:
Alright, this is a good step forward. Gonna merge for now, and follow up on additional questions soon.

payments.md
```diff
@@ -26,67 +26,7 @@ For example:
...
-```
-
-Channel IDs should be memory efficient and generated is such a way that reordering does not change the channelID. This may be solved by first indexing by the target address internally and then using the nonce of the message that invoked the create channel method.
-
-
-
-### Multi-Lane Payment Channel (WIP)
+### Multi-Lane Payment Channel
 
 The filecoin storage market may require a way to do incremental payments between two parties, over time, for multiple different transactions. The primary motivating usecase for this is to provide payment for file storage over time, for each file stored. An additional requirement is the ability to have less than one message on chain per transaction 'lane', meaning that payments for multiple files should be aggregateable (Note: its okay if this aggregation is an interactive process).
```
nicola: what does interactive process means here?

payments.md
```diff
@@ -110,16 +47,173 @@ type SpendVoucher struct {
...
+    Redeemed TokenAmount
+}
+
+func (paych *PaymentChannel) validateSignature(sv SpendVoucher) {
+    if msg.From == paych.From {
+        ValidateSignature(sv.SerializeNoSig(), sv.Signature, paych.To)
+    } else if msg.From == paych.To {
+        ValidateSignature(sv.SerializeNoSig(), sv.Signature, paych.From)
+    } else {
+        Fatal("bad programmer")
```
nicola: uhmm..

payments.md
```diff
@@ -110,16 +47,173 @@ type SpendVoucher struct {
...
+    }
+}
+
+func (paych *PaymentChannel) UpdateChannelState(sv SpendVoucher, secret []byte, pip *PieceInclusionProof) {
```
nicola: I am not sure I like the idea of hard-coding the `pip` here. In fact, payment channels should be completely agnostic from the rest of filecoin.

In other words, we could rename this payment channel actor the retrieval market payment channel, but that would be a bad idea.

Proposal: define the method of another contract that should be called and return true in order to unclock the funds, in other words, making fund unlocking before timeout to be contingent on another contract execution.

payments.md
```diff
@@ -110,16 +47,173 @@ type SpendVoucher struct {
...
+        Fatal("cannot use this voucher yet!")
+    }
+
+    if sv.SecretPreimage != nil {
+        if Hash(secret) != sv.SecretPreimage {
+            Fatal("Incorrect secret!")
+        }
+    }
+  
+  	if sv.DataCommitment != nil {
```
nicola: see my previous comment, my opinion is that either this contract is renamed or it should be more generic.

payments.md
```diff
@@ -110,16 +47,173 @@ type SpendVoucher struct {
...
+      // Checks that the piece inclusion proof is valid, and that the referenced sector
+      // is correctly being stored
+      if !ValidateInclusion(pip, sv.DataCommitment) {
```
nicola: Is this defined anywhere?

payments.md
```diff
@@ -110,16 +47,173 @@ type SpendVoucher struct {
...
+        }
+        if paych.MinCloseHeight < sv.MinCloseHeight {
+            paych.MinCloseHeight = sv.MinCloseHeight
+        }
+    }
+}
+
+func (paych *PaymentChannel) Withdraw(upTo TokenAmount) {
+    // TODO: this ones tricky, withdraw funds without closing it out entirely...
+}
```
nicola: is this the same as collect?

payments.md
```diff
@@ -110,16 +47,173 @@ type SpendVoucher struct {
...
+        ValidateSignature(sv.SerializeNoSig(), sv.Signature, paych.To)
+    } else if msg.From == paych.To {
+        ValidateSignature(sv.SerializeNoSig(), sv.Signature, paych.From)
+    } else {
+        Fatal("bad programmer")
+    }
+}
+
+func (paych *PaymentChannel) UpdateChannelState(sv SpendVoucher, secret []byte, pip *PieceInclusionProof) {
+    if !paych.validateSignature(sv) {
```
dignifiedquire: where is `validateSignature` defined?

payments.md
```diff
@@ -110,16 +47,173 @@ type SpendVoucher struct {
...
+        Fatal("Signature Invalid")
+    }
+
+    if chain.Now() < sv.TimeLock {
+        Fatal("cannot use this voucher yet!")
+    }
+
+    if sv.SecretPreimage != nil {
+        if Hash(secret) != sv.SecretPreimage {
```
dignifiedquire: what `Hash`?

payments.md
```diff
@@ -110,16 +47,173 @@ type SpendVoucher struct {
...
+  	if sv.DataCommitment != nil {
+      // Checks that the piece inclusion proof is valid, and that the referenced sector
+      // is correctly being stored
+      if !ValidateInclusion(pip, sv.DataCommitment) {
+        Fatal("PieceInclusionProof was invalid")
+      }
+  	}
+
+    if sv.RequiredSector != nil {
+        miner, found := GetMiner(msg.From)
```
dignifiedquire: Where is `GetMiner` defined ?

payments.md
```diff
@@ -110,16 +47,173 @@ type SpendVoucher struct {
...
+        if !found {
+            Fatal("Redeemer is not a miner")
+        }
+
+        if !miner.HasSector(sv.RequiredSector) {
```
dignifiedquire: Where is `HasSector` defined ?

payments.md
whyrusleeping: can change to  "illegal state"

payments.md
dignifiedquire: I am blind..

payments.md
whyrusleeping: probably going to delete this one anyways. its difficult to do right, and we can just close channels out

PR #122, opened 2019-01-29
CommitSector accepts SectorId and CommRStar

laser:
- sector id is an input to the PoRep-generating operation
- fix typo which omitted CommRStar from CommitSector parameter-list


pooja:
@whyrusleeping Is this blocked or can it be merged?

laser:
It can be closed. All three commitment-parameters were merged into `Commitment`.

pooja:
Thanks @laser

PR #135, opened 2019-02-08
Create Address spec

frrist:
### Description
Initial write up of Address definition

### DIFF
--- a/address.md
+++ b/address.md
@@ -11,58 +11,48 @@ A Filecoin address is an identifier that refers to an actor in the Filecoin stat
 
 ## Specification
 
-A Filecoin address is a byte string consisting of:
-* A **network indicator** that identifies which network this address belongs to.
-* A **protocol indicator** that identifies the type and version of this address.
+A Filecoin address encoded to a byte array consists of:
+* A **protocol indicator** byte that identifies the type and version of this address.
 * The **payload** used to uniquely identify the actor according to the protocol.
-* A **checksum** verifies the address is valid.
+```
+|----------|---------|
+| protocol | payload |
+|----------|---------|
+|  1 byte  | n bytes |
+```
+
+A Filecoin address encoded to a string consistes of:
+* A **network prefix** character that identifies the network the address belongs to.
+* A **protocol indicator** byte that identifies the type and version of this address.
+* A **payload** used to uniquely identify the actor according to the protocol..
+* A **checksum** used to validet the address. 
 
 ```
-|---------|----------|---------|----------|
-| network | protocol | payload | checksum |
-|---------|----------|---------|----------|
-| 1 byte  |  1 byte  | n bytes |  4 bytes |
+|------------|----------|---------|----------|
+|  network   | protocol | payload | checksum |
+|------------|----------|---------|----------|
+| 'f' or 't' |  1 byte  | n bytes | 4 bytes  |
 ```
 
 An example of a Filecoin address in golang:
 
 ```go
 type Address struct {
-    // 0: main-net
-    // 1: test-net
-    network byte
     
     // 0: ID
     // 1: Blake2b160-Hash of secp256k1 Public Key
     // 2: Blake2b160-Hash of Actor creation data
     // 3: BLS Public Key
-    protocol byte
+    Protocol byte
     
     // raw bytes containing the data associated with protocol
-    payload []byte
+    Payload []byte
 }
 ```
 
-#### Network Indicator
+#### Network Prefix
 
-The **network** byte identifies the network the address belongs to:
-
-- `0` : Mainnet
-- `1` : Testnet
-
-An example description in golang:
-
-```go
-// String Prefix
-Mainnet_Prefix = 'f'
-Testnet_Prefix = 't'
-
-type Protocol = byte
-const (
-    Mainnet Network = iota
-    Testnet
-)
-```
+The **network prefix** is prepended to an address when encoding to a string. The network prefix indicates which network an address belongs in. The network prefix may either be `f` for filecoin mainnet or `t` for filecoin testnet. Is it worth noting that a network prefix will never appear on chain and is only used when encoding an address to a human readable format.
 
 #### Protocol Indicator
 
@@ -88,27 +78,91 @@ const (
 
 ###### Protocol 0: IDs
 
-**Protocol 0** addresses are simple ids. These addresses are expected to remain within the network. All actors have a numeric ID even if they don't have public keys. The key part of an ID address is the decimal representation of the id only characters '0'-'9'. IDs are not hashed and do not have a checksum.
+**Protocol 0** addresses are simple IDs. These addresses are expected to remain within the network. All actors have a numeric ID even if they don't have public keys. The payload of an ID address is encoded as decimal numbers '0'-'9'. IDs are not hashed and do not have a checksum.
+
+**Bytes**
+
+```
+|----------|---------|
+| protocol | payload |
+|----------|---------|
+|    0     | '0'-'9' |
+```
+
+**String**
+
+```
+|------------|----------|---------|
+|  network   | protocol | payload |
+|------------|----------|---------|
+| 'f' or 't' |     0    | '0'-'9' |
+```
 
 ###### Protocol 1: libsecpk1 Elliptic Curve Public Keys
 
 **Protocol 1** addresses represent secp256k1 public encryption keys. The payload field contains the [Blake2b 160](https://blake2.net/) hash of the public key. 
 
+**Bytes**
+
+```
+|----------|---------------------|
+| protocol |        payload      |
+|----------|---------------------|
+|    1     | blake2b-160(PubKey) |
+```
+
+**String**
+
+```
+|------------|----------|---------------------|----------|
+|  network   | protocol |      payload        | checksum |
+|------------|----------|---------------------|----------|
+| 'f' or 't' |     1    | blake2b-160(PubKey) |  4 bytes |
+                  base32[................................]
+```
+
 ###### Protocol 2: Actor
 
 **Protocol 2** addresses representing an Actor. The payload field contains the [Blake2b 160](https://blake2.net/) hash of meaningful data produced as a result of creating the actor. 
 
+```
+|------------|----------|---------------------|----------|
+|  network   | protocol |      payload        | checksum |
+|------------|----------|---------------------|----------|
+| 'f' or 't' |     2    |  blake2b-160(TODO)  |  4 bytes |
+                  base32[................................]
+```
+
 ###### Protocol 3: BLS
 
-**Protocol 3** addresses represent BLS public encryptions keys. The payload filed contains the BLS public key.
+**Protocol 3** addresses represent BLS public encryption keys. The payload field contains the BLS public key.
+
+**Bytes**
+
+```
+|----------|---------------------|
+| protocol |        payload      |
+|----------|---------------------|
+|    1     | 48 byte BLS PubKey  |
+```
+
+**String**
+
+```
+|------------|----------|---------------------|----------|
+|  network   | protocol |      payload        | checksum |
+|------------|----------|---------------------|----------|
+| 'f' or 't' |     3    |  48 byte BLS PubKey |  4 bytes |
+                  base32[................................]
+```
 
 #### Payload
 
-The payload represents the data specified by the protocol. All payloads excecpt the payload of the ID protocol are base32 encoded when seralized to their human readable format.
+The payload represents the data specified by the protocol. All payloads except the payload of the ID protocol are [base32](https://tools.ietf.org/html/rfc4648) encoded when seralized to their human readable format. The base32 encoding uses the lowercase alphabet.
 
 #### Checksum
 
-The checksum is calculated using the first 4 bytes of the blake2b-160 hash of the addresses network, protocol, and payload bytes. Checksums are only used when seralizing and deseralizing an address to its human readable format.
+The checksum is calculated using blake2b-4 hash of the addresses protocol, and payload bytes. Checksums are base32 encoded  and only used when encoding and decoding an address to its human readable format -- a string. Checksums will never appear on chain.
 
 
 #### Expected Methods
@@ -117,28 +171,24 @@ All implementations in Filecoin must have methods for creating, encoding, and de
 
 ```go
 type Address interface {
-    New(n byte, p byte, payload []byte) Address
-    Encode(a Adress) string
+    New(p byte, payload []byte) Address
+    Encode(network Network, a Adress) string
     Decode(s string) Address
     Checksum(a Address) []byte
-    Validate(a Address) bool
+    ValidateChecksum(a Address) bool
 }
 ```
 ##### New()
 
-New returns an Address for the specified network and protocol encapsulating corresponding payload. New fails for unknown network or protocol.
+New returns an Address for the specified protocol encapsulating corresponding payload. New fails for unknown protocol.
 
 ```go
-func New(network byte, protocol byte, payload []byte) Address {
-	if network != Mainnet && network != Testnet {
-        Fatal(ErrUnknownNetwork)
-	}
+func New(protocol byte, payload []byte) Address {
 	if protocol < SECP256K1 || protocol > BLS {
-        Fatal(ErrUnknownType)
+		Fatal(ErrUnknownType)
 	}
     return Address{
-        Network:  network,
-        protocol: protocol,
+        Protocol: protocol,
         Payload:  payload,
     }
 }
@@ -152,28 +202,22 @@ Software encoding a Filecoin address must:
 - produce an address encoded to a known protocol
 - produce an address with a valid checksum
 
-Encodes an Address as a string, converting the network byte to the corresponding prefix and encoding the payload to [base32](https://tools.ietf.org/html/rfc4648).
+Encodes an Address as a string, prepending the network prefix, calculating the checksum and encoding the payload and checksum to [base32](https://tools.ietf.org/html/rfc4648).
 
 ```go
-func Encode(a Address) string {
-	var prefix string
-    switch a.network {
-    case Mainnet:
-        prefix = Mainnet_Prefix // "f"
-    case Testnet:
-        prefix = Testnet_Prefix // "t"
-    default:
-        Fatal("invalid address network")
-    }
-    
-    switch a.protocol {
-        case SECP256K1:
-        case Actor:
-        case BLS:
-        	cksm := Checksum(a)
-            return prefix + a.protocol + base32.Encode(a.payload + cksm)
-        case ID:
-        	return prefix + a.protocol + base32.Encode(a.payload)
+func Encode(network string, a Address) string {
+	if network != "f" && network != "t" {
+        Fatal("Invalid Network")
+	}
+	
+    switch a.Protocol {
+    	case SECP256K1:
+    	case Actor:
+    	case BLS:
+    		cksm := Checksum(a)
+    		return network + a.Protocol + base32.Encode(a.Payload + cksm)
+    	case ID:
+        	return network + a.Protocol + base32.Encode(a.Payload)
         default:
         	Fatal("invalid address protocol")
     }
@@ -188,7 +232,7 @@ Software decoding a Filecoin address must:
 
 * verify the checksum is valid
 
-Decode an Address from a string, converting the network prefix and protocol prefix to their corresponding byte values, decoding the payload from base32, and validating the checksum.
+Decode an Address from a string by, removing the network prefix validating the the address is of a know protocol prefix to theidecoding the payload from base32, and validating the checksum.
 
 ```go
 func DecodeString(a string) Address {
@@ -196,60 +240,57 @@ func DecodeString(a string) Address {
         Fatal(ErrInvalidLength)
     }
     
-    var ntwk byte
-    switch a[:2] {
-    case Mainnet_Prefix:
-        ntwk = Mainnet
-    case Testnet_Prefix:
-    	ntwk = Testnet
-    default:
+    if a[0] != "f" && a[0] != "t" {
         Fatal(ErrUnknownNetwork)
     }
     
-    protocol := a[3]
-    payload := a[3:]
-    
+    protocol := a[1]
+    raw := a[2:] 
     if protocol == ID {
         return Address{
-            network: ntwk,
-            protocol: protocol
-            payload: payload
+            Protocol: protocol,
+            Payload: raw,
         }
     }
     
-    cksm := payload[len(payload)-CksmLen:len(payload)]
+    payload = raw[:len(raw)-CksmLen]
+    if protocol == SECP256K1 || protocol == Actor {
+        if len(payload) != 20 {
+            Fatal(ErrInvalidBytes)
+        }
+    }
     
-    if !Validate(a, cksm) {
+    cksm := payload[len(payload)-CksmLen : len(payload)]
+    if !ValidateChecksum(a, cksm) {
         Fatal(ErrInvalidChecksum)
     }
     
     return Address{
-        network: ntwk,
-        protocol: protocol,
-        payload: payload,
+        Protocol: protocol,
+        Payload: payload,
     }
 }
 ```
 
 ##### Checksum()
 
-Checksum produces a byte array of the blake2b-16 checksum of an address, returning the first 4 bytes of the digest.
+Checksum produces a byte array of the blake2b-4 checksum of an addresses protocol and payload.
 
 ```go
 const CksmLen = 4
 
-func Checksum(a Address) [CKSM_LEN]byte {
-    digest = blake2b(a)
+func Checksum(a Address) [CksmLen]byte {
+    digest = blake2b(a.Protocol + a.Payload)
     return digest[:4]
 }
 ```
 
-##### Validate()
+##### ValidateChecksum()
 
-Validate returns true if the Checksum of data matches the expected checksum.
+ValidateChecksum returns true if the Checksum of data matches the expected checksum.
 
 ```go
-func Validate(data, expected []byte) bool {
+func ValidateChecksum(data, expected []byte) bool {
     digest := Checksum(data)
     return digest == expected
 }
@@ -259,4 +300,4 @@ func Validate(data, expected []byte) bool {
 
 > TODO a better definition of checksums
 
-Filecoin checksums are the first 4 bytes of the blake2b-160 hash of an Address. Checksums are only added to an address when encoding the address to its human readable format. Addresses using Protocol 0 do not have a checksum.
+Filecoin checksums are calculated over the address protocol and payload using blake2b-4. Checksums are only added to an address when encoding the address to its human readable format. Addresses under the ID Protocol do not have a checksum.
--- a/address.md
+++ b/address.md
@@ -2,6 +2,22 @@
 
 An address is an identifier that refers to an actor in the Filecoin state. All actors (miner actors, the storage market actor, account actors) have an address. An address encodes information about the network it belongs to, the type of data it contains, the data itself, and depending on the type, a checksum.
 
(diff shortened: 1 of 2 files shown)


address.md
```diff
@@ -0,0 +1,147 @@
...
+    return digest[len(digest)-CKSM_LEN:]
+}
+
+// validChecksum returns true if cksm is valid for data and false otherwise.
+func validChecksum(data, cksm []byte) bool {
+    digest := Checksum(data)
+    return digest == cksm
+}
+```
+
```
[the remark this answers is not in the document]
frrist: Sounds good, I have added an interface to match the implementation.

frrist:
Couple questions I'd like to talk about:
- Are we alright with base58 encoding or should we use something that is case insensitive instead, such as base32?
- Do we want addresses to use multibase so that we may represent the same address with many different encodings?

bvohaska:
Ping @whyrusleeping on @frrist comment about baseX encodings. I have no strong feelings either way. What is the industry standard on this? Since this effort follows Multiformats, is there a prefered base or should we go with multibase?

dignifiedquire:
we should use base32, to avoid having sth that is case sensitive. the issues are well documented about this. multibase adds overhead so that depends on if we are okay with that. also ipfs is switching to base32 as the default

address.md
```diff
@@ -0,0 +1,162 @@
...
+An exampple implementation of this interface is shown below:
+
+```go
+type Address struct {
+    // 0: main-net
+    // 1: test-net
+    network byte
```
dignifiedquire: when did we decide to stop differentiating between testnets?

address.md
```diff
@@ -0,0 +1,162 @@
...
+    
+    return Address{
+        network: ntwk,
+        typ: raw[0],
+        data: raw[1: len(raw)- CKSM_LEN]
+    }
+}
+
+// Checksum is the last CKSM_LEN bytes the sha256 of Address Type and its data.
+func Checksum(a Address) [CKSM_LEN]byte {
```
dignifiedquire: why are we using sha256 suddenly as checksums? we had a proper scheme with error detections in the previous spec version

address.md
frrist: I don't think we have decided that, what do you mean?

address.md
frrist: What scheme were we previously using? I assume you mean [this one](https://github.com/filecoin-project/go-filecoin/blob/efdc601098fec6d71f1543327090c1e468d0836b/address/address.go#L327-L373), if so we continue to use it -- `sha256` was chosen here for simplicity in writing as I wasn't sure on the details of the aforementioned code.

address.md
whyrusleeping: @dignifiedquire I don't really think its useful. If many people disagree with that, we could put it in the user visible encoding, but not include it in the bytes representation.

address.md
whyrusleeping: @dignifiedquire the previous scheme was just a complicated checksum, its far simpler to use a few bytes from a well known hash function.

address.md
```diff
@@ -0,0 +1,403 @@
...
+
+Learning from the Ethereum:
+
+* Hex encoding is inefficient with respect to storage. Representing addresses as Hex increases the address size by 100%.
+* Checksums require mixed cases, resulting in the same issues as above.
+
+## Specification
+
+A Filecoin address is a byte string consisting of:
+* The **letter 'f'** to identify the address as belonging to a Filecoin network.
```
frrist: It is unclear to me if this `f` exists in addition to the network indicator or if it is a network indicator. The current implementation treats f as a network indicator.

address.md
```diff
@@ -0,0 +1,403 @@
...
+	SECP256K1 = 0
+	ID = 1
+	Actor = 2
+	BLS = 3
+)
+```
+
+###### Protocol 0: libsecpk1 Elliptic Curve Public Keys
+
+**Protocol 0** addresses represent public encryption keys. These addresses represent Filecoin user accounts and are likely to be transmitted outside the network over noisy channels (e.g. hand written on a napkin). As such, they require robust error detection. The multi-base encoded key contains a hash and a checksum. The hash is a [Blake2b 160](https://blake2.net/) hash of the public key. The checksum is a six character BCH checksum. The algorithm details can be found in Appendix - 9.2. Algorithms - Checksums.
```
frrist: I believe we either need to add a second protocol for the case of Actors without public keys, or we need to state that the address represent public encryption keys **or** "random data" used to generate an address. I think "random data" is a pretty poor description, thoughts?

frrist: Also, I believe we decided that we are **not** using multi-base encoding, correct?

frrist: I believe we decided that we are **not** using multi-base encoding, correct?

frrist: Should we have a bit here on secp256k1 public key recovery since we are using the hash of the public key?

whyrusleeping:
Just to state it clearly, we should use a truncated common hash (sha256 or blake2) as the checksum, its much simpler for others to implement, and easier to reason about, and also doesnt have to be targeted to a particular size of input.

bvohaska:
@whyrusleeping: BCH encoding will allow a receiver to (1) check if the data is correct (2) fix some set of errors if there was a problem. Truncated hashes will provide some data integrity but no correction. BCH codes are also blazing fast (not that we expect this to be an issue).

As far as reasonability, BCH should be just as easy to reason about as hashing for error detection. We can also provide some clearer context as it’s spec’d out a bit more. I believe BCH codes are also used in QR Codes so there is also some external reference for devs. Putting addresses into QR codes could be a pretty cool feature in any case.

whyrusleeping:
I posit that the error correction is not super useful, people are not going to be typing these, they will be copy pasted. If someone makes an error in a copy pasted address, and we're just okay with it because the error gets corrected, then that address is assumed to be correct, when its not. This can lead to multiple unique strings that actually correct to the same address.

As far as "easy" goes, the fact that we have to implement BCH codes ourselves and its not just `import "crypto/sha"` makes it far far less easy. The speed of hashing is really not a concern, as its only done when presenting things to the user.

dignifiedquire:
implementation wise we already have it implemented and used in the current address format, so there is not really much to do there.

address.md
frrist: Based on other discussion we have had we are going with the first 4 bytes of the blake2b hash of an address

address.md
frrist: We will have support for Mainnet and Testnet in the initial address spec

address.md
```diff
@@ -0,0 +1,262 @@
...
+- `1` : Testnet
+
+An example description in golang:
+
+```go
+// String Prefix
+Mainnet_Prefix = 'f'
+Testnet_Prefix = 't'
+
+type Protocol = byte
```
rkowalick: this should be `Network`

address.md
```diff
@@ -0,0 +1,262 @@
...
+    ID Protocol = iota
+    SECP256K1
+    Actor
+    BLS
+)
+```
+
+###### Protocol 0: IDs
+
+**Protocol 0** addresses are simple ids. These addresses are expected to remain within the network. All actors have a numeric ID even if they don't have public keys. The key part of an ID address is the decimal representation of the id only characters '0'-'9'. IDs are not hashed and do not have a checksum.
```
rkowalick: This sentence is confusing to me:
```
The key part of an ID address is the decimal representation of the id only characters '0'-'9'.
```

address.md
```diff
@@ -0,0 +1,262 @@
...
+
+**Protocol 1** addresses represent secp256k1 public encryption keys. The payload field contains the [Blake2b 160](https://blake2.net/) hash of the public key. 
+
+###### Protocol 2: Actor
+
+**Protocol 2** addresses representing an Actor. The payload field contains the [Blake2b 160](https://blake2.net/) hash of meaningful data produced as a result of creating the actor. 
+
+###### Protocol 3: BLS
+
+**Protocol 3** addresses represent BLS public encryptions keys. The payload filed contains the BLS public key.
```
rkowalick: typo `filed` => `field`

address.md
```diff
@@ -0,0 +1,262 @@
...
+
+#### Payload
+
+The payload represents the data specified by the protocol. All payloads excecpt the payload of the ID protocol are base32 encoded when seralized to their human readable format.
```
rkowalick: typo `excecpt` => `except`

address.md
```diff
@@ -0,0 +1,262 @@
...
+}
+```
+##### New()
+
+New returns an Address for the specified network and protocol encapsulating corresponding payload. New fails for unknown network or protocol.
+
+```go
+func New(network byte, protocol byte, payload []byte) Address {
+	if network != Mainnet && network != Testnet {
+        Fatal(ErrUnknownNetwork)
```
rkowalick: indent

address.md
```diff
@@ -0,0 +1,262 @@
...
+	}
+	if protocol < SECP256K1 || protocol > BLS {
+        Fatal(ErrUnknownType)
```
rkowalick: indent

address.md
```diff
@@ -0,0 +1,262 @@
...
+	}
+    return Address{
+        Network:  network,
```
rkowalick: `Network` and `Payload` are capitalized here and lowercase in the struct definition.

address.md
```diff
@@ -0,0 +1,262 @@
...
+Software encoding a Filecoin address must:
+
+- produce an address encoded to a known network
+- produce an address encoded to a known protocol
+- produce an address with a valid checksum
+
+Encodes an Address as a string, converting the network byte to the corresponding prefix and encoding the payload to [base32](https://tools.ietf.org/html/rfc4648).
+
+```go
+func Encode(a Address) string {
```
rkowalick: indentation in this func is kind of jank

address.md
```diff
@@ -0,0 +1,262 @@
...
+    switch a[:2] {
+    case Mainnet_Prefix:
+        ntwk = Mainnet
+    case Testnet_Prefix:
+    	ntwk = Testnet
+    default:
+        Fatal(ErrUnknownNetwork)
+    }
+    
+    protocol := a[3]
```
rkowalick: I think this should be `a[2]`

address.md
```diff
@@ -0,0 +1,262 @@
...
+    
+    if protocol == ID {
+        return Address{
+            network: ntwk,
+            protocol: protocol
+            payload: payload
+        }
+    }
+    
+    cksm := payload[len(payload)-CksmLen:len(payload)]
```
rkowalick: can be simplified to `cksm := payload[len(payload)-CksmLen:]

rkowalick: Also `ChksmLen` is used before it is defined.

address.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+##### Checksum()
+
+Checksum produces a byte array of the blake2b-16 checksum of an address, returning the first 4 bytes of the digest.
+
+```go
+const CksmLen = 4
+
+func Checksum(a Address) [CKSM_LEN]byte {
```
rkowalick: The const above is not clearly being used here. Also, are we hashing the entire address struct in this function?

address.md
```diff
@@ -0,0 +1,262 @@
...
+| 1 byte  |  1 byte  | n bytes |  4 bytes |
+```
+
+An example of a Filecoin address in golang:
+
+```go
+type Address struct {
+    // 0: main-net
+    // 1: test-net
+    network byte
```
dignifiedquire: unclear if we need to actually store this here

address.md
```diff
@@ -0,0 +1,262 @@
...
+    payload []byte
+}
+```
+
+#### Network Indicator
+
+The **network** byte identifies the network the address belongs to:
+
+- `0` : Mainnet
+- `1` : Testnet
```
dignifiedquire: We don't need byte identifiers for these, these are just relevant to encoded addresses

address.md
dignifiedquire: maybe saying that they are encoded as decimal numbers is more straightforward

address.md
```diff
@@ -0,0 +1,262 @@
...
+
+**Protocol 3** addresses represent BLS public encryptions keys. The payload filed contains the BLS public key.
+
+#### Payload
+
+The payload represents the data specified by the protocol. All payloads excecpt the payload of the ID protocol are base32 encoded when seralized to their human readable format.
+
+#### Checksum
+
+The checksum is calculated using the first 4 bytes of the blake2b-160 hash of the addresses network, protocol, and payload bytes. Checksums are only used when seralizing and deseralizing an address to its human readable format.
```
dignifiedquire: the network should not be part of the hash if we hash the byte format

address.md
```diff
@@ -0,0 +1,262 @@
...
+
+```go
+type Address interface {
+    New(n byte, p byte, payload []byte) Address
+    Encode(a Adress) string
+    Decode(s string) Address
+    Checksum(a Address) []byte
+    Validate(a Address) bool
+}
+```
```
dignifiedquire: no error handling?

address.md
```diff
@@ -0,0 +1,262 @@
...
+
+##### Encode()
+
+Software encoding a Filecoin address must:
+
+- produce an address encoded to a known network
+- produce an address encoded to a known protocol
+- produce an address with a valid checksum
+
+Encodes an Address as a string, converting the network byte to the corresponding prefix and encoding the payload to [base32](https://tools.ietf.org/html/rfc4648).
```
dignifiedquire: missing the lower case part

address.md
```diff
@@ -0,0 +1,262 @@
...
+Decode an Address from a string, converting the network prefix and protocol prefix to their corresponding byte values, decoding the payload from base32, and validating the checksum.
+
+```go
+func DecodeString(a string) Address {
+    if len(a) < 4 {
+        Fatal(ErrInvalidLength)
+    }
+    
+    var ntwk byte
+    switch a[:2] {
```
dignifiedquire: this should be `a[0]`

address.md
dignifiedquire: actually this should be `a[1]`

address.md
```diff
@@ -0,0 +1,262 @@
...
+    case Mainnet_Prefix:
+        ntwk = Mainnet
+    case Testnet_Prefix:
+    	ntwk = Testnet
+    default:
+        Fatal(ErrUnknownNetwork)
+    }
+    
+    protocol := a[3]
+    payload := a[3:]
```
dignifiedquire: this should be `a[2:]`

address.md
```diff
@@ -0,0 +1,262 @@
...
+
+##### Checksum()
+
+Checksum produces a byte array of the blake2b-16 checksum of an address, returning the first 4 bytes of the digest.
+
+```go
+const CksmLen = 4
+
+func Checksum(a Address) [CKSM_LEN]byte {
+    digest = blake2b(a)
```
dignifiedquire: did you see my comment regarding using `Blake2b-4` instead?

data-structures.md
```diff
@@ -4,6 +4,27 @@ This document serves as an entry point for understanding all of the data structu
...
+type Address struct {
+    // 0: main-net
+    // 1: test-net
+    network leb128
+
+    // 0: SECP256K1 Public Key
+    // 1: Actor
+    // 2: BLS Public Key
+    // 3: ID
+    typ leb128
```
dignifiedquire: these are bytes, not leb128, according to our last agreement

data-structures.md
```diff
@@ -4,6 +4,27 @@ This document serves as an entry point for understanding all of the data structu
...
 
+## Address
+
+An address is an identifier that refers to an actor in the Filecoin state. All actors (miner actors, the storage market actor, account actors) have an address. An address encodes information about the network it belongs to, the type of data it contains, the data itself, and depending on the type, a checksum.
+
+```go
+type Address struct {
+    // 0: main-net
+    // 1: test-net
+    network leb128
```
dignifiedquire: these are bytes, not leb128, according to our last agreement

address.md
whyrusleeping: It doesnt hurt, might make some things easier (that way addresses don’t have to depend on some other thing to get this info)

address.md
```diff
@@ -0,0 +1,262 @@
...
+    protocol byte
+    
+    // raw bytes containing the data associated with protocol
+    payload []byte
+}
+```
+
+#### Network Indicator
+
+The **network** byte identifies the network the address belongs to:
```
whyrusleeping: i’d call it a network prefix instead of a network btye

address.md
whyrusleeping: Yeah, worth just saying its encoded base 10

address.md
```diff
@@ -0,0 +1,262 @@
...
+)
+```
+
+###### Protocol 0: IDs
+
+**Protocol 0** addresses are simple ids. These addresses are expected to remain within the network. All actors have a numeric ID even if they don't have public keys. The key part of an ID address is the decimal representation of the id only characters '0'-'9'. IDs are not hashed and do not have a checksum.
+
+###### Protocol 1: libsecpk1 Elliptic Curve Public Keys
+
+**Protocol 1** addresses represent secp256k1 public encryption keys. The payload field contains the [Blake2b 160](https://blake2.net/) hash of the public key.
```
whyrusleeping: mention how these are encoded for human readability

whyrusleeping: oh i see, its mentioned later

address.md
```diff
@@ -0,0 +1,303 @@
...
+
+###### Protocol 2: Actor
+
+**Protocol 2** addresses representing an Actor. The payload field contains the [Blake2b 160](https://blake2.net/) hash of meaningful data produced as a result of creating the actor. 
+
+```
+|------------|----------|---------------------|----------|
+|  network   | protocol |      payload        | checksum |
+|------------|----------|---------------------|----------|
+| 'f' or 't' |     2    |  blake2b-160(TODO)  |  4 bytes |
```
frrist: Need to address this TODO cc @whyrusleeping, what values are we hasing here?

address.md
```diff
@@ -0,0 +1,303 @@
...
+    ID Protocol = iota
+    SECP256K1
+    Actor
+    BLS
+)
+```
+
+###### Protocol 0: IDs
+
+**Protocol 0** addresses are simple IDs. These addresses are expected to remain within the network. All actors have a numeric ID even if they don't have public keys. The payload of an ID address is encoded as decimal numbers '0'-'9'. IDs are not hashed and do not have a checksum.
```
frrist: What encoding are we using for ID's?
I believe the following are all valid filecoin addresses under this current definition, correct?
`f00`
`f010`
`f0123`

address.md
frrist: We use `Fatal` when there is an error, basically a panic. Validation has been added here: https://github.com/filecoin-project/specs/pull/135/files#diff-c6ea617359921b8070df6dd947c8af6cR257

address.md
```diff
@@ -0,0 +1,303 @@
...
+                  base32[................................]
+```
+
+#### Payload
+
+The payload represents the data specified by the protocol. All payloads except the payload of the ID protocol are [base32](https://tools.ietf.org/html/rfc4648) encoded when seralized to their human readable format. The base32 encoding uses the lowercase alphabet.
+
+#### Checksum
+
+The checksum is calculated using blake2b-4 hash of the addresses protocol, and payload bytes. Checksums are base32 encoded  and only used when encoding and decoding an address to its human readable format -- a string. Checksums will never appear on chain.
```
frrist: @dignifiedquire we have explicitly mentioned blake2b-4 here.

address.md
```diff
@@ -0,0 +1,303 @@
...
+##### Checksum()
+
+Checksum produces a byte array of the blake2b-4 checksum of an addresses protocol and payload.
+
+```go
+const CksmLen = 4
+
+func Checksum(a Address) [CksmLen]byte {
+    digest = blake2b(a.Protocol + a.Payload)
+    return digest[:4]
```
dignifiedquire: @frrist this is not blake2b-4, the `4` needs to be passed to the hashing function as an argument, like here: https://github.com/filecoin-project/go-filecoin/blob/6914061e22bb8725182e5496b1286c914b5b9c26/address/address.go#L15

frrist: Ahh yes, I missed this bit. Seeing as this code isn't expected to compile and is serving here as an example, do we need to add the hash function configuration code? or could we just do something like
`return blake2b4(a.Protocol + a.Payload)`

dignifiedquire: you can do that, and add a comment maybe saying what this is supposed to mean?

frrist: Sounds good to me, will do that.

address.md
```diff
@@ -0,0 +1,300 @@
...
+| protocol | payload |
+|----------|---------|
+|  1 byte  | n bytes |
+```
+
+A Filecoin address encoded to a string consistes of:
+* A **network prefix** character that identifies the network the address belongs to.
+* A **protocol indicator** byte that identifies the type and version of this address.
+* A **payload** used to uniquely identify the actor according to the protocol..
+* A **checksum** used to validet the address.
```
whyrusleeping: typo on validate, and also maybe note that the checksum isnt always present?

address.md
whyrusleeping: What do you mean by 'these addresses are expected to remain within the network.' ?

address.md
```diff
@@ -0,0 +1,300 @@
...
+|    0     | '0'-'9' |
+```
+
+**String**
+
+```
+|------------|----------|---------|
+|  network   | protocol | payload |
+|------------|----------|---------|
+| 'f' or 't' |     0    | '0'-'9' |
```
whyrusleeping: probably put '0' in quotes to indicate we're putting it in as a string literal

whyrusleeping: also this is a little misleading. Reading this, I would expect the bytes for the payload to be the same for the raw and stringed versions, when its not. We should clarify that in the stringed version its base10 encoded

address.md
whyrusleeping: We arent hashing anything, its just random bytes that don't mean anything.

whyrusleeping: (technically, we get those random bytes from hashing something, but thats an implementation detail that I think can be left out of the spec)

address.md
whyrusleeping: New is a method on an address that returns an address?

address.md
```diff
@@ -0,0 +1,300 @@
...
+Encodes an Address as a string, prepending the network prefix, calculating the checksum and encoding the payload and checksum to [base32](https://tools.ietf.org/html/rfc4648).
+
+```go
+func Encode(network string, a Address) string {
+	if network != "f" && network != "t" {
+        Fatal("Invalid Network")
+	}
+	
+    switch a.Protocol {
+    	case SECP256K1:
```
whyrusleeping: note that if using go-like syntax, this is broken. This function will no-op for Secp and Actor. either explicitly say 'fallthrough' or do `case BLS, Actor, SECP256k1:`

address.md
```diff
@@ -0,0 +1,300 @@
...
+    if a[0] != "f" && a[0] != "t" {
+        Fatal(ErrUnknownNetwork)
+    }
+    
+    protocol := a[1]
+    raw := a[2:] 
+    if protocol == ID {
+        return Address{
+            Protocol: protocol,
+            Payload: raw,
```
whyrusleeping: this will need to be decoded from base10

address.md
```diff
@@ -0,0 +1,300 @@
...
+    }
+    
+    payload = raw[:len(raw)-CksmLen]
+    if protocol == SECP256K1 || protocol == Actor {
+        if len(payload) != 20 {
+            Fatal(ErrInvalidBytes)
+        }
+    }
+    
+    cksm := payload[len(payload)-CksmLen : len(payload)]
```
whyrusleeping: protip:
`slice[6:]` means "from index 6 to the end"

address.md
```diff
@@ -0,0 +1,300 @@
...
+    
+    protocol := a[1]
+    raw := a[2:] 
+    if protocol == ID {
+        return Address{
+            Protocol: protocol,
+            Payload: raw,
+        }
+    }
+
```
whyrusleeping: you will need to base32 decode these bits

address.md
```diff
@@ -0,0 +1,300 @@
...
+	}
+	
+    switch a.Protocol {
+    	case SECP256K1:
+    	case Actor:
+    	case BLS:
+    		cksm := Checksum(a)
+    		return network + a.Protocol + base32.Encode(a.Payload + cksm)
+    	case ID:
+        	return network + a.Protocol + base32.Encode(a.Payload)
```
whyrusleeping: we don't use base32 encoding for addresses, we encode to base10

data-structures.md
```diff
@@ -4,6 +4,24 @@ This document serves as an entry point for understanding all of the data structu
...
+    // 0: ID
+    // 1: SECP256K1 Public Key
+    // 2: Actor
+    // 3: BLS Public Key
+    protocol byte
+
+    // raw bytes containing the data associated with protocol
+    payload []byte
+}
+```
```
whyrusleeping: link to the address spec here to discuss serialization.

address.md
frrist: A user could use the same priv/pub key pair on the testnet as they do on the mainnet, this means the same actor could be referenced by the same address on either network (not advocating this, just a possibility). This is not true for ID addresses as they are not deterministically assigned.

This is probably more confusing than helpful, should I take it out?

address.md
frrist: I was trying to indicate that the network and checksum won't always be present by have a "Filecoin address encoded to a byte" section and a "Filecoin address encoded to a string". What would make this more clear? Stating that only the bytes encoding will appear on chain?

address.md
frrist: moved it off the interface

address.md
```diff
@@ -0,0 +1,317 @@
...
+}
+```
+
+##### Encode()
+
+Software encoding a Filecoin address must:
+
+- produce an address encoded to a known network
+- produce an address encoded to a known protocol
+- produce an address with a valid checksum
```
whyrusleeping:
```suggestion
- produce an address with a valid checksum (if applicable)
```

address.md
```diff
@@ -0,0 +1,317 @@
...
+	if network != "f" && network != "t" {
+		Fatal("Invalid Network")
+	}
+
+	switch a.Protocol {
+		case SECP256K1, Actor, BLS:
+			cksm := Checksum(a)
+			return network + a.Protocol + base32.Encode(a.Payload + cksm)
+		case ID:
+			return network + a.Protocol + base10.Encode(a.Payload)
```
whyrusleeping:
```suggestion
			return network + a.Protocol + base10.Encode(varint.Decode(a.Payload))
```

address.md
```diff
@@ -0,0 +1,317 @@
...
+
+	protocol := a[1]
+	raw := a[2:]
+	if protocol == ID {
+		return Address{
+			Protocol: protocol,
+			Payload:  base10.Decode(raw),
+		}
+	}
+
```
whyrusleeping:
```suggestion
raw = base32.Decode(raw)
```

whyrusleeping: the checksum is base32 encoded as well, we have to decode the entire thing before slicing it off

address.md
```diff
@@ -0,0 +1,317 @@
...
+	payload = raw[:len(raw)-CksmLen]
+	if protocol == SECP256K1 || protocol == Actor {
+		if len(payload) != 20 {
+			Fatal(ErrInvalidBytes)
+		}
+	}
+
+	cksm := base32.Decode(payload[len(payload)-CksmLen:])
```
whyrusleeping:
```suggestion
	cksm := payload[len(payload)-CksmLen:]
```

address.md
```diff
@@ -0,0 +1,317 @@
...
+	if !ValidateChecksum(a, cksm) {
+		Fatal(ErrInvalidChecksum)
+	}
+
+	return Address{
+		Protocol: protocol,
+		Payload:  base32.Decode(payload),
```
whyrusleeping:
```suggestion
		Payload:  payload,
```

PR #139, opened 2019-02-12
Add spec for multisig actor wallet

whyrusleeping:
In order to provide a reasonable solution to holding filecoin at launch, we add a multisig account actor.

The design is modeled after the [gnosis multisig wallet](https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWallet.sol), but with only the bare minimum of features (no spend limits or anything funky).

Some design choices that may be worth discussing:

* Transactions are kept around after being executed or canceled, and cleared later by an explicit call to clear transactions. This is done to make auditing easier, seeing transaction history simply requires a lookup in the current state tree, as opposed to crawling back through the chain (being more friendly to lite clients). The alternative is to remove the transaction from the actor state as soon as it is finished. I'm on the fence about which option is best.

* Transactions may be canceled by any of the signers. This seems the simplest approach, though if a signer were to be 'rogue' then they could make it very difficult to spend any funds, or even remove them as a signer. This could be circumvented by asking a miner to mine their transactions all at the same time, preventing the rogue signer from cancelling it. One alternative is to just not have cancelations, and let the 'bad' transaction just hang out forever (bad solution IMO). Another would be to implement some 'required threshold' for canceling a request, this feels a little heavy to me but I could be convinced.

* Self modification methods (add/rm signer, change req) are implemented by doing a multisig transaction invoking the desired method on the contract itself. This means we only have to implement the 'signature threshold' logic in one place, and is generally simpler to reason about. If someone has strong feelings about this, i'm open to discuss.

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -771,6 +771,11 @@ func Propose(to Address, value TokenAmount, method string, params []byte) uint64
 
     self.Transactions.Append(tx)
 
+  if self.Required == 1 {
+    Send(tx.To, tx.Value, tx.Method, tx.Params)
+    tx.Complete = true
+  }
+
     return txid
 }
 ```
@@ -783,7 +788,7 @@ func Approve(txid uint64) {
         Fatal("not authorized")
     }
 
-    tx := getTransction(txid)
+    tx := getTransaction(txid)
     if tx.Complete {
         Fatal("transaction already completed")
     }
@@ -799,7 +804,7 @@ func Approve(txid uint64) {
 
     tx.Approved.Append(msg.From)
 
-    if len(tx.Approved) > self.Required {
+    if len(tx.Approved) >= self.Required {
         Send(tx.To, tx.Value, tx.Method, tx.Params)
         tx.Complete = true
     }
@@ -830,7 +835,7 @@ func Cancel(txid uint64) {
 
 ```go
 func ClearCompleted() {
-    if !isOwner(msg.From) {
+    if !isSigner(msg.From) {
         Fatal("not authorized")
     }
 
@@ -842,8 +847,6 @@ func ClearCompleted() {
 }
 ```
 
-
-
 ### AddSigner
 
 ```go
@@ -852,7 +855,7 @@ func AddSigner(signer Address) {
         Fatal("add signer must be called by wallet itself")
     }
     if isSigner(signer) {
-        Fatal("new address is already an owner")
+        Fatal("new address is already a signer")
     }
 
     self.Signers.Append(signer)
@@ -878,10 +881,14 @@ func RemoveSigner(signer Address) {
 
 ```go
 func ChangeRequirement(req int) {
-    if msg.From != self.Address {
-        Fatal("change requirement must be called by wallet itself")
-    }
+  if msg.From != self.Address {
+    Fatal("change requirement must be called by wallet itself")
+  }
+  if req < 1 {
+	Fatal("requirement must be at least 1")
+  }
 
-    self.Required = req
+  self.Required = req
 }
 ```
+
--- a/actors.md
+++ b/actors.md
@@ -711,7 +711,13 @@ TODO
 
 ## Multisig Account Actor
 
-A basic multisig account actor. Allows sending of messages like a normal account actor, but with the requirement of M of N parties agreeing to the operation. Any party can veto any operation (which is a potentially debateable design decision). Completed and/or cancelled operations stick around in the actors state until explicitly cleared out.
+A basic multisig account actor. Allows sending of messages like a normal account actor, but with the requirement of M of N parties agreeing to the operation. Completed and/or cancelled operations stick around in the actors state until explicitly cleared out. Proposers may cancel transactions they propose, or transactions by proposers who are no longer approved signers.
+
+Self modification methods (add/remove signer, change requirement) are called by
+doing a multisig transaction invoking the desired method on the contract itself. This means the 'signature 
+threshold' logic only needs to be implemented once, in one place.
+
+The [init actor](#init-actor) is used to create new instances of the multisig.
 
 #### State
 
@@ -725,6 +731,7 @@ type Multisig struct {
 }
 
 type Transaction struct {
+  	Created uint64
     TxID uint64
     To Address
     Value TokenAmount
@@ -790,7 +797,7 @@ func Propose(to Address, value TokenAmount, method string, params []byte) uint64
     self.Transactions.Append(tx)
 
   if self.Required == 1 {
-    Send(tx.To, tx.Value, tx.Method, tx.Params)
+    vm.Send(tx.To, tx.Value, tx.Method, tx.Params)
     tx.Complete = true
   }
 
@@ -809,11 +816,11 @@ func Propose(to Address, value TokenAmount, method string, params []byte) uint64
 
 ```go
 func Approve(txid uint64) {
-    if !isSigner(msg.From) {
+    if !self.isSigner(msg.From) {
         Fatal("not authorized")
     }
 
-    tx := getTransaction(txid)
+    tx := self.getTransaction(txid)
     if tx.Complete {
         Fatal("transaction already completed")
     }
@@ -840,17 +847,22 @@ func Approve(txid uint64) {
 
 ```go
 func Cancel(txid uint64) {
-    if !isSigner(msg.From) {
+    if !self.isSigner(msg.From) {
         Fatal("not authorized")
     }
 
-    tx := getTransaction(txid)
+    tx := self.getTransaction(txid)
     if tx.Complete {
         Fatal("cannot cancel completed transaction")
     }
     if tx.Canceled {
         Fatal("transaction already canceled")
     }
+  
+  proposer := tx.Approved[0]
+  if proposer != msg.From && isSigner(proposer) {
+    Fatal("cannot cancel another signers transaction")
+  }
 
     tx.Canceled = true
 }
@@ -860,7 +872,7 @@ func Cancel(txid uint64) {
 
 ```go
 func ClearCompleted() {
-    if !isSigner(msg.From) {
+    if !self.isSigner(msg.From) {
         Fatal("not authorized")
     }
 
@@ -879,7 +891,7 @@ func AddSigner(signer Address) {
     if msg.From != self.Address {
         Fatal("add signer must be called by wallet itself")
     }
-    if isSigner(signer) {
+    if self.isSigner(signer) {
         Fatal("new address is already a signer")
     }
 
@@ -894,7 +906,7 @@ func RemoveSigner(signer Address) {
     if msg.From != self.Address {
         Fatal("remove signer must be called by wallet itself")
     }
-    if !isSigner(signer) {
+    if !self.isSigner(signer) {
         Fatal("given address was not a signer")
     }
 
@@ -909,10 +921,10 @@ func SwapSigner(old, new Address) {
   if msg.From != self.Address {
     Fatal("swap signer must be called by wallet itself")
   }
-  if !isSigner(old) {
+  if !self.isSigner(old) {
     Fatal("given old address was not a signer")
   }
-  if isSigner(new) {
+  if self.isSigner(new) {
     Fatal("given new address was already a signer")
   }
 
@@ -920,8 +932,6 @@ func SwapSigner(old, new Address) {
   self.Signers.Append(new)
 }
 ```
-
-
 
 ### ChangeRequirement
 
@@ -935,6 +945,30 @@ func ChangeRequirement(req int) {
   }
 
   self.Required = req
+}
+```
+
+### Helper Methods
+
+The various helper methods called above are defined here.
+
+```go
+func isSigner(a Address) bool {
+  for signer := range self.Signers {
+    if a == signer {
+      return true
+    }
+  }
+  return false
+}
+
+func getTransaction(txid int) Transaction {
+  tx, ok := self.Transactions[txid]
+  if !ok {
+    Fatal("no such transaction")
+  }
+  
+  return tx
 }
 ```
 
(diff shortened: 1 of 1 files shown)


actors.md
```diff
@@ -478,6 +478,185 @@ func UpdatePeerID(pid PeerID) {
...
+    
+    for _, signer := range tx.Approved {
+        if signer == msg.From {
+            Fatal("already signed this message")
+        }
+    }
+    
+    tx.Approved.Append(msg.From)
+    
+    if len(tx.Approved) > self.Required {
```
dignifiedquire: shouldn’t this be >= ?

actors.md
```diff
@@ -478,6 +478,185 @@ func UpdatePeerID(pid PeerID) {
...
+
+### AddSigner
+
+```go
+func AddSigner(signer Address) {
+    if msg.From != self.Address {
+        Fatal("add signer must be called by wallet itself")
+    }
+    if isSigner(signer) {
+        Fatal("new address is already an owner")
```
dignifiedquire: typo: already a signer

actors.md
```diff
@@ -478,6 +478,185 @@ func UpdatePeerID(pid PeerID) {
...
+    
+    tx.Canceled = true
+}
+```
+
+### ClearCompleted
+
+```go
+func ClearCompleted() {
+    if !isOwner(msg.From) {
```
dignifiedquire: what is the difference between owner and signer? is owner the creator of the wallte? can the ownership be changed?

actors.md
```diff
@@ -478,6 +478,185 @@ func UpdatePeerID(pid PeerID) {
...
+    self.Signers.Append(signer)
+}
+```
+
+### RemoveSigner
+
+```go
+func RemoveSigner(signer Address) {
+    if msg.From != self.Address {
+        Fatal("remove signer must be called by wallet itself")
```
dignifiedquire: this error message confuses me,

actors.md
```diff
@@ -478,6 +478,185 @@ func UpdatePeerID(pid PeerID) {
...
+    
+    self.Signers.Remove(signer)
+}
+```
+
+### ChangeRequirement
+
+```go
+func ChangeRequirement(req int) {
+    if msg.From != self.Address {
```
dignifiedquire: how should this be called? is Address the general address of an actor?

actors.md
whyrusleeping: good catch.

- [ ] Fix

actors.md
whyrusleeping: there is none, i was using the phrasing owner at one point, and switched it to signer. Signer is the appropriate thing to use here

actors.md
whyrusleeping: The way this works is you tell the wallet to send a multisig transaction to itself calling this method. that way we don't have to implement the signature threshold stuff differently for each of these changes.

actors.md
```diff
@@ -478,6 +478,185 @@ func UpdatePeerID(pid PeerID) {
...
+
+### Approve
+
+```go
+func Approve(txid uint64) {
+    if !isSigner(msg.From) {
+        Fatal("not authorized")
+    }
+    
+    tx := getTransction(txid)
```
sternhenri: typo

actors.md
```diff
@@ -478,6 +478,185 @@ func UpdatePeerID(pid PeerID) {
...
+    }
+}
+```
+
+
+
+### AddSigner
+
+```go
+func AddSigner(signer Address) {
```
sternhenri: Forgot to check is function is called by a signer (or arguably should be by owner)

actors.md
```diff
@@ -478,6 +478,185 @@ func UpdatePeerID(pid PeerID) {
...
+
+### ChangeRequirement
+
+```go
+func ChangeRequirement(req int) {
+    if msg.From != self.Address {
+        Fatal("change requirement must be called by wallet itself")
+    }
+    
+    self.Required = req
```
sternhenri: need some sanity checks on the int, i.e. >= 1...

sternhenri:
This is very nifty. Numbering your design considerations, some thoughts to each:
1) I think this is fine indeed. How big an issue is this for lite clients though? It seems like you'd want to keep pretty little in memory and do the actual crawling just once...
2) You have adding and removing of signers done by the wallet address, i.e. by a threshold of signers... Why not apply the same rule to canceling transactions? It seems like the rogue signer issue devolves to a question of who gets to pick signers, so this approach would make sense to me... Make it an m-of-n thing as well.
3) +1

actors.md
sternhenri: from reading the PR notes (not obvious in code itself) i am guessing the self.address is a well-signed threshold, ie m-of-n signers.

whyrusleeping:
Alright @dignifiedquire @sternhenri Now that the init actor PR has merged, i feel this is unblocked. I've rebased it, addressed the concerns raised, and cleaned up a few things. Give it another look over please!

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
 ## Payment Channel Broker Actor
 
 TODO
+
+## Multisig Account Actor
+
+A basic multisig account actor. Allows sending of messages like a normal account actor, but with the requirement of M of N parties agreeing to the operation. Any party can veto any operation (which is a potentially debateable design decision). Completed and/or cancelled operations stick around in the actors state until explicitly cleared out.
+
```
nicola: Meta note, this kind of multi-sig is ok, but since we use BLS signature already anyway, why don't we use threshold signatures instead? This would reduce multisig on-chain footprint. (I might be missing context on threshold sig vs on-chain multi-sig).

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+|  `method` | `string` | The method that will be invoked on the proposed transactions target. |
+| `params` | `[]byte` | The parameters that will be passed to the method invocation on the proposed transactions target. |
+
+```go
+func Propose(to Address, value TokenAmount, method string, params []byte) uint64 {
+    if !isSigner(msg.From) {
+        Fatal("not authorized")
+    }
+
+    txid := self.NextTxID
```
nicola: where do we keep the nextTxID?

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+    self.Signers = signers
+    self.Required = required
+}
+```
+
+
+
+### Propose
+
+> Propose is used to propose a new transaction to be sent by this multisig. The proposer must be a signer, and the proposal also serves as implicit approval from the proposer. If only a single signature is required, then the transaction is executed immediately.
```
nicola: can you specify if this is happening on-chain or off-chain?

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+    tx.Complete = true
+  }
+
+    return txid
+}
+```
+
+### Approve
+
+> Approve is called by a signer to approve a given transaction. If their approval pushes the approvals for this transaction over the threshold, the transaction is executed.
```
nicola: on-chain or off-chain here too

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
 ## Payment Channel Broker Actor
 
 TODO
+
+## Multisig Account Actor
```
nicola: How do I create one of those actors?

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+
+A basic multisig account actor. Allows sending of messages like a normal account actor, but with the requirement of M of N parties agreeing to the operation. Any party can veto any operation (which is a potentially debateable design decision). Completed and/or cancelled operations stick around in the actors state until explicitly cleared out.
```
nicola: technically "any party can veto" is not correct. If you pick n of m and n approve and one veto, transaction gets exectured.

actors.md
nicola: If the answer is "on-chain", a lot of signature collection can definitely happen offchain.

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+    if tx.Complete {
+        Fatal("cannot cancel completed transaction")
+    }
+    if tx.Canceled {
+        Fatal("transaction already canceled")
+    }
+
+    tx.Canceled = true
+}
+```
```
nicola: Can you explain how this can happen?
This sounds like the "veto" power, if n of m required people sign, why would one of the remaining party have power to stop the transaction? this sounds like a setting where is not always desirable

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+    }
+
+    self.Signers.Remove(signer)
+}
+```
+
+### SwapSigner
+
+```go
+func SwapSigner(old, new Address) {
```
nicola: shouldn't be this something approved by the required set of people?

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+  }
+
+  self.Signers.Remove(old)
+  self.Signers.Append(new)
+}
+```
+
+
+
+### ChangeRequirement
```
nicola: shouldn't be this something approved by the required set of people?

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+    TxID uint64
+    To Address
+    Value TokenAmount
+    Method string
+    Params []byte
+    Approved []Address
+    Completed bool
+    Canceled bool
+}
+```
```
dignifiedquire: It would be nice to have a method that queries the current state of a transaction, so I can know if I need/want to do sth about it

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+            self.Transactions.Remove(tx)
+        }
+    }
+}
+```
+
+### AddSigner
+
+```go
+func AddSigner(signer Address) {
```
dignifiedquire: why does this not need approval from others?

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+        tx.Complete = true
+    }
+}
+```
+
+### Cancel
+
+```go
+func Cancel(txid uint64) {
+    if !isSigner(msg.From) {
```
dignifiedquire: why does this not need approval from others?

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+> Approve is called by a signer to approve a given transaction. If their approval pushes the approvals for this transaction over the threshold, the transaction is executed.
+
+| Name     | Type       | Description  |
+| --- | --- | --- |
+| `txid` |  `uint64` | The ID of the transaction to approve. |
+
+
+```go
+func Approve(txid uint64) {
+    if !isSigner(msg.From) {
```
dignifiedquire: `isSigner` is not defined

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+        Fatal("not authorized")
+    }
+
+    tx := getTransaction(txid)
```
dignifiedquire: `getTransaction` is not defined

actors.md
whyrusleeping: Using the init actor.

- [ ] detail how this works

actors.md
whyrusleeping: lets go with "you can cancel your own transaction, or the transactions of anyone who is no longer a signer", and/or "any transaction that hasnt been approved in X blocks can be removed"

whyrusleeping:
TODO: add roughly the following to the spec: "Self modification methods (add/rm signer, change req) are implemented by doing a multisig transaction invoking the desired method on the contract itself. This means we only have to implement the 'signature threshold' logic in one place, and is generally simpler to reason about."

actors.md
sternhenri: On that note, rather than have 
`Completed bool
Canceled bool`

We might just have a State int, with a few potential states: ongoing, completed, canceled...

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+| `to` |  `Address` | The address of the target of the proposed transaction. |
+|  `value` | `TokenAmount` | The amount of funds to send with the proposed transaction |
+|  `method` | `string` | The method that will be invoked on the proposed transactions target. |
+| `params` | `[]byte` | The parameters that will be passed to the method invocation on the proposed transactions target. |
+
+```go
+func Propose(to Address, value TokenAmount, method string, params []byte) uint64 {
+    if !isSigner(msg.From) {
+        Fatal("not authorized")
+    }
```
sternhenri: Shouldn't you conceivably be allowed to propose a transaction for which you are not a signer?

actors.md
```diff
@@ -708,3 +708,233 @@ func UpdatePeerID(pid PeerID) {
...
+}
+```
+
+### Cancel
+
+```go
+func Cancel(txid uint64) {
+    if !isSigner(msg.From) {
+        Fatal("not authorized")
+    }
```
sternhenri: Alternative would be splitting out a `isProposer` permission so only proposer can cancel, as opposed to any signer. I don't think it matters though, somewhat equivalent.

actors.md
sternhenri: IF n have signed, transaction goes through, however someone can cancel ahead of then. This does mean that an evil signer can DoS all m-1 other parties, though the proposer can just remove them from the signer set and try again.

actors.md
sternhenri: +1

actors.md
whyrusleeping: I don't think thats useful. Then any random person can spam your multisig with bad transactions that we have to clear out, and can be misleading (making a typo on an approval transaction much more risky)

sternhenri:
Understood and agreed

PR #151, opened 2019-02-20
faster chain sync protocol, interem solution

whyrusleeping:
We don't yet have graphsync, which would help us avoid having to write our own custom protocol here. But its not that much overhead. The changes to the sync process will still work fine once we have graphsync too.

This isnt as fast as some other chain sync protocols, but its quite simple, and low-trust (downloading a potentially large snapshot is how other protocols gain speed here, but that requires a bit of trust in the other party to not send bad data). There are many improvements we can make, but I think we should opt for simple and fast to implement for now.

### DIFF
--- a/network-protocols.md
+++ b/network-protocols.md
@@ -375,3 +375,48 @@ type RetrievePieceChunk struct {
 ```
 
 TODO: document the query deal interaction
+
+# BlockSync
+
+The blocksync protocol is a small protocol that allows Filecoin nodes to request ranges of blocks from each other. It is a simple request/response protocol with a protocol ID of `/fil/sync/blk/0.0.0`. It uses CBOR-RPC.
+
+```go
+type BlockSyncRequest struct {
+    Start Cid
+    RequestLength uint64
+}
+```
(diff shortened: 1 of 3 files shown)


network-protocols.md
```diff
@@ -375,3 +375,48 @@ type RetrievePieceChunk struct {
...
 TODO: document the query deal interaction
+
+# BlockSync
+
+The blocksync protocol is a small protocol that allows Filecoin nodes to request ranges of blocks from each other. It is a simple request/response protocol with a protocol ID of `/fil/sync/blk/0.0.0`. It uses CBOR-RPC.
+
+```go
+type BlockSyncRequest struct {
+    Start Cid
+    RequestLength uint64
```
frrist: nit: can we call this `NumBlocks` or something that indicates we are talking about some number of blocks here?

whyrusleeping: Maybe chain length would be better. NumBlocks would be more misleading due to EC not giving us a fixed number of blocks for a chain of given length

PR #160, opened 2019-02-28
Feat/consensus review

whyrusleeping:
Cleaning up the mining and EC docs and addressing feedback from review. Still a bit to go though.

@sternhenri could you help out here?


whyrusleeping:
@sternhenri no. there are todos still

sternhenri:
@whyrusleeping took a pass at EC and mining (starting at 'Mining Blocks'), doing the following:
- fixed spacing in code segments with explanatory text
- added some defs to ec
- disambiguated block N, round N, height N
- replaced all mentions of null blocks with losing tickets and new framework for thinking through this.
- removed all mentions of "we" and "you"
- added details on case null blocks extend past k (ie you mine your own tix)
- added details on which losing tix to use if multiple in a round (i.e. use that which leads to winning block rather than smallest in that round).
- reworded slashing rule #2.

I picked up on an interesting issue that affects the question of whether when picking a ticket from a round in which no block was mined we use (1) the smallest ticket in that round or (2) the ticket that led to the eventual block with the smallest ticket. We thought we should do the latter because it made implementation easier. There may be another super subtle (also negligible proba, but interesting) reason.

Let's see how both strategies play out in the case you run out of old tickets to mine on (because of a series of rounds where all miners lose their election longer than K) and start mining atop your own tickets generated mining unsuccessfully. That is all miners mining on private state as opposed to public state. 
For (1): there may be a chance that if multiple blocks are eventually found (rather than just one), none will be valid: a miner would have had to only mined the smallest ticket at every intermediary round (which grows ever more unlikely with the length of the null chain). I don't think there's a way out of this state: all future blocks will be invalid (even creating sybils isn't possible). We lose liveness?
For (2): if multiple blocks are eventually found, only one will be valid: that with the smallest ticket. This does mean wasted work for all other miners who couldn't have known this, but we have liveness, and lost work with a highly negligible proba.

Thus, I think we have to use the solution we landed upon. Lmk if you follow/agree, and if you think this is worth spelling out...

sternhenri:
Also check out the added TipSet condition. I believe it's necessary

expected-consensus.md
```diff
@@ -171,60 +236,53 @@ if winning(electionProof) {
...
+Just as there can have 0 miners win in a round, multiple miners can be elected in a given round . This in turn means multiple blocks can be created in a round. In order to avoid wasting valid work done by miners, EC makes use of all valid blocks generated in a round
 
 ### Tipsets
 
-All valid blocks generated in a round form a 'Tipset' that participants will attempt to mine off of in the subsequent round (see above). Tipsets are valid so long as:
+All valid blocks generated in a round form a `TipSet` that participants will attempt to mine off of in the subsequent round (see above). TipSets are valid so long as:
 
-- All blocks in a Tipset must have the same parents
+- All blocks in a TipSet have the same parent TipSet
+- All blocks in a TipSet have the same number of tickets in their `Tickets` array
```
sternhenri: @whyrusleeping @ZenGround0 spot-check on that extra condition. This is to ensure blocks at a given height were generated in the same number of rounds.

sternhenri:
ignore my earlier rambling. I didn't understand an aspect of how ticket histories are treated. Blocks are independently valid

expected-consensus.md
whyrusleeping: this is to say, they have the same 'round' correct?

sternhenri:
Correct. A ticket is produced at each round. 

On Feb 28, 2019, at 17:30, Whyrusleeping <notifications@github.com> wrote:

@whyrusleeping commented on this pull request.

In expected-consensus.md:

>  
-- All blocks in a Tipset must have the same parents
+- All blocks in a TipSet have the same parent TipSet
+- All blocks in a TipSet have the same number of tickets in their `Tickets` array
this is to say, they have the same 'round' correct?

mining.md
```diff
@@ -200,47 +198,47 @@ func VerifyTicket(b Block) error {
...
 
-If no new blocks appear in the round, you may attempt to mine the same `Tipset` again. In order to do this, simply We call this mining a null block (i.e. mining atop the failed ticket you generated in your previous attempt). 
+### Scratching a losing ticket
 
-To start, you should insert your losing ticket into the `Tickets` array, then repeat the above process (from `Ticket Generation`) using your failed ticket from the previous round rather than the smallest ticket from the parent tipset (multiple null blocks in a row may be found).  This will generate a new ticket.
+If a miner fails to generate a valid `ElectionProof` using their loopback ticket, they may not yet publish a new block at height `H+1`. The miner will likely hear about other blocks being mined on the network at height `H+1` and can thus assemble a new TipSet to mine off of with these blocks (as above).
```
whyrusleeping: s/loopback/lookback/

expected-consensus.md
```diff
@@ -204,7 +204,7 @@ In the case that everybody draws a losing ticket in a given round (i.e. no miner
...
 
-TODO: FIX As was stated above, in the case where the tickets at round `N-K` are all losing tickets, i.e. if no block was generated in round `N-K`, miners should attempt to generate their `ElectionProof` from that ticket which eventually leads to the creation of the block in the next TipSet with the smallest winning ticket, rather than the smallest block at round `N-K` (both are equivalent but this makes implementation simpler).
+TODO: @why to review As was stated above, in the case where there are multiple tickets to choose from at round `N-K` (i.e. if the TipSet eventually created has multiple blocks), miners should attempt to generate their `ElectionProof` from the ticket generated by the block with the smallest final ticket (i.e. not necessarily the smallest ticket generated at that round).
```
whyrusleeping: this sounds technically correct, let's add a TODO to have a diagram to illustrate the problem

whyrusleeping:
one small typo and a TODO to add, otherwise this is looking good, thanks @sternhenri !

expected-consensus.md
sternhenri: yes. a ticket is produced at each round. invariant. will explicit.

sternhenri:
amended. Check the TODOs on EC but otherwise I did a pretty thorough pass. Mining still needs a lot of work.

data-propagation.md
```diff
@@ -0,0 +1,25 @@
+# Filecoin Data Propogation
```
sternhenri: @whyrusleeping WHYYYYYYY :b

data-propagation.md
```diff
@@ -0,0 +1,25 @@
...
+
+The filecoin network needs to broadcast blocks and messages to all peers in the network. This document details how that process works.
+
+Both blocks and messages are propogated using the gossipsub libp2p pubsub router. The pubsub messages are authenticated. For blocks, the pubsub hop validation function is set to check that the block is valid before re-propogating. For messages, a similar validity check is run, the signature must be valid, and the account in question must have enough funds to cover the actions specified.
```
sternhenri: s/propogating/propagating

data-propagation.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Links
+
+- [Gossipsub Spec](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub)
+- [Block Validity Check](mining.md#chain-validation)
+- TODO: Link to message validity check function
+
+## Block Propogation
+
+Blocks are propogated over the libp2p pubsub channel `/fil/blocks`. The block is [serialized](data-structures.md#block) and the raw bytes are sent as the content of the pubsub message. No messages, computed state, or other additional information is sent along with the block message.
```
sternhenri: ibid

data-propagation.md
whyrusleeping: PRs welcome. (somehow this file missed out on my global find and replace...)

sternhenri: https://github.com/filecoin-project/specs/pull/161 :D <3

PR #162, opened 2019-03-01
update slashing rules

ZenGround0:
Motivation [here](https://github.com/filecoin-project/consensus/issues/45).

Pretty simple to state given all of the text generated getting here :)


whyrusleeping:
@ZenGround0 Thanks! This looks good to me. Do you think you could update the pseudocode in actors.md for SlashConsensusFault while we're at it?

sternhenri:
and faults.md :D

whyrusleeping:
additionally, there is some text in validation.md that we should update, or just delete it and link here.

faults.md
```diff
@@ -7,10 +7,10 @@ A fault is what happens when partcipants in the protocol are behaving incorrectl
 ### Consensus Faults
 
 - **Duplicate Block Submission Slashing:**
-  - **Condition:** If any miner posts two different blocks for the same chain height.
+  - **Condition:** If any miner posts two blocks satisfying the slashing conditions defined in [Expected Consensus](./expected-consensus.md).
```
ZenGround0: Attempting to avoid duplication across files.

actors.md
```diff
@@ -57,16 +57,10 @@ Return: None
 
 ```go
 func SlashConsensusFault(block1, block2 BlockHeader) {
-	if block1.Height != block2.Height {
```
ZenGround0: Attempting to avoid duplication across files as fault.md states these things.  If pseudocode is preferable maybe we can move it over there?

actors.md
```diff
@@ -57,16 +57,10 @@ Return: None
...
-    
-    if !ValidateSignature(block1.Signature) || !ValidateSignature(block2.Signature) {
-        Fatal("Invalid blocks")
-    }
-    
-    if AuthorOf(block1) != AuthorOf(block2) {
-        Fatal("blocks must be from the same miner")
+    // see the "Consensus Faults" section of the faults spec (faults.md)
+    // for details.
+    if shouldSlash(block1, block2) {
```
sternhenri: typo here: should read `if !shouldSlash`

expected-consensus.md
```diff
@@ -292,12 +292,11 @@ See the [Faults spec](./faults.md) for implementation details.
...
 
-- A miner submits two blocks for the same round.
-- A miner mines atop a TipSet that should have contained their own block (same height, same parents) but does not.
-  - A miner may omit their own block if they find a winning Ticket in the parent TipSet though they themselves have submitted a smaller (losing) ticket. While it cannot be proven that block omission from a TipSet is malicious (i.e. network latency could simply mean the miner did not receive a particular block), for obvious reasons, a miner must be are aware of the block they previously mined.
-- Also: https://github.com/filecoin-project/consensus/issues/45#issuecomment-468035739
+(1) one block contains at least one ticket in its ticket array generated at the same round as one of the tickets in the other block's ticket array.
+(2) one block's parent is a tipset that does not, but could validly, include the other block.
+  - While it cannot be proven that a miner omits known blocks from a TipSet in general (i.e. network latency could simply mean the miner did not receive a particular block) in this case it can be proven because a miner must be are aware of a block they mined in a previous round.
```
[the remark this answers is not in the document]
sternhenri: +1 on general changes, the condition 2 here is less clear for me than it used to be. 

Will try to touch up

faults.md
sternhenri: should the reference point be ec spec or faults spec @whyrusleeping 

I can quickly see the faults spec becoming a series of links to other docs (which might be fine)

sternhenri:
Last q prior to merge here for me are (for @whyrusleeping):
- Should the conditions be stated in EC.md as opposed to faults.md in attempted dedup effort
- Do we want to include removed pseudocode elsewhere?

expected-consensus.md
```diff
@@ -295,8 +295,9 @@ Due to the existence of potential forks, a miner can try to unduly influence pro
 This is detectable when a miner submits two blocks that satisfy either of the following "slashing conditions":
 
 (1) one block contains at least one ticket in its ticket array generated at the same round as one of the tickets in the other block's ticket array.
-(2) one block's parent is a tipset that does not, but could validly, include the other block.
-  - While it cannot be proven that a miner omits known blocks from a TipSet in general (i.e. network latency could simply mean the miner did not receive a particular block) in this case it can be proven because a miner must be are aware of a block they mined in a previous round.
+(2) one block's ancestor TipSet does not include the other block when it should have.
```
ZenGround0: Ah I misunderstood that this holds for all ancestors instead of just the parent, good catch.

However if you replace ancestor for parent in my comment I still prefer the modification I pushed.  While the comment here uses simpler language and is less dense I believe that it is less exact and meaningful to an implementer of these rules than the words I used.  I will not block the PR but please consider that "should" is laden with assumptions that are not stated.

ZenGround0: Maybe a better phrasing is "one block's ancestor is a tipset that could have validly included the other block according to tipset validity rules, however that ancestor of the first block does not include the other block"

sternhenri: Ok works for me. What do you think of just adding "(in spite of the fact that both were mined by the same miner)" at the end.

If you don't have the fact that the two blocks are from the same miner (which you plainly state up top) cached in, this reads as a much broader condition: i.e. any forgotten block means slash, which makes the below confusing.

Feel free to ignore. Maybe I'm too careful here.

actors.md
whyrusleeping: I think the duplication here is okay. I would leave the pseudocode here.

expected-consensus.md
ZenGround0: This is a genuinely interesting discussion to have about writing style.  My strong belief is that attempting to write no more than what is necessary actually helps clarify communication in contexts like this where readers are looking for precision and paying close attention.  I definitely tend more towards this than a lot of people and maybe too much so.

Fighting for style decisions doesn't feel exactly in scope for me yet (though I hope it will soon as we converge on correctness).  I trust your sense of the spec's style as a whole so by all means add the extra reminder if you feel inspired.

actors.md
ZenGround0: Ok adding it in.  It might get a little confusing to write out new slashing rules but will give it a shot.

ZenGround0: See issue #176 for some thoughts about needed modifications to the spec to make these functions computable from within `SlashConsensusFault`.

expected-consensus.md
sternhenri: no. I agree with you. I have the inverse problem, and i believe synthesis is always better. Having no proof that this is more/less clear to anyone but myself. Leaving as is.

Thanks for engaging earnestly with this.

ZenGround0:
ping @whyrusleeping @sternhenri I am under the assumption that this is ready to merge.  Please let me know anything I need to do to finish this.

expected-consensus.md
```diff
@@ -292,12 +292,12 @@ See the [Faults spec](./faults.md) for implementation details.
...
 
-This happens in the following instances: 
+This is detectable when a miner submits two blocks that satisfy either of the following "slashing conditions":
 
-- A miner submits two blocks for the same round.
-- A miner mines atop a TipSet that should have contained their own block (same height, same parents) but does not.
-  - A miner may omit their own block if they find a winning Ticket in the parent TipSet though they themselves have submitted a smaller (losing) ticket. While it cannot be proven that block omission from a TipSet is malicious (i.e. network latency could simply mean the miner did not receive a particular block), for obvious reasons, a miner must be are aware of the block they previously mined.
-- Also: https://github.com/filecoin-project/consensus/issues/45#issuecomment-468035739
+(1) one block contains at least one ticket in its ticket array generated at the same round as one of the tickets in the other block's ticket array.
+(2) one block's ancestor is a TipSet that could have validly included the other block according to TipSet validity rules, however that ancestor of the first block does not include the other block.
```
whyrusleeping: Just want to make sure that by 'ancestor' we mean explicitly *direct* ancestor. If miner A mines on top of X to produce Y, and another miner mines on top of X to produce Z, then another miner mines on top of Z to produce W, it should not be slashable for miner A to mine on top of W. Mining on top of Z without including Y in their parent tipset is what we're targeting here, correct?

whyrusleeping:
@ZenGround0 one clarification, the rest LGTM

expected-consensus.md
ZenGround0: should be fixed now, thanks for pointing this out

sternhenri:
Sorry, I still don't think parent is right @ZenGround0. I think it is direct ancestor but not any ancestor, ie could be grandparent or grand-grandparent, just not great uncle. Lmk if you disagree @whyrusleeping with what you were saying...

ZenGround0:
@sternhenri I believe you and @whyrusleeping are giving me conflicting directives here.  Specifically I think that by *direct ancestor* @whyrusleeping means what you are calling *parent*.  In the above example @whyrusleeping is saying that you cannot slash a miner who mines and releases both Y and V in this chain:

`X -- Y` this is our miner's attempt to mine on their own chain

`X -- Z -- W -- V` this is our miner giving up on mining their own chain and switching to a chain with a grandparent tipset `Z` that does not include `Y` even though tipset rules would have allowed it.  

As I understand it @whyrusleeping's argument is that we can't penalize this miner from switching because if we did that we would block this miner from switching to the heaviest chain and forever halt this miner's progress.

sternhenri:
You are right and I am wrong! Fuzzy brain, thank you. And all set from me.

ZenGround0:
Alright, I believe this is approved. I cleaned up the commit history. I can't merge myself so I will assume someone else will do it for me.  Please ping me if there is anything else I need to do.  Thanks for the careful review.

actors.md
```diff
@@ -56,6 +56,30 @@ Parameters:
...
+     }
+     
+     // Case B: block1 could have been in block2's parent set but is not
+     block2ParentTipSet := parentOf(block2)
+     if !block2Parent.contains(block1) &&
+     	block2ParentTipSet.Height == block1.Height &&
+	block2ParentTipSet.ParentCids == block1.ParentCids {
+	return true	
+     }
+}
```
sternhenri: Does Go return false by default if you don't specify?

ZenGround0: no I'm dumb

PR #163, opened 2019-03-02
add pseudocode for state machine apply message

whyrusleeping:
working through to make sure every detail of block processing is documented. This bits next in line.

### DIFF
--- a/mining.md
+++ b/mining.md
@@ -250,7 +250,7 @@ An eligible miner can start by filling out `Parents`, `Tickets` and `ElectionPro
 
 Next, they compute the aggregate state of their selected parent blocks, the `ParentState`. This is done by taking the aggregate parent state of the blocks' parent TipSet, sorting the parent blocks by their tickets, and applying each message in each block to that state. Any message whose nonce is already used (duplicate message) in an earlier block should be skipped (application of this message should fail anyway). Note that re-applied messages may result in different receipts than they produced in their original blocks, an open question is how to represent the receipt trie of this tipsets 'virtual block'. For more details on message execution and state transitions, see the [Filecoin state machine](state-machine.md) document.
 
-Once the miner has the aggregate `ParentState`, they must apply the mining reward. This is done by adding the correct amount to the miner owner's account balance in the state tree. (TODO: link to block reward calculation. Currently, the block reward is a fixed 1000 filecoin).
+Once the miner has the aggregate `ParentState`, they must apply the mining reward. This is done by adding the correct amount to the miner owner's account balance in the state tree. See [block reward](#block-rewards) for details.
 
 Now, a set of messages is selected to put into the block. For each message, the miner subtracts `msg.GasPrice * msg.GasLimit` from the sender's account balance, returning a fatal processing error if the sender does not have enough funds (this message should not be included in the chain).
 
@@ -270,6 +270,7 @@ An eligible miner broadcasts the completed block to the network (via [block prop
 
 ### Block Rewards
 
+Currently the block reward is a fixed 1000 FIL. This will change soon.
 
 ### Open Questions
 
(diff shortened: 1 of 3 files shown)


mining.md
```diff
@@ -250,7 +250,7 @@ An eligible miner can start by filling out `Parents`, `Tickets` and `ElectionPro
...
 
-Once the miner has the aggregate `ParentState`, they must apply the mining reward. This is done by adding the correct amount to the miner owner's account balance in the state tree. (TODO: link to block reward calculation. Currently, the block reward is a fixed 1000 filecoin).
+Once the miner has the aggregate `ParentState`, they must apply the mining reward. This is done by adding the correct amount to the miner owner's account balance in the state tree. See [block reward](#block-rewards) for details.
```
acruikshank: In go-filecoin we currently transfer the mining reward before applying the state transformations from the messages. This replicates the old behavior where we sorted the reward message first. It's easy either way, but we either need to change it in the code or change it here.

state-machine.md
```diff
@@ -51,6 +45,44 @@ type VMContext interface {
 
 If the execution completes successfully, changes to the state tree are saved. Otherwise, the message is marked as failed, and any state changes are reverted.
 
+```go
+func ApplyMessage(st StateTree, msg Message) MessageReceipt {
```
acruikshank: This gets the gist of everything `ApplyMessage` does except tracking block gas limits and validation. It's possible that an implementation would implement these outside `ApplyMessage`, but it would be tricky. We track the gas consumed so far in the block and skip any message where `blockGasUsed + message.GasLimit > BlockGasLimit`. We validate:

1. Signature is valid
2. To != From
3. From is AccountActor
4. Sufficient Funds
5. Nonce isn't too low.
6. Nonce isn't too high.

state-machine.md
```diff
@@ -51,6 +45,44 @@ type VMContext interface {
...
+    st.Snapshot()
+    fromActor := st.GetActor(msg.From)
+    
+    totalCost := msg.Value + (msg.GasLimit * msg.GasPrice)
+    if fromActor.Balance < totalCost {
+        Fatal("not enough funds")
+    }
+    
+    st.DeductFunds(msg.From, totalCost)
```
acruikshank: Deducting maximum gas ahead of time and then refunding what's not spent is significantly more complicated to implement correctly than simply deducting spent gas after execution. Since a message execution cannot trigger an account actor to spend more funds, the outcome will always be the same. I've never understood the justification for this.

mining.md
whyrusleeping: I think it largely doesnt matter all that much, but since ethereum applies it at the end of the block processing, i think it would be best to keep things the same as they do it.

state-machine.md
whyrusleeping: We shouldnt be requiring to != from, thats a perfectly valid thing to do.

explicitly checking that 'from' is an account actor feels redundant to checking the validity of the signature.

state-machine.md
whyrusleeping: I don't see how this is significantly more complicated to do. It actually seems about the same complexity, with some nice future proofing guarantees that make it easier to reason about the safety of execution.

acruikshank: Ok. Slightly more complicated (charge limit then either revert and re-charge or refund is more complicated than simply charging at the end). What I don't get is the future proofing guarantees, and why it's easier to reason about safety?

whyrusleeping: For one, it allows other actors you interact with in your calls to reason about account balances.

re: safety, If you withdraw the funds up-front, then you can be sure the actor will be able to pay for its execution no matter what. In the event that its possible for actors to internally choose the amounts that they send to others (its not right now) then withdrawing up front strictly prevents those actors from spending money that would otherwise be used to pay for execution.

whyrusleeping:
Merging so we can use this as the base for the next spec review session. @acruikshank if you want to keep discussing things here please do, I'm happy to continue our conversation.

Issue #126, opened 2019-02-01
Block reward application needs updating and rethinking


whyrusleeping:
We initially had the block reward just be a message from the network actor to the miner. This created an odd special case when validating transactions, and also caused issues with EC transaction grouping (nonce collisions on the message needed an extra special case).

We got rid of all that, and now just directly update the miners balance in the state tree. This is a decent solution, but moving forward we will want to add a constraint that the miner can't spend their block reward funds for some time period. That adds some non-zero complexity.

We should come up with a solution that lets us do this.

One potential idea is to have the miner submit call a 'redeemBlockReward' method on their miner actor, and have this track when each reward becomes spendable. Then later, once spendable, 'withdrawMiningReward' may be called to actually get the funds.

decentralion:
Why do we need the constraint that prevents miners from spending their block reward for some time period? I have a hunch that whatever goal we are aiming for with that mechanism, we could also achieve by adding rules about how the miner can withdraw pledge collateral, and that it will be more convenient to implement it by adding constraints on pledge collateral withdrawal.

cc/ @zixuanzh

sternhenri:
It seems there must be interplay between amount of collateral and block reward (eg block reward can't be higher than collateral, obvs). I understood @whyrusleeping's proposal to be about ensuring that block reward is subject to some of the same withdrawal limits so it can get slashed if abuse is discovered and miners don't run away with ill-gotten gains.

Maybe this can indeed be done away with by designing collateral appropriately (ie taking into account expected reward). Not sure which is better: both are complex in different ways...

Another bad solution here would to be to add block reward to the same holding period as collateral (so presumably it would get slashed as well)... but that seems overly stringent for miners (and a really good incentive to stop mining).

decentralion:
Please take a look at [this design doc][1] on pledge collateral. Our current proposal is that each miner's pledge collateral requirement is a point estimate of their expected block rewards over the coming period `T`, with `T` starting low (e.g. 1 month) during a boostrapping grace period, and then growing to a larger value (eg 12-24 months) in the steady state. 

[...]

PR #165, opened 2019-03-04
write out block reward logic

whyrusleeping:


mining.md
```diff
@@ -270,6 +270,31 @@ An eligible miner broadcasts the completed block to the network (via [block prop
 
 ### Block Rewards
 
+Over the entire lifetime of the protocol, 1,400,000,000 FIL (`TotalIssuance`) will be given out to miners. These funds are initially held by the network account actor, and are transferred to miners in blocks that they mine. The reward amount remains fixed for a period of 1 week (20,160 blocks, the `AdjustmentPeriod`) and is then adjusted.
+
+The equation for the current block reward is of the form:
+
+```
+Reward = IV * (Decay ^ (BlockHeight / 20160))
```
acruikshank: I don't think this is right.

```
HalvingPeriodBlocks = 6 * 365 * 24 * 60 * 2 = 6307200
Decay = e^(ln(0.5) / (6307200 / 20160)) = 0.9977869135615522
IV = 1400000000 * (1 - 0.9977869135615522) = 3098321
```
For the first week BlockHeight / 20160 is 0, so the first block reward is:

```
Reward = IV * Decay^0 = IV = 3098321
```

This is too high. We run out of FIL after 452 blocks (452*3098321 > 1400000000). That's less than 4 hours.

It seems the goal is to halve the FIL in the network actor roughly every 6 years. I'm assuming this based on variable names. It would be nice if that were stated. If this is the case, we want the sum of all block rewards to equal half of the remaining network actor funds each halving period. That is:

∫f(x)_dx_ = TI*(1 - .5 ^ (_x_/HPB))
f(x) = _d/dx_(TI*(1 - .5 ^ (_x_/HPB)))
f(x) = -(TI * ln(.5) / HPB) * .5 ^ (x/HPB)

Where f(x) is the block reward and x is the block height. Summing over this formula for some multiple of the halving period gives:

```
HPB*1: 700000038
HPB*2: 1050000057
HPB*3: 1225000067
```
Which is really close to the intended values.

Holding the reward constant for a week just means replacing x with u=AP*floor(x/AP). sum(-(TI * ln(.5) / HPB) * .5 ^ (u/HPB),0,N) gives:
```
HPB*1: 700775514
HPB*2: 1051163411
HPB*3: 1225000067
```
Which is probably close enough given the other sources of error.

The original is:
```
IV * (D^(x/AP))
TI * (1 - D) * (D^(x/AP))
TI * (1 - .5^(AP/HPB)) * (.5 ^ (AP/HPB)) ^ (u/HPB)
TI * (1 - .5^(AP/HPB)) * .5 ^ (u/HPB)
```
vs.
```
-(TI * ln(.5) / HPB) * .5 ^ (u/HPB)
```
So it's only different by a constant.

whyrusleeping: Oh yeah, forgot to copy a division over from my notes. The Reward here needs to be divided by the AdjustmentPeriod

whyrusleeping:
Here is my derivations: https://observablehq.com/d/e1fdc5a75843b74d

mining.md
```diff
@@ -270,6 +270,31 @@ An eligible miner broadcasts the completed block to the network (via [block prop
 
 ### Block Rewards
 
+Over the entire lifetime of the protocol, 1,400,000,000 FIL (`TotalIssuance`) will be given out to miners. These funds are initially held by the network account actor, and are transferred to miners in blocks that they mine. The reward amount remains fixed for a period of 1 week (20,160 blocks, the `AdjustmentPeriod`) and is then adjusted.
```
sternhenri: 1) The 20,160 assumes 30 sec block time. May be worth mentioning this.

2) Moreover, Unlike with BTC where difficulty is tuned for the block time to remain constant on expectation, we do not do this (yet). Should we leave a TODO here regarding difficulty adjustments pertaining to VDFs? If so, I feel comfortable having the period be counted in time, else we should probably have it be a number of blocks (this would not be as good imo).

3) Why 1 week? I haven't thought this through much, but default value that comes to mind is 2 for obvious reasons. I get why you would want to shorten this, any reason to make it longer?

mining.md
sternhenri: More generally, can we explicitly state intent here (i.e. half every sex years), and then go into how that is executed upon. It may help others reading the spec generally catch things or provide suggestions for improvements of the function (obvs locked in post-testnet but may be useful until then).

mining.md
```diff
@@ -270,6 +270,31 @@ An eligible miner broadcasts the completed block to the network (via [block prop
...
+```
+
+`Decay` is computed by:
+
+```
+Decay = e^(ln(0.5) / (HalvingPeriodBlocks / AdjustmentPeriod))
+```
+
+```
+HalvingPeriodBlocks = 6 * 365 * 24 * 60 * 2
```
sternhenri: same point about the 30 secs assumption. Worth an overall note for the entire section.

sternhenri:
My main asks for touch up would be:
1) Stating intent more broadly in the intro (i.e. divide by 6, will eventually stop yiedling block reward when denomination becomes fractional, etc)
2) Stating assumptions around block time (and potential to-do re difficulty adjustment)
3) Adding details (or TODO) wrt the block reward mechanism and how you actually collect your coinbase (see https://github.com/filecoin-project/specs/issues/126)

mining.md
acruikshank: It turns out these two formulations are approximately equal when HPB >> AP because limit(a->0, (x^a-1)/a) = ln(x). TIL

Looking at your derivations, I don't get how you derived `Decay` or got `IV = TotalIssuance*(1-Decay)/AdjustmentPeriod`. `IV=TotalIssuance*ln(2)/HalvingPeriodBlocks` seems a little simpler and intuitive to me, but they're both correct within tolerances.

I like the way you express the scale as the initial value. 
I'd prefer it if you expressed the rest of function as a power of 1/2:
```
QuantizedBlockHeight = AdjustmentPeriod*⌊BlockHeight/AdjustmentPeriod⌋
Reward = IV * .5 ^ (QuantizedBlockHeight/HalvingPeriodBlocks)
```

At the very least, `Decay` should be expressed as `.5 ^ (AdjustmentPeriod/HalvingPeriodBlocks)`.

mining.md
whyrusleeping: I have no strong reasoning one way or the other on why 1 week. Its somewhere between 'often enough' and 'not too often'. And its long enough that any variance from EC should get averaged out pretty well. 

Thoughts?

whyrusleeping:
Thanks for the review @sternhenri @acruikshank @zixuanzh :)

mining.md
sternhenri: Yeah I think that's fine. Might have said 2 weeks since that's what BTC does, but we can adjust in testing as needed. It seems like a good starting point.

PR #166, opened 2019-03-04
VDF PoSt generation and verification over fixed sectors count

laser:
Our VDP PoSt construction allows us to generate and verify proofs for a fixed number of sectors (`POST_SECTORS_COUNT`). This number is referred to [in the spec](https://github.com/filecoin-project/specs/blob/master/drafts/proof-of-spacetime.md#parameters) as `sectors_count`.

We attempt to hide this detail from the protocol by modifying the `submitPoSt` message to accept a variable number of SNARK proofs which were produced by calling into the PoSt generation routine once per POST_SECTORS_COUNT-sized set of CommRs. The `submitPoSt` message will include a number of proofs equal to `CEIL(LEN(COMM_RS)/POST_SECTORS_COUNT)`.

This modification provides forward compatibility of the `submitPoSt` message if/when we solve the "dynamic sectors count" problem.

CC @porcuquine @dignifiedquire


laser:
Nice. Thanks :)

PR #169, opened 2019-03-05
Init actor sketch

whyrusleeping:
Working on the init actor (we've only been talking about it for two years at this point). This will be the new way to create actors (answering questions for other PRs that needs to be able to create actors), it will also track and allocate actor IDs (though there may be some arguments for not tracking it here, i'll go through that later).

With this, we also remove the 'create actor' from the VM context, cleaning that logic up a bit. The weirdness now is that the init actor is a 'priveleged' actor, in that it can make calls that no other actor can make.

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -13,36 +13,81 @@ The init actor is responsible for creating new actors on the filecoin network. T
 
 ```go
 type InitActor struct {
-    NextActorID uint64
+    // Mapping from Address to ID, for lookups.
+		AddressMap map[Address]BigInt
 }
 ```
 
-#### Exec
+### `Exec(code Cid, params []Param) Address`
 
-Parameters:
+>  This method is the core of the `Init Actor`. It handles instantiating new actors and assigning them their IDs.
 
-- code Cid
-- params []byte
+#### Parameters
 
-TODO: the type of params should probably just be variadic set of arguments. Making it be a `[]byte` here kinda sucks
+| Name     | Type       | Description                                                  |
+| -------- | ---------- | ------------------------------------------------------------ |
+| `code`   | `Cid`      | A pointer to the location at which the code of the actor to create is stored. |
+| `params` | `[] Param` | The parameters passed to the constructor of the actor.       |
+
+
+
+`Param` is the type representing any valid arugment that can be passed to a function.
+
+TODO: Find a better place for this definition.
+
+
 
 ```go
 func Exec(code Cid, params []byte) Address {
-    actorID := self.NextActorID
-    self.NextActorID++
+    // Get the actor ID for this actor.
+    actorID = len(self.AddressMap)
 
-    if code == InitActorCodeCid {
-        Fatal("cannot launch additional init actor instance")
+    // Make sure that only the actors defined in the spec can be launched.
+    if !IsBuiltinActor(code) {
+        Fatal("cannot launch actor instance that is not a builtin actor")
     }
 
-    addr := VM.CreateActor(code, msg.Value, DecodeParams(params))
+    // Ensure that singeltons can be only launched once.
+    // TODO: do we want to enforce this? If so how should actors be marked as such?
+    if IsSingletonActor(code) && self.hasInstance(code) {
+      Fatal("cannot launch singelton, which is already launched")
+    }
 
-    // TODO: where is this mapping stored?
-    SetIDMapping(actorID, addr)
+  	// This call will insert the actor into the global state tree, under the provided ID.
+    addr := VM.CreateNewActor(actorID, code, params)
+
+    // Store the mapping of address to actor ID.
+    self.AddressMap[addr] = actorID
 
     return addr
 }
 ```
+
+### `GetIdForAddress(addr Address) BigInt`
+
+> This method allows for fetching the corresponding ID of a given Address
+
+#### Parameters
+
+| Name   | Type      | Description           |
+| ------ | --------- | --------------------- |
+| `addr` | `Address` | The address to lookup |
+
+
+
+```go
+func GetIdForAddress(addr Address) BigInt {
+    id := self.AddressMap[addr]
+    if id == nil {
+        Fault("unknown address")
+    }
+    return id
+}
+```
+
+
+
+
 
 ## Storage Market Actor
 
--- a/actors.md
+++ b/actors.md
@@ -4,8 +4,45 @@ Any implementations of the Filecoin actors must be exactly byte for byte compati
 
 This spec decsribes a set of actors that operate within the [Filecoin State Machine](state-machine.md). All types are defined in [the basic type encoding spec](data-structures.md#basic-type-encodings).
 
+
 - [Storage Market Actor](#storage-market-actor)
 - [Storage Miner Actor](#storage-miner-actor)
+
+## Init Actor
+The init actor is responsible for creating new actors on the filecoin network. This is a built-in actor and cannot be replicated. In the future, this actor will be responsible for loading new code into the system (for user programmable actors).
+
+```go
+type InitActor struct {
+    NextActorID uint64
+}
+```
+
+#### Exec
+
+Parameters:
+
+- code Cid
+- params []byte
+
+TODO: the type of params should probably just be variadic set of arguments. Making it be a `[]byte` here kinda sucks
+
+```go
+func Exec(code Cid, params []byte) Address {
+    actorID := self.NextActorID
+    self.NextActorID++
+
+    if code == InitActorCodeCid {
+        Fatal("cannot launch additional init actor instance")
+    }
+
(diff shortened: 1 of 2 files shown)


actors.md
```diff
@@ -4,6 +4,42 @@ Any implementations of the Filecoin actors must be exactly byte for byte compati
...
+type InitActor struct {
+    NextActorID uint64
+}
+```
+
+#### Exec
+
+Parameters:
+
+- code Cid
```
dignifiedquire: where do I get this Cid from?

actors.md
```diff
@@ -4,6 +4,42 @@ Any implementations of the Filecoin actors must be exactly byte for byte compati
...
+    actorID := self.NextActorID
+    self.NextActorID++
+    
+    if code == InitActorCodeCid {
+        Fatal("cannot launch additional init actor instance")
+    }
+    
+    addr := VM.CreateActor(code, msg.Value, DecodeParams(params))
+    
+    // TODO: where is this mapping stored?
```
dignifiedquire: might as well store it in this actor?

actors.md
```diff
@@ -4,6 +4,42 @@ Any implementations of the Filecoin actors must be exactly byte for byte compati
...
+
+Parameters:
+
+- code Cid
+- params []byte
+
+TODO: the type of params should probably just be variadic set of arguments. Making it be a `[]byte` here kinda sucks
+
+```go
+func Exec(code Cid, params []byte) Address {
```
dignifiedquire: is the gas cost for this clear?

actors.md
```diff
@@ -4,6 +4,42 @@ Any implementations of the Filecoin actors must be exactly byte for byte compati
...
+    if code == InitActorCodeCid {
+        Fatal("cannot launch additional init actor instance")
+    }
+    
+    addr := VM.CreateActor(code, msg.Value, DecodeParams(params))
+    
+    // TODO: where is this mapping stored?
+    SetIDMapping(actorID, addr)
+    
+    return addr
```
dignifiedquire: how do I get the ID from the addr?

actors.md
```diff
@@ -4,6 +4,42 @@ Any implementations of the Filecoin actors must be exactly byte for byte compati
...
+- params []byte
+
+TODO: the type of params should probably just be variadic set of arguments. Making it be a `[]byte` here kinda sucks
+
+```go
+func Exec(code Cid, params []byte) Address {
+    actorID := self.NextActorID
+    self.NextActorID++
+    
+    if code == InitActorCodeCid {
```
dignifiedquire: I think this should be a white list for now, given there are many actors we don't want other to launch

whyrusleeping: I actually don't mind people launching their own storage market with its own miners. I think that would be interesting. There could be other storage markets for users who don't want to participate in block production, or something.

actors.md
whyrusleeping: I was thinking that nodes would just create and store a reverse index locally. Storing a two-way mapping on-chain seems expensive.

whyrusleeping: alternatively, I was talking with @frrist about a setup where the primary state tree is a `map[ID]Actor`, and separately we store a `map[Address]ID`. This would allow users to shift around keys for their accounts, and maybe other things.

actors.md
whyrusleeping: yeah... that doesnt seem like the worst thing to do. I'll come up with some pros and cons in a bit

actors.md
whyrusleeping: We should have a list of well known CIDs somewhere for the builtin actors

actors.md
frrist: I think we said a `map[PubKey]ID` when we initially talked about this, is that what we mean by `Address` here?

actors.md
```diff
@@ -38,7 +74,7 @@ func CreateStorageMiner(pubkey PublicKey, pledge BytesAmount, pid PeerID) Addres
         Fatal("not enough funds to cover required collateral")
     }
     
-    newminer := VM.CreateActor(MinerActor, msg.Value, pubkey, pledge, pid)
+    newminer := InitActor.Exec(MinerActorCodeCid, EncodeParams(pubkey, pledge, pid))
```
anorth: How about retaining the descriptive `InitActor.CreateActor` name? `Exec` is opaque and also limits future change.

whyrusleeping: We really wanted to stick to unix terminology here, where `exec` is the accepted terminology for taking a given piece of code and instantiating a new process running it.

actors.md
```diff
@@ -4,6 +4,42 @@ Any implementations of the Filecoin actors must be exactly byte for byte compati
...
+```
+
+#### Exec
+
+Parameters:
+
+- code Cid
+- params []byte
+
+TODO: the type of params should probably just be variadic set of arguments. Making it be a `[]byte` here kinda sucks
```
dignifiedquire: @whyrusleeping depends on the level this spec is written, on a low level this has to be `[]byte` serialized in some form to pass through variadic arguments. But if this is meant to describe a high level I would agree

actors.md
dignifiedquire: I think using the `ID` as primary identifier in the state tree would be neat. Having a `map[Address]ID` around, should also address any reordering concerns that could appear due to different `ID` assignments. We could store this mapping in the `InitActor`, given it is the actor that actually adds entries to this mapping. 

Allowing key changes would need some more thought, because if I simply change my key there is no information about my new address (the one derived from my new key) in the `Addresss -> ID` map. So this would require having at least an `AddMapping(from Address, to ID)` call to the `InitActor` (or the place that manages this mapping)

actors.md
dignifiedquire: Should this be added, or should all gas related things be added in one go later?

actors.md
dignifiedquire: If we allow that, then we should have a way to ask the chain directly for the official versions, to reduce chances that anyone is scammed/or accidentally uses a wrong version

actors.md
whyrusleeping: Let's do gas related stuff later in a big gas push

actors.md
whyrusleeping: Yeah, i'm in support of IDs being used as the primary mapping.

actors.md
whyrusleeping: yeah... maybe it makes the most sense to not allow other storage markets for now then too. Can always re-enable it later

actors.md
whyrusleeping: Yeah, I think this can be a little higher level, how things get encoded precisely is discussed elsewhere. But this document should match closely the actual actors that get written.

actors.md
```diff
@@ -2,10 +2,94 @@
...
+
+`Param` is the type representing any valid arugment that can be passed to a function.
+
+TODO: Find a better place for this definition.
+
+
+```go
+func Exec(code Cid, params []byte) Address {
+    // Get the actor ID for this actor.
+    actorID = len(self.AddressMap)
```
whyrusleeping: I think relying on the size of this map to keep track of the next ID may end up being error prone, i'm worried that if we allow actors to be deleted, this will result in unexpected behavior

actors.md
```diff
@@ -2,10 +2,94 @@
...
+
+    // Make sure that only the actors defined in the spec can be launched.
+    if !IsBuiltinActor(code) {
+        Fatal("cannot launch actor instance that is not a builtin actor")
+    }
+
+    // Ensure that singeltons can be only launched once.
+    // TODO: do we want to enforce this? If so how should actors be marked as such?
+    if IsSingletonActor(code) && self.hasInstance(code) {
```
whyrusleeping: tricky.  I see what youre doing here, making it so that we can use the init actor itself to launch the other protected actors at network launch. I think maybe we can fake that all for now, and just have a simple whitelist here.

actors.md
dignifiedquire: done

actors.md
dignifiedquire: done

actors.md
dignifiedquire: done

actors.md
dignifiedquire: You are not wrong, it just felt nicer to use this given the fact that we don't have deletion atm, but I am fine reverting to keeping track of the id manually

actors.md
dignifiedquire: I just realized something, do we have a way to get the code Cid from the actor currently? If not we should have this, such that I can check if a certain actor is spawned from this particular code cid

whyrusleeping: we should definitely have that, i don't think its exposed right now in an easy to use way

whyrusleeping:
@dignifiedquire thanks for the help! I expanded on the `CreateActor` call (broke it into the individual steps to make it clearer, since thats not documented anywhere else), and switched back to using the the explicitly tracked actor ID in the state.

PR #183, opened 2019-03-09
replace lists of checks with pseudocode

whyrusleeping:
Had @jbenet review these docs and his main comment was "more pseudocode, less bulleted lists". Primarily because bulleted lists leave more room for interpretation that doesnt tend to exist in pseudocode.


expected-consensus.md
```diff
@@ -39,22 +40,65 @@ Expected Consensus is a probabilistic Byzantine fault-tolerant consensus protoco
...
-- To check if a miner has won for round `N`:
-  - Sample the secure chain randomness at round `N-K` to receive a ticket (see [Tickets](#tickets) for details). `K` is called the randomness lookback parameter, currently set to 1.
-  - Compute an `ElectionProof` using this ticket as input.
-  - If the hash of the `ElectionProof` is less than the ratio of the miner's power to the total power in the network at the end of round `N-L`, the miner has a winning ticket. `L` is called the committee lookback parameter, currently set to `1`.
-  - The `ElectionProof` will be included in the new block header.
-- If the miner is a winner, generate a new block:
-  - Select the smallest ticket from the heaviest TipSet at round `N-1`.
-  - Generate a new ticket from it for inclusion in the new block.
-  - Generate a new block as shown in [the mining spec](./mining.md#block-creation).
+For each block received over the network, `OnBlockReceived` is called. `IsValid`
```
sternhenri: I think you might be meaning to end the sentence after
> `IsValid`
?

expected-consensus.md
```diff
@@ -39,22 +40,65 @@ Expected Consensus is a probabilistic Byzantine fault-tolerant consensus protoco
...
+func Mine(minerKey PrivateKey) {
+    for r := range rounds { // for each round
+        bestTipset, tickets := ChainTipsMgr.GetBestTipsetAtRound(r-1)
+        
+        ticket := GenerateTicket(minerKey, bestTipset)
+        tickets.Append(ticket)
+        
+        // Generate an election proof and check if we win
+        // Note: even if we don't win, we want to make a ticket
+        // in case we need to mine a null round
```
sternhenri: I might amend "in case we need to mine a null round" to "to respect the invariant that there are always at least as many tickets as blocks", ie every ticket consumed must lead to a new ticket creation...

De facto the same, but i feel like my suggestion is a bit more of a general intuition

expected-consensus.md
```diff
@@ -39,22 +40,65 @@ Expected Consensus is a probabilistic Byzantine fault-tolerant consensus protoco
...
+        win, proof := CheckIfWinnerAtRound(minerKey, r, bestTipset)
+        if win {
+            GenerateAndBroadcastBlock(bestTipset, tickets, proof)
+        } else {
+            // Even if we don't win, add our ticket to the tracker in 
+            // case we need to mine on it later.
+            ChainTipsMgr.AddFailedTicket(bestTipset, tickets)
```
sternhenri: nit: this may be clearer if you actually clear the tickets array on a win (ie line 74), and otherwise have it be initialized outside the for loop, which is to say the miner always maintains a running ticket array. Whenever we can include these in a block (ie on a win) or a new block comes in, it gets reset, otherwise it grows.

Nvm i've convinced myself that capturing all of that logic in Mine is bad. Better to do it through the chainTipsMgr.

But in that case I may nonetheless have it clear tickets after the win, (just as you would taking in a block)

mining.md
```diff
@@ -107,26 +107,54 @@ Chain selection is a crucial component of how the Filecoin blockchain works. Eve
...
-   - Each of the blocks in the block's `ParentTipset` must have the same `ParentWeight`.
-   - The new block's `ParentWeight` must have been properly calculated using the [chain weighting function](./expected-consensus#chain-weighting).
-3. Validate `StateRoot`
-   - In order to do this, a miner must ensure that applying the block's messages to the `ParentState` (not included in block header) appropriately yields the `StateRoot` and the receipts in the `ReceiptRoot`.
-4. Validate that the tickets in the `Tickets` array are valid, thereby proving appropriate delay for this block creation.
-   - Ensure that the new ticket was generated from the smallest ticket in the block's parent TipSet (at round `N-1`).
-   - Recompute the new ticket, using the miner's public key, ensuring it was computed appropriately.
-5. In a case where the block contains multiple values in the `Tickets` array
-   - Ensure that all tickets are signed by the same key.
-   - Ensure that each ticket was used to generate the next, with the first generated from the smallest ticket in the Parent tipset (at `N-1`, as above).
```
sternhenri: either the text of 4 and 5 (ie tickets verif details) should be included as a comment here, or `verifyTickets` should be written out. It isn't defined in EC currently. Though it can be derived from it, we should be explicit here.

sternhenri:
I think this is generally a good idea, with the caveats to pseudocode i've made. a few questions:

- do we lose any intuition from removing the bullets? i.e. short of having both (likely overkill), anything we want to move inside of comments to make sure the point comes across as to **why** things are a certain way? General thought, only instance I actually saw of this was in mining.md for checking Ticket validity on a received block
- EC doc may be missing how receiving blocks and mining new blocks interacts (e.g. with discarding a block if you're mining and you receive a valid one from the network). would be nice to add a bit of detail there.

expected-consensus.md
whyrusleeping: I think this one is a question of audience, the audience here is supposed to be an implementer. Does that change things for you?

expected-consensus.md
whyrusleeping: I think the logic of when to clear stored 'losing' tickets should be left up to the chaintipsmgr, it has to do other maintenance of that set, so doing some here some there feels odd

mining.md
whyrusleeping: VerifyTickets is below, in the ticket validation section.

expected-consensus.md
sternhenri: It does. Your way works

expected-consensus.md
sternhenri: Fair enough, though it's one of the main questions that comes to my mind when reading the spec. Maybe worth mentioning this explicitly (ie that logic is handled by the CTSMgr) or giving a quick written intuition here?

mining.md
sternhenri: perfect. apologies

expected-consensus.md
```diff
@@ -39,22 +40,65 @@ Expected Consensus is a probabilistic Byzantine fault-tolerant consensus protoco
...
-- To check if a miner has won for round `N`:
-  - Sample the secure chain randomness at round `N-K` to receive a ticket (see [Tickets](#tickets) for details). `K` is called the randomness lookback parameter, currently set to 1.
-  - Compute an `ElectionProof` using this ticket as input.
-  - If the hash of the `ElectionProof` is less than the ratio of the miner's power to the total power in the network at the end of round `N-L`, the miner has a winning ticket. `L` is called the committee lookback parameter, currently set to `1`.
-  - The `ElectionProof` will be included in the new block header.
-- If the miner is a winner, generate a new block:
-  - Select the smallest ticket from the heaviest TipSet at round `N-1`.
-  - Generate a new ticket from it for inclusion in the new block.
-  - Generate a new block as shown in [the mining spec](./mining.md#block-creation).
+For each block received over the network, `OnBlockReceived` is called. `VerifyBlock` is defined in the mining spec.
```
nicola: add link plz

expected-consensus.md
```diff
@@ -224,7 +268,7 @@ Tickets = append(Tickets, newTicket)
 
 // If the current ticket isn't a winner and the block isn't found by another miner,
 // derive a ticket from the last ticket
-for !winning(electionProof) && !blockFound()) {
+for !IsProofAWinner(electionProof) && !blockFound()) {
  newTicket = VRF(VDF(H(newTicket)))
  newElectionProof = Sig(H(ticketFromRound(curRound-K)))
```
nicola: shouldn't this use the newTicket?

sternhenri: newTicket is generated from one round back (to prove proper delay) whereas electionProof is from k blocks back (with k the lookback parameter for randomness). That's why we use electionProof here. Does that make sense?

Issue #118, opened 2019-01-25
Host compiled spec output for public consumption


anorth:
When specs are released to the public, we should host the compiled Gitbook HTML and PDF outputs so that readers don't need to install npm, gitbook etc to get a good navigation experience. This is broken out from #113.

Possibly something as simple as pointing to CircleCI build outputs would suffice.

## Acceptance
- Up-to-date HTML and PDF outputs are available on the public internet with a stable URL.
- Not manual step is required to publish new outputs when the non-draft specs are updated.

pooja:
Hey all. I put together an early version of this here: https://filecoin-1.gitbook.io/spec/

Once we finalize the structure for the Filecoin website and the documentation site, I will set this up at a custom domain.

Please provide feedback so we can make this better!

gnunicorn:
>Hey all. I put together an early version of this here: https://filecoin-1.gitbook.io/spec/
Thanks! Am I mistaken or doesn't this provide PDF yet?

pooja:
Sorry for the delay in response @gnunicorn. You are correct, it does not provide PDF 🤦‍♀️. Will look into that!

dignifiedquire:
we now host the spec at https://filecoin-project.github.io/specs/

dignifiedquire:
and the pdf output is tracked here: #266

PR #188, opened 2019-03-12
feat: basic support for bikeshed

dignifiedquire:
Build using `make bikeshed && open docs/index.html`

Preview at https://ipfs.io/ipfs/QmbcjRWUabVHvVFVJsrJBFJu4rj8tv6VUJ9cZqb71jpMdA

There are many improvements that can be done, but this gets the basics working. Main thing to fix would be the links. (I fixed some in `INTRO.md` for demonstration purposes). I can do that if this is something we want to merge and continue with.


dignifiedquire:
I added the rendered version to `docs/index.html`, this way we could host it directly through github pages for now

dignifiedquire:
@whyrusleeping ping, do we want to move forward with this or not?

bvohaska:
@dignifiedquire: for people who may not know why we are doing this, can you provide some context/exposition about why we are moving to bikeshed over gitbook?

dignifiedquire:
Why bikeshed can be better for us than gitbook:

- supported (gitbook cli is deprecated/unsupported, they moved to a hosted version)
- bikeshed is actually made for writing specs, which adds features like
  - automatic linking
  - auto highlighting of elements like `Example`, `Note` and `Issue`s 
  - proper support for definitions, with Index generation
  - proper bibliography support
  - many more things: https://tabatkins.github.io/bikeshed

whyrusleeping:
@dignifiedquire yeah, we should probably move forward with this, sorry for the delay

dignifiedquire:
- I update the branch to the latest version
- My suggestion would be to merge this, so that I don't have to keep maintaining this branch, and we can fix up things in follow up  PRs. 

One unfortunate thing I noticed is that there is no table support yet https://github.com/tabatkins/bikeshed/issues/1128, so we will need to convert those to html tables to render properly for now.

pooja:
@dignifiedquire I wasn't aware of this PR, but I just set up a hosted version of the spec on Gitbook: https://filecoin-1.gitbook.io/spec/

I'm guessing that if we move forward with bikeshed, we won't be able to use Gitbook for hosting anymore, but we still need to have a hosted version of the spec (as is described in #118). That is the most reasonable way for people to engage with the spec longer-term. Gitbook fortunately made this hosting process really easy (and it's free) to integrate/sync between GitHub and Gitbook.

Not sure how it will work for Bikeshed, seems like it can be done but maybe it's a bit of additional work? Also Gitbook's hosted output is WAY prettier than what I've seen with Bikeshed, and it feels much more in line with modern design sensibilities, whereas the Bikeshed outputs seem extremely old-school (not in a good way). Unless there is a really strong need to move to Bikeshed (@whyrusleeping can you weigh in here?) maybe we can stick with Gitbook?

If we really want this, happy to explore how to have hosted output for the spec on Bikeshed as well.

pooja:
Then again, maybe we want a different look for the spec than we would for other types of documentation? One other consideration is that eventually, it would be great to link all of our different types of documentation -- rust-proofs, go-filecoin, developer API, spec, etc through a Filecoin docs interface. I don't think this means that they necessarily all have to share the same design style, but I think it would be nice!

dignifiedquire:
Hosting is easier and we can do it directly from github with bikeshed, the html that is part of this branch can be automatically hosted through github pages if we enable it.

dignifiedquire:
> Also Gitbook's hosted output is WAY prettier than what I've seen with Bikeshed, and it feels much more in line with modern design sensibilities, whereas the Bikeshed outputs seem extremely old-school (not in a good way).

Gitbook might be visually simpler, but it is missing many parts that bikeshed helps us with interms of organizing the spec, as it is more than just a markdown renderer. Things like proper deeplinking for definitions will be very hard to accomplish with just markdown. 

We can always change the css of bikeshed if we want to give it a more modern feeling.

dignifiedquire:
Also, as comparison the most widely used specs are RFCS which look event simpler: https://tools.ietf.org/html/rfc4880 and the ecmascript (javascript) spec looks like this: https://www.ecma-international.org/ecma-262/9.0/index.html

gnunicorn:
I like bikeshed, it looks a lot more practical for what we are attempting do. However, I do share the concerns mentioned: the design is inline with classic specs/RFCs but not really that clean or modern - most annoyingly it is a single file :( . Or did you find any way to have it rendered into multiple files?

Secondly, this seems to target html-output only. I am not sure how much we relied on or wanted to support for ebook and PDF rendered outputs, but I didn't see this being supported by bikeshed...

The last concern is just about the way it is approached in this PR, where it is concatenating all files into a big spec-file, though I found [bikeshed supports including files](https://tabatkins.github.io/bikeshed/#including) and I think we should use that instead.

dignifiedquire:
> The last concern is just about the way it is approached in this PR, where it is concatenating all files into a big spec-file, though I found [bikeshed supports including files](https://tabatkins.github.io/bikeshed/#including) and I think we should use that instead.

The pr is not doing that actually, it just does that when generating the spec, in a script to make it easier for bikeshed to process, the source remains in separate files

dignifiedquire:
Closing in favor of #262

Issue #174, opened 2019-03-05
The diagrams issue


mhammersley:
Diagrams/illustrations will improve understanding of the spec; this issue should serve as the running list for all of them.
@dignifiedquire pointed to this tool as being particularly easy for creating diagrams (flow diagrams in particular): https://mermaidjs.github.io/sequenceDiagram.html

- [ ] state machine diagram for actors.md
- [ ] things that are on-chain vs things that are off-chain (related to storage-market.md)
- [ ] ticket chain growth and sampling as part of expected consensus
- [ ] slashing condition visualizations

bvohaska:
@meiqimichelle for awareness

sternhenri:
All tied to EC:
- [ ] Diagram for min ticket selection from Tipsets
- [ ] Diagram for tie breaking for Tipsets of equal weight (on different forks) iterating through tickets to find min
- [ ] Diagram for ticket production per round vs block production

meiqimichelle:
Curious if you ended up trying Mermaid and have feels -- thinking about using it for some Cluster diagrams (well, the ones that need to be easily updated on a regular basis).

sternhenri:
I did not alas :/

PR #190, opened 2019-03-13
added a tie-breaker to equal weighted, same min ticket tipsets

sternhenri:
May require a drawing. Otherwise, might be able to simplify language...

Feedback welcome.


pooja:
@sternhenri Is this PR blocked for any reason or can it be merged? Looks like you haven't gotten any comments on it. Can you tag people you would like to review this in particular? 

Also, can you please add a note about what type of diagram you'd need to issue #174 so we can keep track and find someone to help with that?

sternhenri:
Need review to merge but all set on my end. @whyrusleeping

Issue #130, opened 2019-02-06
Discussion: Soft block delay enforcement


whyrusleeping:
Even if its susceptible to miners colluding to circumvent it, we should consider adding in block delay enforcement to the protocol. How this roughly works is each miner includes a timestamp in their block, and upon receiving blocks from other miners checks two things:

- the timestamp on the block is not in the future
- the timestamp on the block is at least 30 seconds higher than the latest of its parents

If the received block does not meet both conditions, it should be thrown away.

Pros: 
- Enforces block time as long as miners don't collude and agree to ignore the rule
- Works even without VDFs
Cons:
- Can be circumvented by rational miners who collude and agree to ignore the rule
- Requires miners have very well synchronized clocks

The alternative (or complementary) option is to have a 'good' VDF help enforce the delay.

sternhenri:
We should do this!

whyrusleeping:
@sternhenri yes, we should. Even with a VDF, this provides us decent base level security (especially with a weak vdf that may be broken easily).

Not super applicable here, but heres how bitcoin uses timestamps: https://medium.com/@lopp/bitcoin-timestamp-security-8dcfc3914da6

pooja:
Closed via #191

PR #191, opened 2019-03-14
added soft block delay

sternhenri:
Referencing https://github.com/filecoin-project/specs/issues/130 for most of the context.

Some things to think about in this PR:
1) We must try to keep the protocol simple. Does this update meet the standard of usefulness to warrant making Filecoin mining more complex and take space on the block? (I posit it does under certain network compositions with enough honest miners).
2) Should there be a greater explanation of why we choose to do this or of the limits of its usefulness (collusion)?
3) Should we give miners more leeway in case they are not fully synchronized with the network?
  - I expect we can answer this question in testnet
  - [Bitcoin](https://medium.com/@lopp/bitcoin-timestamp-security-8dcfc3914da6) does this by checking that blocks were not generated more than MAX_FUTURE_BLOCK_TIME from now, thereby giving some cushion for miners with bad synchrony.
  - This however implies some more complexity to the timestamp rules: it has Bitcoin nodes check that the block's timestamp is greater than the median timestamp from the last 11 blocks, and I believe it is related to some of the rules around how block timestamps are updated in the mempool (eg see TxProbe for details -- blocks get buffer added to them to account for this).


mining.md
```diff
@@ -119,17 +119,27 @@ func VerifyBlock(blk Block) {
         Fatal("invalid block signature")
     }
     
-    // 2. Verify ParentWeight
+    // 2. Verify Timestamp
+    // first check that it is not in the future
+    if blk.GetTime() > nAdjustedTime {
```
whyrusleeping: where is nAdjustedTime defined?

mining.md
```diff
@@ -119,17 +119,27 @@ func VerifyBlock(blk Block) {
...
+        Fatal("block was generated too far in the future")
+    }
+    // next check that it is appropriately delayed from its parents
+    if blk.GetTime() <= blk.maxParentTime() + BLOCK_DELAY {
```
whyrusleeping: Hrm... I think its probably fine to use minParentTime here

mining.md
sternhenri: In the BTC code base :b 

If you like the concept, I'd just say implementation-dependent but could simply be a call to an NTP server...

mining.md
sternhenri: Can you discuss pros and cons here?

It's strictly safer to use maxParentTime, though maybe it incentivizes people to not include all the blocks they hear about...

minParentTime doesn't have this issue, but could lead folks to include invalid future blocks in their TipSet that wouldn't get caught? (I'm not fully paged in at this point) which would be a deal breaker...

mining.md
whyrusleeping: i would be happy if this just said `blk.GetTime() > time.Now()`

mining.md
whyrusleeping: > it incentivizes people to not include all the blocks they hear about...

Right, I think this incentive is a bit weird, and we should avoid it if possible.

Also note: We don't have to restart block generation if we hear about a new parent that doesnt have a smaller ticket, its harmless to include it in our parent set while we're running the vdf.

>  but could lead folks to include invalid future blocks in their TipSet that wouldn't get caught?

I'm not sure I follow here, what exactly could happen?

mining.md
sternhenri: ok, I do think asking folks to refer to NTP would help though, no. 

Is that a broader point we can make to miners somwhere (ie sync your clocks to network time)?

mining.md
sternhenri: nvm. I'm not sure where I was going with this. Would you also include some margin of error as BTC does, like
` minTime + BLOCK_DELAY + MARGIN` or no. 

I tend to no. We can fix in testnet if needed.

sternhenri: Sorry, I've been completely confused about this. Please disregard all other comments (save the one about margin perhaps)

`maxParentTime` gives the miner more time to include new blocks. I expect it's better than `minParentTime` which might be too stringent a rule (in case of network delay eg).

pooja:
@sternhenri It seems like there is still some open discussion that needs to happen surrounding this issue. Do you and @whyrusleeping need to sync on this or is this waiting for another type of review?

sternhenri:
Thanks for jumping in @pooja. In this case, left options in comments but ready to go on my end, just waiting for @whyrusleeping 's notes to merge or adapt...

mining.md
```diff
@@ -113,49 +113,60 @@ In order to validate a block coming in from the network at round `N` was well mi
...
-	for i, msg := range blk.Messages {
-		receipt := ApplyMessage(state, msg)
-		if receipt != blk.MessageReceipts[i] {
-			Fatal("message receipt mismatch")
-		}
-	}
-	if state.Cid() != blk.StateRoot {
-		Fatal("state roots mismatch")
-	}
+<<<<<<< HEAD
```
whyrusleeping: you got a merge conflict here

PR #193, opened 2019-03-15
StorageDeal: transfer mechanisms and assurances

whyrusleeping:
I'm working on defining exactly how the references in the storage deal proposal translate into data that goes into the sector, and also working through how some of the interactions with PoRep (and payments) need to work.

@porcuquine It would be much appreciated if you could take a look at this, and maybe then we could discuss in a call.

### DIFF
--- a/network-protocols.md
+++ b/network-protocols.md
@@ -49,10 +49,10 @@ type HelloMessage struct {
 
 ```go
 func SayHello(p PeerID) {
-    s := OpenStream(p)
-    mes := GetHelloMessage()
+	s := OpenStream(p)
+	mes := GetHelloMessage()
 	WriteCborRPC(s, mes)
-    s.Close()
+	s.Close()
 }
 ```
 
@@ -72,13 +72,13 @@ type StorageDealProposal struct {
 	// certifying the data transfer
 	PieceRef Cid
 
-    // SerializationMode specifies how the graph referenced by 'PieceRef' gets transformed
-    // into the data that will be packed into a sector.
-    SerializationMode string
+	// SerializationMode specifies how the graph referenced by 'PieceRef' gets transformed
+	// into the data that will be packed into a sector.
+	SerializationMode string
 
-    // TranslatedRef is the data hashed in a form that is compatible with the proofs system
-    // TODO: this *could* possibly be combined with the PieceRef
-    TranslatedRef Cid
+	// TranslatedRef is the data hashed in a form that is compatible with the proofs system
+	// TODO: this *could* possibly be combined with the PieceRef
+	TranslatedRef Cid
 
 	Size NumBytes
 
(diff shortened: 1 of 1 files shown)


porcuquine:
@whyrusleeping I've read this now so could discuss in the next few days.

network-protocols.md
```diff
@@ -174,25 +178,35 @@ func SendStorageProposal(miner Address, file Cid, duration NumBlocks, price Toke
...
 
     prop := StorageDealProposal{
-	PieceRef: file,
-	TranslatedRef: commitment,
-	TotalPrice: price * size, // Maybe just leave this to the payment info?
-	Duration: duration,
-	Size: size,
-	Payment: payInfo,
+			PieceRef: file,
+			TranslatedRef: commitment,
```
whyrusleeping: TODO: add SerializationMode to this

network-protocols.md
```diff
@@ -276,9 +293,37 @@ If `response.State` is `Accepted` then the client should proceed to transfer the
...
+    Proposal: prop.Cid(),
+    PieceConfirmation: pieceConfirmation,
+  }
+  miner.Sign(resp)
+  miner.SetDealState(resp)
+}
+```
+
+```go
+func OnSectorSealed(prop StorageDealProposal) {
```
whyrusleeping: note: called once for each deal in the sector

network-protocols.md
```diff
@@ -276,9 +293,37 @@ If `response.State` is `Accepted` then the client should proceed to transfer the
...
+      	Message: "TranslatedRef was invalid, reconstructed data did not match",
+    }
+
+    miner.Sign(resp)
+    miner.SetDealState(resp)
+    return
+  }
+  
+  // TODO: is TranslatedRef actually needed? How does it tie in?
+  pieceConfirmation := SectorBuilder.PackData(prop.PieceRef, prop.SerializationMode)
```
laser: At this point, the `SectorBuilder` will have added a user's piece to a yet-to-be-sealed sector. The miner doesn't have to seal it immediately. How will the miner be able to generate a piece inclusion proof for this piece if the miner hasn't yet filled up and kicked off the sealing process?

network-protocols.md
```diff
@@ -307,30 +307,41 @@ func OnDataReceived(prop StorageDealProposal) {
...
-    PieceConfirmation: pieceConfirmation,
+    PieceConfirmation: pieceConf,
   }
+  
   miner.Sign(resp)
   miner.SetDealState(resp)
 }
 ```
 
+Once the deal makes it to this state, the client should be able to query and get the `PieceConfirmation` that they need to verify that the miner is indeed storing their data.
```
laser: I'm Lovin' It.

porcuquine: @whyrusleeping Although this should be obvious, I want to make explicit that if we do this in the simplest way, `PieceConfirmation`s will take some time to generate because we also have to generate `commD`. So the likely flow is:
 - sector is completely packed.
 - `commD` and `PieceConfirmation`s are generated.
 - `OnSectorPacked` is called.
 -  Sealing beings/continues.

We could also incrementally generate `commD` as pieces are packed, but that's probably not the effort, at least up-front. Like I said, should be obvious, but there's likely to be a delay between when the sector is completely packed and when `OnSectorPacked` can be called. I can't think of a reason why this would matter.

PR #194, opened 2019-03-15
Proofs parameter design notes

nicola:
This is a design doc, eventually it should merge with the proof performance doc.
The proof performance doc, currently does not address the security requirements mentioned in this doc.

One of the next plans is to integrate this into our calculators. Ideally we would have a single document that describes all the security parameters and security targets we chose.

This is part of the effort to find the polling time: https://github.com/filecoin-project/research/issues/94


pooja:
@whyrusleeping @porcuquine Can you review this when you have a chance?

PR #195, opened 2019-03-15
ZigZag Offline circuit spec

porcuquine:


nicola:
Note that this spec has currently a bug: it uses Inclusion proofs instead of ReplicaInclusion proofs. In other words, Inclusion should only be used at the first layer and ReplicaInclusion at all the other layers.

I think there is a way to simplify all of this, which is: only have a single InclusionProof variable and make CommD CommR_0 or CommR_{-1}

nicola:
I gave a last pass at the spec, it should be correct now (solved the issue I mentioned above).

Left a single TODO, but it's minor

porcuquine:
Thank you, @nicola.

zigzag-circuit.md
```diff
@@ -49,37 +49,37 @@ This circuit proves that given a Merkle root `CommD`, `CommR_l`, and `commRStar`
```
porcuquine: It think `ParentInclusionHash` and `ParentInclusionLeaf` should have `{0..EXPANSION_DEGREE+BASE_DEGREE}`.





Also, if we're separating the leaves from the paths, shouldn't those paths  be shorter by 1 — so `TREE_DEPTH-2`?

Related: Below it looks like we are using `{0..TREE_DEPTH}`  in `MerkleTreeVerify`. It should be the same as here (whatever that ends up being).

porcuquine: @nicola

nicola: 1) `ParentInclusionHash` and `ParentInclusionPaths` have `{0..EXPANSION_DEGREE+BASE_DEGREE}`, what do you mean exactly?
2) TREE_DEPTH-1 is including the path being shorter by 1
3) Thanks will update!

porcuquine: Sorry, I think Github hid the context. See lines 45 and 48.

45: `ParentInclusionHash_{l=0..LAYERS}_{c=0..LAYER_CHALLENGES[i]}_{EXPANSION_DEGREE+ BASE_DEGREE}_{0..TREE_DEPTH-1}`

The part with `{EXPANSION_DEGREE+BASE_DEGREE}` isn't specifying a range in our syntax. It should be `{0..EXPANSION_DEGREE +BASE_DEGREE}`.

porcuquine: 2. Maybe we are defining depth differently.

I consider a tree with four nodes to have depth 3. It has three rows with 1, 2, and 4 nodes each.

So an inclusion proof starting at the leaves only needs 3-1=2 path bits and elements. That is, the whole path is TREE_DEPTH-1 in length. That path is composed of 1 leaf and TREE_DEPTH-2 hashes.

I could be seeing this wrong or getting the terminology wrong, but that is what I meant.

nicola: solved point 1, on point 2: TREE_DEPTH-1 means we don't hash the leaf, it would be TREE_DEPTH otherwise (or am I missing something?)

porcuquine: No, I see it now. I was thinking that since we reuse the root we don't have to hash it, but actually we still have to hash the previous row to compute the root. So I think everything is resolved now.

porcuquine:
I can't approve this, since I created the PR. But this comment indicates my approval of @nicola's changes to it. This is work we did together, so if @whyrusleeping wants to cast an eye and give final approval, that would probably be the best way. We think this is ready to merge and would like to get it into the spec now.

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+ZigZag is the Proof of Replication used in Filecoin. The prover encodes the original data into a replica and commits to it. An offline PoRep proves that the commitment to the replica is a valid commitment of the encoded original data.
+
+ZigZag PoRep is based on layering DRG graphs `l` times. The data represented in each DRG layer is the data encoded in the previous layer. The final layer is the replica (which in Filecoin terms is the sealed sector).
+
+- `replicaId` is a unique replica identifier (see the Filecoin Proofs spec for details)
+- `CommD` is the Merkle Tree root hash of the input data to the first layer
+- `CommR_{i}` is the Merkle Tree hash of the output of the DRG encoding at each layer `i` 
+- `CommRStar` is the hash of the concatenation of the `replicaId` and all the `commR`s.
+
+In the PoRep offline proof as presented in Fisch2019, the proof is large for blockchain usage. We use SNARKs to generate a proof of knowledge of an offline proof. In other words, we use encode the ZigZag proof verification algorithm in an arithmetic circuit and use SNARKs to prove that it was evaluated correctly.
```
schomatis: maybe a link to the paper?

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
+# ZigZag: Offline PoRep Circuit Spec
+
+ZigZag is the Proof of Replication used in Filecoin. The prover encodes the original data into a replica and commits to it. An offline PoRep proves that the commitment to the replica is a valid commitment of the encoded original data.
+
+#### ZigZag Overview
+
+ZigZag is the Proof of Replication used in Filecoin. The prover encodes the original data into a replica and commits to it. An offline PoRep proves that the commitment to the replica is a valid commitment of the encoded original data.
```
schomatis: duplicated paragraph

zigzag-circuit.md
schomatis: typo: "use encode" -> "encode" (i think)

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+This circuit proves that given a Merkle root `CommD`, `CommR_l`, and `commRStar`, that the prover knew the correct replicated data at each layer.
+
+## Offline PoRep circuit
+
+**Public Parameters**: *Parameters that are embeded in the circuits or used to generate the circuit*
+
+- `LAYERS : UInt`: Number of layers
+- `LAYER_CHALLENGES[0..LAYERS]`: Number of challenges per layer
+- `EXPANSION_DEGREE: UInt`: Degree of the bipartite expander graph to extend dependencies between layers
+- `TREE_DEPTH: UInt`: Height of the Merkle tree
```
schomatis: nit: I'm assuming that *height* and *depth* are used interchangeably here but it would be nice to unify the terminology

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+
+**Public Inputs**: *Inputs that the prover uses to generate a SNARK proof and that the verifier uses to verify it*
+
+- `ReplicaId : Fr`: A unique identifier for the replica Id
```
schomatis: Is the `Fr` element defined elsewhere?

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+- `CommD : Fr`: the Merkle tree root hash of the original data (input to the first layer)
+- `CommRLast : Fr`: the Merkle tree root hash of the output of the last layer 
+- `CommRStar : Fr`: the aggregate of each layer's Merkle tree root hash
+- Inclusion paths: Binary representation of the Merkle tree path that must be proven packed into a single `Fr` element. We have the following inclusion paths:
+  - `InclusionPaths_{i=0..LAYERS}_{0..LAYER_CHALLENGES[i]}`: At each layer `i` we have `LAYER_CHALLENGES[i]` inclusion paths.
+  - `ParentsInclusionPaths_{l=0..LAYERS}_{c=0..LAYER_CHALLENGES[i]}_{0..EXPANSION_DEGREE+ BASE_DEGREE}`: At each layer `l` we have `LAYER_CHALLENGES[i]` an inclusion path for each parent node of the corresponding `InclusionPaths_{l}_{c}`.
```
schomatis: nit: `l=0` -> `i=0` (same for the other `l` references, we seem to use that letter for the total and `i` for each one)

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+
+**Private Inputs**: *Inputs that the prover uses to generate a SNARK proof, these are not needed by the verifier to verify the proof*
+
+- `CommR_{i=0..LAYERS-1}`: Commitment of the the encoded data at layer `i`. 
+- Inclusion Proof: For each inclusion path in the public inputs, we provide a Merkle Tree path
+  - `InclusionHash_{i=0..LAYERS}_{0..LAYER_CHALLENGES[i]}_{0..TREE_DEPTH-1}`
+  - `ReplicaInclusionHash_{i=0..LAYERS}_{0..LAYER_CHALLENGES[i]}_{0..TREE_DEPTH-1}`
+  - `ParentInclusionHash_{l=0..LAYERS}_{c=0..LAYER_CHALLENGES[i]}_{0..EXPANSION_DEGREE+ BASE_DEGREE}_{0..TREE_DEPTH-1}`
+  - `InclusionLeaf_{i=0..LAYERS}_{0..LAYER_CHALLENGES[i]}`
```
schomatis: could we add some text to the private inputs?

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+        Check MerkleTreeVerify(ReplicaInclusionHash_{l}_{c}_{0..TREE_DEPTH-1})
+        
+        For p = 0..(EXPANSION_DEGREE + BASE_DEGREE):
+        	Check MerkleTreeVerify(ParentInclusionHash_{l}_{c}_{p}_{0..TREE_DEPTH-1})
+        ```
+
+      - Correct layer: **Check** that `CommR_{l}` is matching the Replica Inclusion proofs root hash and `CommR_{l-1}` is matching the Inclusion proof root hash. (If `l=0`, use `CommD` instead, if `l=LAYER-1` use `CommRLast` instead).
+
+        ```
+        Check CommR_{l-1} === InclusionHash_{l}_{c}_{0}
```
schomatis: `_{0}` is the root right?

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+**Circuit:**
+
+- **Check** `ReplicaId` is equal to its bit representation
+
+  ```
+  Assign replica_id_bits = Fr_to_bits(ReplicaId)
+  Check Packed(replica_id_bits) == ReplicaId
+  ```
+
+- For each `l = 0..LAYERS`:
```
schomatis: is `LAYERS-1` the last layer which `CommRLast` references?

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+        ```
+        For each parent = ParentInclusionLeaf_{l}_{c}_{p}`:
+          Assign ParentBits_{l}_{c}_{p} = Fr_to_bits(parent)
+          Check Packed(ParentBits_{l}_{c}_{p}) == parent
+        ```
+
+      - **Check** that the KDF was run correctly:
+
+        ```
+        Assign leaf_bits = ParentBits_{l}_{c}_{0} || .. || ParentBits_{l}_{c}_{EXPANSION_DEGREE+BASE_DEGREE}
```
schomatis: maybe define a `TOTAL_DEGREE = EXPANSION_DEGREE + BASE_DEGREE` to avoid repeating the addition all over the place? ("total" may not be the best term here)

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+
+      - **Check** that each parent has a correct bit representation:
+
+        ```
+        For each parent = ParentInclusionLeaf_{l}_{c}_{p}`:
+          Assign ParentBits_{l}_{c}_{p} = Fr_to_bits(parent)
+          Check Packed(ParentBits_{l}_{c}_{p}) == parent
+        ```
+
+      - **Check** that the KDF was run correctly:
```
schomatis: could we define the KDF? maybe mention digest or pedersen?

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+      - **Check** correct encoding: the decoded leaf equals the input data leaf
+
+        ```
+        Check ReplicationInclusionValue_{l}_{c} == InclusionValue_{l}_{c} + k
+        ```
+
+- CommRStar check: **Check** that CommRStar is computed by concatenating the replica identifier and the commR at each layer (specified in the aux inputs)
+
+  ```
+  Check CommRStar == PedersenHash(ReplicaId || CommR_{0} || .. || CommR_{LAYERS-1} || CommRLast)
```
schomatis: aren't `CommR_{LAYERS-1}` and `CommRLast` the same thing?

zigzag-circuit.md
```diff
@@ -0,0 +1,122 @@
...
+
+        ```
+        Check MerkleTreeVerify(InclusionHash_{l}_{c}_{0..TREE_DEPTH-1})
+        Check MerkleTreeVerify(ReplicaInclusionHash_{l}_{c}_{0..TREE_DEPTH-1})
+        
+        For p = 0..(EXPANSION_DEGREE + BASE_DEGREE):
+        	Check MerkleTreeVerify(ParentInclusionHash_{l}_{c}_{p}_{0..TREE_DEPTH-1})
+        ```
+
+      - Correct layer: **Check** that `CommR_{l}` is matching the Replica Inclusion proofs root hash and `CommR_{l-1}` is matching the Inclusion proof root hash. (If `l=0`, use `CommD` instead, if `l=LAYER-1` use `CommRLast` instead).
```
schomatis: I get the clarification about the `0` and `LAYER-1` corner cases but I think it's a bit ambiguous how it's phrased because we replace `CommD` only in the `CommR_{l-1}` line and `CommRLast` in the `CommR_{l}` line (if I'm following this correctly).

zigzag-circuit.md
nicola: fixing!

zigzag-circuit.md
nicola: fixing!

zigzag-circuit.md
nicola: fixing!

zigzag-circuit.md
nicola: fixing!

zigzag-circuit.md
nicola: fixing!

zigzag-circuit.md
nicola: fixing!

zigzag-circuit.md
nicola: will do~!

zigzag-circuit.md
nicola: that should be LAYERS-2, thank you!

zigzag-circuit.md
nicola: that's right.. I will make it explicit

zigzag-circuit.md
nicola: fixing

zigzag-circuit.md
nicola: fixing

zigzag-circuit.md
nicola: yes

zigzag-circuit.md
nicola: yes

nicola:
@schomatis your comments were really helpful, thank you!

zigzag-circuit.md
nicola: added a glossary at the top

nicola:
Found a couple of bugs, will update the spec (and updating notation too)

nicola:
In a new PR: https://github.com/filecoin-project/specs/pull/207/files

I have:
- Fixed some errors
- Renamed some vars
- Written it in easier to read code-form (that reads like go)

Issue #106, opened 2019-01-23
Filecoin bootstrapping proposal


whyrusleeping:
# Filecoin Bootstrapping Routine
## Connect
If a node has never connected to the network before, They will need to connect to initial set of trusted peers. This will generally be a list of bootstrap peers from the config file, though each individual could come up with a set of peers they explicitly trust.

On successive connects, we can augment the ‘initial connect set’ with randomly selected peers remembered from the previous session.

## Peer Set Expansion
Once connected into some portion of the network, a node will need to expand their peer set in order to get a good sample of the network. This can be accomplished many different ways, the simplest for us will be to use the DHT to do random ‘FindPeer’ requests.

Once the node is connected to a sufficiently diverse set of peers, they should then move on to chain selection.

## Chain Selection
Sample what each of your peers thinks of as the latest chain head. If there are no significant disagreements (an insignificant disagreement would be if two peers have different heads, but one is the parent of the other, or some other similar ancestry chain) then that chain should be given to the syncer routine.

If there are disagreements, then the node should watch both chains for some time, syncing enough data to validate them moving forward, but not historically yet. After a ‘trial period’ has elapsed, the node should select the chain that has included valid proofs for the most storage. The other chain(s) should be marked as bad. The selected chain should be passed to the syncer, and the ‘initial sync’ should begin.

whyrusleeping:
some amount of this has been added in https://github.com/filecoin-project/specs/pull/151

whyrusleeping:
closed by #196

PR #196, opened 2019-03-16
Bootstrap Spec

sternhenri:
This PR is tied to:
- https://github.com/filecoin-project/specs/issues/106
- https://github.com/filecoin-project/specs/pull/151
- https://github.com/filecoin-project/research/issues/49
and is meant to kick off a small iterative process leading to an updated spec (ie do not merge yet. It's not ready).

I've gone ahead and pointed out areas of doubt in comments.

The PR encompasses two changes:
1) the addition of a bootstrap spec, showing off how we initially build up the syncer PeerSet
2) an update to how the peerset is maintained by the syncer, to include refreshing of the PeerSet at fixed intervals, and round-robinning nodes out to ensure we are connected to a representative network over time.


bootstrap.md
```diff
@@ -0,0 +1,136 @@
...
+}
+```
+
+
+
+## Peer Set Expansion
+
+Once connected to some portion of the network, a node will need to expand their peer set in order to get a good sample of the network and a live latest head. To do so, use the DHT to do random `FindPeer` requests. We do so until we have requested 5 new peers from each peer in the PeerSet.
+
+For each of our 30 peers, we perform hello handshakes, and run the following verifications:
```
sternhenri: number here is debatable, but I think it's right order of magnitude

bootstrap.md
```diff
@@ -0,0 +1,136 @@
...
+
+- Ensure that each of our bootstrapper's `trustedHeads.keys` are included as ancestors of our peers' latestHeads.
+- If they are not, include that peer in our syncer's `BadTipsetCache`.
+- If they are, add these peers and latestHeads to our syncer's peerSet/peerHeads (with offset initialConnections)
```
sternhenri: I can fill the above in with pseudocode, provided we agree on this being a good process and way the Syncer is built up by bootstrapping

bootstrap.md
```diff
@@ -0,0 +1,136 @@
...
+
+Repeat this process requesting new peers until your PeerHeads contains at least 25 valid peers, to be used to select a latest head to mine off of.
+
+## Chain Selection
+
+We now have 25 peers mining off of the same trusted ancestor heads. Disagreements in their chains have to do with the fact that these may not yet be confirmed blocks. We simply want to start mining off of the heaviest of their chains to sync up to and mine off of. 
+
+```go
```
sternhenri: Moved this from syncer.

Could be worth merging both specs into there...

network-protocols.md
```diff
@@ -62,6 +62,34 @@ func SayHello(p PeerID) {
 
 Upon receiving a 'hello' stream from another node, you should read off the CBOR RPC message, and then check that the genesis hash matches you genesis hash. If it does not, that node is not part of your network, and should probably be disconnected from. Next, the `HeaviestTipSet`, claimed `HeaviestTipSetHeight`, and peerID of the other node should be passed to the chain sync subsystem.
 
+# Heartbeat
```
sternhenri: Unclear that this should not just be the hello handshake

network-protocols.md
```diff
@@ -62,6 +62,34 @@ func SayHello(p PeerID) {
 
 Upon receiving a 'hello' stream from another node, you should read off the CBOR RPC message, and then check that the genesis hash matches you genesis hash. If it does not, that node is not part of your network, and should probably be disconnected from. Next, the `HeaviestTipSet`, claimed `HeaviestTipSetHeight`, and peerID of the other node should be passed to the chain sync subsystem.
```
sternhenri: I'm super unclear as to why the `HeaviestTipSetHeight` matters. I believe we should only care about the `HeaviestTipSetWeight`.

network-protocols.md
```diff
@@ -62,6 +62,34 @@ func SayHello(p PeerID) {
...
+    serialized := cbor.Marshal(mes)
+
+    WriteUVarint(s, len(serialized))
+
+    s.Write(serialized)
+    s.Close()
+}
+```
+
+Upon receiving a 'heartbeat' stream from another node, you should read off the CBOR RPC message. If no heartbeat is received within HEARTBEAT_TIMEOUT, you should remove the peer from your PeerSet. Otherwise, you can just update their heaviest Tipset in your chain sync subsystem and call `InformNewHead` if their weight is greater than your own.
```
sternhenri: I want this to return a (reponse, error) tuple. Am unclear on libp2p interfaces so leaving as is for now.

sync.md
```diff
@@ -27,14 +27,24 @@ type Syncer struct {
     // handle to the block sync service
     bsync BlockSync
     
+    //peer set
+    peerSet map[PeerID]PeerInfo
+    
     // peer heads
     // Note: clear cache on disconnects
-    peerHeads map[PeerID]Cid
+    peerHeads map[PeerID][]Cid
```
sternhenri: @ZenGround0 i think that was an error?

sync.md
```diff
@@ -181,29 +164,61 @@ func (syncer *Syncer) collectChainCaughtUp(maybeHead TipSet) (Chain, error) {
...
 
-During `syncing` mode a node learns about the newest head of the blockchain through the secure bootstrapping protocol. The syncing protocol then syncs the block headers for that entire chain, and validates their linking. It then fetches all the messages for the chain, and checks all the state transitions between the blocks and that the blocks were correctly created.  If validation passes the node's head is updated to the head tipset received from bootstrapping.
+During `syncing` mode a node learns about the newest head of the blockchain through the secure bootstrapping protocol. The syncing protocol then syncs the block headers for that entire chain, and validates their linking. It then fetches all the messages for the chain, and checks all the state transitions between the blocks and that the blocks were correctly created.  If validation passes the node's head is updated to the head TipSet received from bootstrapping.
```
sternhenri: Will need to update this if we agree on bootstrapping routine.

bootstrap.md
```diff
@@ -0,0 +1,136 @@
...
+type HelloMessage struct {
+HeaviestTipSet []Cid
+HeaviestTipSetHeight uint64
+GenesisHash Cid
+}
+```
+
+followed by a `GetPeers` request, getting back
```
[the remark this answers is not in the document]
ZenGround0: Yes as you suspect right now we do not receive our peers' peers by querying directly like this.  Instead [we](https://github.com/filecoin-project/go-filecoin/blob/master/net/bootstrap.go#L111) do [kademlia DHT bootstrapping](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/dht_bootstrap.go#L149).

sync.md
ZenGround0: Not my code, but looks like it.

bootstrap.md
```diff
@@ -0,0 +1,136 @@
...
+
+## What is bootstrapping in Filecoin?
+
+When a Filecoin node first comes online, it must find the latest head for the chain and then initiate the syncing routine in order to catch up to the chain state and maintain it thereafter as new blocks come in. This is covered in the [sync spec](./sync.md). Bootstrapping is the process through which a node finds the chain's latest head in the first place (thereby enabling it to sync), through other peers in the network.
+
+At its simplest, bootstrapping can be thought of as secure network joining. It must enable a new node to get the latest head/chain that the majority of the network power is mining off and must be both:
+
+- Robust — able to do so in adversarial networks (so long as honest nodes make up the majority of the network).
+
+- Consistent — able to reliably get the canonical chain (even given poor latency or state loss in our initial peer set).
```
whyrusleeping: what do you mean by state loss?

bootstrap.md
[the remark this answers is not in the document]
whyrusleeping: this is an interesting thought. Are the peers kademlia is returning the 'proper' set of peers? 

If so we can continue using it, if not we may need a separate protocol.

If we keep using kademlia for this, we should expose the single 'getPeers' request (kademlia `FIND_NODE` or whatever) so that we can use it here. 

Alternatively, we should ensure that kademlia bootstrapping follows this algorithm exactly.

bootstrap.md
```diff
@@ -0,0 +1,136 @@
...
+    for trustedPeer := range b.trustedPeers {
+        trustedChain := sayHello(trustedPeer)
+        otherPeers := getPeers(trustedPeer).PeerSet
+        
+        b.trustedHeads[trustedChain.HeaviestTipSet] = trustedChain.HeaviestTipSetHeight
+    	
+        if b.genesisHash == nil {
+	        b.genesisHash = trustedChain.GenesisHash           
+        } else if b.genesisHash != trustedChain.GenesisHash {
+            panic("One of your trusted peers is bad.")
```
whyrusleeping: Prefer using `Fatal()` in spec pseudocode, its less language specific and makes it clear that implementers should do whats best for their language

bootstrap.md
```diff
@@ -0,0 +1,136 @@
...
+        }
+        
+        b.bootstrapPeers = append(b.bootstrapPeers, otherPeers)
+    }
+    
+    s.genesis = b.genesisHash
```
whyrusleeping: note: we will know our expected genesis block before starting bootstrapping

bootstrap.md
```diff
@@ -0,0 +1,136 @@
...
+}
+```
+
+
+
+## Peer Set Expansion
+
+Once connected to some portion of the network, a node will need to expand their peer set in order to get a good sample of the network and a live latest head. To do so, use the DHT to do random `FindPeer` requests. We do so until we have requested 5 new peers from each peer in the PeerSet.
```
whyrusleeping: I'm thinking that instead of using "random FindPeer requests", we should have a special dht query method called bootstrap that uses the existing underlying query code, but takes as input some 'trusted set' of peers, and uses this algorithm internally to bootstrap, perhaps even having it accept some function to help implement the filecoin specific functionality.

cc @raulk @anacrolix @stebalien

network-protocols.md
```diff
@@ -62,6 +62,34 @@ func SayHello(p PeerID) {
 
 Upon receiving a 'hello' stream from another node, you should read off the CBOR RPC message, and then check that the genesis hash matches you genesis hash. If it does not, that node is not part of your network, and should probably be disconnected from. Next, the `HeaviestTipSet`, claimed `HeaviestTipSetHeight`, and peerID of the other node should be passed to the chain sync subsystem.
 
+# Heartbeat
+
+The Heartbeat protocol is used to ensure that nodes in a Filecoin node's PeerSet is still running. The libp2p protocol ID for this protocol is `/fil/heartbeat/1.0.0`.
```
whyrusleeping: What is this for exactly? libp2p already gives us whether or not peers are still online and connected, we get notifications from libp2p when peers connect and disconnect.

bootstrap.md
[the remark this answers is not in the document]
anacrolix: I'm unsure what peers you're trying to get? You want the peers of the peer you just connected to, but with respect to what? In DHT you want some target, so you might get the contents of the closest bucket in the routing table. Or do you just want whomever the peer is connected to, in which case it's entirely unrelated to the DHT?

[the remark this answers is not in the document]
anacrolix: It does feel like you're trying to get a random set of peers, in which case the only value the DHT is providing is that its routing table properties and basic operations should ensure a reasonably distributed interconnectivity of all participants. That is that by knowing one node in the DHT, one can traverse to any other node in the DHT.

bootstrap.md
anacrolix: I'm not a blockchain person, but that sounds like something you will publish and will be universally known once it goes into production right? Kind of like the empty hash, or a seed.

bootstrap.md
sternhenri: I was trying to bucket all other non-byzantine reasons other than latency why some node may not be up-to-date...

Better way to phrase?

bootstrap.md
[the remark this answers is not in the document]
sternhenri: Yes, I'm unclear what you mean by "'proper' set of peers" @whyrusleeping, but I'm on the same page @anacrolix for all intents and purposes, I'd want to be able to get back n random peers to get some representative sample of the network calling this enough times.

bootstrap.md
sternhenri: Yes, that is right @anacrolix. My thought initially @whyrusleeping was that your set of trusted peers may actually determine which chain you want to sync to, as in if they collectively decide to abandon a Filecoin for another, you'd want to follow them. But that may clearly be the wrong trust model (you see, I'm a Stellar fanboy).

Should I just hardcode GENESIS in there?

bootstrap.md
sternhenri: It seems to me that this bootstrapping should be implemented at the FIL level actually: it seems to be pretty specific for something like libp2p (eg even the notion of a trusted set of peers). So FindPeer would be a way to do non-content-based peer retrieval (i.e. `FindRandomPeer`) composed in Filecoin (or whatever other user of libp2p) to make up a bootstrap (eg).

network-protocols.md
sternhenri: Perfect. I didn't know this. Is there some sort of callback or event-listener we can put on this in libp2p: would be useful to ensure that nodes get cycles out of the peerset appropriately as they become unresponsive (in addition to doing it naturally to ensure you have wide network connectivity)

sternhenri: ie I want a replacement node every time one times out.

network-protocols.md
whyrusleeping: I don't think we need an explicit heartbeat thing

network-protocols.md
whyrusleeping: Youre right, weight matters not height

bootstrap.md
whyrusleeping: eh, i think thats okay. maybe could specify 'non byzantine state loss' ?

bootstrap.md
```diff
@@ -0,0 +1,136 @@
...
+
+This is obviously impractical, so we choose to start the bootstrap sequence with nodes by connecting with an inital set of trusted peers. These can be nodes run by well-known entities (the Filecoin foundation, large exchanges, good actors in the community, etc) or manually specified by the node operator in the config file. We leave trustless (or fully decentralized) bootstrapping for future work.
+
+A strawman bootstrapping sequence would have nodes use these initial peers to sync up to the canonical chain. However, new connection requests might effecitvely DoS these peers, making the Filecoin network unavailable to new nodes. Therefore, Filecoin bootstrapping seeks to "load-balance" initial requests to ensure the network remains widely available.
+
+We connect to each of these initial peers using the `hello handshake` (see in [network protocols](./network-protocols.md#hello-handshake)), getting back
```
whyrusleeping: minor clarification: we don't connect to them with the hello handshake, we connect and then handshake.

bootstrap.md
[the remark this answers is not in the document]
whyrusleeping: @sternhenri yeah, its not clear to me what exactly 'proper' means here. You tell me. All i'm saying is that if you provide a definition of 'proper' and the DHT satisfies that, then thats great! otherwise we will need to implement a specific protocol for this.

bootstrap.md
whyrusleeping: Yeah, hardcode it.

Also, I'm pretty sure that even in stellar what you describe cannot happen.

bootstrap.md
whyrusleeping: My point was actually that this doesnt seem filecoin specific at all, and is a good general method for connecting to a distributed network given an initial set of peers you know you trust, a way of determining if there are divergent sets, and a mechanism for getting more peers.

bootstrap.md
whyrusleeping: what happens if we don't find 30 peers?

bootstrap.md
whyrusleeping: where does trustedHeads come from?

bootstrap.md
```diff
@@ -0,0 +1,136 @@
...
+
+Once connected to some portion of the network, a node will need to expand their peer set in order to get a good sample of the network and a live latest head. To do so, use the DHT to do random `FindPeer` requests. We do so until we have requested 5 new peers from each peer in the PeerSet.
+
+For each of our 30 peers, we perform hello handshakes, and run the following verifications:
+
+- Ensure that each of our bootstrapper's `trustedHeads.keys` are included as ancestors of our peers' latestHeads.
+- If they are not, include that peer in our syncer's `BadTipsetCache`.
+- If they are, add these peers and latestHeads to our syncer's peerSet/peerHeads (with offset initialConnections)
+
+Repeat this process requesting new peers until your PeerHeads contains at least 25 valid peers, to be used to select a latest head to mine off of.
```
whyrusleeping: where does 25 come from?

network-protocols.md
whyrusleeping: Yeah, you can register for a disconnect notification.

whyrusleeping: The callback for disconnect notifications gives you which peer disconnected, so you just say "tell me about all the disconnects that happen" and its up to you to determine what to do about it.

sync.md
```diff
@@ -181,29 +164,61 @@ func (syncer *Syncer) collectChainCaughtUp(maybeHead TipSet) (Chain, error) {
...
+            }
+        }
+    }
+}
+
+func (syncer *Syncer) replacePeer(peer PeerID) {
+    delete(syncer.PeerSet, peer)
+    delete(syncer.PeerHeads, peer)
+            
+    syncer.getNewPeer()
```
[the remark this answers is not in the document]
whyrusleeping: We should define the requirements here. What should this method do?

bootstrap.md
sternhenri: We assume you're getting eclipsed and abandon the bootstrap. 30 may be the wrong number, but having a fallback in case of failure defeats the point of trying to secure bootstrapping through these heuristics.

bootstrap.md
sternhenri: Ship from a config file with Filecoin nodes. Can be changed manually by clients running non-vanilla filecoin (eg mining pools or the like).

bootstrap.md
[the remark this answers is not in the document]
sternhenri: After a conversation with @anacrolix, we have a strategy. First, to define *proper*, we mean a random peer. Basically, we want to query peers to make up a "representative" set of the network, without risking getting peers from given clusters. Specifically, we are looking for this set to be pretty random (and not tied to network topography, latency, etc) to
1) avoid hotspots
2) make eclipse harder to build

So we want to treat the Filecoin nodes and underlying libp2p network as two distinct networks for bootstrapping purposes. While we'll use the DHT later for routing, we don't allow it to affect choice of peers early on.

Typical DHT bootstrapping would mean querying your own ID (to get those nodes closest to you). Instead here, we will successively query random IDs (25 times) round-robinning our trusted nodes as starting points. We'll keep the nodes with IDs closest to our random one in our peer set.

We use libp2p's [GetClosestPeers](https://godoc.org/github.com/libp2p/go-libp2p-kad-dht#IpfsDHT.GetClosestPeers) fn.

All this implies Fil node IDs are randomly generated (so we don't accidentally get back a cluster).

network-protocols.md
sternhenri: Agreed perhaps in form, though it would be helpful to have some functionality to periodically query peers to make sure they aren't tracking a heavier chain than you... Thoughts?

sync.md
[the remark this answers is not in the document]
sternhenri: My point here was to keep the peer set chaning through time. Cycle out nodes on an untrusted basis and replace peers as you go to keep making sure you're tracking the appropriate chain head. 

This could be overkill to start. Without a well defined heartbeat, it may be worth just ripping out altogether. Kill or keep?

bootstrap.md
sternhenri: fair enough. implemented here in case, but happy to help move this to libp2p if better.

bootstrap.md
sternhenri: seemed a good number. up to you!

pooja:
@sternhenri @whyrusleeping How can I help move this PR forward? Looks like there are still some things that need to be resolved, but not sure what complete state is.

sternhenri:
It's close but not quite there yet. Needs a couple more passes from @whyrusleeping or @ZenGround0, or a 40-min live meeting.

raulk:
@sternhenri @whyrusleeping sorry I’m late to the party. I’m catching up now, but I wanted to leave a passer-by note. I suspect a DHT is the wrong tool for the job. It sounds like you may be looking for _uniform, random membership sampling_.

Brahms is an interesting solution that Polkadot is evaluating (it was Pierre who put it in my radar), and Eth2.0 has also considered it: http://webee.technion.ac.il/~idish/ftp/brahms_tr.pdf. I’d love to offer it in go-libp2p, as we have the pubsub primitives to do so.

During my research I also dug up RankSlicing: https://www.researchgate.net/publication/267211103_RankSlicing_A_decentralized_protocol_for_supernode_selection. They cite a mechanism called WPSS for maintaining an overlay. I forgot how it works, but it looked interesting.

One can imagine coalescing various mechanisms to provide the guarantees you need.

raulk:
^^ feel free to disregard those pointers if you’re looking for an approach that operates atop the status quo of libp2p (with relatively minor changes). I thought it was worth bringing these up to cross-pollinate.

bootstrap.md
sternhenri: bootstrap issue on beginning/how people connect:
Sol: sync with trusted peers directly (we assume they reject the sync requests if they can see enough participants in the network). WIll work at network launch and if network dips below n miners.

network-protocols.md
sternhenri: silly q. block propagation takes care of this

sternhenri:
Hey @raulk! @whyrusleeping and I discussed Brahms and agree it's a promising option!

I'm of the mind that in the interim DHTs will do, but if libp2p begins to offer this, we would touch up the spec accordingly in a heartbeat...

sternhenri:
Chose to leave bootstrap.md and sync.md separate to not have an overly long spec, but they are highly interdependent.

Review needed from @whyrusleeping or @ZenGround0 .

sync.md
```diff
@@ -1,131 +1,153 @@
...
+    for head.Height() > 0 {
+        // NB: GetBlocks validates that the blocks are in-fact the ones we
+        // requested, and that they are correctly linked to each other. It does
+        // not validate any state transitions
+        blks := syncer.bsync.GetBlocks(head, RequestWidth)
+        blockSet.Insert(blks)
+            
+        head = blks.Last().Parents()
+    }
+
```
whyrusleeping: you dropped checking if the synced chain links back to the expected genesis block. Thats an important part

sternhenri: `SyncBootstrap` only gets called by 
- `InformNewHead`, itself called by `addPeerToSet` --> which checks that the new peer's genesis is GENESIS
- `ExpandPeerSet` upon peer set expansion failure --> in which case we sync up to the `TrustedNodes` all of whose genesis have been checked in `InitialConnect`

So it actually gets checked elsewhere which is why I removed it.

With that said, that is so non-obvious and likely to lead to a regression that I will re-add.

I'd rather not remove the checks elsewhere though as they logically make sense there. The duplication is NBD, comparatively, imo. LMK if you disagree!

sternhenri: This is getting a bit too in the weeds for a spec imo: I have fallen into the trap of writing code instead of saying "how" things should work.

Going to use an assertion here instead just to make clear that we didn't forget.

whyrusleeping: @sternhenri It doesnt matter if the peer claims that X is their genesis block. They can still claim that, and then give us a completely different chain.

sternhenri: That's a v good point. Well taken!

With that said, does [l114 in syncer](https://github.com/filecoin-project/specs/pull/196/files#diff-3d25bc8d97eb696d79b9ab02ade7a634R114) not cover it?

PR #207, opened 2019-03-19
Simpler spec for zigzag

nicola:


zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+
+  Note: Size of proof per layer is `TREE_DEPTH-2` because it excludes (1) the roothash, and (2) the leaf value: (1) the root hash of the proof will be the corresponding unencoded data commitment for the current layer `l` (either `CommD` or `CommR[l-1]`), (2) the leaf value is `DataValue[l][c]`
+
+- `ReplicaProof : [LAYERS][][TREE_DEPTH-2]Fr`: Merkle tree inclusion proof for the current layer encoded challenged nodes.
+
+  Note: Size of proof per layer is `TREE_DEPTH-2` because it excludes (1) the roothash, and (2) the leaf value: (1) the root hash of the proof will be the corresponding unencoded data commitment for the current layer  (either `CommR[l]` or `CommRLast`), (2) the leaf value is `ReplicaValue[l][c]`
+
+- `ParentProof : [LAYERS][][PARENT_NODES][TREE_DEPTH-2]Fr`: Pedersen hashes of the Merkle inclusion proofs of the parent nodes for each challenged node at layer `l`.
+
+  Note: Size of proof per layer is `TREE_DEPTH-2` because it excludes (1) the roothash, and (2) the leaf value: (1) the root hash of the proof will be the corresponding unencoded data commitment for the current layer  (either `CommR[l]` or `CommRLast`), (2) the leaf value is `ParentValue[l][c][p]`
```
schomatis: yeah, let's definitely collapse this 3 very similar notes if possible (I get that a spec has to be exhaustive and I don't mean to lose information in the collapse but having the first two lines of the 3 notes being identical it's very easy to pass them by not realizing the difference that lies at the end)

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+
+- `ReplicaValue : [LAYERS][]Fr`: Value of the encoded nodes for each challenged node at layer `l`.
+
+- `ParentValue : [LAYERS][][PARENT_NODES]Fr`: Value of the parent nodes for each challenged node at layer `l`.
+
+##### Design notes
+
+#### Circuit
+
+##### High Level
```
schomatis: niiiiiice

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+for l in range LAYERS {
+  
+  if l == 0 {
+    DataRoot = CommD
+  } else {
+    DataRoot = CommR[l-1]
+  }
+
+  if l == LAYERS-1 {
+    ReplicaRoot = CommRLast
```
schomatis: similar as above note, if this was processed the same as the rest of the `CommR`s there's no need for this `if`, we can just always access `CommR[l]` (if not, ignore this)

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+  } else {
+    ReplicaRoot = CommR[l]
+  }
+  
+  for c in range LAYERS_CHALLENGES[l] {
+    // 2: Inclusion Proofs Checks
+    // 2.1: Check inclusion proofs for data nodes are correct
+    assert(MerkleTreeVerify(DataRoot, InclusionPath[l][c], DataProof[l][c], DataValue[l][c]))
```
schomatis: let's define `MerkleTreeVerify` or link it to the high level check it's in charge of
(edit: I realize now there's the number linking it, so this is fine as is)

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+    // 3: Encoding checks - Check that replica nodes have been correctly encoded
+    let ParentBits [PARENT_NODES][255]Fr
+    for p in range PARENT_NODES {
+      // 3.1: Check that each ParentValue is equal to its bit representation
+      let parent = ParentValue[l][c][p]
+      ParentBits[p] = Fr_to_bits(parent)
+      assert(Packed(ParentBits[p]) == parent)
+    }
+
+		// 3.2: Check that each key has generated correctly
```
schomatis: indent

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+    assert(PedersenHash(PreImage) == key)
+    // 3.3: Check that the data has been encoded to a replica with the right key
+    assert(ReplicaValue[l][c] == DataValue[l][c] + k)
+
+    // 4: CommRStar check - Check that the CommRStar constructed correctly
+    let hash = ReplicaId
+    for l in range LAYERS-1 {
+      hash.Append(CommR[l])
+    }
+    hash.Append(CommRLast)
```
schomatis: same as before, if possible put everything in the `for`

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+
+		// 3.2: Check that each key has generated correctly
+		let PreImage = ReplicaIdBits
+		for parentbits in ParentBits[l][c] {
+ 	   PreImage.Append(parentbits)
+		}
+    let key Fr = PedersenHash(PreImage)
+    assert(PedersenHash(PreImage) == key)
+    // 3.3: Check that the data has been encoded to a replica with the right key
+    assert(ReplicaValue[l][c] == DataValue[l][c] + k)
```
schomatis: is this a concatenation?

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+}
+```
+
+
+
+#### Verification of offline porep proof
+
+- SNARK proof check: **Check** that given the SNARK proof and the public inputs, the SNARK verification outputs true
+- Parent checks: For each `node = InclusionPaths_{l}_{c}`:
+  - **Check** that all `ParentsInclusionPaths_{l}_{c}_{0..PARENT_NODES}` are the correct parent nodes of `node` in the DRG graph.
```
schomatis: could we put this vector/matrix  also in the new format?

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
+# ZigZag: Offline PoRep Circuit Spec
+
+ZigZag is the Proof of Replication used in Filecoin. The prover encodes the original data into a replica and commits to it. An offline PoRep proves that the commitment to the replica is a valid commitment of the encoded original data.
+
+ZigZag has been presented by [Ben Fisch at EUROCRYPT19](https://eprint.iacr.org/2018/702.pdf).
+
+#### ZigZag Overview
+
+ZigZag PoRep is based on layering DRG graphs `l` times. The data represented in each DRG layer is the data encoded in the previous layer. The final layer is the replica (which in Filecoin terms is the sealed sector).
```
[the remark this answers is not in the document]
nicola: fixing

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+
+#### Public Parameters
+
+*Parameters that are embeded in the circuits or used to generate the circuit*
+
+- `LAYERS : UInt`: Number of DRG layers.
+- `LAYER_CHALLENGES : [LAYERS]UInt`: Number of challenges per layer.
+- `EXPANSION_DEGREE: UInt`: Degree of each bipartite expander graph to extend dependencies between layers.
+- `BASE_DEGREE: UInt`: Degree of each Depth Robust Graph.
+- `TREE_DEPTH: UInt`: Depth of the Merkle tree. Note, this is (log_2(Size of original data in bytes)).
```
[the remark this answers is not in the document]
nicola: fixing

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+
+- `CommRStar : Fr`: A commitment to each `l` layer's Merkle tree root hash `CommR[l]` and `ReplicaId`.
+
+- `InclusionPath : [LAYERS][]Fr`: Inclusion path for the challenged data and replica node.
+
+  Note: Each layer `l` has `LAYER_CHALLENGES[l]` inclusion paths.
+
+- `ParentInclusionPath : [LAYERS][][PARENT_NODES]Fr`:  Inclusion path for the parent nodes of the corresponding `InclusionPath[l][c]` nodes.
+
+  Note: Each layer `l` has `LAYER_CHALLENGES[l]` inclusion paths.
```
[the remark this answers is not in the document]
nicola: fixing

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+
+  Note: Size is `LAYERS-1` since the commitment to the last layer is `CommRLast`
+
+- `DataProof : [LAYERS][][TREE_DEPTH-2]Fr`: Merkle tree inclusion proof for the current layer unencoded challenged nodes.
+
+  Note: Size of proof per layer is `TREE_DEPTH-2` because it excludes (1) the roothash, and (2) the leaf value: (1) the root hash of the proof will be the corresponding unencoded data commitment for the current layer `l` (either `CommD` or `CommR[l-1]`), (2) the leaf value is `DataValue[l][c]`
+
+- `ReplicaProof : [LAYERS][][TREE_DEPTH-2]Fr`: Merkle tree inclusion proof for the current layer encoded challenged nodes.
+
+  Note: Size of proof per layer is `TREE_DEPTH-2` because it excludes (1) the roothash, and (2) the leaf value: (1) the root hash of the proof will be the corresponding unencoded data commitment for the current layer  (either `CommR[l]` or `CommRLast`), (2) the leaf value is `ReplicaValue[l][c]`
```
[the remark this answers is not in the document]
nicola: fixing

zigzag-circuit.md
nicola: fixing

zigzag-circuit.md
nicola: fixing

zigzag-circuit.md
nicola: it is the sum of two Fr elements, it should return an Fr element, changed the `k` with `key`

zigzag-circuit.md
nicola: fixing

zigzag-circuit.md
nicola: will leave it like this for now

zigzag-circuit.md
nicola: will leave it like this for now

zigzag-circuit.md
```diff
@@ -0,0 +1,179 @@
...
+
+- `CommRLast` is a public input, since we will be using it during Proof-of-Spacetime
+
+#### Private Inputs
+
+*Inputs that the prover uses to generate a SNARK proof, these are not needed by the verifier to verify the proof*
+
+- `CommR : [LAYERS-1]Fr`: Commitment of the the encoded data at each layer. 
+
+  Note: Size is `LAYERS-1` since the commitment to the last layer is `CommRLast`
```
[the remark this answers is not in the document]
nicola: the issue is that CommRLast and CommD must be in public inputs, while all the other ones must be private.

zigzag-circuit.md
```diff
@@ -0,0 +1,168 @@
...
+- `CommD : Fr`: the Merkle tree root hash of the original data (input to the first layer).
+- `CommRLast : Fr`: The Merkle tree root hash of the final replica (output of the last layer).
+- `CommRStar : Fr`: A commitment to each `l` layer's Merkle tree root hash `CommR[l]` and `ReplicaId`.
+- `InclusionPath : [LAYERS][]Fr`: Inclusion path for the challenged data and replica node.
+- `ParentInclusionPath : [LAYERS][][PARENT_NODES]Fr`:  Inclusion path for the parent nodes of the corresponding `InclusionPath[l][c]` nodes.
+
+##### Design notes
+
+- `CommRLast` is a public input, since we will be using it during Proof-of-Spacetime.
+- `InclusionPath` and `ParentInclusionPath`: Each layer `l` has `LAYER_CHALLENGES[l]` inclusion paths.
```
porcuquine: It's a bit confusing that we explain the number of challenges per layer here, but don't mention it below. I understand why you didn't, but I think it would be worth either a slight change to notation or else an annotation in the descriptions wherever we have `[LAYERS][]…`.

zigzag-circuit.md
```diff
@@ -0,0 +1,168 @@
...
+      PreImage.Append(parentbits)
+    }
+    let key Fr = PedersenHash(PreImage)
+    assert(PedersenHash(PreImage) == key)
+    // 3.3: Check that the data has been encoded to a replica with the right key
+    assert(ReplicaValue[l][c] == DataValue[l][c] + key)
+
+    // 4: CommRStar check - Check that the CommRStar constructed correctly
+    let hash = ReplicaId
+    for l in range LAYERS-1 {
```
porcuquine: This loop shouldn't be inside the enclosing loop as it is. You could move the whole inner loop out, or accumulate the hash without another loop.

zigzag-circuit.md
```diff
@@ -0,0 +1,168 @@
...
+      hash.Append(CommR[l])
+    }
+    hash.Append(CommRLast)
+
+    assert(CommRStar == PedersenHash(hash))
+    // TODO check if we need to do packing/unpacking
+  }
+}
```
porcuquine: I did not see the kdf or encoding checks above.

zigzag-circuit.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+
+
+#### Verification of offline porep proof
+
+- SNARK proof check: **Check** that given the SNARK proof and the public inputs, the SNARK verification outputs true
+- Parent checks: For each `node = InclusionPath[l][c]`:
+  - **Check** that all `ParentsInclusionPaths_[l][c][0..PARENT_NODES}` are the correct parent nodes of `node` in the DRG graph.
+  - **Check** that the parent nodes are in numerical order.
```
porcuquine: We also need to check for padding, since we always need the same number of parents.

zigzag-circuit.md
```diff
@@ -0,0 +1,168 @@
...
+#### ZigZag Overview
+
+ZigZag PoRep is based on layering DRG graphs `LAYERS` times. The data represented in each DRG layer is the data encoded in the previous layer. The final layer is the replica (which in Filecoin terms is the sealed sector).
+
+- `ReplicaId` is a unique replica identifier (see the Filecoin Proofs spec for details)
+- `CommD` is the Merkle Tree root hash of the input data to the first layer
+- `CommR[l]` is the Merkle Tree hash of the output of the DRG encoding at each layer `l` 
+- `CommRStar` is the hash of the concatenation of the `ReplicaId` and all the `CommR`s.
+
+The (offline) proof size in the ZigZag is too large for blockchain usage (~3MB). We use SNARKs to generate a proof of knowledge of a correct ZigZag proof. In other words, we implement the ZigZag proof verification algorithm in an arithmetic circuit and use SNARKs to prove that it was evaluated correctly.
```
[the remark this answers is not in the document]
nicola: Current calc: (8000*(15*31*32 + 15*32))/(1024*1024)

zigzag-circuit.md
nicola: good!

Issue #123, opened 2019-01-29
(Draft Spec) Signing & Signature Aggregation


bvohaska:
This issue serves to provide:
---
- [x] A plan for adding a signature aggregation to Filecoin
- [x] An understanding of what specs will be affected by the addition of signature aggregation
- [x] A high-level description on [why](https://github.com/filecoin-project/research/issues/1) signature aggregation will be included in Filecoin
- [x] A high-level description of the signature aggregation technique (BLS Signature Aggregation)

The output of this effort will be,

- [x] A [PR](https://github.com/filecoin-project/specs/pull/210) that amends the signature spec to include signature aggregation and BLS signature support
- [x] Enumerate all specs that need updating in order to add support for signature aggregation and BLS signatures
- [x] A PR for that amends all specs that require updating to support signature aggregation and BLS signatures

**Updated list of specs that require updating**

- [x] [signatures.md](https://github.com/filecoin-project/specs/blob/master/signatures.md) (`Signature`, `Verify`, `Sign`, `SignedMessage`, `Aggreagate`)
- [ ] [actors.md](https://github.com/filecoin-project/specs/blob/master/actors.md) (`StorageMarketActor`, `CreateStorageMiner`)
- [ ] [mining.md](https://github.com/filecoin-project/specs/blob/master/mining.md) (`VerifyBlock`, `VerifyBlock.ValidateSignature`, `VerifyTickets.ValidateSignature`, `VerifyTickets..getPublicKeyForMiner`, `Block Creation`)
- [ ] [expected-consensus.md](https://github.com/filecoin-project/specs/blob/master/expected-consensus.md) (`tickets`)
- [ ] [data-structures.md](https://github.com/filecoin-project/specs/blob/master/data-structures.md) (`Block`, `Tickets.Signature`, `SignedMessage`)

**Possible dependencies for the go-filecoin team**

- [x] In [actors.md](https://github.com/filecoin-project/specs/blob/master/actors.md), `SlashConsensusFault.ValidateSignature(blockX.Signature)` may need updating in order to more clearly add support for this Signature Aggregation effort.

**Support Links**

- [go-filecoin BLS Signature Effort](https://github.com/filecoin-project/go-filecoin/issues/1606)

bvohaska:
**Short primer on signature aggregation in Filecoin: [link](https://github.com/filecoin-project/research/issues/1)**

[...]

PR #210, opened 2019-03-22
Signature Aggregation

bvohaska:
Initial changes to Signature spec
- Updated Sign, Verify, Recover
- Added aggregate
- Added a section for Recovery
- Added a subsection for BLS signature public key recovery


signatures.md
```diff
@@ -10,40 +10,39 @@ We use signatures in filecoin to verify *something* was done by *someone*. For e
...
 What uses them
 
-- Messages [TODO: Link to messages spec]
+- [Messages](https://github.com/filecoin-project/specs/blob/master/data-structures.md)
 - Block validation
-- Tickets which inform leader election in Expected Consensus (EC) [TODO: link to EC spec]
+- Tickets [TODO: link to EC spec; inform leader election in Expected Consensus (EC)]
 
-Note that messages between actors are not signed this is because messages between actors are always spawned by a message from a user -- which are singed by that user. 
+Note that messages between actors are not signed. This is an unfortunate namespace collision. Messages between actors are always spawned by a message from a user. In this way, messages between actors are containers for signed messages.
```
whyrusleeping: I think this is reversed: "In this way, messages between actors are containers for signed messages.". Signed messages contain a 'cross actor' message. The signature is the way of getting it into the system in the first place, from that point its just a message from the account actor to another actor

signatures.md
```diff
@@ -52,59 +51,89 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
         // In:
-        //    m - a series of bytes representing the signed message
-        //    pk - the public key belonging to the signer `M`
-        //    sig - a series of bytes representing a signature usually `r`|`s`
+        //   msgs - an array of bytes representing signed messages
+        //   addrs - an array of Address {address_sig_1, address_sig_2..., address_sig_n}
+        //   sigs - an array of SignatureBytes {sig_1, sig_2..., sig_n}
         //
-        Verify(m Messgage, pk PublicKey, sig SignatureBytes) (valid bool, err error)
+        Verify(msgs Messgages, addrs Addresses, sigs SignatureBytes) (valid bool, err error)
```
whyrusleeping: use arrays to denote multiple things, so `msgs []Message`

signatures.md
bvohaska: What do you suggest for better wording?

signatures.md
bvohaska: Was trying to avoid go-lang syntax but that's probably silly at this point. Will change.

signatures.md
whyrusleeping: “In this way, signed messages are containers for messages between actors”

whyrusleeping:
@bvohaska what sort of review are you looking for right now?

bvohaska:
>@bvohaska what sort of review are you looking for right now?

@whyrusleeping: I’m looking for any feedback to make sure this spec is heading in the right direction. No deep review needed yet. Should have most or all of this spec ready for deep review by 28 March.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

bvohaska:
@whyrusleeping added a _very_ rough description for BLS public key recovery based on [this](https://github.com/filecoin-project/specs/issues/123#issuecomment-477255245). Please update with corrections.

This update should represent the last update in the draft process. All reviewers commence deep review.

signatures.md
```diff
@@ -52,163 +49,359 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
         // In:
-        //    m - a series of bytes representing the signed message
-        //    pk - the public key belonging to the signer `M`
-        //    sig - a series of bytes representing a signature usually `r`|`s`
+        //   smsgs - a SignedMessage which contains {[]Message, []Address, SignatureBytes}
+        //   pks - an array of Public Keys of the appropriate type associated with a message
+  			// 	 at index `i` in []Message.
         //
-        Verify(m Messgage, pk PublicKey, sig SignatureBytes) (valid bool, err error)
+  			VerifyAggregate(smsgs SignedMessgage,  pks []PublicKeys) (valid bool, err error)
```
whyrusleeping: are there multiple signed messages here? also does the signature get passed in? or how does this work?

signatures.md
```diff
@@ -52,163 +49,359 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
-        //    m - a series of bytes representing a message to be signed
-        //    sk - a private key which cryptographically links `M` to `sig`
+        //   msg - a series of bytes representing a message to be signed
+  		  //   addr - the Filecoin `Address` of the signer
+        //   sk - a private key which cryptographically links `M` to `sig`
         //
-        Sign(m Message, sk PrivateKey) (sig SignatureBytes, err error)
-
-        // Verify validates the statement: only `M` could have generated `sig`
+        Sign(msg Message, addr Address, sk PrivateKey) (sig SignatureBytes, err error)
```
whyrusleeping: this feels kinda weird. the sign method is on the signature itself. So i need to have a signature to sign something? seems like the sign method should be on a private key, or some other 'signaturescheme' or 'privkeymanager' type thing (those names are probably bad, but i'm sure theres something better here)

signatures.md
```diff
@@ -52,163 +49,359 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
 ```
 
+### Signed Message
 
-# Selected Signature Scheme
+This is the container for a signature that combines a message with an address and signature. 
 
-Currently, Filecoin uses secp256k1 signatures to fulfill the above interface. All signatures on messages, blocks, and tickets currently use the same scheme and format.
+```go
+Type SignedMessage Struct {
```
whyrusleeping: where does this type actually get used?  This is not the signed message that gets sent over the network for message propogation, its something else, right?

signatures.md
```diff
@@ -52,163 +49,359 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
+[go-filecoin SignedMessage](https://github.com/filecoin-project/go-filecoin/blob/master/types/signed_message.go)
 
-  - - Easy to check add; use (mod2) or check last bit
+[go-filecoin signer](https://github.com/filecoin-project/go-filecoin/blob/master/types/signer.go)
 
-  - Sig = (r,s) - s is standard but r could be (rx,ry) or rx or r all of these can be used to generate r*G
+## References & Notes
 
-  - Recovery() can be used for (1) finding r*G from (r,s) and (2) pk = sk*G from SignatureBytes
+Signatures are not sent over the wire on their own, instead they are encapsulated in in a [SignedMessage](https://github.com/filecoin-project/go-filecoin/blob/master/types/signed_message.go#L22-L28). Each `SignesMessage` contains a `Message`, `Address`, and a signature. `SignedMessage` is sent over the wire by marshaling it to cbor and send, once it is received it is unmarshaled form [cbor](http://cbor.io/) back into the `SignedMessage` Structure.
```
whyrusleeping: typo `SignesMessage`

signatures.md
```diff
@@ -52,163 +49,359 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
+[Code for marshaling and unmarshaling SignedMessages](https://github.com/filecoin-project/go-filecoin/blob/master/types/signed_message.go#L22-L38)
```
whyrusleeping: let's drop any references to go-filecoin code, in favor of pointing to things in the spec.

whyrusleeping:
good detail on the base cryptographic primitives (are there RFCs for some of that? seems like stuff that should go in the standard)

A bit more still needed before completion:
- [ ] Spec for how message signatures are aggregated by the miner when putting together the block
- [ ] spec for how a block with aggregated signatures gets verified
  - [ ] Where do the public keys come from here? (they come from the chain somehow, but that needs to be detailed)
  - [ ] The account actor needs to be upgraded to hold public keys for bls based accounts
- [ ] What does the array of signed messages in the block look like now that some of the messages have had their signatures aggregated?
- [ ] How exactly do we disambiguate signatures?
- [ ] add the signature type to the data-structures doc, along with pointers to how things are serialized.

probably more questions, but once the above is done, let's get dev to review. A sign-off from them is what we're after.

signatures.md
bvohaska: Updating name "smgs" --> "smsg" to reduce confusion. Any `SignedMessage` containing a signature aggregate will contain the signature aggregate as well as all messages contained therein. So we only need the one `SignedMessage` as input.

Setting aside golang syntax, suppose there is `X = SignedMessage`. To verify a signature aggregate, VerifyAggregate will perform something like: 

- For all messages `m_i` in `X.Msgs`, `h_i = hash2curve(m_i)`
- Compute `S = deserialize(X.SignatureBytes)`
- Compute `T = e(h_1, pk_i)*...*e(h_n, pk_n)`
- If `T = e(S, P2)` => valid

signatures.md
bvohaska: Agree. Definitely feels weird. This should be captured in a `privateKey` method for BLS signatures. I don't believe we have a specification for private keys. Maybe this should be in data-structures.md? Context: If I recall correctly, I put this method here (last year) as a placeholder so we captured the`Sign` function somewhere.

Alternatively, we could place the `privateKey` specification here instead of data-structures.md. Open to suggestions.

(FYI: @dignifiedquire)

signatures.md
bvohaska: Will update.

signatures.md
bvohaska: removed

signatures.md
bvohaska: @frrist: As one of the core devs, do you have a strong preference on where this new method should be documented?

frrist: I think documenting this in the spec is fine for now. Once we implement it the documentation can be added to the implementation as well.

signatures.md
```diff
@@ -10,40 +10,37 @@ We use signatures in filecoin to verify *something* was done by *someone*. For e
...
-- Block validation
-- Tickets which inform leader election in Expected Consensus (EC) [TODO: link to EC spec]
+Note that messages between actors are not signed. This is an unfortunate namespace collision. Messages between actors are always spawned by a message from a user. In this way, signed messages are containers for messages between actors.
 
-Note that messages between actors are not signed this is because messages between actors are always spawned by a message from a user -- which are singed by that user. 
+## Dependencies
 
-Filecoin specific reliance
+- Elliptic curve implementations
+  - ECDSA with secp256k1 (BitCoin-style elliptic curve)
```
dignifiedquire: Would be good to have some reference link for this.

signatures.md
```diff
@@ -10,40 +10,37 @@ We use signatures in filecoin to verify *something* was done by *someone*. For e
...
+    - Described as *Variant 1* in [IETF BLS Signature Scheme](https://tools.ietf.org/html/draft-boneh-bls-signature-00)
+    - Small signature size
+    - Larger public keys
+- Signature size limits
+  - ECDSA - 64 bytes + 1 byte for public key recovery
+  - BLS Signatures - 48 Bytes
+- Avg. processing power available to a CPU 
+  - Affects ability to verify signature aggregates
+  - Each signature or collection of signatures should be validated with no impact to the consumption of the chain
+- `Gas` cost - `Gas(ECDSA) > Gas(BLS)`
```
dignifiedquire: please clarify which operations are more costly than others, eg `Gas(ECDSA_Sign)`

signatures.md
```diff
@@ -52,163 +49,350 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
+        // Out:
+        // 	 sig - a series of bytes representing a single BLS signature
+        //   err - a standard error message indicating any process issues
+        // In:
+     		//   sigs - an array of SignatureBytes {sig_1, sig_2..., sig_n}
+    		//   addrs - an array of Addresses {address_sig_1, address_sig_2..., address_sig_n}
+  			//
+  			Aggregate(sigs []SignatureBytes, addrs []Address) (sig SignatureBytes, err error)
+  
+  			// Verify validates the statement: only `M` could have generated `sig`
```
dignifiedquire: whuiii, I think @whyrusleeping wrote a tool to format code blocks in markdown :)

signatures.md
```diff
@@ -52,163 +49,350 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
+```go
+Type SignedMessage Struct {
 
-Currently, Filecoin uses secp256k1 signatures to fulfill the above interface. All signatures on messages, blocks, and tickets currently use the same scheme and format.
+		// Addrs is an array of Addresses associated with a particular message in Msgs 
+  	// at the same index. Each Address tracks the ID of the actor responsible for 
+  	// signing a particular Msg. An ID can be converted into a public key via 
+  	// Address.GetPublicKey(ID).
+		//
+		Addrs []Address `JSON`
```
dignifiedquire: I think you can leave out the `JSON`

signatures.md
```diff
@@ -52,163 +49,350 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
+    Sig SignatureBytes `JSON` 
+    
+}
+```
 
-## Wire Format
+## Key Generation & Elliptic Curve Parameters
 
-What bits are on the wire and in what order/format. We are currently adhering to libsecp256k1 serialization which is laid out as follows. Note that this format description may not be accurate. See the github link below for an authoritative format description.
+This specification assumes the implementer uses a strong cryptographic pseudorandom generator `RANDOM`. The following was adapted from [IETF BLS Signature Scheme - Keygen](https://tools.ietf.org/html/draft-boneh-bls-signature-00#section-2.2).
```
dignifiedquire: do you have a good link for a definition of a strong cryptographic pseudorandom [number] generator?

signatures.md
```diff
@@ -52,163 +49,350 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
+   3.  Output PK, SK
 ```
-sig SignatureBytes = [0x30][len][0x02][r][indicator][s][indicator][recovery]
+
+For ECDSA, `P` is the generator for `E`. For BLS, we choose `P` to be a generator in `E2` which is the larger of the elliptic curve groups.
+
+### BLS12-381 Parameter Choices
+
+We have chosen to use the same elliptic curve parameters as [zkcrypto/pairing](https://github.com/zkcrypto/pairing/tree/master/src/bls12_381) with the exception that we are signing in G1 and generating public keys in G2. Note: there is a namespace collision between [zkcrypto/pairing](https://github.com/zkcrypto/pairing/tree/master/src/bls12_381) and [IETF BLS Signature Scheme](https://tools.ietf.org/html/draft-boneh-bls-signature-00) concerning G1 and G2. In zkcrypto/paring, G1 and G2 are points defined in the largest prime order subgroup of the elliptic curve defined by BLS12-381. In the IETF BLS Signature Scheme, G1 and G2 represent the largest prime order subgroups in BLS12-381. We will comply with the IETF standard. We will denote the generators of G1 and G2 as P1 and P2 respectively. E1 and E2 will denote the elliptic curve defined by BLS12-381.
```
dignifiedquire: > G1 and G2 are points defined in the largest prime order subgroup

This is not quite as I read this: https://github.com/zkcrypto/pairing/tree/master/src/bls12_381#generators
To quote
> The generators of G1 and G2 are computed by finding the lexicographically...

So the use `G1` and `G2` for both the subgroup, as well as the generators.

signatures.md
```diff
@@ -52,163 +49,350 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
+y = 1339506544944476473020471379941921221584933875938349620426543736416511423956333506472724655353366534992391756441569
+
+P2:
+x = 3059144344244213709971259814753781636986470325476647558659373206291635324768958432433509563104347017837885763365758*u + 352701069587466618187139116011060144890029952792775240219908644239793785735715026873347600343865175952761926303160
+y = 927553665492332455747201965776037880757740193453592970025027978793976877002675564980949289727957565575433344219582*u + 1985150602287291935568054521177171638300868978215655730859378665066344726373823718423869104263333984641494340347905
+
 ```
 
-`s` = Scalar of size 32 bytes
+[TODO: add notes on point compression and effect on serialization once a standard has been set]
```
dignifiedquire: Why not document the one from https://github.com/zkcrypto/pairing/tree/master/src/bls12_381#serialization for now, as that is what we are using?

signatures.md
```diff
@@ -52,163 +49,350 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
-- Should elliptic curve points always be compressed in Marshal?
+```go
+Input: Address.payload, ROOT
+Output: PK
+
+1. If Address.protocol = 3 output string_to_E2(Address.payload) and stop
+2. If Address.protocol != 0 output "INVALID" and stop
+3. Actor = ROOT.GetActor(ID)
+4. PK = Actor.GetPublicKey()
+4. Output PK
```
dignifiedquire: double `4.`

signatures.md
bvohaska: I don't have this information. Ping @whyrusleeping : can you answer this?

signatures.md
bvohaska: @dignifiedquire: You are correct. This is a language difference between the IETF standard and ZK-crypto. We are using the language from the IETF standard and not ZK-crypto. I'm going to leave the language as is.

signatures.md
bvohaska: Will update.

signatures.md
bvohaska: There is no standard way of doing this yet. I would prefer to include a standard once it is set instead of using the point compression in ZK-crypto and extra bits needed to indicate compression (3 bits for ZK-Crypto, I think?). If we just want to put something here that I can include that method for point compression and representation.

I have no particularly strong feelings on this as long as we properly indicate that we will adhere to the IETF standard (or companion) once it adds a method for point compression.

signatures.md
bvohaska: Will update.

signatures.md
bvohaska: Will add link to [NIST](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90Ar1.pdf). Incidentally, the authors Elaine Barker and John Kelsey are fairly responsive if we have further questions or need clarifications.

dignifiedquire: Maybe also https://en.wikipedia.org/wiki/Cryptographically_secure_pseudorandom_number_generator which is likely easier to understand from a high level persepective

signatures.md
dignifiedquire: Maybe just a link, to reference what we are doing in the meantime?

bvohaska: Will add link.

signatures.md
bvohaska: Will add link in References section

signatures.md
bvohaska: Will add links to standards

signatures.md
bvohaska: Added an interface for `PrivateKey`. This will require further work. Please review.

signatures.md
bvohaska: There should be only one `SignedMessage` --> https://github.com/filecoin-project/go-filecoin/blob/master/types/signed_message.go#L22-L28

If there are more than one, then we should do some disambiguation.

signatures.md
```diff
@@ -52,163 +49,350 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
-    
-}
+If there is more then one message in `SignedMessage.Msgs` then the `SignedMessage` is assumed to be a signature aggregate. If `Address` is empty then `Verify` will assume the `SignedMessage` is a signature aggregate. The procedure below uses `len()` to indicate a function that returns the number of elements in an array.
+
+```go
+Input: SignedMessage
+Output: "Aggregate" or "INVALID"
+	    
+	 1.  If SignedMessage.Address[i].protocol != {0 OR 3} for all i output "INVALID" and stop
+	 2.  If len(SignedMessage.Msgs) > 1 output "Aggregate" else output "INVALID"
```
frrist: Would the case `len(SignedMessage.Msgs) == 1` indicate the the `SignedMessage` contains a ECDSA signature?

signatures.md
frrist: Adding a `PrivateKey` interface sounds reasonable to me.

signatures.md
```diff
@@ -52,163 +49,350 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
+**Determining Signature Type**
 
-**Signed Message**
+`SignedMessage.Address` will inform `Verify` if the `SignedMessage.SignatureBytes` is a BLS or ECDSA signature. `Verify` will infer this data from the `Address.protocol` field. 
 
+```go
+Input: Address
+Output: "BLS", "ECDSA", or "INVALID"
+   
+   1.  If Address.protocol = 0 output "BLS" and stop
```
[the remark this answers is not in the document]
frrist: I see you have covered this case here: https://github.com/filecoin-project/specs/issues/123#issuecomment-478766240

> Does Address.protocol specify 0? If so, fetch actor with ID = Address.payload, return BLS or ECDSA. <--this method needs to be specified in actors.md and should distinguish between BLS and ECDSA signatures.

signatures.md
```diff
@@ -52,163 +49,350 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
+        //    sig - a series of bytes representing a signature
         //
-        Recover(m Message, sig SignatureBytes) (pk PublicKey, err error)
+        Recover(msg Message, addr Address, sig SignatureBytes) (pk PublicKey, err error)
 }
 ```
 
+### Signed Message
+
+This is the container for a signature that combines a message with an address and signature.
```
anorth: I'm confused by the plurality of addresses and messages in the structure given this description.

I understand the use case for multiple addresses, but what is the case for multiple different messages? This seems like redundant data in the case of multiple signers of an identical message.

A brief outline of use cases and capabilities at the top of this doc would help greatly if my understanding of the motivations and context is lacking.

signatures.md
```diff
@@ -52,163 +49,350 @@ Note: `Message` is used here as the object being signed, but this interface shou
...
-Type SignedMessage Struct {
 
-    Message bytes `JSON`   
+**Determining if the signature is a Signature Aggregate**
 
-    Signature bytes `JSON` 
-    
-}
+If there is more then one message in `SignedMessage.Msgs` then the `SignedMessage` is assumed to be a signature aggregate. If `Address` is empty then `Verify` will assume the `SignedMessage` is a signature aggregate. The procedure below uses `len()` to indicate a function that returns the number of elements in an array.
+
```
anorth: As a practical matter, I recommend an explicit representation of the "type" of signed message rather than assuming from the number of messages and/or addresses. An explicit byte identifying the semantics will simplify future changes and make code easier to understand and more likely to be correct.

signatures.md
anorth: What about the case of no addresses and one message? Is an aggregate of one explicitly invalid? While that's what the pseudocode implied, explicit clarification would help.

What about the case of multiple addresses and multiple messages? I'm guessing this means an aggregate (but should it check the lengths are equal)? An enumeration of all possible cases of {1, many}-messages ⨯ {0, many}-addresses would help (perhaps in the prose above).

signatures.md
bvohaska: Awesome tool. Thanks!

signatures.md
bvohaska: @frrist: yes. ECDSA signatures can't be aggregated in this scheme so if we encounter them we should error. Maybe "INVALID" is not the right wording. We should probably specify input validation. Alex alludes to this below. 

@anorth: Good point. Request someone from go-filecoin to work on this (input validation).

bvohaska: Added note about ECDSA not being aggregatable in the "Aggregating Signed Messages" section

signatures.md
bvohaska: I'm note sure I quite understand but I've reworked the description. Agree it is very confusing. Will push. Let me know if that answers your questions.

signatures.md
bvohaska: @whyrusleeping: ^ . It would make things easier to understand but would cost a bit or two on-chain. What do you think?

signatures.md
```diff
@@ -10,205 +10,403 @@ We use signatures in filecoin to verify *something* was done by *someone*. For e
...
+	//    **
+	// In:
+	//    msg - a series of bytes representing the signed message
+	//    sig - a series of bytes representing a signature
+	//
+	Recover(msg Message, addr Address, sig SignatureBytes) (pk PublicKey, err error)
+  
+  //
+  //
+  Marshall(smsg SignedMessage) (SmsgBytes []byte, err error)
```
bvohaska: @anorth: Request someone from go-filecoin help with `Marshall` and `Unmarshall`. I believe this should be CBOR encoding.

anorth: I would be happy, nay happier(!) if these interfaces were pseudocode rather than Go code. That is to say, I think it would be fine and desirable for the spec to leave language-specifics to implementation, and instead just describe the functionality generally required primarily in prose (the comments above) and illustrative pseudocode. Would that decoupling make it easier for you?

The spec should certainly define the encoding, but it's sufficient to say "CBOR".

I realise that this may be a controversial position. If rejected, you could leave placeholders for the Go code here but you shouldn't be blocked on the go-filecoin team. Neither should this pre-empt that team's priorities and focus.

bvohaska: I think if you're willing to commit people to working on this, I'm for it :-) 

If I recall, the reason we went with Go-style code blocks was b/c it was much faster for us to use at the time--since most of this work was first described in go-filecoin. I don't really have any objections to moving to a different structure as long as we realize that it'll be a lot of work for whoever has that undertaking.

anorth: To be fully transparent, I am not intending to commit any people to working on it until the go-filecoin team needs to implement it. We can then backport form implementation to spec. If this doesn't work for the specs team, I'd suggest pseudocode.

whyrusleeping: we use go-style pseudo code in the spec, it should not try to hard to be real working go code (i don’t want to see any if err blocks in the spec...). @bvohaska just do your best to write in pseudo code what should be done, ‘illustrative pseudocode’ as @anorth says.

mining.md
```diff
@@ -107,7 +107,7 @@ Chain selection is a crucial component of how the Filecoin blockchain works. Eve
 
 ### Block Validation
 
-The block structure and serialization is detailed in [the datastructures spec](data-structures.md#block). Check there for details on fields and types.
+The block structure and serialization is detailed in [the datastructures spec](data-structures.md#block). Check there for details on fields and types. All references to `VaidateSignature` must implement the `Signature` interface found [here](https://github.com/filecoin-project/specs/blob/master/signatures.md).
```
whyrusleeping: This addition is confusing. Why not just say that signature validation is covered in detail in the signature spec?

bvohaska: This is a good point. I've been mulling over how to best represent that `ValidateSignature` must use the verification method in signatures.md. The current phrasing is confusing b/c it should read: "`Block` implements the `Signature` interface" and `ValidateSignature` --> `Block.Verify(SignedMessage_Block, PublicKey)`. in this case, the internal block messages would be represented as a single value (hashed), `SignatureBytes` would be the actual signature of the message and `Address` would be the address of the block proposer.

bvohaska:
@anorth, @whyrusleeping: This PR is suffering a bit of scope creep. Let's close it out and push to master. I'd like to ship this so we can move a bit faster. We can create a new PR for other concerns captured in this PR. I believe most of the concerns are captured in [this](https://github.com/filecoin-project/specs/issues/123) GH issue.

whyrusleeping:
@bvohaska I don't think its scope creeping. The scope of signature aggregation should be every change needed to the protocol to support it. I don't want to merge a partial spec change which would leave the spec in a weird place here.

signatures.md
anorth: I still don't understand why there are multiple messages in the container. The comment uses the singular "a message and an address" which disagrees with the code directly below. Please clarify the intended role of a `SignedMessage`. E.g. you could write: 

```A container for one or multiple messages and an aggregation of the signatures of those messages by corresponding addresses. In the case of a single message and address, the signature is a vanilla signature, not an aggregate. In the case of multiple messages, the signature is assumed to be an aggregate. See *Determining signature type* below for more detail.```

But I don't know whether that is in fact correct or complete.

File: signatures.md (hunk 16 above)
Comment (anorth) [reply signatures.md]: I would love us to work on input validation, but we need you to specify what inputs should be valid. Please enumerate all combinations that are semantically valid and explicitly state that other combinations are meaningless.

**Comment by mhammersley** [comment]
suggestion: have a set of diagrams describing the constituent elements as well as the process flow with concise, precise definitions

File: data-structures.md [hunk 22]
```diff
@@ -79,18 +104,19 @@ Blocks are currently serialized simply by CBOR marshaling them, using lower-came
...
 	// the message must match the expected nonce in the from actor.
 	// This prevents replay attacks.
 	Nonce Integer
 
 	Value Integer
-    
-    GasPrice Integer
-    GasLimit Integer
+
+	GasPrice Integer
```
bvohaska: @whyrusleeping, @dignifiedquire: is there a canonical description for `Integer`? In other words, what exactly do we mean by integer here? Is this a 32 byte binary quantity? Perhaps it's a bigInt in form LEB128?

data-structures.md
```diff
@@ -29,22 +29,47 @@ For most objects referenced by Filecoin, a Content Identifier (CID for short) is
...
+- Ensure leader secrecy -- meaning a block producer will not be known until they release their block to the network.
+- Prove leader election -- meaning a block producer can be verified by any participant in the network.
+- Prove appropriate delay between drawings — thereby preventing leaders from "rushing" the protocol by releasing blocks early (at the expense of fairness for miners with worse connectivity).
+- Ensure a single drawing per round — derived in part from the above, thereby preventing miners from grinding on tickets (e.g. by repeatedly drawing new tickets in the hopes of winning) within a round.
+
+```go
+type Ticket struct {
+  
+	// The VDF Result is derived from the prior ticket in the ticket chain
+	VDFResult BigInteger
```
bvohaska: @sternhenri, @ZenGround0: Do we have a language-independent description of BigInteger here? What is the binary representation? Maybe we could point to a spec that defines this for us.

bvohaska:
Moved content of this comment to the tracking issue [here](https://github.com/filecoin-project/specs/issues/123)

data-structures.md
```diff
@@ -56,17 +81,25 @@ type Block struct {
...
+  BLSSignatureAggregate SignedMessage
+  
+  // BLSMessages is the set of messages with associated BLS signatures contained
+  // in BLSSignatureAggregate.
+  BLSMessages []Message
 
 	// MessageReceipts is a set of receipts matching to the sending of the `Messages`.
-    // TODO: should be the same type of merkletree-list thing that the messages are
+	// TODO: should be the same type of merkletree-list thing that the messages are
 	MessageReceipts []MessageReceipt
```
bvohaska: @whyrusleeping: what are `MessageReceipts`?

data-structures.md
dignifiedquire: That is what LEB128 is for

data-structures.md
dignifiedquire: They are defined below and store the result of applying the message

data-structures.md
```diff
@@ -29,22 +29,47 @@ For most objects referenced by Filecoin, a Content Identifier (CID for short) is
...
 	// Tickets are the winning ticket that were submitted with this block.
 	Tickets []Ticket
-    
-    // ElectionProof is a signature over the final ticket that proves this miner
-    // is the leader at this round
-    ElectionProof Signature
+
+	// ElectionProof is a signature over the final ticket that proves this miner
+	// is the leader at this round
+	ElectionProof SignatureBytes
```
dignifiedquire: Is it defined somewhere what kind of signature this is cc @sternhenri

mining.md
dignifiedquire: we also need a place that implements verification of the aggregated signatures, e.g. a method like `VerfiyAggregates([]Message, SignatureBytes) bool` for use inside the block validation.

data-structures.md
```diff
@@ -29,22 +29,47 @@ For most objects referenced by Filecoin, a Content Identifier (CID for short) is
...
 
 ```go
 // Block is a block in the blockchain.
 type Block struct {
+  
 	// Miner is the address of the miner actor that mined this block.
 	Miner Address
 
 	// Tickets are the winning ticket that were submitted with this block.
 	Tickets []Ticket
```
dignifiedquire: Could we aggregate the signatures on the tickets as well?

data-structures.md
```diff
@@ -56,17 +81,25 @@ type Block struct {
...
-	Messages []SignedMessage
+	ECDSAMessages []SignedMessage
+  
+  // BLSSignatureAggregate is the SignedMessage contianing the aggreate signature of 
+  // all messages signed with the BLS signature scheme.
+  BLSSignatureAggregate SignedMessage
+  
+  // BLSMessages is the set of messages with associated BLS signatures contained
+  // in BLSSignatureAggregate.
+  BLSMessages []Message
```
dignifiedquire: In general I like this structure, but we are now unable to present an ordering of messages in the form of 

```
- bls message
- ecdsa message
- bls message
```
In regards to execution order. 

I would argue this is fine, and miners are simply asked to process these not together, but first one kind, then the other kind. But this is an explicit design decision that might have other implications.

data-structures.md
bvohaska: Do we have a size for these bigInts?

data-structures.md
bvohaska: Do we know the size of these Integers?

data-structures.md
bvohaska: I see the `struct` definition. Do we have any exposition as to why we need them? Given that they are an on-chain quantity, I would claim that it's important to declare exactly why we have them. This would be the best place to put that explanation.

data-structures.md
bvohaska: ^ I took a guess at this. Will need @sternhenri to verify

data-structures.md
bvohaska: If the EC spec can be updated to support it, I don't see why not. We'll need to make sure that the EC spec is using `SignedMessages` and the `Signature` interface.

^ thoughts @sternhenri?

bvohaska:
@whyrusleeping: from mining.md (https://github.com/filecoin-project/specs/blob/master/mining.md#block-validation), it doesn't look like on-chain message signatures are validated. Maybe I'm missing something?

data-structures.md
dignifiedquire: They are variably length sized, to support arbitrary numbers. But if you have a max range you want to represent then I can tell you the size.

bvohaska: Understood. Someone will have to update the spec comments to expound on this. As for the typing, we are using inconsistent type nomenclature for big integers. It would be nice to go through and standardize these. So far I've seen `Integer`, `BigInteger`, `BigInt`; they no doubt mean the same thing --> LEB128 encoded integers of arbitrary size but we should specifically call that out here. 

If the spec owners acknowledge that this is what they mean when they write an element from the list above, I'll go ahead and change it. Don't want to misinterpret the authors intent.

sternhenri:
outdated

PR #211, opened 2019-03-22
add some notes on how the chaintipsmanager works

whyrusleeping:


whyrusleeping:
merging this for now, will get more details later (a correct implementation can be inferred here, so its less necessary to add details now)

PR #212, opened 2019-03-25
Update based on actual implementation

steven004:


whyrusleeping:
Hey @steven004 thanks for bringing this up. In this case, the go-filecoin implementation is out of date and needs to be updated. There are many such places where this is the case, please do bring them up as they are found

PR #213, opened 2019-03-28
fix problems in apply message

whyrusleeping:
Not sure how I managed to sneak these through review... Let's hope its right this time around.

### DIFF
--- a/state-machine.md
+++ b/state-machine.md
@@ -44,8 +44,11 @@ If the execution completes successfully, changes to the state tree are saved. Ot
 ```go
 func ApplyMessage(st StateTree, msg Message) MessageReceipt {
 	st.Snapshot()
-	fromActor := st.GetActor(msg.From)
-
+	fromActor, found := st.GetActor(msg.From)
+  if !found {
+    Fatal("no such from actor")
+  }
+  
 	totalCost := msg.Value + (msg.GasLimit * msg.GasPrice)
 	if fromActor.Balance < totalCost {
 		Fatal("not enough funds")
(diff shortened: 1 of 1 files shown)


state-machine.md
```diff
@@ -55,30 +58,50 @@ func ApplyMessage(st StateTree, msg Message) MessageReceipt {
...
-	if errcode != 0 {
-		// revert all state changes since snapshot
-		st.Revert()
-		st.DeductFunds(msg.From, vmctx.GasUsed() * msg.GasPrice)
-	} else {
-		// refund unused gas
-		st.DepositFunds(msg.From, (msg.GasLimit - vmctx.GasUsed()) * msg.GasPrice)
-	}
+  if msg.Method != "" {
+		ret, errcode := toActor.Invoke(vmctx, msg.Method, msg.Params)
```
whyrusleeping: this s/from/to/ is one of the main fixes

state-machine.md
```diff
@@ -55,30 +58,50 @@ func ApplyMessage(st StateTree, msg Message) MessageReceipt {
...
+			st.DeductFunds(msg.From, vmctx.GasUsed() * msg.GasPrice)
+		} else {
+			// refund unused gas
+			st.DepositFunds(msg.From, (msg.GasLimit - vmctx.GasUsed()) * msg.GasPrice)
+		}
+  }
 
 	// reward miner gas fees
-	st.DepositFunds(msg.To, msg.GasPrice * vmctx.GasUsed())
+	st.DepositFunds(BlockMiner, msg.GasPrice * vmctx.GasUsed())
```
whyrusleeping: other big fix was giving the gas reward to the miner, not the actor youre sending the message to

whyrusleeping:
@bvohaska this is how actors will get created with BLS public keys. The details are left out here, hoping you can fill that in as part of the signature aggregation work

state-machine.md
```diff
@@ -44,8 +44,11 @@ If the execution completes successfully, changes to the state tree are saved. Ot
 ```go
 func ApplyMessage(st StateTree, msg Message) MessageReceipt {
 	st.Snapshot()
-	fromActor := st.GetActor(msg.From)
-
+	fromActor, found := st.GetActor(msg.From)
+  if !found {
```
acruikshank: This indentation seems off here and elsewhere.

state-machine.md
```diff
@@ -55,30 +58,50 @@ func ApplyMessage(st StateTree, msg Message) MessageReceipt {
...
+	st.DepositFunds(BlockMiner, msg.GasPrice * vmctx.GasUsed())
 
 	return MessageReceipt{
 		ExitCode: errcode,
 		Return: ret,
 		GasUsed: vmctx.GasUsed(),
 	}
 }
+
+func TryCreateAccountActor(st StateTree, addr Address) Actor {
```
acruikshank: For clarity's sake, it might be best to leave this definition out of this spec (in the way the methods on the state tre are undefined). Maybe it it's worth mentioning in the address section.

state-machine.md
acruikshank: There is a lot of message validation that needs to happen on the message. The existence of the from actor is just one of those. This is probably fine though.

whyrusleeping: mind taking a stab at adding more of that validation? Its worth having here. Just submit another PR whenever

whyrusleeping:
Yeah, on the indentation, i just need to find some tool that does code formatting inside markdown docs. Typora definitely does not do it well.

PR #217, opened 2019-04-02
Chain Weighting clarification 

steven004:
Based on the latest implementation of expected.go


whyrusleeping:
@steven004 as I said in the last PR you submitted with these changes:

In this case, the go-filecoin implementation is out of date and needs to be updated. There are many such places where this is the case, please do bring them up as they are found

steven004:
OK, got you. I'm closing this.

PR #219, opened 2019-04-03
Moving drg porep from the main spec

nicola:
This is not needed anymore, it's all written in the zigzag spec


pooja:
@nicola This looks like it's good to go. Please merge at will!

PR #221, opened 2019-04-04
(proposal) actor updates to support multiple sector sizes

laser:
## What's in this PR?

This PR proposes updates to the `StorageMarketActor` and `StorageMinerActor`s to support multiple sector sizes.

## Summary of Changes

1. `CreateStorageMiner` modified to accept a sector size. Sector size cannot be changed after miner actor has been created. Rationale: One sector size per storage miner implies one proving period per miner.
1. `StorageMarketActor.TotalStorage` tracked as `BytesAmount` instead of `Integer` (`UpdateStorage` and `GetTotalStorage` have been updated, too). Rationale: Miners will commit sectors of different sizes. A miner who commits a single, 10GB sector to the network must have the same power as a miner who has committed ten, 1GB sectors.
1. `CollateralForSector` now accepts a sector size.
1.  `ProvingPeriodDuration` and `GenerationAttackTime` now accept a sector size. Rationale: Small sectors take less time to prove than large sectors. To prevent generation attacks, the proving period for small sectors must be smaller than the proving period for large sectors.
1. `CheckPostProofs` and `ValidatePoRep` now accept a sector size. Rationale: Choice of verifying key will depend on sector size.

## Additional Talking Points

1. This proposal does add sector size to the ask. If a storage client wishes to make a deal with a miner offering sectors of a particular size, they can query the owner of the ask to obtain the miner actor's sector size.
1. This proposal represents `StorageMarketActor.TotalStorage` as `BytesAmount`. Power in the network should probably be tracked in some other unit, e.g. multiple-of-smallest-sector-available-to-network.

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -160,7 +160,7 @@ func CreateStorageMiner(pubkey PublicKey, pledge, sectorSize BytesAmount, pid Pe
 		Fatal("not enough funds to cover required collateral")
 	}
 
-	newminer := InitActor.Exec(MinerActorCodeCid, EncodeParams(pubkey, pledge, pid, sectorSize))
+	newminer := InitActor.Exec(MinerActorCodeCid, EncodeParams(pubkey, pledge, sectorSize, pid))
 
 	self.Miners.Add(newminer)
 
--- a/actors.md
+++ b/actors.md
@@ -405,9 +405,7 @@ func CommitSector(comm Commitment, proof *SealProof) SectorID {
 	}
 
 	// make sure the miner has enough collateral to add more storage
-	// currently, all sectors are the same size, and require the same collateral
-	// in the future, we may have differently sized sectors and need special handling
-	coll = CollateralForSector()
+	coll = CollateralForSector(miner.SectorSize)
 
 	if coll < miner.Collateral {
 		Fatal("not enough collateral")
--- a/actors.md
+++ b/actors.md
@@ -487,14 +487,14 @@ func SubmitPost(proofs []PoStProof, faults []FaultSet, recovered BitField, done 
 	permLostSet = AggregateBitfields(faults).Subtract(recovered)
 
 	// adjust collateral for 'done' sectors
-	miner.ActiveCollateral -= CollateralForSectors(miner.NextDoneSet)
-	miner.Collateral += CollateralForSectors(miner.NextDoneSet)
+	miner.ActiveCollateral -= CollateralForSectors(miner.SectorSize, miner.NextDoneSet)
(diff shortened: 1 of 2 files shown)


whyrusleeping:
> to prevent generation attacks, the proving period for small sectors must be smaller than the proving period for large sectors

Hrm... thats technically true, but tricky. cc @nicola

porcuquine:
When I talked to @nicola today, he seemed to agree that timing differences contribute to why mixing sector sizes in a single PoSt is a non-trivial problem. i.e. not *necessarily* impossible, but not something we should commit to doing without expecting significant design (which might or might not yield fruit). I'll let him elaborate in case I got that wrong, but we've discussed this before so I think we're in broad agreement.

nicola:
**Sealing & PoSt relation**
The time to encode a sectors determines the polling time.

The polling time is the time between two online proofs of replication (which directly maps to what we should tune the VDF to).

The smaller the sector, the smaller the polling time.

**Does this impact the proving period?**
In theory, since the proving period is just determined by how many times we are willing to repeat VDFs, then this would not impact proving period.

In practice, a smaller polling time implies more proofs, which implies larger snark circuits, which unless we have SNARK batching or merkle tree replacement, then this implies larger proofs.

**So, should the proving period be smaller for smaller sectors?**

Yes:
- unless, we are fine to have larger proofs
- unless, we have better SNARKs (either batching of merkle trees or SNARKs)

**Can I do a proof of spacetime on sectors of different sizes?**
In theory, yes. For example, the simplest way to do this is to assume the smallest polling time for all sectors - in other words the strictest security for all sectors 

In practice, it is tricky to reason this through. We did not analyze this mixed case yet and I would be worried on going with this type of solution until we are confident 

Given the other priorities, it's in my opinion more costly to think about possible edge cases, then having multiple miners running.

laser:
Thanks, @nicola and @porcuquine . 

In summary:

- `ProvingPeriodDuration` must accept a sector size as input
- `GenerationAttackTime` must accept a sector size as input
- Further work is needed to support generating a PoSt over sectors of differing sizes. For now (for purposes of this PR), a PoSt must be generated over sectors which are of the same size.

@whyrusleeping - Let me know if you have any further questions/objections/whatever about the content of this PR.

nicola:
I have update my reply, where:

> `ProvingPeriodDuration` must accept a sector size as input

This only true if we choose solution 3, with solution 1 and 2, it will be fine.
The most natural solution is to choose solution 1 since it will require minimal change to proofs, and go-filecoin.

whyrusleeping:
Alright, looks like the consensus is that miners get to pick a sector size per actor up front. Lets ship it and get on with it!

PR #232, opened 2019-04-09
Rename PieceConfirmation and TranslatedRef.

porcuquine:
This standardizes some terminology, renaming `PieceConfirmation` to `PieceInclusionProof` and `TranslatedRef` to `CommP`.


porcuquine:
cc @laser @sidke

PR #233, opened 2019-04-10
Piece Inclusion Proof constructions

acruikshank:
This is a note on possible constructions for Piece Inclusion Proofs and how they trade off proof size and structure complexity for storage efficiency.

It contains very little mathematical rigor. I'm confident I'm correct enough to talk effectively about trade offs. I suspect I'm wrong in some details.

It discusses the 4 best constructions I know of. Better constructions could definitely exists.


pooja:
@whyrusleeping This looks like it needs your comments!

whyrusleeping:
@acruikshank good notes, i'll merge this and we can iterate on it for the final proof construction.

PR #236, opened 2019-04-17
added some definitions related to generation attack threshold

sternhenri:
Added
- Generation Attack Threshold
- Polling Time
- Power Table Lookback
- Randomness Lookback


whyrusleeping:
@nicola can you review?

definitions.md
```diff
@@ -175,6 +191,12 @@ Used in the Filecoin system by a storage miner to prove that client data was kep
 
 TODO add a note to distinguish predictability from randomness
 
+#### Randomness Lookback
```
nicola: I suggest a different name which is more specific, e.g. Leader Randomness Lookback, or Election Randomness Lookback.

definitions.md
```diff
@@ -141,10 +147,20 @@ A piece is a portion of a file that gets fitted into a sector.
 
 ****The initial commitment of a storage miner to provide a number of sectors to the system.
 
+#### Polling Time
+
+Security Parameter. Number of rounds between every proving period.
```
nicola: this is a bit misleading since the proving period is the duration of an entire post.

Polling time is the time between two online proofs of replication in a PoSt proof.

We can be more specific to what it means in proof of space, but this will be ok for spec purposes

PR #238, opened 2019-04-21
 a proposal on filecoin pledge mechanism

arsstone:


whyrusleeping:
Hey @arsstone, thanks for the proposal. I'll hopefully get to reviewing this today or tomorrow

whyrusleeping:
So my understanding of this proposal is to add a 'middleman' entity that manages penalizing miners, and allows some amount of leniency in whether or not they have a fault?

I think some things in the spec might not be clear, so let me explain how things currently work.

If a miner loses data in the current system, they have some small amount of slack time to recover the data. This isnt long (hasnt been decided concretely yet) but it will likely be less than an hour. If the fault gets repaired in that slack time, then there is no penalty.

If the fault cannot be repaired in time, the miner has two options. They can report the sectors that have failed, and pay just the collateral for the failed sectors, and mark them as 'temporary failures'. These temporarily failed sectors can be recovered at any point during the next post.

The second option the miner has is to submit their PoSt late. Late submission of PoSts requires paying some fee proportional to how late the miner is.

These penalties are meant to be quite small, the amount of collateral actually being penalized here shouldnt break the bank,it should just make things unprofitable for attackers who get caught.

It seems to me like you might be under the impression that a miner loses all of their collateral if they miss proving a single sector, which is not the case.

arsstone:
Hi whyrusleeping. I fully understand what you mean by the current Filecoin mortgage mechanism. I know that the current mortgage mechanism will change in time.
I am proposing a new mortgage mechanism from another dimension. Our two mechanisms are not contradictory, but synergistic. Let me give some examples of the actual operation and maintenance of distributed cloud storage.
If the miner wants to move data from the A machine to the B machine. Or the big miners migrated their data centers and moved from the A-site to the B-room. The current Filecoin mortgage mechanism is not possible.
My intention is not to add a middleman entity, but to the file contract described in the filecoin white paper. Through the rules constitute a virtual middleman - the big miners. The big miners have a collection of small miners.
  https://github.com/arsstone/specs/blob/master/drafts/proposal-on-filecoin-pledge-mechanism/proposal-on-filecoin-pledge-mechanism.md
[caption](http://185.239.68.188/static/20190426.jpg)
My intention is to build the filecoin on-chain information system and filecoin off-chain information system. Separating the Sealing Layer from the Mortgage Layer gives flexibility. This program gives Filecoin metabolic abilities so that it can be consistent with a systematic theory. A reliable overall can be achieved with a large number of unreliable components. This will allow FIlecoin to build a stable and robust whole with a cheaper storage server. This reduces storage costs within the Filecoin network.
With my mechanism, it is easy to implement mining.md
Future Work
There are many ideas for improving upon the storage miner, here are ideas that may be potentially implemented in the future.
Sector Resealing: Miners should be able to 're-seal' sectors, to allow them to take a set of sectors with mostly expired pieces, and combine the not-yet-expired pieces into a single (or multiple) sectors.
Sector Transfer: Miners should be able to re-delegate the responsibility of storing data to another miner. This is tricky for many reasons, and will not be implemented in the initial release of Filecoin, but could provide 
The current timeout penalty mechanism is not friendly to the home version of miners. Family power outages, network disconnects, accidents, etc. One day or more than 10 hours is a high probability event in many countries. In this case, the current timeout penalty and the proportional deduction deposit are not good solutions. With 99% reliability, there will be downtime of 3.65 days in a year. If the number of mining machines in the home version is large enough, it can be accurately measured, and all home version mining machines will fail within one year. Filecoin will be detained. If reliability is required to reach 99.9%, 8.76 hours of downtime will occur within one year. All home mining machines will fail 8.76 hours a year. Three 9-rated home versions are impossible to achieve. Just think about the router at home, and you have to restart it a lot to know.
The solution I proposed can give off-chain information system service providers the ability to play their faults at the miners.

arsstone:
Hi Why 。
      My intention is not just to propose this mortgage mechanism. I want Filecoin to rethink whether the FIlecoin project should include an on-chain information system and an off-chain information system . Then think about it, the two systems work together.
     The distinction and cooperation between the on-chain information system  and the off-chain information system  is also lacking in all current blockchain projects. If you are also thinking about the difference between these two systems. Then the boundaries of the two systems are there. How to present it in the Filecoin project. I think one of the ways is that the two systems interact through the interface API.

whyrusleeping:
@arsstone i'm not exactly sure what you mean by off-chain information system. What information is stored there? how is it verified? Who stores it? Why do they store it?

anorth:
@whyrusleeping @porcuquine  this raises a interesting question to which I don't think the answer is obvious (to me, anyway). If a miner irrecoverably loses a sealed sector, but has access to the client data that was sealed into it (e.g. on a tape backup), is it possible to reconstruct the sector and resume proving? Or will the proofs be incompatible with the original sealed sector?

porcuquine:
@anorth You can always reseal a sector if you have the original inputs. Sealing is pure and should take just as long each time.

arsstone:
@whyrusleeping I mean the off-chain information system is actually similar to Ethereum's two-layer chain system.Off-chain information systems do not require chain verification, they are the supporting layer of the chain system. For example, the reconstruction of data, redundant coding, mining system detection is the off- chain system. 
       For example, when the off-chain information system detects the failure of the mining machine hard disk, it actively reports to the chain. (This is similarly described in Filecoin Fault spec). 
       Filecoin does not need to check the authenticity of the off-chain information system, and the blockchain only checks the data it cares about. In this specific column, Filecoin only checks if the data between the mortgage layer and the seal layer is correct. As for why the off-chain information system is motivated, where the data is stored, who saves it, how to verify it completely ignores it. 
     Filecoin is a complex system, and complex systems are divided into many sub-modules. The off-chain information system is another layer.

arsstone:
@anorth    Yes, you fully understand the meaning of one of me. I mean, when the miner loses the seal, he can rebuild the seal. And the chain can verify whether the seal he reconstructed is correct or not. My proposal is to require Filecoin to decouple the seal layer from the mortgage layer. Only when these two layers are decoupled, it is possible for miners to report failures and download data from other data sources to reseal, and then have a specific set of mortgage layers point to the new seal.

arsstone:
@porcuquine    Yes. Of course, each time the seal is rebuilt, a seal needs to be regenerated. Both need to be sealed for a long time. My requirement is that the resealed data can point to the layer of mortgage that the miner belongs to. Once you understand the decoupling of the mortgage layer and the sealing layer, you will get a very flexible data redundancy mechanism.
   For example, the repair protocol is combined with this pledge mechanism, the repair protocol gives the client the ability to protect data, and the mortgage mechanism is to give the miner the ability to guarantee data. The combination of the two will be very good. For a specific example, the current miner A's sealed data b is broken due to the hard disk. First, the off-chian information system immediately captures the fault, and it can start the reconstruction of the data. At the same time, the on-chain information system on the chain also captures the fault. There is a question who will first fix this data . The obvious fact is that if both the miner and the client are repaired at the same time, resources will be wasted. At this time, you can define it so that the miner can fix the fault within a limited time; if there is no repair fault within a limited time, the client initiates a repair agreement to re-employ a new miner to fix the fault.

whyrusleeping:
@arsstone a resealed sector is going to be exactly the same as the original sealed sector. Additionally, sectors don't 'point' at anything, so i'm not sure what youre saying here. I think that you're saying the collateral for sectors should be pooled in some way?

arsstone:
Yes. I found that the two of us are discussing different levels of things. This creates our confusion and understanding errors. We should clarify the differences in language between us.
      In my proposal, I actually proposed a whole set of mortgage mechanisms. It contains abstract conceptual categories and specific implementation categories.My abstract concept category is to propose two new objects: the mortgage layer and the seal layer. How the mortgage layer and the sealing layer work is the specific implementation category. 
      I know that Filecoin currently has resealed, this resealed is what you said (Filecoin Fault.md), which is the specific implementation category I understand.
 
   https://github.com/filecoin-project/specs/blob/master/faults.md Reported storage fault penalty
  The fault report here includes two actions, one is to reseal the data and the other is to resubmit the sector.
 Note: This implies that miners cannot re-seal data into different sectors. We could come up with a protocol where the client gives the miner explicit consent to re-seal, but that is more complicated and can be done later.

      At Filecoin Fault.md, you are aware of these issues. But at the moment, I think Filecoin treats both actions as a whole when thinking about sealing and mortgages. Then a seal and a mortgages the two actions are coupling  together. 
     Obviously, sealing and mortgage are two acts, and there is a time between them. I just abstracted these two specific behaviors and proposed the sealing layer and the mortgage layer. They should all be reflected in the file contract. Filecoin's file contract is equivalent to Ethereum's ERC20 token contract.

   Let me give a concrete example. Customer A stores file F in Big Miner B.
   Customer A, the file is F
   File Contract E: A->B(F:P , / M)
   Explanation: E is a file contract, customer A has document F in Miner B, P is the price payment rule, and M is the mortgage rule.
   Big miner B
   Small miners b, c
   Initial state: small miner b is sealed for file F as f
    Fault status

   At present, because the miner b (mine machine node) is broken, the miner's u off-chain information system detects that the miner b is broken. The big miner started the repair process.
   Repair status: The big miner designated the miner c to rebuild the document F and seal it to f'. At this time, M needs to be able to be reprogrammable. At this time, we can understand M as the miner's fault repair contract in the file contract.
   M:b.F. f->c.F.f'/ f=f' F-link-f=f' is required to pass the chain check.

    After understanding the idea of decoupling the mortgage layer from the seal layer, we revisited Filecoin's file contract.
   A file contract can be understood as a collection of rules and contracts that build a storage market and a search market.
   When we understand the Filecoin file contract above, we can understand that my proposed mortgage proposal is actually a programmable repair contract for the miner under the file contract.
   We can also understand the current Filecoin https://github.com/filecoin-project/specs/blob/master/notes/repair-market.md

    This repair miner is equivalent to proposing a programmable repair contract from the client under the file contract.
    The repair of the miners is nothing more than to ensure that the customer's data is not lost. If you take my advice, Filecoin doesn't have to distinguish between storage miners and repair miners. Because the properties of the repair miners have been included in the rules of the file contract.

   The big miners are constructing  the off-chain information system and running the on-chain file contract to construct a reliable blockchain distributed cloud storage service by constructing different file contracts. This is what I call the collaboration of off-chain and on-chain.

 After accepting my proposal, there will be more details to add.

arsstone:
@whyrusleeping     Yes. I found that the two of us are discussing different levels of things. This creates our confusion and understanding errors. We should clarify the differences in language between us.
      In my proposal, I actually proposed a whole set of mortgage mechanisms. It contains abstract conceptual categories and specific implementation categories.My abstract concept category is to propose two new objects: the mortgage layer and the seal layer. How the mortgage layer and the sealing layer work is the specific implementation category. 
      I know that Filecoin currently has resealed, this resealed is what you said (Filecoin Fault.md), which is the specific implementation category I understand.
 
  Https://github.com/filecoin-project/specs/blob/master/faults.md Reported storage fault penalty
  The fault report here includes two actions, one is to reseal the data and the other is to resubmit the sector.
 Note: This implies that miners cannot re-seal data into different sectors. We could come up with a protocol where the client gives the miner explicit consent to re-seal, but that is more complicated and can be done later.

      At Filecoin Fault.md, you are aware of these issues. But at the moment, I think Filecoin treats both actions as a whole when thinking about sealing and mortgages. Then a seal and a mortgages the two actions are coupling  together. 
Obviously, sealing and mortgage are two acts, and there is a time between them. I just abstracted these two specific behaviors and proposed the sealing layer and the mortgage layer. They should all be reflected in the file contract. Filecoin's file contract is equivalent to Ethereum's ERC20 token contract.
Let me give a concrete example. Customer A stores file F in Big Miner B.
   Customer A, the file is F
   File Contract E: A->B(F:P , / M)

Explanation: E is a file contract, customer A has document F in Miner B, P is the price payment rule, and M is the mortgage rule.
   Big miner B
   Small miners b, c
   Initial state: small miner b is sealed for file F as f
    Fault status
At present, because the miner b (mine machine node) is broken, the miner's u off-chain information system detects that the miner b is broken. The big miner started the repair process.
Repair status: The big miner designated the miner c to rebuild the document F and seal it to f'. At this time, M needs to be able to be reprogrammable. At this time, we can understand M as the miner's fault repair contract in the file contract.
M:b.F. f->c.F.f'/ f=f' F-link-f=f' is required to pass the chain check.

After understanding the idea of decoupling the mortgage layer from the seal layer, we revisited Filecoin's file contract.
A document contract can be understood as a collection of rules and contracts that build a storage market and a search market.
   When we understand the Filecoin file contract above, we can understand that my proposed mortgage proposal is actually a programmable repair contract for the miner under the file contract.
   We can also understand the current Filecoin https://github.com/filecoin-project/specs/blob/master/notes/repair-market.md

  This repair miner is equivalent to proposing a programmable repair contract from the client under the file contract.
The repair of the miners is nothing more than to ensure that the customer's data is not lost. If you take my advice, Filecoin doesn't have to distinguish between storage miners and repair miners. Because the properties of the repair miners have been included in the rules of the file contract.

The big miners are constructing  the off-chain information system and running the on-chain file contract to construct a reliable blockchain distributed cloud storage service by constructing different file contracts. This is what I call the collaboration of off-chain and on-chain.

After accepting my proposal, there will be more details to add.

sternhenri:
@arsstone Just an update. I think this could be a good candidate for a Filecoin Improvement Proposal down the line (I haven't fully dug into it but reading rapidly). Note that we haven't overlooked this, just got crunched in finalizing pledge collateral for v1

mishmosh:
Closing this as a PR. Could be good candidate for a Filecoin Improvement Proposal after v1 (process will be described in [1.4 Filecoin Spec Process](https://filecoin-project.github.io/specs/#intro__process)).

PR #242, opened 2019-04-23
initial document describing how client data pre-processing works

whyrusleeping:
Discussing specifically how clients are expected to pre-process their data before making a deal with a storage miner.

cc @porcuquine @laser @nicola

### DIFF
--- a/client-data.md
+++ b/client-data.md
@@ -31,4 +31,41 @@ For details on how UnixfsV1 works, see its spec [here](https://github.com/ipfs/s
 
 Storage Encoding is computed by building a merkle tree out of the data using `RepHash`.
 
-TODO(proofs team): Fill in how exactly this works.
+Inputs to `RepHash` must first be preprocessed then padded.
+
+__*Preprocessing*__ adds two zero bits after every 254 bits of original data, yielding a sequence of 32-byte blocks, each of which contains two zeroes in the most-significant bits, when interpreted as a little-endian number. That is, for each block, `0x11000000 & block[31] == 0`.
+
+Next, __*piece padding*__ is added: blocks of 32 zero bytes are added so that the total number of blocks (including __*piece padding*__) is a power of two.
+
+`RepHash` constructs a binary merkle tree from the resulting blocks, designated as the *leaves* — by applying the __*RepHash Compression Function*__, `RepCompress`, to adjacent pairs of leaves. The final result is the merkle root of the constructed tree.
+
+```go
+type node [32]uint8
+
+// Create and return the root of a binary merkle tree.
+// len(leaves) must be a power of 2.
+func RepHash(leaves node) node {
+	currentRow := leaves
+	for height := 0; len(currentRow) > 1; height += 1 {
+		var nextRow []node
+
+		for i := 0; i < len(row)/2; i += 2 {
+			left := row[i]
+			right := row[i+1]
+
+			// NOTE: Depending on choice of RepCompress, heightPart may be trimmed to fewer than 8 bits.
+			heightPart := []uint8{height}
+
+			input1 := append(heightPart, left...)
+			input := append(input1, right...)
+			hashed = RepCompress(input)
+			nextRow = append(nextRow, hashed)
+		}
+
+		currentRow = nextRow
+	}
+
+	return currentRow[0]
+}
+
+```
--- a/client-data.md
+++ b/client-data.md
@@ -1,0 +1,34 @@
+# Filecoin Client Data Processing
+
+Data that clients wish to store must be pre-processed before it can be sent to miners and have a storage deal made for it.
+
+This processing is comprised of two top-level pieces:
+
+1. **Transfer Encoding**
+2. **Storage Encoding**
+
+Transfer Encoding is a pre-processing for the file to allow it to be transferred over the network more easily. Both for the initial transfer to the storage miner, and for later retrievals. This mechanism can be done in any way that is supported by the client and the miner (alternative strategies may be implemented).
+
+Storage Encoding is a pre-processing for the file that allows the Filecoin storage market and proofs to reference the data. Filecoin operates on a special type of merkle tree generated with a function we call `RepHash`. The root hash of this merkle tree is called the `Piece Commitment` or `CommP` and is included in the deal when it is proposed to a miner.
+
+The next sections will go into some additional detail on how each of these encodings works.
+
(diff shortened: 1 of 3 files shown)


whyrusleeping:
I'm not convinced that 'encoding' is the right word to use here, so please suggest something better

client-data.md
```diff
@@ -0,0 +1,34 @@
...
+- Leaf Format: Raw
+- Max Branch Width: 1024
+
+For details on how UnixfsV1 works, see its spec [here](https://github.com/ipfs/specs/tree/master/unixfs).
+
+## Storage Encoding
+
+Storage Encoding is computed by building a merkle tree out of the data using `RepHash`. 
+
+TODO(proofs team): Fill in how exactly this works.
```
whyrusleeping: @dignifiedquire do you think you could handle filling this in?

whyrusleeping:
Also related:
- https://github.com/ipfs/specs/pull/199
- https://github.com/ipld/specs/pull/114

client-data.md
porcuquine: I could also take a shot at this when I have a little time, but in considering it a few more details came to mind.

The way I see it, `RepHash` operates on field elements and is used both to generate `CommP` and as a component of PoRep itself. So the pre-processing (bit-padding) step precedes `RepHash` and is separate from it.

We could name the aggregate something new if we wanted to, but it might be simpler just to explain it. We'll need to describe the bit-processing somewhere anyway. So the question is whether this should be the canonical place or should refer to a description elsewhere.

We can always add here then move elsewhere if appropriate. Also note that although clients will need to perform pre-processing in order to generate `CommP`, the data they actually send shouldn't be preprocessed — since preprocessing is currently applied after sealing. (This is somewhat wasteful, but it is how things currently work.)

arsstone: I think it is important to distinguish the different roles of the Filecoin network to discuss the data pre-processing needs of customers. First of all, I think there will be three roles in the Filecoin network in the future, End user, broker, miner. It is not possible if Filecoin wants to design both the End user and the miner p2p.
    When Filecoin only wants to design End users and miners for point-to-point transactions. In the first case, the miner has already implied the role of an intermediary. For example, the big miners, the big miners include countless small miners. When the big miners completed how to manage and dispatch small miners, the big miners had already acted as intermediaries. In the second case, the client implies the role of the intermediary, the client will layer the data, one layer is directly used by the end user; the other layer is the multi-miner scheduling, based on the  filecoin pledge mechanism I mentioned, when the client wants to implement When the fault-tolerant miners end, it has actually changed and turned into a large miner dispatch.
  To become popular, Filecoin must be using low in cost. If each customer does a set of such a mechanism, it cannot be quickly promoted. So I assert that Filecoin must have an intermediary.
       So after we got this understanding, we started to discuss how the client's data is pre-processed. Since Filecoin uses Proof-of-Replication this makes it unsynchronized when the user transfers data to the miner seal. Therefore, there must be a stage of pre-processing in the Filecoin network.
When we envision an actual user usage process: a Filecoin based network disk. Once the user has uploaded the photo, they want to see their photos and thumbnails in real time. According to the current Filecoin design, due to the seal of theProof-of-Replication. It is impossible to view in real time. At this time, FIlecoin's intermediary must do the pre-processing layer action, on the one hand, to arrange data to seal; on the other hand, when the data is not sealed, respond to the user data that has been received, such as viewing.
I think Filecoin needs to abstract out the  on-chain  information system and off-chain information system. As far as I understand, FIlecoin is to be a public trading market, then FIlecoin's on-chain information system should be the infrastructure layer of this publicly traded market, and the intermediary, user, and miner in this open market have both behaviors that is on-chain behavior and off-chain behavior. The intermediary service providers that survive on the Filecoin network must have an off-chain information system, and this off-chain information system is the unique skill of each intermediary service provider. Filecoin needs to provide an interface between the off-chain information system and the off-chain information system. For example, the pre-processing behavior should be coordinated by the off-chain information system and the off-chain information system.

client-data.md
```diff
@@ -0,0 +1,34 @@
...
+
+Transfer Encoding is a pre-processing for the file to allow it to be transferred over the network more easily. Both for the initial transfer to the storage miner, and for later retrievals. This mechanism can be done in any way that is supported by the client and the miner (alternative strategies may be implemented).
+
+Storage Encoding is a pre-processing for the file that allows the Filecoin storage market and proofs to reference the data. Filecoin operates on a special type of merkle tree generated with a function we call `RepHash`. The root hash of this merkle tree is called the `Piece Commitment` or `CommP` and is included in the deal when it is proposed to a miner.
+
+The next sections will go into some additional detail on how each of these encodings works.
+
+## Transfer Encoding
+
+Transfer Encoding may be computed in a variety of ways, as long as it is supported by both the client and miner involved. Here, we document the `default` transfer encoding.
```
laser: Dumb question: Does our reference implementation use Unixfsv1 today? I have not yet delved into the guts of the blockservice / bitswap combo (the current client-to-miner transfer mechanism).

client-data.md
laser: @porcuquine 

> since preprocessing is currently applied after sealing

Bit padding is applied before sealing, when the piece is being added to the staged sector. Which preprocessing are you referring to here?

porcuquine: Sorry, word glitch: updated my comment. Not after sealing — after data is received from the client.

whyrusleeping: @arsstone I'm not sure exactly what youre asking, but the purpose of the pre-processing specified here is to prepare data to be verifiably stored. 

The concept of an intermediary won't be part of the network at launch, be we are absolutely working on it, and it will be coming in a later release. For more information on our thoughts there, take a look at the repair market docs: https://github.com/filecoin-project/specs/blob/master/notes/repair-market.md

client-data.md
```diff
@@ -0,0 +1,71 @@
...
+
+### `default`
```
anorth: I suggest seeking a more meaningful name for this, maybe `UnixfsV1` or something derived from it. Explicitness will aid understanding of what's going on. Also I think the mechanism will surprise someone not already intimately familiar with IPFS, who might otherwise expect more-or-less raw bytes. It would be appropriate to repeat a 1-sentence justification for this more complex scheme here, and a better name will help set expectations.

client-data.md
whyrusleeping: yeah, thats what we currently use.

PR #248, opened 2019-04-29
Initial cut at zigzag spec.

porcuquine:
This is still very rough and needs reworking. In particular, it's not clear how much detail to include (for now) when describing component algorithms.

On the one hand, we have working code which the spec should agree with completely. On the other hand, it's written in Rust and contains artifacts of the implementation.

I need to figure out how to split the difference between translating into an imaginary Go implementation, translating to some pidgin pseudocode, or just leaving high-level english stubs for now.

@whyrusleeping @nicola Input welcome. I'm going to rest this for the moment and come back to it fresh later.


zigzap-porep.md
```diff
@@ -0,0 +1,281 @@
...
+__*Filecoin ZigZag Proof of Replication*__ is the process by which raw data is transformed into a replica and a proof of replication.
+
+This comprises the following steps:
+ - Preprocessing
+ - Sector Padding
+ - Replication
+ - Proof Generation
+ - Circuit Proof Generation
+ 
+ Together, __*Replication*__, __*Proof Generation*__, and __*Circuit Proof Generation*__ constitute the `Seal` operation, as described in [Filecoin Proofs](proofs.md).
```
nicola: up until this point this is general glossary, not hash functions; I would use a different heading.

dignifiedquire:
@porcuquine even if this is not a 100% right, might be useful to get into the spec, such that we at least have sth that describes the current construction.

porcuquine:
Fair point, @dignifiedquire.

@whyrusleeping Could you take a quick pass over this and provide me with targeted feedback I can use to get this to V0.1 merge-readiness? We could merge as-is or with a bit more marking for WIP — or we can clean it up a little more. I will complete this properly but don't have time for that quite yet.

zigzap-porep.md
```diff
@@ -0,0 +1,284 @@
...
+```
+
+`RepHash` constructs a binary merkle tree from the resulting blocks, designated as the __*leaves*__ — by applying the __*RepHash Compression Function*__, `RepCompress`, to adjacent pairs of leaves. The final result is the merkle root of the constructed tree.
+
+```go
+type node [32]uint8
+
+// Create and return a binary merkle tree and its root commitment.
+// len(leaves) must be a power of 2.
+func RepHash(leaves node) node {
```
whyrusleeping: leaves should be of type []node, right?

porcuquine: Yeah, let me change that.

porcuquine:
Will do, @whyrusleeping.

PR #249, opened 2019-04-30
Initial work toward PieceInclusionProof (merkle inclusion proof spec).

porcuquine:
This PR creates a directory for TLA+ specs and adds one for Piece Inclusion Proof.

That spec is a work in progress and so far only specifies `RepHash` in terms of an abstract `RepCompress`.


porcuquine:
Ack comments.

PR #250, opened 2019-05-01
add sharray spec and make block use it for messages and receipts

whyrusleeping:


sharray.md
```diff
@@ -0,0 +1,90 @@
+# Sharray
+
+> Sharded IPLD Array
+
+The Sharray is a fixed width IPLD tree structure used to store an array of items.
+
+## Overview
+
+Each node has a height, and a number of items. The number of items must not exceed the trees given width.
```
Stebalien: Width == degree?

sharray.md
```diff
@@ -0,0 +1,90 @@
...
+
+The tree must not be sparse, if the tree represents an array of N items, then the left `N/Width` leaves must contain the first `N` items. If `N` is not evenly divisible by `Width` then the final leaf must contain the final remainder. Every node in the tree must have its height set to one less than the node above it. Nodes with a height of 0 contain array values, and nodes with heights greater than zero contain the cids of their child nodes.
```
Stebalien: I'm having trouble parsing this. Should this be, "each node must contain (height * width) items"? From the text, it sounds like we need to fit `N` items into `N/Width` slots.

sharray.md
```diff
@@ -0,0 +1,90 @@
+# Sharray
+
+> Sharded IPLD Array
+
+The Sharray is a fixed width IPLD tree structure used to store an array of items.
```
Stebalien: Just to be clear, this _is_ for fixed sized arrays. O(n) insertion isn't going to be an issue.

sharray.md
whyrusleeping: yes

sharray.md
whyrusleeping: yeah. This is fixed size arrays, whose items are all known at creation time. Build-once, use many.

sharray.md
whyrusleeping: Yeah... I had a hard time writing this. I'm basically trying to say every node has at most Width items. And all the items go on the left side of the tree (no gaps). suggestions?

sharray.md
```diff
@@ -0,0 +1,90 @@
...
+
+> Create a sharray from a given set of items
+
+```go
+func create(items []Item) Cid {
+  var layer []cid.Cid
+  
+  for len(items) > 0 {
+    // get the next 'Width' items from the input items
+    vals, items = getNextN(Width, items)
```
dignifiedquire: maybe call it `popN` that makes it clearer that things get removed, otherwise this loop doesn't terminate

dignifiedquire: wait are you reassigning `items` here? hmm this is a pretty confusing line :/ 
maybe change the loop to run over a counter `i` and just increase it by `width` every round? would make the intent much easier to understand

sharray.md
```diff
@@ -0,0 +1,90 @@
...
+
+```go
+func (n node) get(i int) Item {
+  if n.Height == 0 {
+    return n.Array[i]
+  }
+  
+  childWidth := Pow(Width, n.Height)
+  
+  child := loadNode(n.Array[i / childWidth])
```
dignifiedquire: how exactly does `loadNode` work? Also we don't ever call `storeNode` in the creation code, so this would technically not work out

sharray.md
```diff
@@ -0,0 +1,90 @@
...
+  height Int
+  items [Item]
+} representation tuple
+```
+
+`Item` may be either a direct value, if `height == 0`, or the Cid of a child node if `height > 0`.
+
+(For details on IPLD Schemas, see the [IPLD Schema Spec (draft)](https://github.com/ipld/specs/blob/dcbfb25468092be796bab90e90e3f2535fdeddc7/schema/representations.md))
+
+We use DAG-CBOR for serialization, and blake2b-256 for hashing.
```
whyrusleeping: should change to "To construct CIDs for sharray nodes, use DAG-CBOR for serialization and hash with blake2b-256"

sharray.md
whyrusleeping: Yeah, i left those details out because they arent necessarily relevant to the algorithm. Mostly trying to show structure here. If you think its important I can put those calls in.

anorth:
I asked some questions in https://github.com/filecoin-project/specs/issues/255, but in brief:
- This removes message from blocks without restoring a way for them to be fetched, e.g. during chain sync. Should there be a `Block`/`BlockHeader` distinction made explicitly here? We do not want to add more round-trips to chain syncing.
- Is it intended that a merkle proof of membership is implementable for future light clients? That should at least be stated, if not the mechanism detailed.

whyrusleeping:
@anorth ah yes, the FetchMessages call made during SyncBootstrap is not also called during SyncCaughtUp. Thats a mistake.

Also, re: round trips, yes. The current description in the spec says to use bitswap, which would incur annoying extra round trips. We're hoping to have graphsync soon, But I think that for the short term we should just add in message fetching to the block sync protocol to avoid those round trips.

data-structures.md
```diff
@@ -327,7 +331,7 @@ Additionally, CBOR Major type 5 is not used. If an FCS object contains it, that
 
 ## IPLD Considerations
 
-Cids for FCS objects should use the FCS multicodec (`0x1f`).
+Cids for FCS objects should use the FCS multicodec (`0x1f`), and should be hashed with blake2b-256.
```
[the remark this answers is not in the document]
dignifiedquire: maybe use `with hash function blake2b-256`?

sharray.md
dignifiedquire: The tree must not be sparse. 
Given an array of size `N` and a fixed width of `W`.
- The left `floor(N/W)` leaves contain the first `N` items.
- If `N % W != 0` the final leaf contains the final remainder.
- The tree is perfectly balanced.
- The height is the distance from the leaves, not the root.
- Leaves (nodes with a height of 0) contain array values.
- Inner nodes (nodes with height greater than zero) contain the cids of their child nodes.

sharray.md
dignifiedquire: I think it is important to show the call to `storeNode`, and give like a three word description of what `store/loadNode` do.

anorth:
Clarifying my understanding here: please tell me if I'm wrong.
- Right now, the sharray serialisation defined solely for the purpose of computing a CID. #261 adds a flat array of messages to the interim blocksync protocol 👍 which could be used by a node either syncing or after receiving a gossipped block (header). The serialised sharray would never be transmitted.
- When [graphsync](https://github.com/ipfs/go-graphsync) is used for syncing, the serialised sharray can be included in response to a query ([selector](https://github.com/ipld/specs/blob/master/selectors/selectors.md)) for a block without requiring additional roundtrips. Graphsync removes the need that other blockchains have for a `Block` structure than joins a header with body (messages + receipts) since a response to a single query can traverse a large IPLD graph.

That would make sense to me. The implicit dependence on graphsync's capabilities here may have contributed to this not being obvious to myself and the likes of @gnunicorn.

whyrusleeping:
> The serialised sharray would never be transmitted.

The sharray (or any merkle-tree like structure we might use) is used for hashing the messages together, but also used to make merkle proofs. In the case of light clients, the serialized sharray may be transmitted to (for example) prove to a light client that a given message was included in the block, and prove its receipt without requiring the client download the entire block contents.  

Graphsync is a nice to have, but we can (and already have, in the case of the blocksync protocol) write up simple protocols that perform the queries we want to use graphsync for.

PR #251, opened 2019-05-03
Add rle+ datastructure

### LINKED ISSUE #133: sector bitfield proposal
### Problem
Filecoin miners need a compact way to specify which of a (large) set of sectors is either bad or complete. The number of sectors in the set may be in the tens of millions, and a bitset selecting a subset of them must be much much smaller than that.

If sectors are selected randomly, then it is unlikely that we can gain much over one bit per sector. However, it is very likely that miners will fill disks up one at a time, and so sectors added around the same time would likely be physically near (on the same disk). If a disk failure occurred, then an entire range of sectors would be lost, which is really easy to represent with something like run length encoding. 

## Proposal 0 - Simple RLE
Encode the bitset as follows:

```go
func encodeBitset(bitsetQ) []byte {
  var out []byte
  
  var countType bool
  for !bitsetQ.Empty() {
    var count int
    for bitsetQ.Pop() == countType {
      count++
    }
    out = append(out, varint.Encode(count))
  }
  return out
} 
```

This works pretty well in most ideal cases, but breaks down if bits are set randomly (or in some pattern, like 011011011011). 

## Proposal 1 - RLE + Extension
To address the cases in which the previous proposal is inefficient, we can give the encoding the ability to specify bitfields directly. To do this, we can say that while reading the encoding, if a negative varint is parsed, that many bytes should be read and treated as the exact bitfield for the duration. 

For example:

`000000011111100000011111110110110001010101`
would encode as
`varint<7> varint<6> varint<6> varint<7> varint<-2>[]byte{108, 85}`

(implementation note: it is sufficient just to implement a correct parser for this, and leave the optimization of the encoding up to the individuals)

dignifiedquire:
Note: I haven't figured out the identification prefix yet, I haven't forgotten about it.

Closes #133

### DIFF
--- a/data-structures.md
+++ b/data-structures.md
@@ -380,7 +380,8 @@ The format can be expressed as the following [BNF](https://en.wikipedia.org/wiki
 
 ```
     <encoding> ::= <header> <blocks>
-      <header> ::= <bit>
+      <header> ::= <version> <bit>
+     <version> ::= "00"
       <blocks> ::= <block_single> | <block_short> | <block_long>
 <block_single> ::= "1"
  <block_short> ::= "01" <bit> <bit> <bit> <bit>
--- a/data-structures.md
+++ b/data-structures.md
@@ -368,7 +368,7 @@ MessageReceipts: []MessageReceipt{}
 
 RLE+ is a lossless compression format based on [RLE](https://en.wikipedia.org/wiki/Run-length_encoding).
 It's primary goal is to reduce the size in the case of many individual bits, where RLE breaks down quickly,
-while keeping the same level of compression for large sets of contigous bits.
+while keeping the same level of compression for large sets of contiugous bits.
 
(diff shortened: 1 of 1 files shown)


data-structures.md
```diff
@@ -363,3 +363,58 @@ StateRoot:       Cid("zDPWYqFD5abn4FyknPm1PibXdJ2kwRNVPDabKyzfdXVJGjnDuq4B")
 Messages:        []SignedMessage{}
 MessageReceipts: []MessageReceipt{}
 ```
+
+## RLE+ Bitset Encoding
+
+RLE+ is a lossless compression format based on [RLE](https://en.wikipedia.org/wiki/Run-length_encoding).
+It's primary goal is to reduce the size in the case of many individual bits, where RLE breaks down quickly,
+while keeping the same level of compression for large sets of contigous bits.
```
ZenGround0: nit: contiguous

dignifiedquire:
@whyrusleeping added version bits

data-structures.md
dignifiedquire: fixed, thank you

whyrusleeping:
🆗

PR #258, opened 2019-05-09
Slight restructure of documents

gnunicorn:
Generally move from the bottom up, consolidate everything basic under a "core", the on-chain processing under actors and move spec related things into its own section at the end.

Cleans up the intro and fixes some style, too.

Looks like this:
Screenshot-20190509155237-404x1566


INTRO.md
```diff
@@ -4,15 +4,6 @@ This is the official Filecoin protocol specification. It is a work in progress.
 
 This collection of pages specify the protocols comprising the Filecoin network. The goal of these specs is to provide sufficient detail that another implementation, written using only this document as reference, can be fully compatible with `go-filecoin` peers, but this is still a work in progress.
 
-
-### Style
```
whyrusleeping: I don't think we should remove this. We should adhere to this strctly, and address the (many) places that we don't use this correctly.

whyrusleeping:
+1 to the intro restructure, but let's avoid dropping the RFC2119 bits.

INTRO.md
gnunicorn: sorry, wasn't meant to be dropped, but moved. then forgot to add the file. will add it now.

gnunicorn: Added. The reasoning for splitting that out were the following two:

1. I want to add more (like the just added hints in #259 ), making this a lot more verbose and though
2. it is often said in the beginning of IETF RFCs, I think it hinders the reading flow here.

I would see the argument about having it as a sup-entry under "Overview" on the left (rather than at "SPEC" at the end) but I'd argue, it is less important for usual reader.

gnunicorn:
Fixed the missing styles file and added a section for the hints.

whyrusleeping:
It looks like theres a conflict in the summary doc... not sure why. Could you take a look?

gnunicorn:
@whyrusleeping fixed.

whyrusleeping:
Thanks!

PR #261, opened 2019-05-10
allow blocksync to sync messages

whyrusleeping:


network-protocols.md
```diff
@@ -443,27 +443,41 @@ TODO: document the query deal interaction
 
 # BlockSync
 
-The blocksync protocol is a small protocol that allows Filecoin nodes to request ranges of blocks from each other. It is a simple request/response protocol with a protocol ID of `/fil/sync/blk/0.0.0`. It uses CBOR-RPC.
+The blocksync protocol is a small protocol that allows Filecoin nodes to request ranges of blocks from each other. It is a simple request/response protocol with a protocol ID of `/fil/sync/blk/0.0.1`. It uses CBOR-RPC.
 
 ```go
 type BlockSyncRequest struct {
-	Start         Cid
+	Start         []Cid
```
dignifiedquire: can you add a comment that this means the tipset, otherwise not really clear

network-protocols.md
```diff
@@ -443,27 +443,41 @@ TODO: document the query deal interaction
...
+
+type TipSetBundle struct {
+  Blocks []Blocks
+  Messages []Message
+  MsgIncludes [][]int
+}
 ```
 
-The response contains the chain of requested blocks, in reverse iteration order, the zero'th block should be the block referenced by `request.Start`, and the following N blocks should be its N parents, and so on. This is done to streamline validation.
+The response contains the requested chain in reverse iteration order. Each item in the `Chain` array is contains the blocks for that tipset if the `Blocks` option bit in the request was set, and if the `Messages` bit was set, the messages across all blocks in that tipset. The `MsgIncludes` array contains one array of integers for each block in the `Blocks` array. Each of the arrays in `MsgIncludes` contains a list of indexes of messages from the `Messages` array that are in each `Block` in the blocks array.
```
dignifiedquire: "is contains" is really hard to parse

dignifiedquire: might be good to have a tiny example to make sure it is clear how this indexing works

```
Blocks: [b0, b1]
Messages: [m0, m1, m2]
MsgIncludes: [[0, 1], [1, 2, 0]]
```

matches to 

```
Block 0: [m0, m1]
Block 1: [m1, m2, m0]
```

whyrusleeping:
feedback addressed and merged!

PR #274, opened 2019-05-14
CommitSector takes FaultSet, not FailureSet

laser:
FailureSet isn't a specified type, but FaultSet is.


dignifiedquire:
thank you

PR #275, opened 2019-05-14
drop extra "Staged" state and correct "Complete" value

laser:
## Why does this PR exist?

- `Staged` is in the table twice
- `Complete` and `Staged` use same value


dignifiedquire:
thank you for fixing my mistakes :)

Issue #268, opened 2019-05-13
clarify distinction and uses of FCS and CBOR-RPC


whyrusleeping:
Filecoin has two different encoding schemes, both are cbor based. The first, FCS (Filecoin Compact Serialization) is used for all data structures that are 'on chain'. This is the Block (header) and everything it links to (messages, state, receipts and so on).

The second scheme, CBOR-RPC, is used for network communications between filecoin peers, where 'saving every byte' matters a bit less. If an object that usually is serialized with FCS is included in a CBOR-RPC message, it should continue to be encoded with FCS.

This should be detailed more clearly in the spec

gnunicorn:
I'd generally argue to have one or at maximum two clean and clearly defined encoding schemes for all types. Right now, encoding definitions are inconsistent and on a type-by-type basis:
 -  [`Address` defines a string-concatenation-scheme](https://filecoin-project.github.io/specs/docs/address/#encode),
 - so [do Signatures](https://filecoin-project.github.io/specs/docs/data-structures/#serialization-1)
 - while [`MessageReceipt`s are CBOR marshalled with lower case named fields](https://filecoin-project.github.io/specs/docs/data-structures/#serialization)
 - and [`Block`, `Message` and `SignedMessage` are encoded as CBOR-Arrays of their fields (in order of appereance in the spec) and prepended with an object type tag](http://localhost:1313/specs/docs/data-structures/#filecoin-compact-serialization)

The reasons for these different schemes are unclear, if intended at all.

I suggest taking a closer look at SCALE ([Section 9.1 in this draft paper](https://github.com/w3f/polkadot-re-spec/blob/master/polkadot_re_spec.pdf)), which is very similar to the concatenating system or FCS but more compact, for everything that is _on chain_ (`Messages`, `SignedMessage`, `MessageReceipts`, potentially `Actor`s as well?) and thus must be super compact. This, of course, could also be used for Network Transfers and messages on the network, but I think there it doesn't matter to have them this compact, and such a slightly more verbose system and flexible codec, like CBOR, could very well be expected standard. 

(But I am wondering, whether that might be something additional that the handshake message might contain and thus different encoding-formats between two peers might be negotiated - [similarly as http content negotiation works](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation). Form a spec perspective we don't really care how the blocks are encoded in transit, thus we can easily allow for easily extending on this level without failing to spec compliant.)

dignifiedquire:
[...]

PR #278, opened 2019-05-15
feat: use fcs for MessageReceipts and Signatures

dignifiedquire:
As suggested in #268


data-structures.md
```diff
@@ -161,18 +157,14 @@ type Signature struct {
...
+| BLS12-381 ECDSA | `2`     |
 
 ### Serialization
 
-`<uvarint(Type)><Data>`
-
-Note: As signatures should always be within wrapper types, length prefixing is not needed here.
-
-
+Uses FCS.
```
whyrusleeping: hrm... allocating a code for signatures in FCS doesnt feel necessary to me, as signatures are never serialized independent of the 'SignedMessage' type.

dignifiedquire: Shouldn't we just inline them then instead? I am worried that we will use them in different places, and then end up with them by themselves, when you eg traverse these structures using ipld

dignifiedquire:
@whyrusleeping update with the idea I mentioned, please take a look

data-structures.md
```diff
@@ -313,11 +305,15 @@ if ((shift <size) && (sign bit of byte is set))
 
 Datastructures in Filecoin are encoded as compactly as is reasonable. At a high level, each object is converted into an ordered array of its fields (ordered by their appearance in the struct declaration), then CBOR marshaled, and prepended with an object type tag.
```
gnunicorn: I had a chat with @pooja last week, and the optimization goals of our encoding-mix came up again and it still remains unclear (to me at least). Not strictly part of this changeset, but this sentence sums it up very well for me:

> each object is converted into an ordered array of its fields (ordered by their appearance in the struct declaration), then CBOR marshaled, and prepended with an object type tag.

We drop the bigger context of the named fields (that CBOR would support), convert it into an unnamed mixed-data array, which then is CBOR'ed adding the meta-information that it is an array as well as the type-information for each field - both unnecessary if we need to hard know the typed fields of the struct anyways. Leading to a weird, seemingly inconsistent mix of sometimes optimizing away type info in the name of "saving data", but sometimes keeping it with the "it's still good to have some context". The same for FCS, which contains some typed info, yet it is not possible to decode without knowing the wider context and struct. I am not sure what we are optimizing for here, not for saving space or we'd drop this - IMHO useless - type info...

whyrusleeping: The type information in CBOR doesnt actually take up much (if any) extra space. For most fields, the type and field length are combined into a single byte. Yes, we could maybe save a couple bytes by switching to a fully custom format, but a big part of the point here is that its not a fully custom format. It's CBOR, you can inspect the fields without necessarily requiring filecoin specific code or context, you can traverse the links with standard IPLD tooling, you can use off-the-shelf tooling for all of this. Yes, you can't convert the bytes of an address to its correct string representation without filecoin specific code, but that doesnt invalidate the usefulness of having the majority of the codec already written for you. 

I'm not entirely against a fully custom codec, but I just don't think its that worth it.

dignifiedquire:
Merging, as the ongoing discussion is tied to a more general issue not this particular change.

PR #281, opened 2019-05-16
CommitSector should accept a sector ID instead of producing a sector ID

laser:
## Why does this PR exist?

The spec [shows that sector id is an input to the seal operation](https://github.com/filecoin-project/specs/blob/master/proofs.md#seal). However, the current actors.md document shows that sector id is returned from the `CommitSector` method. The `CommitSector` method is called _after_ sealing has completed, so it cannot return the sector id. Rather, it needs to accept a sector id.

## What's in this PR?

This PR modifies `CommitSector` to accept a sector ID instead of returning a sector ID.

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -475,10 +475,10 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
 	miner.Collateral -= coll
 	miner.ActiveCollateral += coll
 
-	miner.Sectors.Add(commR)
-	// TODO: sectors IDs might not be that useful. For now, this should just be the number of
-	// the sector within the set of sectors, but that can change as the miner experiences
-	// failures.
+    // Note: There must exist a unique index in the miner's sector set for each
+    // sector ID. The `faults`, `recovered`, and `done` parameters of the
+    // SubmitPoSt method express indices into this sector set.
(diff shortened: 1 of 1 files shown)


actors.md
```diff
@@ -474,7 +475,7 @@ func CommitSector(comm Commitment, proof *SealProof) SectorID {
 	miner.Collateral -= coll
 	miner.ActiveCollateral += coll
 
-	sectorId = miner.Sectors.Add(commR)
+	miner.Sectors.Add(commR)
```
whyrusleeping: we should probably pass the sectorID as a parameter here, so it can insert it into the right place. Also, do we just want to store commR? I think we want to store commD as well for payments.

laser: > we should probably pass the sectorID as a parameter here, so it can insert it into the right place. Also, do we just want to store commR?

Good call. We do want the CommD (for payments) and we do want the SectorID (so we insert things at the correct indices).

How shall I represent this in psuedocode? How about:

```
miner.Sectors.Add((sectorID, commR, commD))
```

laser: Just pushed update to pseudo code + comment.

whyrusleeping:
LGTM, thanks @laser

laser:
my pleasure!

PR #282, opened 2019-05-20
add some notes on actor creation

whyrusleeping:


actors.md
```diff
@@ -162,10 +166,10 @@ type StorageMarketActor struct {
 
 | Index     | Method Name       |
 | -------- | ---------- |
-| 0   | `CreateStorageMiner`     |
```
dignifiedquire: why this change to `1` index?

whyrusleeping: I'm wanting to reserve the 0 value for the constructor

state-machine.md
```diff
@@ -17,6 +17,16 @@ Second, an `actor` may call a method on another actor during the invocation of o
 
 The `global state` is modeled as a map of actor `ID`s to actor structs. This map is implemented by an ipld HAMT (TODO: link to spec for our HAMT) with the 'key' being the serialized ID address (every actor has an ID address that can be looked up via the `InitActor`), and the value is an [`Actor`](data-structures.md#actor) object with the actors information. Within each `Actor` object is a field called `state` that is an ipld pointer to a graph that can be entirely defined by the actor.
 
+### Actor Creation
+
+There are two mechanisms by which an actor can be created. By explicitly invoking `exec` on the `Init` actor, and by sending a message to a public key address (secp256k1 or BLS, address types 1 and 3 respectively).
```
gnunicorn: I'd prefer to have a single place of truth and use definitions rather than repeating specifics and their types in multiple places - makes it a lot more error-prone if you ever update something. Thus I'd replace this with:

```suggestion
There are two mechanisms by which an actor can be created. By explicitly invoking `exec` on the `Init` actor, and by sending a message to the [`Address` of a `Public Key`](address.md#protocol-indicator).
```

This way, if we extend the `Public Key`s `Address` is allowed to hold, we don't even have to update this.

PR #284, opened 2019-05-20
clarifications to libp2p requirements and block validation

whyrusleeping:
cc @gnunicorn, these are a couple of the things we discussed in slack.


INTRO.md
```diff
@@ -31,6 +31,6 @@ The Filecoin protocol itself is really a suite of protocols, including:
...
+Filecoin uses [libp2p](https://libp2p.io) for all network communications. libp2p provides transport-agnostic services for peer discovery, naming, routing, pubsub channels and a distributed record store, and there are full or partial [implementations](https://libp2p.io/implementations/) in a number of languages. This spec assumes the use of libp2p and its services and does not specify transport-level details. That said, to be at least minimally compatible with other Filecoin nodes, we recommend supporting at least the [mplex](https://github.com/libp2p/specs/tree/master/mplex) stream multiplexer, and the [secio](TODO, spec in PR) encrypted transport protocols. For more details on the exact wire protocol of libp2p, refer to the [libp2p specs](https://github.com/libp2p/specs).
```
gnunicorn: more spec-speek maybe? `-we recommend supporting at least`  `+it should support at least`

```suggestion
Filecoin uses [libp2p](https://libp2p.io) for all network communications. libp2p provides transport-agnostic services for peer discovery, naming, routing, pubsub channels and a distributed record store, and there are full or partial [implementations](https://libp2p.io/implementations/) in a number of languages. This spec assumes the use of libp2p and its services and does not specify transport-level details. That said, to be at least minimally compatible with other Filecoin nodes, it should support at least the [mplex](https://github.com/libp2p/specs/tree/master/mplex) stream multiplexer, and the [secio](TODO, spec in PR) encrypted transport protocols. For more details on the exact wire protocol of libp2p, refer to the [libp2p specs](https://github.com/libp2p/specs).
```

or maybe even `must`?


```suggestion
Filecoin uses [libp2p](https://libp2p.io) for all network communications. libp2p provides transport-agnostic services for peer discovery, naming, routing, pubsub channels and a distributed record store, and there are full or partial [implementations](https://libp2p.io/implementations/) in a number of languages. This spec assumes the use of libp2p and its services and does not specify transport-level details. That said, to be at least minimally compatible with other Filecoin nodes, it must support at least the [mplex](https://github.com/libp2p/specs/tree/master/mplex) stream multiplexer, and the [secio](TODO, spec in PR) encrypted transport protocols. For more details on the exact wire protocol of libp2p, refer to the [libp2p specs](https://github.com/libp2p/specs).
```

validation.md
```diff
@@ -21,7 +21,7 @@ A valid block:
 
 
 {{% notice info %}}
-Once the block passes validation, it should be added to the local datastore, even in the case where we don't accept it right now. Future blocks from other miners may be mined on top of it and in that case we will want to have it around to avoid refetching. Blocks a certain distance from the current chain height may be dropped (exact number TBD, but blocks that havent been included after several days may be purged).
+Once the block passes validation, it should be added to the local datastore, even in the case where we don't accept it as our best tip right now. Future blocks from other miners may be mined on top of it and in that case we will want to have it around to avoid refetching. Blocks a certain distance from the current chain height may be dropped (exact number TBD, but blocks that havent been included after several days may be purged).
```
gnunicorn: Phrasing suggestions: 
```suggestion
Once the block passes validation, it must be added to the local datastore, regardless whether it is understood as the best tip at this point. Future blocks from other miners may be mined on top of it and in that case we will want to have it around to avoid refetching. Blocks a certain distance from the current chain height may be dropped (exact number TBD, but blocks that havent been included after several days may be purged).
```

INTRO.md
dignifiedquire: maybe add another todo/note that we would like to switch to TLS 1.3 if it becomes ready/stable?

whyrusleeping: eh, I don't think thats necessary. If we switch, we switch and update it in the spec. If we don't, we don't. There are enough random TODOs laying around

PR #286, opened 2019-05-22
add in steps for miners to aggregate signatures

whyrusleeping:


data-structures.md
```diff
@@ -43,6 +43,10 @@ type Block struct {
 	// Height is the chain height of this block.
 	Height Uint64
 
+	// BlockSig is a signature over the hash of the entire block with the miners
+	// worker key to ensure that it is not tampered with after creation
+	BlockSig Signature
```
dignifiedquire: Given that the signature can't sign itself, it would be good to maybe put it at the end and say it signs everything before? Or how exactly do you imagine this to work?

mining.md
```diff
@@ -265,10 +287,12 @@ To create a block, the eligible miner must compute a few fields:
 - `MsgRoot` - To compute this:
   - Select a set of messages from the mempool to include in the block.
   - Insert them into a Merkle Tree and take its root.
+    - Note: Messages with BLS signatures should be included as raw `Message` types, and not `SignedMessage`. Their signatures should be gathered up and aggregated for the `BLSAggregate` field.
```
dignifiedquire: What about ordering them and/or splitting them. Do you imagine these to be just interspersed as the miners choose?

mining.md
```diff
@@ -113,59 +113,81 @@ In order to validate a block coming in from the network at round `N` was well mi
 
 ```go
 func VerifyBlock(blk Block) {
```
dignifiedquire: Is there a `BuildBlock` method somehwere?

mining.md
```diff
@@ -113,59 +113,81 @@ In order to validate a block coming in from the network at round `N` was well mi
...
+	// 6. Verify Message Signatures
+	messages := LoadMessages(blk.Messages)
+	state := GetParentState(blk.Parents)
+
+	var blsMessages []Message
+	var blsPubKeys []PublicKey
+	for i, msg := range messages {
+		if IsBlsMessage(msg) {
+			blsMessages.append(msg)
+			blsPubKeys.append(state.LookupPublicKey(msg.From))
```
dignifiedquire: Do we have a description of `LookupPublicKey` somewhere?

mining.md
```diff
@@ -113,59 +113,81 @@ In order to validate a block coming in from the network at round `N` was well mi
...
+		} else {
+			if !ValidateSignature(msg) {
```
dignifiedquire: Would be ~~nice~~ clearer to call this `ValidateSecp256k1Signature`.

mining.md
```diff
@@ -113,59 +113,81 @@ In order to validate a block coming in from the network at round `N` was well mi
...
+	}
+
+	ValidateBLSSignature(blk.BLSAggregate, blsMessages, blsPubKeys)
+
+	// 7. Validate State Transitions
+	receipts := LoadReceipts(blk.MessageReceipts)
+	for i, msg := range messages {
+		if IsBlsMessage(msg) {
+			blsMessages = append()
+		}
```
dignifiedquire: Copy past error above? empty `append` doesn't seem that useful ;)

mining.md
```diff
@@ -113,59 +113,81 @@ In order to validate a block coming in from the network at round `N` was well mi
...
+
+	// 4. Verify Tickets
+	if !VerifyTickets(blk) {
+		Fatal("tickets were invalid")
+	}
+
+	// 5. Verify ElectionProof
+	randomnessLookbackTipset := RandomnessLookback(blk)
+	lookbackTicket := minTicket(randomnessLookbackTipset)
+	challenge := sha256.Sum(lookbackTicket)
```
dignifiedquire: Why are we using `sha256` here, I am pretty sure we said we are going for blake2 everywhere?

data-structures.md
whyrusleeping: oh yeah, it should go at the end. 👍

mining.md
whyrusleeping: its up to the miner how they want to order messages. I could have a secp message that sends money to an account, and then next a BLS message that spends from that account. Or visa versa. Forcing one to come before the other doesnt work.

mining.md
whyrusleeping: I think we don't have pseudocode for it right now, just the prose description down below. It might be good to have pseudocode for that tho

mining.md
whyrusleeping: not yet, will add.

mining.md
whyrusleeping: oh whoops

mining.md
whyrusleeping: eh, I didnt add this in this PR, it was just there. We can change, i'm not opposed.

whyrusleeping: I made a note in #290 to address this too

PR #288, opened 2019-05-23
Add ability to generate PDF docs

### LINKED ISSUE #266: Generate PDF spec
Probably using pandoc, based on the generated html is the easiest way.

mslipper:
Closes #266. I've attached the outputted PDF to this PR. I've fixed the following problems referenced in the original issue:

1. Code blocks are line-broken correctly.
2. Invisible UTF-8 chars have been removed.
3. List nesting has been fixed.

I replaced hex strings in tables with a code block so that I could manually wrap lines where necessary. Latex is super finnicky about breaking words; since the hex strings have no whitespace it wasn't wrapping them even after trying several different methods. You can see an example of my solution in the 'Actors' section of the PDF.

The `make pdf` command takes one argument - the version of the spec to generate. If it isn't provided, it uses the current Git hash.

PDF: [filecoin-spec-71c2b6b.pdf](https://github.com/filecoin-project/specs/files/3209974/filecoin-spec-71c2b6b.pdf)


mslipper:
All three things are fixed - note that I'll have to rejigger the headers to get it to match the menu exactly. I'm happy to do that, but it'll affect the way headers are laid out on the website so I thought I'd hold off until checking with you first.

Also, the `publish.sh` script now adds a link to the PDF spec on deploy.

mslipper:
The new PDF is uploaded here: [filecoin-spec-d2dd0ad.pdf](https://github.com/filecoin-project/specs/files/3234239/filecoin-spec-d2dd0ad.pdf). The headers are still sans-serif, but I think that matches the site more closely.

There seemed to be a broken symlink to the `drbg-porep` file, which I removed to get things building again - is this expected, or should I debug further?

dignifiedquire:
the symlink was broken in: https://github.com/filecoin-project/specs/pull/219 so removing it is the right thing to do 👍

dignifiedquire:
stylewise 💯  thank you

whyrusleeping:
thanks @mslipper!

PR #289, opened 2019-05-23
Spec for ordering block parent CIDs by ticket

### LINKED ISSUE #285: Define ordering of parent CIDs in block header
Blocks in a tipset are to be ordered by ticket. Define the order of parent CIDs in Block header to match this. Note that this requires access to the parent blocks in order to validate ordering.

A possible extension is to relax the tipset concept to only require blocks in a tipset to have matching sets of parent CIDs, ignoring ordering. The ordering of parents in a block header could then be intentionally undefined.

Slack [thread](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1558433081202800).

I intend to update the spec in concert with implementation. cc @whyrusleeping @gnunicorn

anorth:
As discussed in slack. If this validation proves onerous in implementation we may elect to declare the ordering irrelevant to tipset equivalence, and order lexicographically in the block header instead (at the cost of having two distinct concepts of parent set).

Closes #285

CC @gnunicorn.


validation.md
```diff
@@ -2,13 +2,28 @@
...
-For every block received, the node must validate it before executing it or passing it on. Before a node can validate a block, it first must ensure the block is structurally correct by decoding it (see [block](data-structures.md#block)) and ensuring that no field contains any illegal values. After that the node must check the signature of the block, then moves on to validate the block itself.
+Validation is split into two stages, syntactic and semantic. The syntactic stage may be validated without reference to additional data (see [block](data-structures.md#block)). The semantic stage requires access to the chain which the block extends.
 
-A valid block:
+A node must decode and perform syntactic validation for every block received before passing it on (e.g. in a lipbp2p pubsub validator). A node must perform semantic validation before accepting the block as an extension to its chain.
+
+A syntactically valid block:
+
+- must include a well-formed miner address
+- must include at least one well-formed ticket
```
whyrusleeping: the chain of tickets should be checkable at this point.

anorth:
Hold off on landing this for me: see discussion in #285

expected-consensus.md
```diff
@@ -321,6 +321,8 @@ All valid blocks generated in a round form a `TipSet` that participants will att
 
 The first condition implies that all blocks in a TipSet were mined at the same height (remember that height refers to block height as opposed to ticket round). This rule is key to helping ensure that EC converges over time. While multiple new blocks can be mined in a round, subsequent blocks all mine off of a TipSet bringing these blocks together. The second rule means blocks in a TipSet are mined in a same round.
 
+The blocks in a tipset are canonically ordered by ticket.
```
whyrusleeping: I had thought we settled on making the order not matter at all?

anorth: This line refers to the logical ordering, the order of processing in consensus state transitions, rather than the "syntactic" requirement (or not) for ordering in the block header. Can I clarify that in the text more?

For the block header, I was waiting for your call in #285 whether to order by CID or not at all.

whyrusleeping: Ah, okay. I think that what we want to say is that there is no ordering enforced for tipsets, but during state computation, we apply the transitions of blocks in a tipset ordered by their tickets.

anorth:
@whyrusleeping ready to land

mining.md
```diff
@@ -303,6 +301,9 @@ To create a block, the eligible miner must compute a few fields:
 - `ReceiptsRoot` - To compute this:
   - Apply the set of messages selected above to the parent state, collecting invocation receipts as this happens.
   - Insert them into a Merkle Tree and take its root.
+- `Timestamp` - A Unix Timestamp generated at block creation. We use an unsigned integer to represent a UTC timestamp (in seconds). The Timestamp in the newly created block must satisfy the following conditions:
```
whyrusleeping: curious why the timestamp was moved. It's not a big deal, but it feels a little odd now being between the message stuff, and the signature aggregation (Which is kinda part of the messages flow). So this now reads "apply all the messages, collect their receipts, GO SET A TIMESTAMP, then take all the signatures from those messages and aggregate them together"

anorth: I moved it to align with the field order in the [block header](https://github.com/filecoin-project/specs/blob/master/data-structures.md#block). I agree its a bit awkward but was aiming for consistency. I'm happy to undo this here, but opine that it should be re-ordered in the block header to come earlier.

anorth:
(P.S. I'm not authorized to merge, so do land it for me if ok, or push over it)

anorth:
Please merge me @whyrusleeping

whyrusleeping:
Sorry for the delay, was a busy week last week

PR #295, opened 2019-05-28
Clean up the networking stack

gnunicorn:
Working on the networking stack, refactoring, cleaning up and - most importantly - consolidating information into single places with a concise and clear language.

Also added the healthy peerset concept - as a suggestion - I discussed with @pooja last week.


data-propagation.md
```diff
@@ -2,13 +2,16 @@
 
 The filecoin network needs to broadcast blocks and messages to all peers in the network. This document details how that process works.
 
-Both blocks and messages are propagated using the gossipsub libp2p pubsub router. The pubsub messages are authenticated. For blocks, the pubsub hop validation function is set to check that the block is valid before re-propagating. For messages, a similar validity check is run, the signature must be valid, and the account in question must have enough funds to cover the actions specified.
+Both block headers and message references are propagated using the [gossipsub libp2p pubsub router](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub). Every full node must implement and run that protocol. All pubsub messages are authenticated and must be validated before being propagated further.
```
whyrusleeping: s/references//

Whole messages are broadcast through the pubsub layer on the messages channel. the blocks channel broadcasts the block headers and message cids.

bitswap is only used in the cases where we missed pubsub messages (around joining the network, or if hiccups occur, such as a miner mining a message that they didnt propagate)

data-propagation.md
```diff
@@ -23,12 +26,20 @@ type BlockMsg struct {
...
+Messages are propagated over the libp2p pubsub channel `/fil/messages`. On this channel, every [serialised `Message`](data-structures.md#messages) ist announced.
+
+Upon receiving the message, its validity must be checked: the signature must be valid, and the account in question must have enough funds to cover the actions specified. If the message is not valid it should be dropped and must not be forwarded.
 
-The pubsub validation function for messages checks that the content of each pubsub message on this topic is, first, under the maximum size limit for a message, and then that it is a properly constructed message. (TODO: discuss checking signatures and account balances, some tricky bits that need consideration). If an invalid message is received from a peer, that peer should be marked as potentially bad.
+{{% notice todo %}}
+discuss checking signatures and account balances, some tricky bits that need consideration. Does the fund check cause improper dropping? E.g. I have a message sending funds then use the newly constructed account to send funds, as long as the previous wasn't executed the second will be considered "invalid" ... though it won't be at the time of execution.
```
whyrusleeping: 👍 this needs discussion, the scenario you describe here should not result in the second message being dropped.

We should have a spec for the message pool, and this check here should just be 'propagate the message if it gets successfully added to the message pool'

network-protocols.md
```diff
@@ -1,14 +1,14 @@
-# Filecoin Network Protocols
+# Filecoin Custom Network Protocols
```
whyrusleeping: I wouldn't call them 'custom', they are *the* Filecoin network protocols.

networking.md
```diff
@@ -0,0 +1,92 @@
...
+ - [gossipsub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) ([for data announcements](./data-propagation.md))
+ - [bitswap] ([for data exchange](./))
+ - Filecoin specific Protocols:
+   - Hello Handshake
+   - StorageDeal
+   - BlockSync
+
+## Transports, Streams & Encryption
+
+Connections between nodes are end-to-end-encrypted and -authenticated, thus every node has a `peerId` associated with it. Any node can also support unencrypted or unauthenticated connection request, but it is not required to do so.
```
whyrusleeping: I would leave off the bit about supporting unencrypted/unauthenticated connections. Thats a debug mode and is not to ever be used in production.

networking.md
```diff
@@ -0,0 +1,92 @@
...
+
+Filecoin protocols are run as multiple streams over a single connection. Only one connection should exist between two nodes at any time, multiple streams within that connections must be used to exchange different protocols.
```
whyrusleeping: Maybe add a note at the end of this, "This is all abstracted away and handled by libp2p, Filecoin implementations with a correct libp2p library need not worry about this"

networking.md
```diff
@@ -0,0 +1,92 @@
...
+
+This document doesn't make any assumptions on which transports and multiplexing features a particular node offers, however it must at least be able to respond and communicate through ABC AND XYZ
+
+{{% notice todo %}}
+Define minimum requirements in terms of network/e2ee-protocols and multiplexing mechanism. Should we directly jump to yamux?
```
whyrusleeping: We should go with secio and mplex for now, if tls 1.3 becomes mature enough all around, we could swap it for secio, but we're not there yet. Mplex also tends to be more performant, and drastically easier to implement than yamux (downside being that it doesnt have great backpressure).

data-propagation.md
```diff
@@ -23,12 +26,20 @@ type BlockMsg struct {
 
 The array of message cids must match the `Messages` field in the block when used to construct a [sharray](sharray.md).
 
-Each Filecoin node sets a 'validation function' for the blocks topic that checks that the block is properly constructed, its ticket is valid, the block signature is valid, the miner is a valid miner, and the block is a child of a known good tipset. (TODO: clarify which of these checks are needed, any slowness here impacts propagation time significantly, this is not a full validity check) If an invalid block is received, the peer it was received from should be marked as potentially bad (TODO: we could blacklist peers who send bad blocks, maybe need support from libp2p for this?)
+Every `BlockMsg` received must be validated before being propagated again. If validation fails, it must not be propagated. The validation must check that is its ticket is valid, the block signature is valid, the miner is a valid miner, and the block is a child of a known good tipset.
+
+
+{{% notice todo %}} 
+clarify which of these checks are needed, any slowness here impacts propagation time significantly, this is not a full validity check.
+{{% /notice %}}
```
anorth: Perhaps reference `validation.md`, where I think I've answered these questions in https://github.com/filecoin-project/specs/pull/289/files#diff-eff160cdcda8f026fba936da7f3fea15

bootstrap.md
```diff
@@ -1,5 +1,10 @@
 # Filecoin Bootstrapping Routine
 
+
+{{% notice warning %}}
+Outdated. The link that got you here should probably link to [`networking.md`](./networking.md) instead.
+{{% /notice %}}
+
```
anorth: I'd prefer we delete the content below that we know is outdated/duplicated. People already familiar with the spec structure may continue to come here but miss this message, leading to confusion.

gnunicorn: wasn't sure about it and there is still stuff, that isn't fully moved yet, but I agree and will probably remove the file before bumping to merge it.

data-propagation.md
gnunicorn: > the scenario you describe here should not result in the second message being dropped.

I agree, but the currently mentioned validation function would do that.

> We should have a spec for the message pool, and this check here should just be 'propagate the message if it gets successfully added to the message pool'

What is the check to be put into the message pool though? Because if that is a light-weight test, yes, but if it tried to do a similar introspection/contextual test, we'd just move the point when it was dropped and it would still mean significant delays before we propagate ourselves.

---

I'd rather vote for super slim structure-only validation check before propagating. however, if the client reports back that a transaction failed, it could stop propagating it from a later point. Not sure, if that would actually have any useful in practice.

network-protocols.md
gnunicorn: Don't really care. Just, we are also using "standard" protocols ... in FIlecoin ... making them "Filecoin Network Protocols", too ... or?

how about 
```suggestion
# Filecoin Specific Network Protocols
```

?

whyrusleeping: Yeah, specific sounds good to me

networking.md
gnunicorn: > I would leave off the bit about supporting unencrypted/unauthenticated connections.

Wasn't sure if that was something we intended to support - libp2p certainly does allow that in general and I wasn't sure if that might be useful for light-client scenarios or unnecessary. I am fine with dropping it.

whyrusleeping: I think its worth taking a hardline stance on 'all connections should be encrypted and authenticated'. leaving the option open for unencrypted comms just sounds like its asking for downgrade attacks.

networking.md
gnunicorn: I am a little hesitant to add sentences like this for two reasons: 1. with libp2p being a modular stack, it doesn't really state whether muxing is implemented, nor is it enforced by the library and 2. there is no real definition of what constitutes the "correct libp2p library"/implementation -- especially as we don't currently run cross-implementation-integration tests...

How about
```suggestion
Filecoin protocols are run as multiple streams over a single connection. Only one connection should exist between two nodes at any time, multiple streams within that connections must be used to exchange different protocols. This is usually handled from within the libp2p stack if configured properly.
```

networking.md
```diff
@@ -0,0 +1,92 @@
...
+   - StorageDeal
+   - BlockSync
+
+## Transports, Streams & Encryption
+
+Connections between nodes are end-to-end-encrypted and -authenticated, thus every node has a `peerId` associated with it. Any node can also support unencrypted or unauthenticated connection request, but it is not required to do so.
+
+Filecoin protocols are run as multiple streams over a single connection. Only one connection should exist between two nodes at any time, multiple streams within that connections must be used to exchange different protocols.
+
+This document doesn't make any assumptions on which transports and multiplexing features a particular node offers, however it must at least be able to respond and communicate through ABC AND XYZ
```
gnunicorn:
```suggestion
This document doesn't make any assumptions on which transports and multiplexing features a particular node offers, however it must at least be able to respond and communicate through secio using mplex channels.
```

networking.md
gnunicorn:
```suggestion
Connections between nodes are end-to-end-encrypted and -authenticated, thus every node has a `peerId` associated with it.
```

networking.md
whyrusleeping: That sounds good to me!

whyrusleeping:
@gnunicorn let me know when you're ready for a review (just switch the PR out of WIP mode)

data-propagation.md
gnunicorn: we should move this discussion to the validation-section.

data-propagation.md
gnunicorn: I removed it here and put it into the validation.md for this PR. we can link to that once both PRs converge.

content/menu/index.md
```diff
@@ -14,7 +14,7 @@ headless: true
   * [Block Validation]({{< relref "/docs/validation.md" >}})
 * [Network]({{< relref "/docs/networking.md" >}})
   * [Data Propagation]({{< relref "/docs/data-propagation.md" >}})
-  * [Custom Protocols]({{< relref "/docs/network-protocols.md" >}})
+  * [FC specific Protocols]({{< relref "/docs/network-protocols.md" >}})
```
anorth: FC -> Filecoin? I don't think we abbreviate it anywhere else.

sync.md
```diff
@@ -1,284 +0,0 @@
-# Chain Syncing
-
```
dignifiedquire: where does all the information from this file live now? seems to be a lot if it was just dropped

content/menu/index.md
gnunicorn: Yeah,  I wasn't sure about this one. We dropped "Custom" in the header and called it "Filecoin specific", but in the sidebar this gets to long. So maybe dropping the reference?
```suggestion
  * [Specific Protocols]({{< relref "/docs/network-protocols.md" >}})
```

But that seems weird somehow. So maybe rather?

```suggestion
  * [Special Protocols]({{< relref "/docs/network-protocols.md" >}})
```

networking.md
```diff
@@ -0,0 +1,88 @@
...
+In order to do that a node must be "bootstrapped" when starting up. At its simplest, bootstrapping can be thought of as secure network joining. It must enable a new node to get the latest head/chain that the majority of the network power is mining off and must be both:
+
+- Robust — able to do so in adversarial networks (so long as honest nodes make up the majority of the network).
+- Consistent — able to reliably get the canonical chain (even given poor latency or state loss in our initial peer set).
+
+The specific bootstrapping method and heuristic - or a mix of them - is left up to the node implementation, but we provide some example ideas of how this could be done and what aspects should be taken into consideration for that.
+
+### Trusted bootnodes
+
+As with many other networks, trusted entities - the Filecoin foundation, large exchanges, good actors in the community, etc)- might publish a set of bootnodes one can connect to in order to establish network connections.
```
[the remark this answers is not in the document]
gnunicorn: replacing the ending `)` with `-`
```suggestion
As with many other networks, trusted entities - e.g. the Filecoin foundation, large exchanges, good actors in the community, etc - might publish a set of bootnodes one can connect to in order to establish network connections.
```

sync.md
gnunicorn: Well, 230 out of the 300 lines were actually code, describing interfaces and even implementation specifics, that shouldn't be in a spec in the first place, as they are left to the implementers (**if at all** they could be put in an appendix of a reference implementation). Same goes for the operating in two modes - if you want to understand it as such, the implementer may, but that isn't strictly necessary. 

Rather I put a brief [syncing description](https://github.com/filecoin-project/specs/pull/295/files/62fa09f0648c12b04878c5f06d2bef8dbf48cb53#diff-f9719a6664c477d6b213f13bc09ac677R34) inspired by this file about sync in the networking.md but cut it down to the essentials and constraints - e.g. while importing, you must not mine. I found most other information in this file to be redundant to what is already said at other, more appropriate places. But please let me know if I missed anything.

gnunicorn:
updated to latest master. ready for review @whyrusleeping

sync.md
dignifiedquire: thanks, will go through it again

content/menu/index.md
dignifiedquire: I think Specific works well enough

network-protocols.md
```diff
@@ -18,7 +18,7 @@ Note: Implementations should limit the maximum number of bytes read during the `
 
 # Hello Handshake
 
-> The Hello protocol is used when two filecoin nodes initially connect to eachother in order to determine information about the other node.
```
dignifiedquire: Hello handshake is defined twice, could you clean that up while you are in here?

networking.md
```diff
@@ -0,0 +1,88 @@
+# Filecoin Network
+
+The filecoin network is built using libp2p building blocks, as transports and protocols, as well as some additional filecoin specific protocols as outlined in [Network Protocols](./network-protocols.md).
+
+## Required Protocols
+
+Every full node must support the following libp2p protocols:
+
+ - [gossipsub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) ([for data announcements](./data-propagation.md))
+ - [bitswap] ([for data exchange](./data-propagation.md))
```
dignifiedquire: Missing link to bitswap

networking.md
```diff
@@ -0,0 +1,88 @@
...
+ - Filecoin specific Protocols:
+   - Hello Handshake
+   - StorageDeal
+   - BlockSync
```
dignifiedquire: What about the other protocols defined?

networking.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Establishing connections
+
+When a filecoin node connects to the network, it may connect to multiple nodes at once. When a new node connects, it must be greeted with the `HelloMessage` through the hello-handshake protocol first. Only after that message has been sent can other streams be opened. If a node receives a `HelloMessage` with a GenesisBlock it doesn't support, it must immediately close the entire connection to that peer.
+
+If the node learns through that handshake about newer blocks it should request these immediately to sync up their chain.
+
+### Syncing
+
+Whenever a node learns about a new `BlockHead` it should attempt to import the block - let that be through the `HelloMessage` or through the [block pubsub protocol](data-propagation.md#block-propagation). For that it may fetches the ancestors of that block through bitswap until it is fully caught up. Importing in this context refers to the node confirming that the block is valid, as described in [Validation](./validation.md) and storing it locally. This mode is known as "syncing". During "syncing" the same node may not mine blocks.
```
dignifiedquire: What about „caught up“ mode? Is that not a thing anymore?

networking.md
```diff
@@ -0,0 +1,88 @@
...
+
+{{% notice info %}}
+Implementers are encouraged to offer and experiment with more protocols and transports, as long as the above mentioned requirements of ee2e, authentication and muxing are still met. In particular the authors encourage to look into TLS1.3 and QUIC as libp2p protocol/transport layers.
+{{% /notice %}}
+
+### Establishing connections
+
+When a filecoin node connects to the network, it may connect to multiple nodes at once. When a new node connects, it must be greeted with the `HelloMessage` through the hello-handshake protocol first. Only after that message has been sent can other streams be opened. If a node receives a `HelloMessage` with a GenesisBlock it doesn't support, it must immediately close the entire connection to that peer.
+
+If the node learns through that handshake about newer blocks it should request these immediately to sync up their chain.
```
sternhenri: I think the situation here may actually be a bit more complex: generally, some "newer blocks" may be invalid (not in themselves but in leading back to genesis -- i.e. DoS) or there may be other required checks to be performed when establishing a new connection with a node, vs receiving blocks from an existing connection. I would defer here: replace "it should request these immediately to sync up their chain." with "it should initiate the syncing sequence" and we'll defer details to there (as is currently done anyways)

networking.md
sternhenri: typo "it may fetche**es**".

networking.md
```diff
@@ -0,0 +1,88 @@
...
+- Robust — able to do so in adversarial networks (so long as honest nodes make up the majority of the network).
+- Consistent — able to reliably get the canonical chain (even given poor latency or state loss in our initial peer set).
+
+The specific bootstrapping method and heuristic - or a mix of them - is left up to the node implementation, but we provide some example ideas of how this could be done and what aspects should be taken into consideration for that.
+
+### Trusted bootnodes
+
+As with many other networks, trusted entities - the Filecoin foundation, large exchanges, good actors in the community, etc)- might publish a set of bootnodes one can connect to in order to establish network connections.
+
+Keep in mind that, because these could easily be DoS'ed, a) large set of different entities should be pre-shipped but only a few should be tried at a time b) they might drop connections rather quickly after establishing (and ban that peer for a short period of time) in order to serve more nodes. Thus you want to make sure to send a discovery request over the bootnodes to have nodes to connect to.
```
sternhenri: may be worth a reminder of what a discovery request is here

networking.md
```diff
@@ -0,0 +1,88 @@
...
+
+A node should not attempt to connect to more than 10 bootnodes at a time. And should diversify its set of nodes as quickly as possible.
```
sternhenri: number seems a bit arbitrary. May change to "should not attempt to connect to more than a handful of nodes (e.g. 10) at a time: these will be used to quickly diversify a set of nodes to build a network off of"

networking.md
```diff
@@ -0,0 +1,88 @@
...
+## Establishing a network
+
+Every node should aim to establish a set of stable connections to the network in order to stay up to date with latest changes and block propagation - aka to sync - as well as to participate in the processes of the network - e.g. mining.
+
+In order to do that a node must be "bootstrapped" when starting up. At its simplest, bootstrapping can be thought of as secure network joining. It must enable a new node to get the latest head/chain that the majority of the network power is mining off and must be both:
+
+- Robust — able to do so in adversarial networks (so long as honest nodes make up the majority of the network).
+- Consistent — able to reliably get the canonical chain (even given poor latency or state loss in our initial peer set).
+
+The specific bootstrapping method and heuristic - or a mix of them - is left up to the node implementation, but we provide some example ideas of how this could be done and what aspects should be taken into consideration for that.
```
sternhenri: may be worth quickly stating the following: "should be taken into consideration for that**, such as relying on both new nodes and new connections to ensure a node is connected to both a reliable and representative slice of the network**."

Which sets up for the following sections

networking.md
```diff
@@ -0,0 +1,88 @@
...
+
+A node attempts to always hold a certain amount of connections ("slots") to the network -- we recommend 25-50 on an on-the-shelf system. Whenever a new node connects, it can check their previously stored reputation or assign a default value and if that reputation is higher than the lowest currently connected nodes, may replace that connection (and thus drop the lowest quality connection) or otherwise refuse to take that connection.
+
+In this system a node might also only record a strong decrease in reputation but not drop a connection to a peer directly even upon a strong violation, because the node may still be more useful than others. It is better to stay connected to a crappy network than to no network. However, this doesn't not free the node from still adhering to the spec itself - it should not forward said violation or its connection might still be rightously dropped. 
+
+We also recommend to regularly check the ranking and drop and clear up the lowest 10% of slots, leaving 5% open for incoming connections and fill up the other 5% by connecting to other nodes it can find through peer discovery.
```
sternhenri: I would pull this notion into an above section: it's a good heuristic even without the notion of "usefulness" and is worth duplicating ie, "We recommend nodes regularly diversify their peerSet by dropping a section of their peers (e.g. 10%) and making new connections."

networking.md
```diff
@@ -0,0 +1,88 @@
...
+
+All this to creating a local view of for that node most useful connections to the network. The reputation may be stored permanently and be available between restarts - thus providing a neat bootstrap start list, too.
```
sternhenri: typo "to **creating**"

sternhenri:
Great work! I agree a lot of this has been simplified with relatively little missing. A couple of missing things from me I would want to see ahead of approving:
- I think we need to distinguish between the regular interplay between sync/caught up modes, and what happens on initial connection when you have no data to start (i,e, whether or not to address the bootnodes, how far to seek out peer expansion, etc.)
- We need to bulk up the block validation piece, as well as tipset validation: a todo right now, but arguably the core part of the protocol here. how do they each work? What precise actions must happen for a block to be validated in FIL and what can be left to impl?

I think more generally there is a q of how descriptive we want the spec to be. I expect @gnunicorn, @anorth and @whyrusleeping can answer this together, but I'm all for moving away from the more normative version of spec-writing so long as devs feel it is as descriptive as the old v. We don't have example code anymore (good) but do we need to provide more defaults for folks?

I wonder if we've abstracted a tad too much, thereby forcing developers to do too much wheel reinventing (e.g. looking into Brahms or Kademlia peer discovery). Should we nonetheless provide a default set of impl choices (clearly marked as such) or is that perhaps the purpose of Go-fil (beyond being one impl)?

network-protocols.md
```diff
@@ -1,14 +1,14 @@
...
-All filecoin network protocols are implemented as libp2p protocols. This document will assume that all data is communicated between peers on a libp2p stream.
 
 ## CBOR RPC
 
 Filecoin uses many pre-existing protocols from ipfs and libp2p, and also implements several new protocols of its own. For these Filecoin specific protocols, we will use the CBOR RPC protocol format, defined below.
 
 This format consists of series of [CBOR](https://tools.ietf.org/html/rfc7049) serialized objects. Whenever a filecoin protocol says "send X", it means "cbor serialize the object X, then write the serialized bytes".
 
-
 ## Hello Handshake
```
gnunicorn:
```suggestion
```

network-protocols.md
```diff
@@ -18,7 +18,7 @@ Note: Implementations should limit the maximum number of bytes read during the `
 
 # Hello Handshake
```
gnunicorn:
```suggestion
```

networking.md
gnunicorn: `[for data exchange](./data-propagation.md)` links generally don't seem to since the hugo setup anymore. @dignifiedquire any idea why or how to do it instead (without breaking regular markdown links?)

networking.md
gnunicorn: In a previous discussion we agreed to not go to much into the specifics of how discovery works in this spec, as it isn't really filecoin specific nor is that a necessity for it to work. Thus the paragraph two below explains options, and this is kept vague on purpose. I added a few words to remind everyone, what a discovery is about though.

networking.md
gnunicorn: Yes, the number it sort-of arbitrary, but not newly defined. It is the combination of the [first 5 bootnodes](https://github.com/filecoin-project/specs/pull/295/files#diff-caf96f80dec81066bbd3e06dfe6b4d25L54)  and the expansion of [another 5 slots via kademlia](https://github.com/filecoin-project/specs/pull/295/files#diff-caf96f80dec81066bbd3e06dfe6b4d25L78) as previously stated.

I stated that specific number here, to allow for efficient and clearly defined spam-protection rules for anyone running the bootnodes farm: if anyone sees 10 connects over their entire cluster they are totally in their right to refuse any further connection, because that client doesn't handle according to spec. It also makes more clear the expectation to implementers that they shouldn't _and can't_ rely on just bootnodes to create their networks (we've seen that behaviour in the past, where clients wouldn't implement queries but just used bootnodes and then swamped the bootnodes and stalled the network)

networking.md
gnunicorn: I was thinking that, too, but I am missing the background to consider it a "good heuristic". I only know this idea of dropping these connection from papers that have some form of ranking or determination of what the "lowest x%" entails. I know of "randomly drop x% of connections" only in the context of brahms, but that is a much lower number and has a much more complex structure to ensure this doesn't degrade the network over time (like certain minimal numbers of connections to always hold and a local network view to draw from).

I am not sure that generally dropping 10% of connections and _attempting_ to fill these slots with new connections (as you aren;'t guaranteed they will succeed) is not - over time - leading to you having just less and less connections until you dropped the last _good_ connection and then ... are on your own ...

networking.md
gnunicorn: I agree with the stated problems, but am not sure how this is relevant here, but it clearly becomes cloudy with us being so specific here. I'll rephrase the last part to leave it more open to `sync` how that actually takes place. It was for sure not meant to "inject" this block into a running sync process.

networking.md
gnunicorn: @dignifiedquire I never quite understood the idea of distinct "modes", it for sure isn't how we have implemented syncing in substrate. I picture this might be a useful architecture in implementing it in certain frameworks or context, but from a spec perspective a node is a node and there aren't really distinct "modes" in which it is but just certain requirements that is has to adhere by when doing certain things. The only thing I found was the there obviously shouldn't be any "mining" while importing. But that could easily be implemented with a lock rather than understanding it as different modes. And that is much easier than thinking about all the cases of transitions between the modes that could occur IMHO.

Or is there anything in particular what I missed about these modes and that we are missing in this version?

dignifiedquire: Afaict there is not more to these modes, but @whyrusleeping might have some additional thoughts here, as he was the one who added them in the first place I believe.

gnunicorn:
thanks for the feedback @sternhenri, highly appreciated!

> I think we need to distinguish between the regular interplay between sync/caught up modes

Why?

I see why thinking of this as "modes" might make sense in certain context or architectures for implementation but there is no necessity for that from the spec perspective - and we in substrate certainly don't implement it that way: without a global view of the network, and as we don't know when a new block comes around, there is never a fully "caught up" state. We could always connect to a node that has a range of newer heads and it turns out we were never caught up in the first place. Or the other way around, even after being disconnected for quite a while, nothing happened and thus we are already caught up without even knowing it. 

And adding specific modes, you'd also have to define all the potential events for each one of the modes as well as all transitions between them. For example, when we receive a new header over pubsub, does this send us into "sync"-mode again? What if we receive another message over pubsub, do we first switch back out into 'caught up`-mode and then transition back? Or is receiving message from pubsub part of being in sync-mode? If so, when does it switch back out into `caught up`-mode? And what really does it mean to be in either mode? Are there different expectations what I do - from the outside ?

In async protocols I'd much rather like to think of activities that must occur upon events and state requirements or constraints directly attached to the activity. Like in this case, where it clearly states: while we sync (which is a clearly defined activities with clean ends, started by two types of events), the client should not be mining.

Aside from that, there was nothing else I could identify that makes these modes distinct from one another. And that as that is perfectly reflected in the spec without any "modes", I'd prefer to do that.

Or did I miss something? Was there any other behaviour or expectation implied with either mode?

----

>and what happens on initial connection when you have no data to start (i,e, whether or not to address the bootnodes, how far to seek out peer expansion, etc.)

You always have  1. a set of bootnodes; 2. the gensis block; 3. the chain-definition (as in, can identify if another node is running on the same chain), right? That and the handshake with any bootnode is enough to catch up as described. I am not sure what you are referring to - warp sync is out of scope for this version.

Or do you mean the expectation of the sequence of events that unfold? That is left up to the developers/implementers IMHO: a spec defines the protocol that nodes can speak to one another but it unless absolutely necessary (like the `Hello Handshake`) I would not constraint the order in which things must happen. One client might find it more fruitful to first do an mDNS query and attempt to sync locally, others fetch the headers and parallelize fetching of messages and blocks accros the entire network, others pick the strongest/fastest connection to fetch as much info from them as possible, yet others (on lower bandwidth devices) round-robin a request at a time -- that should all be valid behaviour IMHO, as this is all tailored to its circumstances. We shouldn't pick how they implement that.

----

> We need to bulk up the block validation piece, as well as tipset validation: a todo right now, but arguably the core part of the protocol here. how do they each work? What precise actions must happen for a block to be validated in FIL and what can be left to impl?

There is a PR pending by @anorth that attempt to address these:
https://github.com/filecoin-project/specs/pull/289/files#diff-eff160cdcda8f026fba936da7f3fea15

----

> I think more generally there is a q of how descriptive we want the spec to be. I expect @gnunicorn, @anorth and @whyrusleeping can answer this together, but I'm all for moving away from the more normative version of spec-writing so long as devs feel it is as descriptive as the old v. We don't have example code anymore (good) but do we need to provide more defaults for folks?
> 
> I wonder if we've abstracted a tad too much, thereby forcing developers to do too much wheel reinventing (e.g. looking into Brahms or Kademlia peer discovery). 

I am generally against re-iterating specs or code from elsewhere in a spec, but rather reference them. They are in the best case superficial and incomplete and in the worst outright outdated and wrong. Specs change much slower than code. And just considering that the libp2p-kademlia-dht spec isn't even there yet, I doubt that we will make a better job writing that spec in here than they will putting their entire focus to it.

That said, I am very much in favor of adding an appendix with "known at the time of writing" further reading and known implementations, also of components and libraries to use _and not to use_ (and why). What do you think of that compromise?

whyrusleeping:
@gnunicorn Sorry for the delay, had a real busy week last week. 

I think the main thing that the different 'modes' are useful for is to make statements about what protocols and activities should be done when. If you are obviously far behind and need to sync a large number of blocks, you want to ensure that certain other processes the node might be doing don't happen (mining, payment channel validation, etc) and you'll likely want to use a different method for syncing (agree that this is up to the implementer, but if you don't spec an 'accepted' mechanism, you end up with 20 different ways to sync the chain, some of which only work for a certain client, and many that arent interoperable like is the current state in ethereum).

data-propagation.md
```diff
@@ -23,12 +26,15 @@ type BlockMsg struct {
...
+Every `BlockMsg` received must be validated [through the basic check](./validation.md#basic-check) before being propagated again. If validation fails, it must not be propagated.
 
-TODO: we should likely be smarter here and track which messages we could send along with each block to improve propagation time
 
 ## Message Propagation
 
-Messages are propagated over the libp2p pubsub channel `/fil/messages`. The message is [serialized](data-structures.md#messages) and the raw bytes are sent as the content of the pubsub message.
+Messages are propagated over the libp2p pubsub channel `/fil/messages`. On this channel, every [serialised `Message`](data-structures.md#messages) ist announced.
```
whyrusleeping: s/ist/is/

networking.md
whyrusleeping: +1 to picking a number for now.

networking.md
whyrusleeping: I'd link these into the relevant network-protocols.md portions

gnunicorn: as said, we have a linking issue, waiting for that to be resolved.

gnunicorn:
@whyrusleeping no worries, all good.

> If you are obviously far behind and need to sync a large number of blocks, you want to ensure that certain other processes the node might be doing don't happen (mining, payment channel validation, etc)

Which ones specifically, because that was not at all clear from the current spec. And what would that mean for switching between these modes? Like, would I have to drop my payment channel protocol while I sync 10 blocks? That's why I'd rather have specific constraints, requirements and dependencies layed out and made explicit.

> and you'll likely want to use a different method for syncing

Wait. What? This spec only gives you two potential ways to sync: pubsub and query via bitswap. What "other method" are you talking about?

> you end up with 20 different ways to sync the chain, some of which only work for a certain client, and many that arent interoperable like is the current state in ethereum

I can't follow. Are you talking about warp-sync? That would still very much be a feature any client could implement as they act outside of this spec to catch up in different ways. I would not forbid that, even if only a subset of nodes provides that. That type of innovation is exactly why you want to have multiple nodes and different teams, if they all built the exact same thing, there would be no point in that. And whatever emerges as a common ground, should become the standard of the next iteration of the spec.

Or are you talking about something else? If so, what are you referring to? And how do they relate to this spec, which _does define clear ways to sync_?

sternhenri:
Sorry to add my piece here, but just to echo a bit of @whyrusleeping 's comment. I expect you're right we don't explicitely need the modes @gnunicorn, but we should in that case give guidance with regards to when sync should be done passively alongside mining, and when the state moves far enough out of sync that sync should become the node's sole/main function. I'm ok thinking this is impl choice, but we should lay out the tension between wasted work and DoSability (how I see it), and give some heuristics for folks...

That said, taking a step back, I believe this could just be a one sentence add, rather than a major change.

anorth:
I am in general 👍 for removing much of the implementation detail from the spec, especially where not strictly necessary for node interoperability, so thank you. It's a bit hard for me to tell whether this patch removes anything that is in fact spec-required, e.g. `ForkLengthThreshold` or the concept of punctuality; these were buried before, so this change is good in that *if* those concepts were included it would be much more clear that they are strictly required.

@sternhenri or @whyrusleeping I think there is some value preserving some content like this somewhere as a guide specifically for go-filecoin implementers. E.g. if you've thought about the network bootstrap process a bunch and have a recommendation for a stable process, that could save implementers some confusion. A [design doc](https://github.com/filecoin-project/designdocs) would be perfect.

whyrusleeping:
> That's why I'd rather have specific constraints, requirements and dependencies layed out and made explicit.

Agree, we should be more explicit here. In general though, i'm sure we can agree that at some degree of being out of sync, miners should not attempt to mine, payment channel updates should not be accepted, and other things that depend on an up to date view should not be done. 

> Wait. What? This spec only gives you two potential ways to sync: pubsub and query via bitswap. What "other method" are you talking about?

What i'm saying is that you don't want to treat the initial bootstrap chain download sync the same as you do receiving the most recent block from the network while (mostly) caught up. 

I think what we should do here is take the pseudocode out of sync.md, and put it into an 'appendix' of some sort that we reference, label it as an implementation guide for now, and drop it later if we still feel its not useful.

data-propagation.md
```diff
@@ -23,12 +26,15 @@ type BlockMsg struct {
...
+Every `BlockMsg` received must be validated [through the syntactical check](./validation.md#syntactical-validation) before being propagated again. If validation fails, it must not be propagated.
 
-TODO: we should likely be smarter here and track which messages we could send along with each block to improve propagation time
 
 ## Message Propagation
 
-Messages are propagated over the libp2p pubsub channel `/fil/messages`. The message is [serialized](data-structures.md#messages) and the raw bytes are sent as the content of the pubsub message.
+Messages are propagated over the libp2p pubsub channel `/fil/messages`. On this channel, every [serialised `Message`](data-structures.md#messages) ist announced.
```
gnunicorn:
```suggestion
Messages are propagated over the libp2p pubsub channel `/fil/messages`. On this channel, every [serialised `Message`](data-structures.md#messages) is announced.
```

gnunicorn:
Updated to latest master, fixed some typos, linked the syntactical validation

---

> i'm sure we can agree that at some degree of being out of sync, miners should not attempt to mine,...
yes, that is being said already [in the current version of this PR](https://github.com/filecoin-project/specs/pull/295/files#diff-f9719a6664c477d6b213f13bc09ac677R36).

One _could_ argue that an on-going mining process doesn't have to be stopped for a single block to be imported, because forks on the same height can be merged while they can't if they have diverged a lot further. However, even in that case, that this new block is far behind, it could still have higher value as calculated by the weight function, so I'd argue you should probably continue building the block, even if we import 10 or more blocks and rather run the weight function before announcing their new block if they notice they have diverged. That would also improve the speed of the entire network, as more mining could/would happen in parallel. 

Therefore I would understand this constraint as one that "suggest": you may not want to _start_ mining while there is an sync happening, because you are wasting resources and it increases the likelyhood of forks. But it doesn't have to be a "must", strictly speaking. We may want to rephrase that.

> ...payment channel updates should not be accepted,
 Well, that is a risk of the node, no? Sure, most nodes wouldn't want to accept payment transfers before they know they are caught up (and will probably not respond before the sync succeeded), but higher volume clients might be configured to eat the risk and accept non-the-less, because it might be worth it. I think we might want to add a phrase in the payment channel spec to keep that in mind that the state could change if a new block is synced.

----

> What i'm saying is that you don't want to treat the initial bootstrap chain download sync the same as you do receiving the most recent block from the network while (mostly) caught up.

Define "mostly". 

Because even if I just missed one block in between, from the perspective of the code, I have to do the same as when I missed 100: I need to query them, going through to the last parent I know of, fetch all of the blocks in order (oldest first) and import them. From the current protocols, I don't see any other way of being able to do that - is there a plan for a warp- or mass-sync? Is that what the graph query is supposed to be for?

Though I agree that there is probably a point at which it just is too high risk to accept payments or just not worth it to mine blocks because the costs outweigh the benefits, I wouldn't know where that number is and would leave that to the implementations or, even, to a tuneble parameter that operators should be able to tweak around with. 

----

> I think what we should do here is take the pseudocode out of sync.md, and put it into an 'appendix' of some sort that we reference, label it as an implementation guide for now, and drop it later if we still feel its not useful.

I went through it again, and I don't see anything in there that isn't something I expect an implementation to come up by themselves, no surprising twists or nifty tricks to be aware of. And I don't expect a specification to give a coding-beginners-how-to-build-the-client-level code base, even in its appendix. If we want to resurrect it, we'll always have it in the git history.

gnunicorn:
@whyrusleeping anything in particular missing for this to merged?

whyrusleeping:
> I don't see any other way of being able to do that - is there a plan for a warp- or mass-sync? 

Yeah, the ['BlockSync'](https://github.com/filecoin-project/specs/blob/master/network-protocols.md#blocksync) protocol is a faster sync algorithm. Graphsync will replace this down the road, but we have this here until then (which may or may not happen by testnet/mainnet)

> anything in particular missing for this to merged?

Adding more hours to the day would be helpful, but not necessary. This all LGTM, merging! (and thank you!)

PR #301, opened 2019-05-28
Identify bitfield params in submit PoSt and specify bitfield encoding

ZenGround0:


whyrusleeping:
perfect, thanks @ZenGround0!

PR #302, opened 2019-05-28
replace spaces with tabs in psuedocode example

laser:
## Why does this PR exist?

The pseudocode examples must use tabs instead of spaces.

## What's in this PR?

Replace groupings of 4 spaces with a tab character, for consistency.


whyrusleeping:
Thaaanks @laser :)

PR #304, opened 2019-05-28
include null blocks in timestamp verification

frrist:


whyrusleeping:
Actually, before we merge, do you mind also fixing the part about Timestamps under block creation in mining.md? Its basically the same change, just on the creating side instead of the verification side.

Issue #116, opened 2019-01-25
Sector set storage


whyrusleeping:
We need a data structure to hold sectors on chain, inside the miner actor. Currently, we just have them all in a big map, which won’t scale well (every change to it creates a new object, meaning we would rewrite the entire thing with every change).

This issue is a place to brainstorm solutions.

#### Requirements
* Ordered Set
	* no values, just keys (TODO: is this true? Is it just storing CommRs?)
	* ~~‘Ordered’ requirement isn’t as strict as it sounds, we really just need to be able to reference elements within a given set by their index within the set. It is okay if indexes of items change with insertions and deletions.~~
* Efficient additions and random removals
* ~O(log(n)) changed objects per mutation
* Efficient lookup by key and index

#### Nice to haves
* Efficient batch remove by index range

### Does the HAMT work?
We could potentially use the HAMT code, but we would need to adapt it to have efficient index lookups. To do this, we could annotate links to children with a count of how many values exist in and under that child, allowing us to know at each node which child the thing we are looking up is in.

porcuquine:
FWIW, replicaID currently includes the sectorID.

whyrusleeping:
@porcuquine I just read your comments, youre basically saying "if miners can find hash function collisions, they can pull off an attack". To which I respond: If you can generate collisions, the hash function is broken.

If we are assuming that the hash function will not be broken anyways, then i see no problem in allowing full miner selection over the ID.

In addition, I believe that it is a hard requirement that miners have full selection over the sector IDs (I could bring this down to full selection within some range) in order for the protocol to work scalably.

porcuquine:
@whyrusleeping I actually agree with you and concluded the same in my edit.

This doesn't apply to the risks of grinding to generate a fake Seal proof. Although this doesn't add to the risk during a generation attack, allowing an infinite sampling of replica IDs does (I believe) reduce attempts to generate a single fake PoRep for a fixed `commD` to being as easy (or hard) as for any other generation attack.

While this doesn't affect security of the power table, it does reduce the security of storage market redundancy to 'only' on what is also provided against generation attacks. Of course, we assume this is sufficient, but I do think this is a reduction — so we should at least be clear about it.

whyrusleeping:
Additional interesting design constraint to consider: **Do we ever allow sector ID reuse?** If we do not, then we have to track which IDs have been used, in order to disallow their use in the future.

[...]

PR #305, opened 2019-05-29
clarify fault set type and details around it in SubmitPost

whyrusleeping:


actors.md
```diff
@@ -580,6 +580,38 @@ func SubmitPost(proofs []PoStProof, faults []FaultSet, recovered BitField, done
 	miner.NextDoneSet = done
 	miner.ArbitratedDeals.Clear()
 }
+
+func ValidateFaultSets(faults []FaultSet, recovered, done BitField) bool {
+	var aggregate BitField
+	for _, fs := range faults {
+		aggregate.Combine(fs.BitField)
```
ZenGround0: nit: `Union` might be a more intuitive name for what's going on with the bitsets here

actors.md
```diff
@@ -580,6 +580,38 @@ func SubmitPost(proofs []PoStProof, faults []FaultSet, recovered BitField, done
 	miner.NextDoneSet = done
 	miner.ArbitratedDeals.Clear()
 }
+
+func ValidateFaultSets(faults []FaultSet, recovered, done BitField) bool {
```
ZenGround0: Checking for understanding: do methods like `IsSubsetOf`, `Intersects` and `Combine` actually operate on the rle+ Bitfield encoding?  I've heard that this encoding is not designed for efficient updates, it is good only at being compact.  

I am guessing that there is an implicit step in `submitPoSt` where the state machine translates the efficiently encoded BitField into an internal representation that can quickly do unions and intersections.  If this is correct is this hypothesized state-machine-memory-resident type in scope of being specced out?  Also if this is correct is it the case that we should be concerned about *its* performance properties?

Hopefully I am just overthinking this...

state-machine.md
```diff
@@ -151,3 +151,7 @@ as the system has resolvers for them).
 The current state of a given actor can be accessed first by calling `Head` to retrieve the CID of the root of the actors state, then by using `Get` to retrieve the actual object being referenced.
 
 To store data, `Put` is used. Any number of objects may be `Put`, but only the object whose CID is committed, or objects that are linked to in some way by the committed object will be kept. All other objects are dropped after the method invocation returns. Objects stored via `Put` are first marshaled to CBOR-IPLD, and then stored, the returned CID is a 32 byte sha2-256 CBOR-IPLD content identifier.
+
+### Burning Funds
```
ZenGround0: 👍

data-structures.md
```diff
@@ -184,7 +184,18 @@ type Signature struct {
...
+FaultSets are used to denote which sectors failed at which block height.
+
+```go
+type FaultSet struct {
+	Index    uint64
+	BitField BitField
+}
+```
 
+The `Index` field is a block height offset from the start of the miners proving period (in order to make it more compact).
```
dignifiedquire: how does this work? unclear how a single index is enough

actors.md
whyrusleeping: yeah, thats a good point

actors.md
whyrusleeping: Yeah, i'm not necessarily getting into implementation details, this is more of an abstract operation. You have a set of items, these are well known operations on sets of items.

That said, you should be able to operate fairly efficiently over the encoded bitsets, combining multiple rle+ bitfields can be done in a single O(n) pass (where n is the number of bit runs). If the implementation decides to parse that into a different structure internally for performance reasons, I think thats fine.

data-structures.md
whyrusleeping: Its an offset from the miners proving period start, which is a number stored on-chain in their miner actor. It corresponds to the set of challenges being sourced from that block height and says that "for this set of proofs, these sectors were not available"

actors.md
laser: @dignifiedquire and I have discussed plumbing a "current block height" value to the PoSt-generation routine (which is updated when the node learns of a new heaviest tipset). @whyrusleeping - Is it your expectation that a single call to `generate_post` could produce multiple `FaultSet`s (each with a different block height)? 

Also: Is it your expectation that a non-empty `FaultSet` produced during the miner's first proving period would be included in a `submitPoSt` message created during a subsequent proving period?

actors.md
```diff
@@ -580,6 +580,38 @@ func SubmitPost(proofs []PoStProof, faults []FaultSet, recovered BitField, done
...
+	}
+
+	// the done set cannot intersect with the aggregated faults
+	// you can't mark a fault as 'done'
+	if aggregate.Intersects(done) {
+		return false
+	}
+
+	for _, bit := range aggregate.Bits() {
+		if !miner.HasSectorByID(bit) {
```
laser: It's my belief that this `aggregate` includes indices into the miner's `SectorSet`. I think that it will be required that we build something which maps a miner's sector id to an index in the `SectorSet`. If so, I think this pseudocode should be updated to call that function and that the function's output should be passed to `miner.HasSectorByID`.

laser: Note: The above recommendation assumes that the sector IDs cannot be used as indices into the `SectorSet`. The current version of the specification permits miners to pick their own sector ids - so there could be a large gap in between any two sector IDs. Unsure what impact, if any, this would have on the `SectorSet`.

actors.md
laser: @ZenGround0 

> I am guessing that there is an implicit step in submitPoSt where the state machine translates the efficiently encoded BitField into an internal representation that can quickly do unions and intersections.

This too is what I assumed. I would have thought we'd use RLE+ when converting `BitField` to `[]byte` and back - and that inside the state machine we'd be working with some fancy type optimized for set operations.

actors.md
whyrusleeping: sector IDs are indexes into the sector set, having gaps is allowable, and should add minimal overhead to the bitfields. This is discussed in #116 (giant issue i know, we should write up a summary)

actors.md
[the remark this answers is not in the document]
whyrusleeping: yes

actors.md
whyrusleeping: > Is it your expectation that a single call to generate_post could produce multiple FaultSets (each with a different block height)?

Yes, the failure could happen part ways through the post. multiple failures could happen at different times during the PoSt

> Also: Is it your expectation that a non-empty FaultSet produced during the miner's first proving period would be included in a submitPoSt message created during a subsequent proving period?

No, at the end of a proving period, any sectors that are not marked as recovered are dropped from the sector set and should not be proved during the next PoSt.

actors.md
ZenGround0: FYI I'm working on a summary of the Bitfield and SectorSet design discussions as tracked [here](https://github.com/filecoin-project/go-filecoin/issues/2824)

Issue #299, opened 2019-05-28
Miscellaneous Miner Actor Spec Improvements


ZenGround0:
The go-filecoin team pointed out these areas in the miner actor's contracts as needing work to facilitate a complete implementation.

- [ ] At some point we need to specify `CollateralForSector` as called by [CommitSector](https://github.com/filecoin-project/specs/blob/master/actors.md#commitsector).  Including how it differs based on pledge amount and sector size.
- [x] The worker / owner distinction makes sense but it is not specified enough to be easily understood or implemented.  
  - Motivation behind worker account in the comments above the field in the miner actor would resolve lots of confusion
  - The worker is not specified [during construction](https://github.com/filecoin-project/specs/blob/master/actors.md#constructor) and I can't find a setting method.  Right now there is now way to assign a worker
  - It is not clear to me what the relationship between the worker, owner, and publickey are.  Is the key unrelated to worker and owner addresses or can it sign messages for one of them?
  - This TODO on `DePledge`: "Do both the worker and the owner have the right to call this?"  
- [x] At some point we need to specify Proving Period length as a function of sector size (and anything else relevant)
- [ ] We need to move asks offchain in the spec

whyrusleeping:
> Motivation behind worker account in the comments above the field in the miner actor would resolve lots of confusion

@ZenGround0 is the owner worker distinction here not sufficient? https://github.com/filecoin-project/specs/blob/master/mining.md#owner-worker-distinction


the rest sgtm, will send a PR

ZenGround0:
@whyrusleeping yup I missed that, just what I was looking for

whyrusleeping:
Made a new issue for offchain asks: https://github.com/filecoin-project/specs/issues/309

and #60 should address the collateral stuff when resolved.

Closing this issue out now

PR #307, opened 2019-05-29
some fixes and clarifications to address #299

whyrusleeping:
This doesnt resolve every point in #299, but gets the easy ones out of the way

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -176,18 +176,15 @@ type StorageMarketActor struct {
 
 Parameters:
 
-- pubkey PublicKey
-
+- worker Address
 - pledge BytesAmount
-
 - sectorSize BytesAmount
-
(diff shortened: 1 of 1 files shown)


actors.md
```diff
@@ -580,6 +579,44 @@ func SubmitPost(proofs []PoStProof, faults []FaultSet, recovered BitField, done
 	miner.NextDoneSet = done
 	miner.ArbitratedDeals.Clear()
 }
+
+func ValidateFaultSets(faults []FaultSet, recovered, done BitField) bool {
```
ZenGround0: nit: looks like a crossover from #305

whyrusleeping: oh weird. yeah... let's get 305 in and i'll rebase this

Issue #293, opened 2019-05-27
Spec for late/failed PoST fees, robust payment mechanism


anorth:
The [miner actor](https://github.com/filecoin-project/specs/blob/master/actors.md#submitpost) `SubmitPoSt` invokes `ComputeLateFee` to compute fees for late submission of a proof, but that method is currently unspecified. See also https://github.com/filecoin-project/specs/blob/master/faults.md (which calls it `CalculateLatePenalty`).

1. The call site implies that the late fee is a pure function of the number of blocks late that the submission is. Should it not also be a function of the miner's pledge collateral or committed bytes?
1. The faults spec implies that the late fee included in the message must exactly match the computed fee. Since it's a function of block height, this will be a brittle check, prone to failing due to propagation delays or censorship, adding yet more fee burden. I propose that this should be made more robust. Two suggestions are:
    1. the message may carry more than the required fee, and the balance is refunded to the offender when the message is processed
    1. the message may carry less than the required fee and still be accepted; the miner remains in an under-collateralised state (which can occur for other non-malicious reasons too)

[Late PoSts might *still* suffer from censorship, but let's not bury this issue in that discussion]

Go-filecoin is implementing this soon. The specific function arithmetic may be changed later, but it would be helpful to be confident in what the parameters to the late fee function are, and the requirements for the network accepting it.

ZenGround0:
3. We also need a spec for the related `ComputeTemporarySectorFailureFee` during Reported Storage Fault Penalties.  As for `ComputeLateFee` the exact arithmetic isn't so important but the interface is.  Also this fee seems to not depend on block time so payment does not suffer from the issue described in bullet (2)

whyrusleeping:
> The faults spec implies that the late fee included in the message must exactly match the computed fee.

It doesnt.

> the message may carry more than the required fee, and the balance is refunded to the offender when the message is processed

yep! It does this, see the code for submitPoSt:
```go
	if msg.Value < feesRequired {
		Fatal("not enough funds to pay post submission fees")
	}

	// we want to ensure that the miner can submit more fees than required, just in case
	if msg.Value > feesRequired {
		Refund(msg.Value - feesRequired)
	}
```

I'll work on a PR for the parameters for the late fee functions.

whyrusleeping:
@anorth does #312 address this for you? we should make new issues for the specific constructions of those fees, but this gets things in the right shape

anorth:
[...]

PR #312, opened 2019-05-30
some clarifications to fees and fixes to collateral tracking

whyrusleeping:
Should address #293 

Added stubs for the fee calculation functions (@decentralion if you could help with the real constructions for these, i'd really appreciate it).

Additionally, I removed the 'pledge' amount that was being used to track how many bytes a miner claims they have available. This isnt really useful, as we either have to force them to put collateral up for this (in which case they won't say how much storage they really have) or theres nothing backing it and people will just put random numbers, making the metric useless.

I've also removed the tracking of free collateral in the miner actor, and am just using the available balance in the account as the 'total collateral pool'. This should simplify a few things.

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -177,24 +177,19 @@ type StorageMarketActor struct {
 Parameters:
 
 - worker Address
-- pledge BytesAmount
 - sectorSize BytesAmount
 - pid PeerID
 
 Return: Address
 
 ```go
-func CreateStorageMiner(worker Address, pledge, sectorSize BytesAmount, pid PeerID) Address {
+func CreateStorageMiner(worker Address, sectorSize BytesAmount, pid PeerID) Address {
 	if !SupportedSectorSize(sectorSize) {
 		Fatal("Unsupported sector size")
 	}
 
 	if pledge < MinimumPledge(sectorSize) {
 		Fatal("Pledge too low")
-	}
-
-	if msg.Value < MinimumCollateral(pledge) {
-		Fatal("not enough funds to cover required collateral")
 	}
 
 	newminer := InitActor.Exec(MinerActorCodeCid, EncodeParams(pubkey, pledge, sectorSize, pid))
@@ -323,18 +318,9 @@ type StorageMiner struct {
 	// to this miner
 	PeerID peer.ID
 
-	// PledgeBytes is the amount of space being offered by this miner to the network
-	PledgeBytes BytesAmount
-
 	// SectorSize is the amount of space in each sector committed to the network
 	// by this miner.
 	SectorSize BytesAmount
-
-	// Collateral is locked up filecoin the miner has available to commit to storage.
-	// When miners commit new sectors, tokens are moved from here to 'ActiveCollateral'
-	// The sum of collateral here and in activecollateral should equal the required amount
-	// for the size of the miners pledge.
-	Collateral TokenAmount
 
 	// ActiveCollateral is the amount of collateral currently committed to live storage
 	ActiveCollateral TokenAmount
@@ -360,9 +346,6 @@ type StorageMiner struct {
 	// ArbitratedDeals is the set of deals this miner has been slashed for since the
 	// last post submission
 	ArbitratedDeals CidSet
-
-	// TODO: maybe this number is redundant with power
-	LockedStorage Integer
 
 	// Power is the amount of power this miner has
 	Power Integer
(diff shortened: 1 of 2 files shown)


decentralion:
I honestly don't have an intuition for how the actual fees should be computed (I've just been focused on what pledge collateral should be overall). Do you? Also, is there a list of all the circumstances under which miners get slashed?

whyrusleeping:
@decentralion all the possible slashings (possibly incomplete, but i think its good) are listed here: https://github.com/filecoin-project/specs/blob/master/faults.md

As for inuition around fees, i don't have a good one. Basically, at the end of the day, the fees need to be quite small for normal well behaved miners who make a mistake, and painful for attackers trying to game the system. beyond that... i havent put much thought into it.

actors.md
```diff
@@ -617,26 +594,13 @@ func ProvingPeriodDuration(sectorSize uint64) Integer {
...
-
-
-```go
-func IncreasePledge(addspace BytesAmount) {
-	// Note: msg.Value is implicitly transferred to the miner actor
-	if miner.Collateral+msg.Value < CollateralForPledge(addspace+miner.PledgeBytes) {
-		Fatal("not enough total collateral for the requested pledge")
-	}
+func ComputeLateFee(power Integer, blocksLate Integer) TokenAmount {
+  return 4 // TODO: real collateral calculation, obviously
```
laser: you missed the trailing 2! :)

actors.md
```diff
@@ -323,19 +318,10 @@ type StorageMiner struct {
...
-
 	// SectorSize is the amount of space in each sector committed to the network
 	// by this miner.
 	SectorSize BytesAmount
 
-	// Collateral is locked up filecoin the miner has available to commit to storage.
-	// When miners commit new sectors, tokens are moved from here to 'ActiveCollateral'
-	// The sum of collateral here and in activecollateral should equal the required amount
-	// for the size of the miners pledge.
-	Collateral TokenAmount
```
laser: Note: The `DePledge` method [checks the `amt` against the `Collateral` field's value](https://github.com/filecoin-project/specs/pull/312/files#diff-8cd5872b87e5b64500c99ea50b3a9241R724). I think you should check `amt` against `vm.MyBalance() - miner.ActiveCollateral` instead.

actors.md
```diff
@@ -471,11 +450,10 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
 	// make sure the miner has enough collateral to add more storage
 	coll = CollateralForSector(miner.SectorSize)
 
-	if coll < miner.Collateral {
+  if coll < vm.MyBalance() - miner.ActiveCollateral {
```
laser: Note: I see that the [`SlashStorageFault`](https://github.com/filecoin-project/specs/pull/312/files#diff-8cd5872b87e5b64500c99ea50b3a9241R634) method consumes the `ConsensusCollateral` field on the miner. I don't see that field specified anywhere. Should this check be rewritten as:

```
if coll < vm.MyBalance() - miner.ActiveCollateral - miner.ConsensusCollateral
```

?

actors.md
whyrusleeping: It's implied :P

ZenGround0: was gonna make a snarky factorial joke but got rekt because 2

actors.md
```diff
@@ -323,19 +318,10 @@ type StorageMiner struct {
 	// to this miner
 	PeerID peer.ID
 
-	// PledgeBytes is the amount of space being offered by this miner to the network
-	PledgeBytes BytesAmount
```
ZenGround0: I know there are some ideas in the running for our collateral function that are not as simple as `collateral = C * PledgeBytes`.  Are all of these candidate collateral functions invertible?  If not user experience will suffer because it would be ambiguous how many bytes a miner has left to store.

I'm aware that we are not actually enforcing the pledge amount so malicious miners could always look available and not accept any requests, but the system shouldn't make it hard for clients to know if honest miners have room left.  Is the plan to move this signal off the blockchain ?

ZenGround0: If we do not track storage and consensus collateral separately then I think ambiguity is always there.

laser: The specification shows that the storage miner actor [does track consensus collateral (`ConsensusCollateral`)](https://github.com/filecoin-project/specs/pull/312/files#diff-8cd5872b87e5b64500c99ea50b3a9241R634). 

`ActiveCollateral` is [divided into storage collateral and pledge collateral](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1559239486004600?thread_ts=1559235221.001500&cid=CHMNDCK9P).

It is also the case that a storage miner actor's FIL balance [is always less than or equal](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1559237847003700?thread_ts=1559235221.001500&cid=CHMNDCK9P) to `ActiveCollateral` plus `ConsensusCollateral`.

whyrusleeping: the construction of the collateral function is separate from the issue of not being able to prove how much storage a given miner has available.

We haven't yet separated pledge collateral and storage collateral, that will happen soon

anorth:
Thanks for the simplifications to the collateral variables – I was working up to a bunch of questions about them too.

PR #314, opened 2019-05-30
Add signed wrappers for StorageDealProposal and StorageDealResponse

acruikshank:
### Problem

Including the signature within the data structure to be signed complicates the signing and verification logic. This makes it more difficult to specify and more difficult to implement.

### Solution

This PR adds SignedStorageDealProposal and SignedStorageDealResponse that wrap StorageDealProposal and StorageDealResponse to pull out the signatures from these objects. This allows us to sign and verify these object consistently by cbor encoding the wrapped object and passing that to either the signing or verification logic without special handling of the Signature field.


whyrusleeping:
Thanks @acruikshank ! this simplifies things a bit

Issue #315, opened 2019-05-30
block validation needs to check that miner is part of storage market


whyrusleeping:
The storage market actor keeps track of all the miners that are actually part of the official filecoin storage market. Technically, other miner actors may be created and have sectors committed to them (though I suspect that may fail). Additionally, miners who are slashed for breaking consensus rules are evicted from the storage market.

Miners who are not part of the storage market should not be able to mine blocks. We should add a check that ensures the miner in question is an actual storage market miner. 

Open question: Do we make this check at the lookback (the point back in the chain where we pull the power table from?) or from the current state? I think we might need both, as if we just do the lookback, then a slashed miner would be able to still keep making blocks until the lookback period elapses. cc @sternhenri

sternhenri:
Mmmh it's a good question.

The rule you point out would be: 

1. Check at lookback L (ie power table lookback) what the miner set is.
2. Ensure no miners have been slashed since L, and exclude them from set if they have.

That seems fair. There might be the following question: what if a slashing ends up not being on the "main chain?" i.e. we include it too early by looking ahead of L vs waiting to see whether it makes it to L.

This wouldn't happen for consensus faults, though it could for storage faults. So all this to comfort your proposal @whyrusleeping: we should use both, but only include slashing for consensus faults ahead of time, not for storage faults (only once they've made it to L, which will be tied to finality).

pooja:
@sternhenri Is it okay if we make you DRI for this issue?

sternhenri:
sorry for lag here. on it!

sternhenri:
So my comment above is silly: if you were slashed somewhere between lookback and present (for any reason) and that slash did not make it to the final chain, then nor would the block we are currently validating (as a descendent of the one in which the slashing occurred), so we needn't worry about this. @whyrusleeping's simpler solution works.

PR #319, opened 2019-05-31
add block validation checks to ensure miner is legitimate

whyrusleeping:
addresses #315

### DIFF
--- a/mining.md
+++ b/mining.md
@@ -130,17 +130,23 @@ func VerifyBlock(blk Block) {
 		Fatal("block was generated too soon")
 	}
 
-	// 3. Verify ParentWeight
+  // 3. Verify miner has not been slashed and is still valid miner
+  curStorageMarket := LoadStorageMarket(blk.State)
+  if !curStorageMarket.IsMiner(blk.Miner) {
+    Fatal("block miner not valid")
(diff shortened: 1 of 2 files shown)


mining.md
```diff
@@ -129,18 +129,24 @@ func VerifyBlock(blk Block) {
 	if blk.GetTime() <= blk.minParentTime() + (BLOCK_DELAY * len(blk.Tickets)) {
 		Fatal("block was generated too soon")
 	}
+  
+  // 3. Verify miner has not been slashed and is still valid miner
+  curStorageMarket := LoadStorageMarket(blk.State)
+  if !curStorageMarket.IsMiner(blk.Miner) {
+    Fatal("block miner not valid")
+  }
```
anorth: Inconsistent indentation here and below.

whyrusleeping:
SubmitPoSt will fail if the miner is not official due to the call to `UpdatePower` checking the miners inclusion in the storage market. I actually think this is unfortunate, and i'd be happy if 'external' miners could still function and prove storage, but just not be able to gain power for it. I'm definitely not going to go out of my way to make that work right now though, maybe in the future we can add 'secondary storage markets'.

Issue #56, opened 2018-12-09
SEAL Replica ID should also include a historical ticket


nicola:
This enforces that the miner is dedicating storage to a particular part of the network

EDIT: Based on discussion below and in spec review, I am changing the title and am adding to the original description here. –@porcuquine

When sealing begins, a ticket should be included in the data hashed to generate the replica id.

This ticket should be from a block which is `FINALITY` rounds back (at seal start time).

When seal proofs are verified, it must be verified that the round from which the ticket was fetched is less than `RECENCY` rounds back (at verification time).

This implies that `RECENCY` must be greater than `FINALITY` (by the total allowable seal time, in rounds).

`RECENCY` and `FINALITY` are both integer constants whose values are yet to be determined.

*NOTE*: This change will require a change to the sealing process because in general, a Filecoin node does not request sealing. Rather sealing is triggered when a piece is added and this results in a sector being full. For this to work, we should pass the correct ticket (`FINALITY` rounds back) whenever a piece is added. Then whenever sealing is triggered, the most recent ticket can be used to generate the replica id. cc:@laser.

---

This requires changes in code by @porcuquine @laser

porcuquine:
@laser Is there now a clear answer to the question of where this 'hash of block' chain randomness would come from and what its characteristics would be? I think we should make the spec change once that's the case, but not before.

nicola:
Note: the spec have this regardless of what is implemented today, otherwise will mislead others reading the spec / interested in doing separate implementations

pooja:
@porcuquine @nicola Can you please update this issue with the latest?

porcuquine:
@nicola @laser @ZenGround0 @sternhenri @whyrusleeping Does anyone one know *exactly* what value is meant to be added to the replica ID? I have not been able to get this answer, but I suspect that one of you knows it.

If none of us know, I believe between us we can decide. Let's do that here, please.

porcuquine:
Maybe this is the answer:

> there should be only one block (technically, we should not use the hash of the block but the winning ticket at the current epoch)

Is this well-defined? Does it mean the current epoch when sealing begins? If so, is there a restriction on how soon the sector must be committed? If not, what prevents generating replica IDs far in advance and using them much later? That is: how does that differ from using very old tickets to generate replica IDs?

nicola:
My latest comment is correct: we should use the ticket of the winning blocks

porcuquine:
Please read my questions again, or consider this:

[...]

PR #323, opened 2019-06-02
miscellaneous proofs updates (proof size, partitions, path/sector access, etc.)

laser:
## Why is this PR needed?

There were [various](https://github.com/filecoin-project/specs/issues/56) conversations which necessitated changes to the proofs portion of the spec. 

## What's in this PR?

- ~replace "ticket" with "roundNumber" as per [this convo](https://github.com/filecoin-project/specs/issues/56)~
- introduce proof partitions and sector size as inputs to `Seal` and `VerifySeal`
- replace "sector access" with path and drop `SectorStore`

### DIFF
--- a/data-structures.md
+++ b/data-structures.md
@@ -233,11 +233,11 @@ BlockHeader is a serialized `Block`.
 
 #### `SealProof`
 
-SealProof is a 384-element array of bytes.
+SealProof is an opaque, dynamically-sized array of bytes. The proof will have a length equal to N * 192, where N is the number of partitions used when creating the proof.
 
 #### `PoStProof`
 
-PoStProof is a 192-element array of bytes.
+PoStProof is an opaque, dynamically-sized array of bytes. The proof will have a length equal to N * 192, where N is the number of partitions used when creating the proof.
 
 #### `TokenAmount`
 
--- a/data-structures.md
+++ b/data-structures.md
@@ -233,11 +233,11 @@ BlockHeader is a serialized `Block`.
 
 #### `SealProof`
 
-SealProof is an opaque, dynamically-sized array of bytes. The proof will have a length equal to N * 192, where N is the number of partitions used when creating the proof.
+SealProof is an opaque, dynamically-sized array of bytes.
 
 #### `PoStProof`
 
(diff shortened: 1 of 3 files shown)


data-structures.md
```diff
@@ -233,11 +233,11 @@ BlockHeader is a serialized `Block`.
 
 #### `SealProof`
 
-SealProof is a 384-element array of bytes.
+SealProof is an opaque, dynamically-sized array of bytes. The proof will have a length equal to N * 192, where N is the number of partitions used when creating the proof.
 
 #### `PoStProof`
 
-PoStProof is a 192-element array of bytes.
+PoStProof is an opaque, dynamically-sized array of bytes. The proof will have a length equal to N * 192, where N is the number of partitions used when creating the proof.
```
whyrusleeping: since this is an opaque data type, its probably going to have a small amount of framing. I don't know if its worth saying a specific size it will be here.

proofs.md
```diff
@@ -69,15 +70,15 @@ Seal
...
-  proof     []byte,                         // returned from Seal
-  proverID  [31]byte,                       // uniquely identifies miner
-  ticket                                    // identifies ticket provided when sealing
-  sectorID  [31]byte,                       // uniquely identifies sector
+  commD       [32]byte, // returned from Seal
+  commR       [32]byte, // returned from Seal [will be removed in future iteration]
+  commRStar   [32]byte, // returned from Seal
+  proof       []byte,   // returned from Seal
+  proverID    [31]byte, // uniquely identifies miner
+  roundNumber uint64,   // round (block height) to which a miner committed when sealing began
```
whyrusleeping: in the interface to the actual function i think the ticket is correct. We just don't want to submit the entire ticket to the chain in a message (let the chain look up the ticket and pass it to this method)

data-structures.md
porcuquine: Yes — both sealing and verification will need the actual ticket bytes, not the round number.

proofs.md
porcuquine: Yes — both sealing and verification will need the actual ticket bytes, not the round number.

laser: > both sealing and verification will need the actual ticket bytes, not the round number.

Ah, yeah - that makes more sense.

data-structures.md
laser: Done and done.

Issue #92, opened 2019-01-15
Resume mining after being slashed


whyrusleeping:
If a miner fails to submit their PoSt in time, they can be slashed. Currently this means they lose all of their (EDIT: storage) collateral. However, it does not mean they have actually lost all of their data. A miner should be able to re-up on their collateral, and continue mining.

A 'simple' way to do this would be to simply require that miners send enough funds along with a normal submit PoSt call to 'resume' mining.

Also, this means that clients cannot call arbitrate deal after a miner has been slashed for not submitting their PoSt. There should be some period of time defined after which miners can be assumed to have actually lost the data.

mishmosh:
This is now possible with temporary faults.

PR #324, opened 2019-06-03
Remove duplication of the storage market actor interface in storage-market.md

anorth:
~This exposes that `SlashStorageFault` is currently missing pseudocode.~

Re: #92, #292

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -156,11 +156,11 @@ The storage market actor is the central point for the Filecoin storage market. I
 
 ```go
 type StorageMarketActor struct {
+	// A set containing the actor address of all storage miner actors.
 	Miners AddressSet
 
-	// TODO: Determine correct unit of measure. Could be denominated in the
-	// smallest sector size supported by the network.
-	TotalStorage BytesAmount
+    // The number of bytes of proven storage in the market.
+	TotalComittedStorage BytesAmount
 }
 ```
 ### Code Cid
@@ -168,8 +168,8 @@ type StorageMarketActor struct {
 `<codec:raw><mhType:identity><"smarket">`
 
 | Index     | Method Name       |
-| -------- | ---------- |
-| 1   | `CreateStorageMiner`     |
+| --------- | ---------- |
+| 1 | `CreateStorageMiner`     |
 | 2 | `SlashConsensusFault` |
 | 3 | `UpdateStorage` |
 | 4 | `GetTotalStorage` |
@@ -178,6 +178,8 @@ type StorageMarketActor struct {
 
 
 ### CreateStorageMiner
+
+> CreateStorageMiner registers a new storage miner with the given public key. The miner's collateral is set by the value in the message. The public key must match the private key used to sign off on blocks created by this miner. This key is the 'worker' key for the miner. The libp2p peer ID specified should reference the libp2p identity that the miner is operating. This is the ID to which clients will connect to propose deals.
 
 Parameters:
 
@@ -193,7 +195,7 @@ func CreateStorageMiner(worker Address, sectorSize BytesAmount, pid PeerID) Addr
 		Fatal("Unsupported sector size")
 	}
 
-	newminer := InitActor.Exec(MinerActorCodeCid, EncodeParams(pubkey, pledge, sectorSize, pid))
+	newminer := InitActor.Exec(MinerActorCodeCid, EncodeParams(worker, sectorSize, pid))
 
 	self.Miners.Add(newminer)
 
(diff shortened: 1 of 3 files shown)


actors.md
```diff
@@ -186,16 +189,16 @@ Parameters:
 Return: Address
 
 ```go
-func CreateStorageMiner(worker Address, sectorSize BytesAmount, pid PeerID) Address {
```
whyrusleeping: this change is incorrect. please keep it as an address.

actors.md
```diff
@@ -265,14 +270,31 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
...
 
-	// Now delete the miner (maybe this is a bit harsh, but i'm okay with it for now)
+	// Remove the actor's state from the network entirely.
 	miner.SelfDestruct()
 }
 ```
 
+### SlashStorageFault
+
+> SlashStorageFault penalizes a storage miner for not submitting a PoSt within the required [time window](#TODO-link-to-faulty-submission). This may be called by anyone who detects the faulty behavior. The slashed miner loses all of their staked collateral and all of their power, and as a result, is no longer a candidate leader for extending the chain. See [faults](faults.md).
```
whyrusleeping: its worth linking directly to the unreported storage fault slashing section that this is for.

Also, what is 'TODO-link-to-faulty-submission' supposed to link to?

actors.md
```diff
@@ -265,14 +270,31 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
...
+
+Parameters: 
+
+- miner Address
+
+Return: None
+
+```go
+func SlashStorageFault(miner Address) {
+	// TODO
```
whyrusleeping: +1, will file an issue, thanks

whyrusleeping: ah, right. I *thought* I had written this already, its a method on the miner itself.

whyrusleeping: It was initially on the storage market, but all it did was call through to a method on the miner, and didnt actually do anything on the storage market, so it was moved to the miner.

It definitely feels weird to have the consensus slashing here, and the storage slashing there though, so I could be convinced that it should be moved back. But we would then be adding network overhead for a small UX improvement.

actors.md
anorth: Roger that. I inferred pubkey from its use at 198, and `storage-market.md`. Address is consistent with the constructor.

actors.md
anorth: Ah, of course, thanks for spotting. I'll remove this.

actors.md
anorth: 1. Linkified.
1. I don't know, I copied this from `storage-market.md`. I'll remove the TODO if you don't know what it's standing for.

storage-market.md
```diff
@@ -115,16 +80,14 @@ TODO: 'complete' isnt really the right word here, as it implies that the deal is
 
 ## The Power Table
 
-The `power table` is exported by the storage market for use by consensus. There isn't actually a concrete object that is the power table, instead, there is a 'total power' exported by the storage market actor, and each individual miner reports its power through their actor.
+The storage market actor maintains a total committed storage amount, aka "total power". The "power table" is a logical construct which is the fraction of this total power that each miner contributes. This is used by the consensus algorithm.
 
-TODO: rephrase the above to make it clear that power is updated only on PoSt submission and when slashed.
+Each individual miner reports its power through its on-chain actor. A miner's power (and hence the total power) is updated only upon PoSt submission (when the set of sectors being proved may change) or when a miner is slashed.
```
dignifiedquire: Might be good to note that this submission is dependent on the individual miner, rather than a global operation

dignifiedquire: Actually this is already clear, as this is talking about a specific miner, nvm

faults.md
```diff
@@ -10,7 +10,7 @@ A fault is what happens when partcipants in the protocol are behaving incorrectl
   - **Condition:** If any miner posts two blocks satisfying the slashing conditions defined in [Expected Consensus](./expected-consensus.md).
   - **Reporting:** Anyone may call `SlashConsensusFault` and pass in the two offending block headers.
   - **Check:** The chain checks that both blocks are valid, correctly signed by the same miner, and satisfy the consensus slashing conditions.
-  - **Penalization:** All of the miner's pledge collateral and all of their power is irrevocably slashed. This miner can never again produce blocks, even if they attempt to repost their collateral.
+  - **Penalization:** All of the miner's pledge collateral and power is slashed and the miner is irrecoverably removed from the market. This miner can never again produce blocks, even if they attempt to repost their collateral.
```
dignifiedquire: I think there is a typo in irrevocably

anorth:
@whyrusleeping ready for merge.

anorth:
It looks like this has rotted and will need re-doing against the changed actors spec.

dignifiedquire:
> It looks like this has rotted and will need re-doing against the changed actors spec.

sorry about that, please ping me when you have rebased onto `master` and I'll make sure we get it in

anorth:
It's a bunch of mechanical work - I don't expect to get to it soon. Feel
free also to close and have someone else do it if this PR is in the way.

Issue #60, opened 2018-12-19
Separate storage collateral and pledge collateral


whyrusleeping:

pooja:
@decentralion Is this something you'd have bandwidth to help document? Thanks!

decentralion:
I can help document if I have some more context on what's expected here. What are the relevant sections of the spec? What info do we need to communicate? Etc. cc @whyrusleeping

pooja:
I'm guessing we want something like this in the spec: https://github.com/filecoin-project/cryptoeconomics/blob/master/docs/filecoin_collateral.md

I'll work with @whyrusleeping to get the relevant context together and post it here!

whyrusleeping:
At a high level (mostly recap, this is what @decentralion and I have talked about before) we have two types of collateral, pledge collateral and storage collateral. Both are needed per sector committed. The pledge collateral needed per sector is set by the network, the storage collateral per miner is set individually per miner on creation (parameter to the constructor).

Concrete changes that should be made:
- actors.md:
  - adjust miner constructor to take storage collateral
  - adjust slashing methods (slashconsensusfault, slashstoragefault, arbitratedeal) to slash the correct collateral
    - also write in the 'slashing auction' mechanism here
- storage-market.md
  - probably want to add a section explaining how collateral works in english

anorth:
[Copied from #294]
The breach of contract dispute in faults states that the penalty for defaulting on a contract is taken from pledge collateral. It's not explicit about whether the fee is paid to the storage client or burned. I gather this is incorrect and should refer to storage collateral.

In the actor spec, ArbitrateDeal pseudocode mentions both burning pledge collateral and transferring storage collateral to the client. How this storage collateral is represented and where it comes from isn't specified.

On being slashed in mining spec opens a discussion and TODO for disambiguating the two collaterals throughout the spec. My impression from conversations elsewhere suggest we're currently intending to identify both.

Assuming we do follow through with storage collateral, we'll need:

- a mechanism for determining how much storage collateral is to be posted (requested by client? network constant?). @whyrusleeping's comment above suggest the miner sets a constant value (per byte/year?), that a client can take or leave
- a mechanism for posting that collateral somewhere (payment channels w/ conditional vouchers? a new account?
- a mechanism for retrieving or repurposing the collateral when a deal expires successfully
- ArbitrateDeal to transfer some of that to the client when a dispute is successful
 - Extend to the "slashing auction" model so that this action can be delegated, if desired

[...]

PR #325, opened 2019-06-03
An initial stab at speccing Pledge Collateral

decentralion:
See context in #60.

Thanks to @whyrusleeping for help translating this into spec.
Significant thanks to an anonymous contributor who helped design and model this proposal.

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -484,10 +484,11 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
 		Fatal("sector already committed!")
 	}
 
-	// make sure the miner has enough collateral to add more storage
-	coll = CollateralForSector(miner.SectorSize)
+  // Power of the miner after adding this sector
+  futurePower = miner.power + miner.SectorSize
+  collateralRequired = collateralForPower(futurePower)
 
-	if coll < vm.MyBalance()-miner.ActiveCollateral {
+  if collateralRequired > vm.MyBalance() {
 		Fatal("not enough collateral")
 	}
 
@@ -510,6 +511,16 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
 		miner.ProvingSet = miner.Sectors
 		miner.ProvingPeriodEnd = chain.Now() + ProvingPeriodDuration(miner.SectorSize)
 	}
+}
+
+func collateralForPower(power BytesAmount) TokenAmount {
+  availableFil = FakeGlobalMethods.GetAvailableFil()
+  totalNetworkPower = StorageMinerActor.GetTotalStorage()
+  numMiners = FakeGlobalMethods.GetTotalNumMiners()
+  powerCollateral = availableFil * NetworkConstants.POWER_COLLATERAL_PROPORTION * power / totalNetworkPower
+  perCapitaCollateral = availableFil * NetworkConstants.PER_CAPITA_COLLATERAL_PROPORTION / numMiners
+  collateralRequired = minerPowerCollateral + minerPerCapitaCollateral
+  return collateralRequired
 }
 ```
 
--- a/actors.md
+++ b/actors.md
@@ -513,13 +513,13 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
 	}
 }
 
-func collateralForPower(power BytesAmount) TokenAmount {
+func CollateralForPower(power BytesAmount) TokenAmount {
   availableFil = FakeGlobalMethods.GetAvailableFil()
   totalNetworkPower = StorageMinerActor.GetTotalStorage()
-  numMiners = FakeGlobalMethods.GetTotalNumMiners()
+  numMiners = StorageMarket.GetMinerCount()
   powerCollateral = availableFil * NetworkConstants.POWER_COLLATERAL_PROPORTION * power / totalNetworkPower
   perCapitaCollateral = availableFil * NetworkConstants.PER_CAPITA_COLLATERAL_PROPORTION / numMiners
-  collateralRequired = minerPowerCollateral + minerPerCapitaCollateral
+  collateralRequired = math.Ceil(minerPowerCollateral + minerPerCapitaCollateral)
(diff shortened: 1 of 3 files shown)


pledge-collateral.md
```diff
@@ -0,0 +1,63 @@
...
+- increases the cost of launching a 51% attack
+
+
+## Computing Pledge Collateral
+
+The total pledge collateral across all miners is a fixed proportion of available FIL.
+Available FIL is computed as the total amount of FIL that has been mined, plus the total amount of FIL that's been vested.
+
+```go
+availableFil := minedFil + vestedFil
```
whyrusleeping: we *could* say `- burnedFil`, otherwise over time this proportion will become effectively bigger than the percentage we set (though that will happen anyways as funds will be 'lost' through ways aside from the official burn mechanism)

actors.md
```diff
@@ -450,10 +450,11 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
...
 	}
 
-	// make sure the miner has enough collateral to add more storage
-	coll = CollateralForSector(miner.SectorSize)
+  // Power of the miner after adding this sector
+  futurePower = miner.power + miner.SectorSize
+  collateralRequired = collateralForPower(futurePower)
 
-  if coll < vm.MyBalance() - miner.ActiveCollateral {
+  if collateralRequired < vm.MyBalance() {
```
anorth: Should ActiveCollateral be removed from the spec entirely at this point? I confess I'm [still confused](https://github.com/filecoin-project/specs/issues/60#issuecomment-497585831) as to which collateral it is intended to represent.

It's still in the state, `SubmitPost` and `DePledge`.

actors.md
```diff
@@ -477,6 +478,16 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
 		miner.ProvingPeriodEnd = chain.Now() + ProvingPeriodDuration(miner.SectorSize)
 	}
 }
+
+func collateralForPower(power BytesAmount) TokenAmount {
```
anorth: Nit: Is the lower case initial character intended to be meaningful? It will be quite opaque to anyone not familiar with Go and it looks like the standard here is to uppercase all function/method names everywhere. Visibility isn't relevant to the spec.

actors.md
```diff
@@ -477,6 +478,16 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
...
+  availableFil = FakeGlobalMethods.GetAvailableFil()
+  totalNetworkPower = StorageMinerActor.GetTotalStorage()
+  numMiners = FakeGlobalMethods.GetTotalNumMiners()
```
anorth: Suggest `StorageMarket.GetMinerCount()` (which implementation is pretty obvious for us to follow-up with).

actors.md
```diff
@@ -477,6 +478,16 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
...
+  powerCollateral = availableFil * NetworkConstants.POWER_COLLATERAL_PROPORTION * power / totalNetworkPower
+  perCapitaCollateral = availableFil * NetworkConstants.PER_CAPITA_COLLATERAL_PROPORTION / numMiners
```
anorth: We should specify truncation towards zero as the rounding mode (assuming that intent).

deltazxm:
Good job,I love this proposal.

pledge-collateral.md
decentralion: I'm fine including burnedFil. It technically becomes impossible to forecast future pledge collateral with precise accuracy, but I don't think that's an issue.

actors.md
decentralion: I don't see any need for active collateral. I agree that we should remove it.

actors.md
decentralion: Done.

actors.md
decentralion: I think it makes sense to round up. If the "true" collateral required is (say) 1.3, then by holding 2 FIL you are over-collateralized (fine) but by holding 1 you are under collateralized (technically disallowed).

We can round after summing the two collateral types.

actors.md
decentralion: Nope... that was just me being javascript developer and naming variables like I'm in JSland

decentralion:
@whyrusleeping I think this is ready for merge.

actors.md
whyrusleeping: yep, it should be removed.

whyrusleeping:
@decentralion one final bit, could you update the `DePledge` method, as @anorth pointed out in slack, it's also affected by this and needs to be updated to match these changes

mining.md
```diff
@@ -306,7 +306,7 @@ To create a block, the eligible miner must compute a few fields:
 - `MsgRoot` - To compute this:
   - Select a set of messages from the mempool to include in the block.
   - Insert them into a Merkle Tree and take its root.
-    - Note: Messages with BLS signatures should be included as raw `Message` types, and not `SignedMessage`. Their signatures should be gathered up and aggregated for the `BLSAggregate` field. 
+    - Note: Messages with BLS signatures should be included as raw `Message` types, and not `SignedMessage`. Their signatures should be gathered up and aggregated for the `BLSAggregate` field.
```
decentralion: (my editor automatically removes trailing whitespace.)

decentralion:
@whyrusleeping updated as requested.

actors.md
anorth: This arithmetic is all integer arithmetic of attofil. The `/` operator means truncation by default almost everywhere. Even if we specified decimal arithmetic, we would still need to define a specific precision and rounding mode. It's ok to specify rounding up, but the `ceil` below doesn't do anything to integers.

Rounding other than truncation towards zero will generally be harder to implement. For example, Go's `big.Int.Div` only truncates (but we can simulate the other modes with `DivMod`).

dignifiedquire:
created an issue for defining math: https://github.com/filecoin-project/specs/issues/365

PR #326, opened 2019-06-03
replace pledge symbol with message value

laser:
## Why is this PR needed?

The `pledge` symbol is not defined in the enclosing function body.

## What's in this PR?

Replace `pledge` symbol with message value. The FIL in the message will be used to set the balance of the newly-created storage miner actor.

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -191,7 +191,12 @@ func CreateStorageMiner(worker Address, sectorSize BytesAmount, pid PeerID) Addr
 		Fatal("Unsupported sector size")
 	}
 
-	newminer := InitActor.Exec(MinerActorCodeCid, EncodeParams(pubkey, msg.Value, sectorSize, pid))
+	newminer := InitActor.Exec(MinerActorCodeCid, EncodeParams(pubkey, sectorSize, pid))
+
+	// Send the newly-created miner the FIL sent to the storage market actor by
+	// the miner's owner. The miner will draw collateral from this FIL when
+	// committing sectors to the network.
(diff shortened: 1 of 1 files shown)


whyrusleeping:
The parameters to the miner actor constructor are the worker address, the sector size, and the peer ID. we should amend this to be:

```go
newminer := InitActor.Exec(MinerActorCodeCid, EncodeParams(worker, sectorSize, pid))
```

laser:
@whyrusleeping 

@whyrusleeping 

> The parameters to the miner actor constructor are the worker address, the sector size, and the peer ID.

OK. How should we, in the psuedocode, communicate the transfer of FIL to the newly-created miner actor from the newly-created storage miner actor's owner?

whyrusleeping:
@laser hrm... maybe just put it as a comment?

laser:
@whyrusleeping Done!

anorth:
I would think this needs to be more than a comment. The value will have been implicitly transferred to the storage miner actor, and needs to be explicitly sent using a primitive available to actors.

Actors clearly need a way of sending a message, with or without a method, and with or without params. The multisig actor spec already invokes a `[vm.]Send()` method. I think this should similarly invoke it explicitly to transfer the value off the storage market actor.

mishmosh:
Closing since outdated. `newminer` is now `createMiner`, and defined in https://filecoin-project.github.io/specs/#systems__filecoin_mining__storage_mining.

Please reopen if you disagree.

PR #332, opened 2019-06-07
Rational PoSt

dignifiedquire:


drafts/proof-of-spacetime.md
```diff
@@ -1,24 +1,141 @@
...
+  - Same as Public parameters.
+- **Public Parameters**
+  - `post_periods_count`: uint
+  - `challenge_count`: number of challenges to be asked at each iteration
+  - `sector_size`: size of the sealed sector in bytes
+  - `sectors_count`: number of sectors over which the proof is performed
+  - `challenge_bits`: number of bits in one challenge (length of a merkle path)
+  - `seed_bits`: number of bits in one challenge
+- **Public Inputs**
+  - `CommRs : [sectors_count]Hash` CommRs must be ordered by their timestamp on the blockchain.
```
laser: > CommRs must be ordered by their timestamp on the blockchain.

Right now, CommRs are ordered (in the dynamic array passed from go-filecoin through the FFI to rust-fil-proofs) by the id of the sector. Nothing enforces that `commitSector` messages originating from a miner are posted to the blockchain ordered by their sector id.

Can you explain why CommRs must be "ordered by their timestamp on the blockchain?"

drafts/proof-of-spacetime.md
```diff
@@ -1,24 +1,141 @@
...
+- **Private Inputs**
+  - `replicas: [sectors_count]SealedSector`: sealed sectors
+- **Proof**
+  - `post_proofs [post_periods_count]PoRepProof`
+
+### Methods
+
+#### `ProveStep(Step, Beacon_t, PublicParameters, PublicInputs, PrivateInputs)`
```
laser: What is a step? How does the caller of `ProveStep` obtain a value to be used as the `Step` parameter here?

drafts/proof-of-spacetime.md
```diff
@@ -1,24 +1,141 @@
...
-- An Extension to PoSt to support challenges taken from a Random Beacon 
+- An Extension to PoSt to support challenges taken from a Random Beacon
 
 ## Syntax
 
 - **PoSt Epoch**: The total time passing between Online PoReps in the PoSt Computation (in VDF, this interval is the time it takes to run a VDF and an Online PoRep prove step). We define the number of epochs as `POST_EPOCHS_COUNT`
-- **PoSt Period**: The total time it takes to run a single PoSt. If a PoSt is repeated multiple times, we define the number of periods as `POST_PERIODS_COUNT`. This can be reasoned as: `PoSt Epoch * POST_EPOCHS_COUNT`. We assume that the best  Post Period time is `MIN_POST_PERIOD_TIME` 
+- **PoSt Period**: The total time it takes to run a single PoSt. If a PoSt is repeated multiple times, we define the number of periods as `POST_PERIODS_COUNT`. This can be reasoned as: `PoSt Epoch * POST_EPOCHS_COUNT`. We assume that the best  Post Period time is `MIN_POST_PERIOD_TIME`
 - **Total proving time**: The time it takes to run a PoSt. Note that a PoSt could be a composition of multiple PoSt. This can be reasoned as: `PoSt Period * POST_PERIODS_COUNT` TODO Check with papers' syntax
```
laser: The current incarnation of the [specification](https://github.com/filecoin-project/specs/blob/master/actors.md#submitpost) shows that "proving period duration" is a function of sector size. I assume that `Total proving time` and proving period duration are the same thing. So, I have some questions (pertaining to our PoSt construction which uses the random beacon):

- Will `POST_PERIODS_COUNT` be fixed for all sector sizes?
- Will `PoSt Period` be a function of sector size?
- Will `PoSt Epoch` be a function of sector size?

drafts/proof-of-spacetime.md
```diff
@@ -1,24 +1,141 @@
...
+  - `sector_size`: size of the sealed sector in bytes
+  - `sectors_count`: number of sectors over which the proof is performed
+  - `challenge_bits`: number of bits in one challenge (length of a merkle path)
+  - `seed_bits`: number of bits in one challenge
+- **Public Inputs**
+  - `CommRs : [sectors_count]Hash` CommRs must be ordered by their timestamp on the blockchain.
+- **Private Inputs**
+  - `replicas: [sectors_count]SealedSector`: sealed sectors
+- **Proof**
+  - `post_proofs [post_periods_count]PoRepProof`
```
laser: Typo? `s/PoRepProof/PoStProof/g`

drafts/proof-of-spacetime.md
```diff
@@ -1,24 +1,141 @@
...
+
+## Base-PoSt
+
+- **Setup Parameters**
+  - Same as Public parameters.
+- **Public Parameters**
+  - `post_periods_count`: uint
+  - `challenge_count`: number of challenges to be asked at each iteration
+  - `sector_size`: size of the sealed sector in bytes
+  - `sectors_count`: number of sectors over which the proof is performed
```
ZenGround0: Here is a practical question: what is the concrete plan for allowing miners to flexibly prove over many sectors without submitting multiple posts to the chain or running two posts at once?

Is it simply to generate a (`pk`, `vk`) pair for every circuit with sector count {1, 2, 3, 4, 5, ... Max Sectors per miner}, and then have provers and verifiers use the correct key during proof generation and verification?  

Since we're probably talking about gigabytes per key what is the feasible limit for distinct sectors count circuits?

drafts/proof-of-spacetime.md
ZenGround0: Additionally our current plan is to allow reuse of deprecated (through faults or completion) sector IDs.  It seems like ordering by sectorID is sufficient for provers and verifiers to have a stable ordering.

drafts/proof-of-spacetime.md
ZenGround0: I'm now remember something about being able to pass in subtrees of zeros as private inputs along with a sector's count to a single circuit.  Does that sound right?

laser: @ZenGround0 

> what is the concrete plan for allowing miners to flexibly prove over many sectors without submitting multiple posts to the chain or running two posts at once?

@nicola had the idea (and I [have put up a PR to match](https://github.com/filecoin-project/specs/pull/348)) to cap the number of sectors which a single miner actor can commit to the network to 16384 (`POST_SECTORS_COUNT`, or 1PiB/64GiB sectors). This number forms a part of the PoSt's public inputs, and specifies the number of sectors for which a single PoSt will be generated. If a single miner operator wishes to commit more than 16384 sectors to the network, they'll have to provision more than one storage miner actor (and generate multiple, _independent_ PoSts in parallel).

Note that the aforementioned PR has not yet been ratified into the spec.

proof-of-spacetime.md
```diff
@@ -0,0 +1,200 @@
...
+                // faulty sector, generate a new challenge
+                faults.Add(sector)
+                attempt += 1
+            } else {
+                // no fault, move on to the next challenge
+                inclusionProofs[n] = inclusionProof
+                break 'inner
+            }
+        }
+    }
```
nicola: I am actually not too sure if I buy this behavior: if you announce faults only when challenged on them, then you might have a lot more of unreported faults, and this is bad, since if the challenges are ~200, then miners would only report at most 200 of them.

Hence, we:
- EITHER challenge every sector at least once (seems pretty expensive since we want to support 16k sectors per post (in case we do so, we must use the CommA trick, instead of passing one CommR per sector)
- OR miners must announce their faults before the POST_CHALLENGE_TIME, in this way they are required to report them all.

proof-of-spacetime.md
```diff
@@ -0,0 +1,200 @@
...
+        attempt := 0
+        'inner: for {
+            challenge := DerivePoStChallenge(seed, n, faults, attempt)
+
+            // check if we landed in a faulty sector
+            sector := challenge % sectorSize
+            if !faults.Contains(sector) {
+                // Valid challenge
+                challenges[n] = challenge
+                break 'inner
```
nicola: is there a simpler way that define how this works without break?

it's very difficult to reason through this

proof-of-spacetime.md
```diff
@@ -0,0 +1,200 @@
...
+for c in range POST_CHALLENGES_COUNT {
+  // Inclusion Proofs Checks
+  assert(MerkleTreeVerify(CommRs[c], InclusionPath[c], IncludionProof[c], InclusionValue[c]))
+}
+```
+
+#### Verification of PoSt proof
+
+- SNARK proof check: **Check** that given the SNARK proof and the public inputs, the SNARK verification outputs true
+- Challenges check: rederive the challenges, based on the seed, and check that they are equal.
```
nicola: no need to check that they are equal.

The verifier does not have any challenges, this step is done just to derive them.

proof-of-spacetime.md
dignifiedquire: fixed

proof-of-spacetime.md
dignifiedquire: fixed

drafts/proof-of-spacetime.md
dignifiedquire: - epoch and periods_count are no more
- very likely period will _not_ be a function of the sector size, but this is still being discussed

drafts/proof-of-spacetime.md
dignifiedquire: as @laser said there will be a maximum of sectors that can be submitted per post (and which will be the limit for a single storage miner actor). Details on how this will work exactly will be added into this doc

proof-of-spacetime.md
dignifiedquire: Faults need to be announced as per the last discussions.

mining.md
```diff
@@ -192,6 +183,13 @@ func (state StateTree) LookupPublicKey(a Address) PubKey {
     }
     Fatal("can only look up public keys for BLS controlled accounts")
 }
+
+// Get the canonical randomness from a block.
+func GetRandFromBlock(blk) []byte {
+	randomnessLookbackTipset := RandomnessLookback(blk)
+	lookbackTicket := minTicket(randomnessLookbackTipset)
+	blake2b(lookbackTicket)
```
whyrusleeping:
```suggestion
	return blake2b(lookbackTicket)
```

proof-of-spacetime.md
```diff
@@ -0,0 +1,185 @@
...
+#### Generation
+
+```go
+func GeneratePoSt(sectorSize BytesAmount, sectors []commR, seed []byte, faults FaultSet) PoStProof {
+    // Generate the Merkle Inclusion Proofs + Faults
+
+    inclusionProofs := []
+    challenges := DerivePoStChallenges(sectorSize, seed, faults)
+
+    for n in 0..POST_CHALLENGES_COUNT {
```
whyrusleeping: try to stick to valid go syntax so we can use the code formatter

whyrusleeping:
overall +1, great to finally have this written down in a spec

proof-of-spacetime.md
```diff
@@ -0,0 +1,187 @@
...
+        challenge := challenges[n]
+        sector := challenge % len(sectors)
+
+        // Leaf index of the selected sector
+        challenge_value = challenge % sectorSize
```
porcuquine: Is it really safe to derive the challenged sector and challenged leaf in this way? This seems to heavily correlate the two. For example, suppose `len(sectors)` and `sector_size / 32` were the same. (I know this is unrealistic.) In that case, the challenged leaf would be entirely determined by the challenged sector (for a given set of sectors). That means a miner whose sectors are static (at capacity) would always receive the same challenges for each sector and could simply cache the proof for one leaf (no data storage required). 

This is obviously an extreme case, but its existence is troubling. A more plausible example: suppose a miner is storing 1TiB, with a sector size of 1GiB. That would make `len(sectors) = 1024` and `sector_size = 1073741824 `, with one being 1048576 times the other. In this case, it seems that there are only 1048576 possible challenged leafs for any given sector. If true, that would allow storage of only these 1048576 proofs (each 32x25 = 800 bytes) for a 22% space reduction.

As far as I can tell, this issue exists any time the sector size is a multiple of the of the number of sectors being proved. If this is true, rather than come up with clever ways to reduce the risk, I think we should probably use entirely independent challenge values to select the sector and the leaf within that sector. (There are other strategies, but I think what I propose is simple and safe.)

To see why the current challenge derivation method is problematic, consider:

When SECTOR_SIZE = M * SECTOR_COUNT, then for any given challenge C such that C % SECTOR_COUNT = CHOSEN_SECTOR, there are only M distinct values of C % SECTOR_SIZE. The possible chosen leaves are: i * M + CHOSEN_SECTOR for i in [1..M]. Any C which yields a different leaf when selecting by C % SECTOR_SIZE will also yield a different sector when selecting by C % SECTOR_COUNT.

Make sense?

cc: @nicola @nikkolasg

dignifiedquire: proposal: given that we already have 32bytes of randomness and only need 64bits of it at the moment we can do 
- take the low 8 bytes to select the sector
- take the high 8 bytes to select the leaf

thoughts?

proof-of-spacetime.md
```diff
@@ -0,0 +1,189 @@
...
+
+    return challenges
+}
+
+// Derive a single challenge for PoSt.
+func DerivePoStChallenge(seed []byte, n Uint, attempt Uint) Uint {
+    n_bytes := WriteUintToLittleEndian(n)
+    data := concat(seed, n_bytes, WriteUintToLittleEndian(attempt))
+    challenge := blake2b(data)
+    ReadUintLittleEndain(challenge)
```
porcuquine: 'Endian' (spelling)

proof-of-spacetime.md
porcuquine: It might be a little cleaner just to take the first 8 bytes for one and the next 8 bytes for the other, but either way would address this issue.

drafts/proof-of-spacetime.md
dignifiedquire: defined ordering in latest version clearly

drafts/proof-of-spacetime.md
dignifiedquire: no steps anymore

drafts/proof-of-spacetime.md
dignifiedquire: most of those words are gone, relationship to sector sizes will come in a future pr

drafts/proof-of-spacetime.md
dignifiedquire: there is no maximum in the latest construction, all sectors can be proving in a single PoSt

proof-of-spacetime.md
dignifiedquire: I have switched to this, makes it clearer how challenge derivation works as well. I am not a 100% sure it is an issue, but the new solution should be fully safe

actors.md
```diff
@@ -530,6 +532,7 @@ type MinerInfo struct {
 | `ChangeWorker` | 13 |
 | `IsSlashed` |  14 |
 | `IsLate` | 15 |
+| `AddFault` | 16 |
```
nicola: you once mentioned that we needed "recoveredFault" or something, do we still need that?

proof-of-spacetime.md
```diff
@@ -0,0 +1,184 @@
+## Proof-of-Spacetime
+
+This document describes Rational-PoSt, the Proof-of-Spacetime used in Filecoin.
+
+## Rational PoSt
+
+### Definitions
+
+- **POST_PROVING_PERIOD**: The time interval in which a PoSt has to be submitted.
```
nicola: put this into a table, and put a number (we can stick to ~1 day for now)

proof-of-spacetime.md
```diff
@@ -0,0 +1,184 @@
...
+- **POST_CHALLENGE_TIME**: The time offset at which the actual work of generating the PoSt should be started. This is some delta before the end of the `Proving Period`, and as such less then a single `Proving Period`.
```
nicola: this should be backed with some research on:
- how long it takes to get a message in other currencies, for now, we can set to 2h
- how long it takes to generate a proof

add this as a todo note or as a design note ^

proof-of-spacetime.md
```diff
@@ -0,0 +1,184 @@
...
+
+### High Level API
+
+#### Fault Detection
+
+Fault detection happens over the course of the life time of a sector. When the sector is for some reason unavailable, the miner is responsible to post an up to date `AddFaults` message to the chain. The PoSt generation then takes the latest available `faults` of the miner to generate a PoSt matching the committed sectors and faults.
```
nicola: `faultSet`

proof-of-spacetime.md
```diff
@@ -0,0 +1,184 @@
...
+
+At the beginning of a new proving period all faults are reset, and if they persist the miner needs to resubmit an `AddFaults` message.
```
nicola: add a note on what happens if they are "lost" sectors - e.g. put them in them in the done set

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+    n_bytes := WriteUintToLittleEndian(n)
+    data := concat(seed, n_bytes, WriteUintToLittleEndian(attempt))
+    challenge_bytes := blake2b(data)
+
+    sector_challenge := ReadUintLittleEndian(challenge_bytes[0..8])
+    leaf_challenge := ReadUintLittleEndian(challenge_bytes[8..16])
+
+    Challenge {
+        Sector: sector_challenge % sectorCount,
+        Leaf: leaf_challenge % (sectorSize / NODE_SIZE),
```
laser: Question: Does the protocol define `NODE_SIZE` anywhere? I know it to be 32, but that's from looking at the software.

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+```go
+type Challenge struct {
+    Sector Uint
+    Leaf Uint
+}
+```
+
+#### Challenge Derivation
+
+```go
```
laser: The code block is marked as `go`, but the code inside it is a mixture of Go and Rust (e.g. `for n in 0..POST_CHALLENGES_COUNT`).

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+
+### PoSt Circuit
+
+#### Public Parameters
+
+*Parameters that are embeded in the circuits or used to generate the circuit*
+
+- `POST_CHALLENGES_COUNT: UInt`: Number of challenges.
+- `POST_TREE_DEPTH: UInt`: Depth of the Merkle tree. Note, this is `(log_2(Size of original data in bytes/32 bytes per leaf))`.
+- `SECTOR_SIZE: UInt`:
```
laser: Nit: Dangling `:`.

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+
+#### Public Inputs
+
+*Inputs that the prover uses to generate a SNARK proof and that the verifier uses to verify it*
+
+- `CommRs: [POST_CHALLENGES_COUNT]Fr`: The Merkle tree root hashes of all CommRs, ordered to match the inclusion paths and challenge order.
```
laser: Nit: I suggest that you change this wording to:

> The Merkle tree root hashes of all replicas.

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+- `InclusionPaths: [POST_CHALLENGES_COUNT]Fr`: Inclusion paths for the replica leafs, ordered to match the `CommRs` and challenge order. (Binary packed bools)
+
+#### Private Inputs
+
+*Inputs that the prover uses to generate a SNARK proof, these are not needed by the verifier to verify the proof*
+
+- `InclusionProofs: [POST_CHALLENGES_COUNT][TREE_DEPTH]Fr`: Merkle tree inclusion proofs, ordered to match the challenge orderd.
```
laser: Nit: `s/orderd/order/g`.

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+In high level, we do 1 check:
+
+1. **Inclusion Proofs Checks**: Check the inclusion proofs
+
+##### Details
+
+```go
+for c in range POST_CHALLENGES_COUNT {
+  // Inclusion Proofs Checks
+  assert(MerkleTreeVerify(CommRs[c], InclusionPath[c], IncludionProof[c], InclusionValue[c]))
```
laser: Nit: `s/IncludionProof/InclusionProof/g`

proofs.md
```diff
@@ -24,20 +24,19 @@ Throughout this document, the following definitions are used:
...
+- __*Multi-SNARK proof:*__ a block of one or more __*SNARK proofs*__, each proving partition of the total set of challenges to be proved.
 - __*Merkle inclusion proof:*__ a proof (whose format is unspecified here) that a given leaf block is contained within a merkle tree whose root is the __*commitment*__ associated with the proof.
 - __*commitment:*__ an opaque block of data to which a prover 'commits', enabling subsequent proofs which cannot be validly constructed unless the __*commitment*__ itself was validly constructed. For example: the output of a suitable pseudorandom collision-resistant hash function may serve as a __*commitment*__ to the data which is the preimage of that hash. Publication of the __*commitment*__ proves that the creator was in possession of the preimage at the time the __*commitment*__ was generated.
 - __*prover:*__ the party who generates a proof, in Filecoin it's always the Storage Miner.
 - __*verifier:*__ the party who verifies a proof generated by a __*prover*__, in Filecoin it's a full node.
-- __*sectors count:*__ the number of sectors over which a proof-of-spacetime is performed (`POST_SECTORS_COUNT`, `16 PiB / 64 GiB = 262144`).
```
laser: Does eliminating `POST_SECTORS_COUNT` mean that we can cap the number of sectors which a miner can commit to the network to MAX(uint64)? If so, would you please update the `actors.md` `CommitSector` method? It checks `if miner.Sectors.Size() >= POST_SECTORS_COUNT { ... fail` now.

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+- `POST_CHALLENGES_COUNT: UInt`: Number of challenges.
+- `POST_TREE_DEPTH: UInt`: Depth of the Merkle tree. Note, this is `(log_2(Size of original data in bytes/32 bytes per leaf))`.
+- `SECTOR_SIZE: UInt`:
+
+#### Public Inputs
+
+*Inputs that the prover uses to generate a SNARK proof and that the verifier uses to verify it*
+
+- `CommRs: [POST_CHALLENGES_COUNT]Fr`: The Merkle tree root hashes of all CommRs, ordered to match the inclusion paths and challenge order.
+- `InclusionPaths: [POST_CHALLENGES_COUNT]Fr`: Inclusion paths for the replica leafs, ordered to match the `CommRs` and challenge order. (Binary packed bools)
```
laser: > (Binary packed bools)

Is this detail relevant to this part of the spec? If not, I suggest terminating it. If so, ignore me.

actors.md
laser: The document uses `AddFaults` elsewhere - but this says `AddFault`.

actors.md
```diff
@@ -641,40 +644,26 @@ func CollateralForPower(power BytesAmount) TokenAmount {
...
 
-{{% notice todo %}}
-TODO: GenerationAttackTime
-{{% /notice %}}
-
 ```go
-func SubmitPost(proofs PoStProof, faults []FaultSet, recovered Bitfield, done Bitfield) {
-	if msg.From != miner.Worker {
+func SubmitPost(proofs PoStProof, doneSet Bitfield) {
+	if msg.From != self.Worker {
```
laser: Nit: Indentation.

actors.md
```diff
@@ -688,26 +677,25 @@ func SubmitPost(proofs PoStProof, faults []FaultSet, recovered Bitfield, done Bi
 		TransferFunds(msg.From, msg.Value-feesRequired)
 	}
 
-	if !CheckPostProof(miner.SectorSize, proof, faults) {
+    seed := GetRandFromBlock(self.ProvingPeriodStart + POST_CHALLENGE_TIME)
+    faultSet := self.faultSet
+
+	if !VerifyPost(self.SectorSize, self.provingSet, seed, proof, faultSet) {
```
laser: Nit: This should be `VerifyPoSt` (note capitalization).

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+    for challenge in challenges {
+        // Leaf index of the selected sector
+        inclusionProof, isFault := GenerateMerkleInclusionProof(challenge.Sector, challenge.Leaf)
+        if isFault {
+            // faulty sector, need to post a fault to the chain and try to recover from it
+            return Fatal("Detected late fault")
+        }
+
+        inclusionProofs[n] = inclusionProof
+		sectorsSorted[i] = sectors[sector]
```
laser: I don't see the `sectors` (in `sectors[sector]`) symbol defined in this function anywhere.

data-structures.md
```diff
@@ -291,8 +291,14 @@ type Bitfield Bytes
 
 ### SectorSet
 
+A sector set stores a mapping of sector IDs to the respective `commR`s.
+
+```sh
+type SectorSet map{Uint:Bytes}
```
laser: Suggestion: The spec uses values of type `SectorID` to index into the miner's sector set. This PR (perhaps by accident) specifies that type to be `Uint`. For consistency, I suggest that you specify the `SectorID` type in the `data-structures.md` document.

actors.md
```diff
@@ -688,26 +677,25 @@ func SubmitPost(proofs PoStProof, faults []FaultSet, recovered Bitfield, done Bi
...
+	// Remove doneSet from the current sectors
+	miner.Sectors.Subtract(doneSet)
 
-	// update miner power to the amount of data actually proved during
-	// the last proving period.
+	// Update miner power to the amount of data actually proved during the last proving period.
 	oldPower := miner.Power
 
-	miner.Power = (miner.ProvingSet.Size() - allFaults.Count()) * miner.SectorSize
+	miner.Power = (miner.ProvingSet.Size() - faultSet.Count()) * miner.SectorSize
```
laser: You just `// clear fault set` above, so the `- faultSet.Count()` isn't going to do much.

actors.md
```diff
@@ -688,26 +677,25 @@ func SubmitPost(proofs PoStProof, faults []FaultSet, recovered Bitfield, done Bi
 		TransferFunds(msg.From, msg.Value-feesRequired)
 	}
 
-	if !CheckPostProof(miner.SectorSize, proof, faults) {
+    seed := GetRandFromBlock(self.ProvingPeriodStart + POST_CHALLENGE_TIME)
+    faultSet := self.faultSet
```
laser: Nit: `self.` is a Rust idiom, but you're in a `go` code block.

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+{{% /notice %}}
+
+### High Level API
+
+#### Fault Detection
+
+Fault detection happens over the course of the life time of a sector. When the sector is for some reason unavailable, the miner is responsible to submit the known `faults` using the `AddFaults` message to the chain.
+The PoSt generation then takes the latest available `faults` of the miner to generate a PoSt matching the committed sectors and faults.
+
+At the beginning of a new proving period all faults are reset, and if they persist the miner needs to resubmit an `AddFaults` message.
```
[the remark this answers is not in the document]
dignifiedquire: It is assumed that they are recovered, so they will potentially get challenged on them.

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+## Rational PoSt
+
+### Definitions
+
+| Name | Value |Description |
+|------|-------|------------|
+| `POST_PROVING_PERIOD` | `2880` blocks  (~24h) | The time interval in which a PoSt has to be submitted. |
+| `POST_CHALLENGE_TIME` | `240` blocks (~2h) | The time offset at which the actual work of generating the PoSt should be started. This is some delta before the end of the `Proving Period`, and as such less then a single `Proving Period`. |
+
+{{% notice todo %}}
```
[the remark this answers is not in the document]
dignifiedquire: we use this to get nice renderings on https://filecoin-project.github.io/specs/

mining.md
```diff
@@ -42,24 +42,15 @@ TODO: sectors need to be globally unique. This can be done either by having the
...
-    var faults []FaultSet
-    for t := 0; t < ProvingPeriod; t += ReseedPeriod {
-        seeds = append(seeds, GetSeedFromBlock(startTime+t))
-        proof, faultset := GenPost(sectors, seeds[t], vdfParams)
-        proofs = append(proofs, proof)
-        faults = append(faults, faultset)
-    }
-    return GenPostSnark(sectorSize, sectors, seeds, proofs), faults
+func ProveStorage(sectorSize BytesAmount, sectors []commR, faults FaultSet) PoStProof {
+    seed := GetRandFromBlock(startTime + POST_CHALLENGE_TIME)
```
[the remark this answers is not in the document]
dignifiedquire: yes, thank you for spotting

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+    Sector Uint
+    Leaf Uint
+}
+```
+
+#### Challenge Derivation
+
+```go
+// Derive the full set of challenges for PoSt.
+func DerivePoStChallenges(seed Seed, faults FaultSet, sectorSize: Uint, sectorCount: Uint) [POST_CHALLENGES_COUNT]Challenge {
```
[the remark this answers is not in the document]
dignifiedquire: rust and go are mixing in my head..

proof-of-spacetime.md
dignifiedquire: yeah it is defined in the zigzag spec

data-structures.md
dignifiedquire: works for me

actors.md
dignifiedquire: faultSet is a local copy, I clear `self.faultSet`

actors.md
dignifiedquire: @whyrusleeping  and I agreed to use `self` to refer to the current actor in the pseudo code. should document this somewhere

proof-of-spacetime.md
dignifiedquire: It is relevant and explained in more detail in the zigzag spec I believe

proofs.md
dignifiedquire: done

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+    challenges := []
+
+    for n in 0..POST_CHALLENGES_COUNT {
+        attempt := 0
+        while challenges[n] == nil {
```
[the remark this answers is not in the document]
dignifiedquire: you are right, should be fixed

actors.md
laser: Ah hah. Thanks.

mining.md
```diff
@@ -42,24 +42,15 @@ TODO: sectors need to be globally unique. This can be done either by having the
...
-    var faults []FaultSet
-    for t := 0; t < ProvingPeriod; t += ReseedPeriod {
-        seeds = append(seeds, GetSeedFromBlock(startTime+t))
-        proof, faultset := GenPost(sectors, seeds[t], vdfParams)
-        proofs = append(proofs, proof)
-        faults = append(faults, faultset)
-    }
-    return GenPostSnark(sectorSize, sectors, seeds, proofs), faults
+func ProveStorage(sectorSize BytesAmount, sectors []commR, faults FaultSet) PoStProof {
+    seed := GetRandFromBlock(miner.ProvingPeriodStart + ProvingPeriodDuratoin(sectorSize) + POST_CHALLENGE_TIME)
```
ingar: Should this be `- POST_CHALLENGE_TIME`?

proof-of-spacetime.md
```diff
@@ -0,0 +1,198 @@
...
+
+{{% notice todo %}}
+**TODO**: The above values are tentative and need both backing from research as well as detailed reasoning why we picked them.
+{{% /notice %}}
+
+### High Level API
+
+#### Fault Detection
+
+Fault detection happens over the course of the life time of a sector. When the sector is for some reason unavailable, the miner is responsible to submit the known `faults` using the `AddFaults` message to the chain.
```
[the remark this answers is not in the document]
ingar: Reading this PR, aren't all faults going to be ones that go-filecoin doesn't know about?  Since faults are reset at the beginning of a proving period?  Or is that we could be challenged on the same (possibly faulted) sector more than once in a proving period...

actors.md
```diff
@@ -466,6 +466,9 @@ type StorageMinerActorState struct {
 	## when a PoSt is submitted (not as each new sector commitment is added).
     provingSet &SectorSet
 
+    ## Current self reported faults.
```
anorth:
```suggestion
    ## Faulty sectors reported since last SubmitPost.
```

actors.md
```diff
@@ -599,11 +602,6 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
...
 	}
 
-	// ensure that the miner cannot commit more sectors than can be proved with a single PoSt
-	if miner.Sectors.Size() >= POST_SECTORS_COUNT {
-		Fatal("too many sectors")
-	}
-
 	// Note: There must exist a unique index in the miner's sector set for each
 	// sector ID. The `faults`, `recovered`, and `done` parameters of the
 	// SubmitPoSt method express indices into this sector set.
```
anorth: Request: could we pick a consistent capitalisation of `SubmitPo[sS]t`?

actors.md
```diff
@@ -688,26 +672,25 @@ func SubmitPost(proofs PoStProof, faults []FaultSet, recovered Bitfield, done Bi
 		TransferFunds(msg.From, msg.Value-feesRequired)
 	}
 
-	if !CheckPostProof(miner.SectorSize, proof, faults) {
+    seed := GetRandFromBlock(self.ProvingPeriodStart + ProvingPeriodDuration(self.SectorSize) - POST_CHALLENGE_TIME)
+    faultSet := self.faultSet
+
+	if !VerifyPoSt(self.SectorSize, self.provingSet, seed, proof, faultSet) {
```
anorth: Something somewhere should document that this will fail if a fault has been declared at any point since `self.ProvingPeriodStart + ProvingPeriodDuration(self.SectorSize) - POST_CHALLENGE_TIME`

In case a miner does experience a fault after the challenge window starts, what mechanism will move their proving period forward to the subsequent window so that they can try again with a new fault set?
This is currently inconsistent with code higher up that prevents any PoSt submission after the generation attack time.

actors.md
anorth: IMO the state updates could be re-arranged into a clearer block at the end of the method:
```
// Update miner power to the amount of data actually proved during the last proving period.
self.Power = (self.ProvingSet.Size() - self.FaultSet.Count()) * self.SectorSize

// Done sectors no longer need to be proven
self.Sectors.Subtract(doneSet)
self.ProvingSet = self.Sectors

// Faults are assumed temporary unless the miner has also declared the sector "done"
self.FaultSet = EmptySectorSet()

... // deal state
```

actors.md
anorth: `self` is not used consistently below, `miner` appears too.

proof-of-spacetime.md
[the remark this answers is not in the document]
anorth: "the miner is responsible to submit the known faults..."  *before the PoSt challenge window begins*. If a miner experiences a fault after the PoSt challenge window opens but before computing the PoSt, they will be unable to submit a PoSt in this period, will lose power, and potentially face other penalties.

proof-of-spacetime.md
porcuquine: Thanks, @dignifiedquire.

actors.md
dignifiedquire: correct, but I am trying to keep changes minimal

dignifiedquire: correct, but I am trying to keep changes minimal

actors.md
dignifiedquire: PoSt is the default, but hard to keep correct everywhere

actors.md
```diff
@@ -1132,6 +1055,25 @@ func IsSlashed() (bool) {
 }
 ```
 
+#### `AddFaults`
+
+**Parameters**
+
+```sh
+type SubmitPost struct {
```
whyrusleeping: wrong type name here

proof-of-spacetime.md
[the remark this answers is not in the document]
whyrusleeping: no, the miner just has to submit the faults before they submit their post.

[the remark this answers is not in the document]
anorth: I was led to believe that the seed for challenges would be sampled at a fixed chain height, and that faults reported before this height would affect the seed/challenges, but those reported afterwards would not be taken into account (and hence might be challenged). I would love to be wrong, but let's be very clear here.

If I'm wrong, a miner could sample for challenges, discover a challenge they don't like, report a fault, sample again for a different value, 🔁 . Is that ok?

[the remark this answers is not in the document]
dignifiedquire: Validation expects the challenge to be taken at a specific block height, including the faults. (I need to update somethings in the PR to make this clearer) So you can _not_ change your challenge in any way once you know about it.

nicola:
@anorth, from this proposal: if a miner loses a file JUST BEFORE the challenge time, then they miss the current PoSt. I can't see a way for which we can allow the miner to pass still pass the proof.

@dignifiedquire tell me if this scenario is correct, but what I am thinking is: if the miner can't submit this post, they could declare this post as "missed", setting all the sectors to faulty, then re-adding them in the next proving period.

dignifiedquire:
> @dignifiedquire tell me if this scenario is correct, but what I am thinking is: if the miner can't submit this post, they could declare this post as "missed", setting all the sectors to faulty, then re-adding them in the next proving period.

This is not correct with master + this PR, this will only be possible once #374 lands

mining.md
dignifiedquire: yes, thank you

actors.md
dignifiedquire: cleaned up

mining.md
dignifiedquire: fixed

actors.md
dignifiedquire: This was actually tricker than I thought, I added more words, as well as more tracking to make this correct. please review again

dignifiedquire:
> Blocking: there is no recovery path if a miner experiences a fault that they cannot report in time. There needs to be some mechanism for a miner to declare their proving period bust and prepare for the next one.

They don't need to declare anything, they fault on that proving period and try to get in a PoSt in the next proving period. 
Solving the fact that this is (too) heavily penalized as of current `master` is work that needs to happen in #374 not in this PR.

actors.md
```diff
@@ -688,68 +675,36 @@ func SubmitPost(proofs PoStProof, faults []FaultSet, recovered Bitfield, done Bi
 		TransferFunds(msg.From, msg.Value-feesRequired)
 	}
 
-	if !CheckPostProof(miner.SectorSize, proof, faults) {
+    seed := GetRandFromBlock(self.ProvingPeriodStart + ProvingPeriodDuration(self.SectorSize) - POST_CHALLENGE_TIME)
```
anorth:
```suggestion
    seed := GetRandFromBlock(self.ProvingPeriodEnd - POST_CHALLENGE_TIME)
```

The miner state doesn't have `ProvingPeriodStart`.

actors.md
```diff
@@ -1132,6 +1059,33 @@ func IsSlashed() (bool) {
...
+type AddFaults struct {
+    faults FaultSet
+} representation tuple
+```
+
+**Algorithm**
+
+```go
+func AddFaults(faults FaultSet) {
+    let challengeBlockHeight = self.ProvingPeriodStart + ProvingPeriodDuration(self.SectorSize) - POST_CHALLENGE_TIME
```
anorth:
```suggestion
    let challengeBlockHeight = self.ProvingPeriodEnd - POST_CHALLENGE_TIME
```

proof-of-spacetime.md
```diff
@@ -0,0 +1,197 @@
...
+This document describes Rational-PoSt, the Proof-of-Spacetime used in Filecoin.
+
+## Rational PoSt
+
+### Definitions
+
+| Name | Value |Description |
+|------|-------|------------|
+| `POST_PROVING_PERIOD` | `2880` blocks  (~24h) | The time interval in which a PoSt has to be submitted. |
+| `POST_CHALLENGE_TIME` | `240` blocks (~2h) | The time offset at which the actual work of generating the PoSt should be started. This is some delta before the end of the `Proving Period`, and as such less then a single `Proving Period`. |
```
nicola: The challenge for the current proving period can only be retrieved after POST_PROVING_PERIOD-POST_CHALLENGE_TIME

nicola: maybe you should define it in the opposite way

proof-of-spacetime.md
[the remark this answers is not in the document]
nicola: Note 1: faults can only be submitted before POST_CHALLENGE_TIME. So miners can't wait for the challenge to appear in order to report faults, and as dig is saying the can't affect the challenge once it appears.
Note 2: one does not know the challenge ahead of time, so they can't really try to minimize the reported faults. So, yes, reporting faults would affect the challenge, but they can't predict how, hence not a problem.

Does this clarify?

proof-of-spacetime.md
```diff
@@ -0,0 +1,197 @@
...
+{{% notice todo %}}
+**TODO**: The above values are tentative and need both backing from research as well as detailed reasoning why we picked them.
+{{% /notice %}}
+
+### High Level API
+
+#### Fault Detection
+
+Fault detection happens over the course of the life time of a sector. When the sector is for some reason unavailable, the miner is responsible to submit the known `faults` using the `AddFaults` message to the chain.
+Only faults which have been reported at challenge time, will be accounted for. If any other faults have occured the miner can not submit a valid PoSt for this proving period.
```
nicola: "Only faults which have been reported at challenge time, will be accounted for. If a miner has faults after the challenge time, then they cannot submit a valid PoSt for this proving period."

proof-of-spacetime.md
```diff
@@ -0,0 +1,197 @@
...
+
+The PoSt generation then takes the latest available `faults` of the miner to generate a PoSt matching the committed sectors and faults.
+
+At the beginning of a new proving period all faults are reset, and if they persist the miner needs to resubmit an `AddFaults` message.
```
nicola: At the beginning of the next proving period all faults are reset and assumed to be recovered. A miner must be able to include those sectors in the next proof submission or re-submit the faults via an`AddFaults` message.

proof-of-spacetime.md
```diff
@@ -0,0 +1,197 @@
...
+	sectorsSorted := []
+    challenges := DerivePoStChallenges(seed, faults, sectorSize, len(sectors))
+
+    for i := 0; i < len(challenges); i++ {
+        challenge := challenges[i]
+
+        // Leaf index of the selected sector
+        inclusionProof, isFault := GenerateMerkleInclusionProof(challenge.Sector, challenge.Leaf)
+        if isFault {
+            // faulty sector, need to post a fault to the chain and try to recover from it
```
nicola: Comment should be more like:

"Detected late fault: The data needed for this post submission is missing, report either as a fault (if you are planning to recover it) and also as a done the missing sector (if you cannot recover)."

proof-of-spacetime.md
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    return challenges
+}
+
+// Derive a single challenge for PoSt.
+func DerivePoStChallenge(seed Seed, n Uint, attempt Uint, sectorSize Uint, sectorCount Uint) Challenge {
+    n_bytes := WriteUintToLittleEndian(n)
+    data := concat(seed, n_bytes, WriteUintToLittleEndian(attempt))
```
nicola: create an issue on whether or not we need to add the "minerId" in this^

proof-of-spacetime.md
```diff
@@ -0,0 +1,197 @@
...
+    challenge_bytes := blake2b(data)
+
+    sector_challenge := ReadUintLittleEndian(challenge_bytes[0..8])
+    leaf_challenge := ReadUintLittleEndian(challenge_bytes[8..16])
+
+    return Challenge {
+        Sector: sector_challenge % sectorCount,
```
nicola: this does not give a uniform distribution, but in practice it should work for this case, worth creating an issue.

If sectorsCount does not divide the 8 bytes, then distribution is not uniform. For example: say that the sector_challenges goes from 0-20, and sectors_count is 17, then sector 0,1,2 are more likely to be queried.

Check how others do this, ping @irenegiacomelli @ben-a-fisch

porcuquine: I think when this came up for EC, we decided it was fine to use division. That would mean dividing the challenge range (0..2^64) into `sector_count` sub-ranges and selecting a sector based on which sub-range the challenge falls in. I don't think introduction of `BigNum` math would be problematic here, especially if we already use that method for consensus.

porcuquine: That said, we use `mod NODES` to derive challenges within sectors already, so we tolerate the bias there.

I can see the argument that it's okay when the potential challenge set is 'large' (like node count) but if we are revisiting here, we might want to consider everywhere and be consistent.

actors.md
```diff
@@ -688,68 +675,36 @@ func SubmitPost(proofs PoStProof, faults []FaultSet, recovered Bitfield, done Bi
...
-	// combine all the fault set bitfields, and subtract out the recovered
-	// ones to get the set of sectors permanently lost
-	permLostSet = allFaults.Subtract(recovered)
+    // The next fault set becomes the current one
+    self.currentFaultSet = self.nextFaultSet
+    self.nextFaultSet = EmptySectorSet()
 
-	// burn funds for fees and collateral penalization
-	self.BurnFunds(CollateralForSize(self.SectorSize*permLostSet.Size()) + feesRequired)
+    // TODO: penalize for faults
```
whyrusleeping: too dooo

actors.md
whyrusleeping:
```suggestion
    challengeBlockHeight := self.ProvingPeriodStart + ProvingPeriodDuration(self.SectorSize) - POST_CHALLENGE_TIME
```

actors.md
```diff
@@ -1132,6 +1059,33 @@ func IsSlashed() (bool) {
...
+```sh
+type AddFaults struct {
+    faults FaultSet
+} representation tuple
+```
+
+**Algorithm**
+
+```go
+func AddFaults(faults FaultSet) {
```
whyrusleeping: I actually think we should simplify this all the way to just:
```
func AddFaults(faults FaultSet) {
  self.currentFaultSet = Merge(self.currentFaultSet, faults)
}
```

If you submit faults before you submit your post, they apply to the post you still havent submitted.

actors.md
whyrusleeping: If this is the `GetRandFromBlock` defined below in the mining.md doc, then this is incorrect. the randomness needs to come from exactly the block at `ProvingPeriodEnd - POST_CHALLENGE_TIME`. Any lookback will give attackers an advantage and degrade the security argument that RationalPoSt makes.

whyrusleeping: BLOCKING ^

mining.md
```diff
@@ -42,24 +42,18 @@ TODO: sectors need to be globally unique. This can be done either by having the
...
-        proofs = append(proofs, proof)
-        faults = append(faults, faultset)
-    }
-    return GenPostSnark(sectorSize, sectors, seeds, proofs), faults
+func ProveStorage(sectorSize BytesAmount, sectors []commR) PoStProof {
+    challengeBlockHeight := miner.ProvingPeriodStart + ProvingPeriodDuratoin(sectorSize) - POST_CHALLENGE_TIME
+
+    // Faults to be used are the currentFaultSet for the miner.
+    faults := miner.currentFaultSet
+    seed := GetRandFromBlock(challengeBlockHeight)
```
whyrusleeping: As stated in the other comment, this is incorrect. The randomness must come from exactly `challengeBlockHeight`, no lookback.

proof-of-spacetime.md
whyrusleeping:
```suggestion
| `POST_CHALLENGE_TIME` | `240` blocks (~2h) | The time offset at which the actual work of generating the PoSt must not be started earlier than. This is some delta before the end of the `Proving Period`, and as such less then a single `Proving Period`. |
```

This phrasing might make the intent a little clearer, but also maybe not. just a suggestion

actors.md
dignifiedquire: yeah, this will stay for now and a new issue for handling post faults will be created

actors.md
dignifiedquire: @sternhenri and I thought actually that the lookback was good to avoid another issue he was concerned about.

dignifiedquire: also how does the lookback give attackers an advantage? could you explain?

proof-of-spacetime.md
dignifiedquire: maybe "can not"?

proof-of-spacetime.md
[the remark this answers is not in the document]
whyrusleeping: @nicola that doesnt work. What happens if a miner experiences a fault after the PoSt deadline? They won't be able to complete their PoSt or submit a new set of faults, and will just miss their PoSt entirely (bad).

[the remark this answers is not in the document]
nicola: We already went through this. The chance that this happens is very small. You should have all of your fault submitted before you see the challenge. If you fault just after you see the challenge and just before you manage to grab the challenged data, then you fail the current post. You get penalized, it’s the equivalent of faulting every sector and then reintroduce them in the next pp.

If we are not fine with this, then PoSt, must get many more challenges, linear in the number of sectors - so quite a lot.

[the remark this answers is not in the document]
whyrusleeping: @nicola thats not an okay solution

what happens if network delays cause the fault submission to get pushed past the deadline?
what if a miner just doesnt notice their data is corrupted until they try to read it (really common)?
"sorry, you lose everything, piss off" is not a good answer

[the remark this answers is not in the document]
nicola: You “lose everything” in this case means: you temporarily loose all power and you get a penalization. This does not trigger storage collateral (would happen if this repeats five times) or pledge collateral loss.

Temporary power reduction seem fine to me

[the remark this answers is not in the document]
nicola: Another note: the SLA that we are giving to the users is that the miners will keep on storing the data, if a miner realised during challenge time that they don’t have the data, then they have been either very unlucky or offered bad service, hence this penalisation ain’t that bad

actors.md
dignifiedquire: If we simplify this, it means you are not allowed to submit any faults between challenge time and post submission. Which becomes an issue if you miss a post, as you can't post new faults anymore.

proof-of-spacetime.md
whyrusleeping: yeah, lower numbers will be `1 / ( 2^len(randomness) / numSectors)` more likely to be challenged.

In practice, that number is effectively zero.

actors.md
dignifiedquire: I have thought about this some more and you are right, we need to take the challenge directly from this block, as otherwise it can be precomputed in advance.

dignifiedquire: fixed

mining.md
dignifiedquire: fixed

proof-of-spacetime.md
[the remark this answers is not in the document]
whyrusleeping: but miners get penalized for the faults they report. I havent gotten a good answer for why this needs to be complicated like this. Let miners submit faults for their post up until they submit the post.

proof-of-spacetime.md
whyrusleeping: you'd rather have them submit no post at all than let people submit faults after the deadline? I think we're on really different pages about how this all works.

whyrusleeping: @acruikshank technically yes, but i think that they are saying that that type of slashing would be removed, which I think is a bad idea.

Having the only downside of not submitting a post for an entire proving period be 'no power for that proving period' really damages a lot of the security that rational post is built on. That basically cuts the cost of the attack by the number of proving periods you can consecutively not submit PoSts for.

actors.md
acruikshank: This would allow a miner to accept storage deals, commit and then discard sectors containing the storage, and then add faults each proving period thereafter. Since the fault set uses RLE, the cost of a the AddFault message is constant over a large number of sectors, and the cost per sector drops without bound. At some point the gas costs for faulting on all deals will drop below the value of the deals.

We're going to need some mechanism to limit the number of times a sector can be faulted or at least a mechanism for timing-out faulted sectors so deal arbitration and payment cancellation can occur.

proof-of-spacetime.md
```diff
@@ -0,0 +1,197 @@
...
+**TODO**: The penalization for faults is not clear yet.
+{{% /notice %}}
+
+#### Generation
+
+`GeneratePoSt` generates a __*Proof of Spacetime*__ over all  __*sealed sectors*__ of a single miner— identified by their `commR` commitments. This is accomplished by performing a series of merkle inclusion proofs (__*Proofs of Retrievability*__). Each proof is of a challenged node in a challenged sector. The challenges are generated pseudo-randomly, based on the provided `seed`. At each time step, a number of __*Proofs of Retrievability*__ are performed.
+
+```go
+// Generate a new PoSt.
+func GeneratePoSt(sectorSize BytesAmount, sectors SectorSet, seed Seed, faults FaultSet) PoStProof {
```
ZenGround0: This document should specify how this seed is extracted from the chain.  Specifically what height and what lookback parameter.  From conversation with @whyrusleeping and @henri I believe it is taken from the first block that begins the challenge period with no lookback parameter.

ZenGround0: If the above is correct one complication for setting `POST_CHALLENGE_TIME` is that the seed value will become invalid after generation has begun with some non-negligible probability due to head churn.  This interval will need to be long enough for an acceptable number of retries.  Equivalently we could add a lookback parameter past the chain's head churn.

whyrusleeping:
Had a chat with @nicola, i'm a bit more onboard with this now, but there are some realllllly important pieces of information that led to that that arent anywhere in this PR.

The reason that we don't want to allow any faults whatsoever after the post challenge time is because, due to the way we currently compute challenges, at that point in time, you can know every sector that will be challenged. The scenario we are trying to prevent with this approach is the following:

- Attacker has a large number of faulty sectors.
- Attacker waits for the post challenge to appear, computes the challenges, and then submits faults only for sectors that will be challenged.
- Attacker computes their post successfully, despite having far more faulty sectors than reported.

This is obviously bad. A better challenge generation construction could solve this, and would work iteratively, generating a single challenge at a time, using the faultset,  and using the output of the successfully generated proof to generate the next challenge. We don't do this as it would make the circuits more complex, and we're expending a lot of effort into making them faster. Here we see that effort externalizing some ugliness onto the protocol.

Additionally, nico and I talked about how allowing miners to just skip an entire PoSt negatively impacts the security of rational PoSt. If we want to move towards the same security as the existing PoSt construction, there must be penalties for every reported fault.

dignifiedquire:
@whyrusleeping good to hear, are there any specific parts about this that you think should be reflected in the spec?

whyrusleeping:
@dignifiedquire let's make sure the rationale for this approach is written somewhere. The spec is a fine place for this, i'd just summarize the challenge sampling stuff in a 'note' somewhere.

Then add in bits about penalties for adding faults (unless we really want to do that in a followup PR, which i'm not opposed to).

dignifiedquire:
> @dignifiedquire let's make sure the rationale for this approach is written somewhere. The spec is a fine place for this, i'd just summarize the challenge sampling stuff in a 'note' somewhere.

Done

> Then add in bits about penalties for adding faults (unless we really want to do that in a followup PR, which i'm not opposed to).

Added a note + todo for this. Lets please do this in conjuction with #379, that will make it easier to define and reason about it.

proof-of-spacetime.md
dignifiedquire: @ZenGround0 yes churn is a possible concern but we need to take the seed from that exact block, see earlier comments from @whyrusleeping.

dignifiedquire:
> I cannot see the mechanism by which this happens. The only path to updating `ProvingPeriodEnd` requires a valid PoSt to have been verified prior to that update, which a miner with a late-discovered faulty sector cannot produce. If they just wait and use the subsequent challenge, the PoSt will be rejected as being past the generation attack grace period. There must be a means to updating `ProvingPeriodEnd` without a valid proof in order for recovery to be possible.

@anorth I believe the latest code and structure allows for this. Please let me know if this works from your understanding. 

The recovery path is submitting as many faults, as necessary, such that a valid PoSt makes it in. (Yes with the penalizations on `master` this is quite harsh at the moment, but this is why we have #379, to fix that issue)

dignifiedquire:
@anorth sorry I was wrong, the code does not handle this yet properly, updating it now

actors.md
```diff
@@ -455,6 +455,12 @@ type StorageMinerActorState struct {
 	## when a PoSt is submitted (not as each new sector commitment is added).
     provingSet &SectorSet
 
+    ## Faulty sectors reported since last SubmitPost, up to the current PoSt challenge time.
+    currentFaultSet BitField
+
+    ## Faults for the next round of PoSt.
```
anorth: We can be more clear: "Faults submitted after the current proving period's challenge time but before the PoSt for that period is submitted. These become the `currentFaultSet` when a PoSt is submitted."

These fields are still a bit hard to describe in the period between when a SubmitPoSt lands and when the subsequent proving period opens, but I think we have the same idea.

actors.md
```diff
@@ -455,6 +455,12 @@ type StorageMinerActorState struct {
 	## when a PoSt is submitted (not as each new sector commitment is added).
     provingSet &SectorSet
 
+    ## Faulty sectors reported since last SubmitPost, up to the current PoSt challenge time.
```
anorth:
```suggestion
    ## Faulty sectors reported since last SubmitPost, up to the current proving period's challenge time.
```

actors.md
```diff
@@ -632,40 +635,26 @@ func CollateralForPower(power BytesAmount) TokenAmount {
...
-		return
+    // TODO: rework fault handling, for now anything later than 2 proving periods is invalid
+    if chain.now() > nextProvingPeriodEnd {
+        Fatal("PoSt submited too late")
 	} else if chain.Now() > self.ProvingPeriodEnd {
-		feesRequired += ComputeLateFee(self.power, chain.Now()-self.provingPeriodEnd)
+		feesRequired += ComputeLateFee(self.power, chain.Now() - self.provingPeriodEnd)
 	}
 
 	feesRequired += ComputeTemporarySectorFailureFee(self.sectorSize, recovered)
```
anorth:
```suggestion
	feesRequired += ComputeTemporarySectorFailureFee(self.sectorSize, self.currentFaultSet)
```

actors.md
```diff
@@ -521,6 +526,9 @@ type MinerInfo struct {
 | `IsLate` | 16 |
 | `PaymentVerifyInclusion` | 17 |
 | `PaymentVerifySector` | 18 |
+| `ReportFaultySectors` | 19 |
+| `RecoverFaultySectors` | 20 |
```
anorth: Remove these two un-specified methods from the table.

proof-of-spacetime.md
```diff
@@ -0,0 +1,211 @@
...
+### High Level API
+
+#### Fault Detection
+
+Fault detection happens over the course of the life time of a sector. When the sector is for some reason unavailable, the miner is responsible to submit the known `faults`, before the PoSt challenge begins. (Using the `AddFaults` message to the chain).
+Only faults which have been reported at challenge time, will be accounted for. If any other faults have occured the miner can not submit a valid PoSt for this proving period.
+
+The PoSt generation then takes the latest available `faults` of the miner to generate a PoSt matching the committed sectors and faults.
+
+At the beginning of a new proving period all faults are reset, and if they persist the miner needs to resubmit an `AddFaults` message.
```
anorth:
```suggestion
When a PoSt is successfully submitted all faults are reset and assumed to be recovered. A miner must either (1) resolve a faulty sector and accept challenges against it in the next proof submission, (2) report a sector faulty again if it persists but is eventually recoverable, (3) report a sector faulty *and done* if the fault cannot be recovered.
```

proof-of-spacetime.md
anorth: We should adjust `POST_CHALLENGE_TIME` to allow ample time for churn to resolve and still leave time for proof calculation.

actors.md
```diff
@@ -679,68 +668,44 @@ func SubmitPost(proofs PoStProof, faults []FaultSet, recovered Bitfield, done Bi
...
 
-	if !CheckPostProof(miner.SectorSize, proof, faults) {
+    var seed
+    if chain.Now() < self.ProvingPeriodEnd {
+      // good case, submitted in time
+      seed = GetRandFromBlock(self.ProvingPeriodEnd - POST_CHALLENGE_TIME)
+    } else {
+      // bad case, submitted late, need to take new proving period end as reference
+      seed = GetRandFromBlock(nextPovingPeriodEnd - POST_CHALLENGE_TIME)
+    }
```
anorth: There are some obvious cases where this could fail if `SubmitPoSt` is submitted too early, or races against the proving period end arriving. I guess they're implicitly fatal, but the spec should eventually be specific.

dignifiedquire:
https://github.com/filecoin-project/specs/issues/449 is the follow up tracking issue, if I missed anything please comment there

PR #334, opened 2019-06-11
ec clarifications and updates

bvohaska:
Updates to definitions, ticket generation, and electionProof generation


sternhenri:
@pooja review plz :D

data-structures.md
```diff
@@ -81,6 +81,38 @@ type BlockHeader struct {
 } representation tuple
 ```
 
+## Ticket
+
+A ticket is a shared random value stapled to a particular block in the chain. Every miner must produce a new ticket each time they run a leader election attempt. In that sense, every new block produced will have one or more associated tickets (in the case the block took multiple leader election attempts to produce).
+
+```go
+type Ticket struct {
```
dignifiedquire: please update this to the new IPLD schema syntax

data-structures.md
```diff
@@ -81,6 +81,38 @@ type BlockHeader struct {
...
+}
+```
+
+## ElectionProof
+
+An election proof is generated from a past ticket (chosen based on public network parameters) by a miner during the leader election process. Its output value determines whether the miner is elected leader and may produce a block. Its inclusion in the block allows other network participants to verify that the block was mined by a valid leader.
+
+```go
+type ElectionProof struct {
+  TODO HS
```
dignifiedquire: maybe make it a bit more clear that this is not an actual field 😆

definitions.md
```diff
@@ -101,6 +105,12 @@ The Generation Attack Threshold is equal to the Polling Time + some Grace Period
 
 #### GHOST
 
+GHOST is an acronym for `Greedy Heaviest Observable SubTree`, a class of blockchain structures in which multiple blocks can validly be included in the chain at any given height or round. GHOSTy protocols produce blockDAGs rather than blockchains and use a weighting function for fork selection, rather than simply picking the longest chain.
```
dignifiedquire: maybe a link to a reference?

definitions.md
```diff
@@ -101,6 +105,12 @@ The Generation Attack Threshold is equal to the Polling Time + some Grace Period
...
+
+#### Height
+
+`Height` refers to the number of `TipSets` that have passed between this `TipSet` and the genesis block (which starts at block height 0). If a `TipSet` contains multiple blocks, each block in the TipSet will have the same `height`.
```
dignifiedquire: please say how this interacts with null blocks here

definitions.md
```diff
@@ -203,7 +217,9 @@ Repair refers to the processes and protocols by which the Filecoin network ensur
 
 #### Round
 
-A round refers to the period over which a new leader election occurs and a block is generated if a leader is found. Typically this means a new block will be mined at every round, though some rounds may see 0 or multiple blocks generated.
+A round refers to the period over which a new leader election occurs and a block is generated if a leader is found. Typically this means a new block will be mined at every round, though
```
dignifiedquire: missing `.` at the end

definitions.md
```diff
@@ -233,21 +249,24 @@ In the context of:
 
 #### Ticket
 
-Some unpredictable element generated by the system for two uses:
-
-- as a random challenge to PoSTs.
-- to elect a leader in Expected Consensus.
-
-See more on [Expected Consensus](expected-consensus.md) and [PoSTs](proofs.md).
+A `ticket` is used as a source of randomness in EC leader election. Every block depends on an `ElectionProof` derived from a `ticket`. At least one new `ticket` is produced with every new block. Ticket creation will be described [here](./expected-consensus.md#Ticket-generation).
```
dignifiedquire: maybe "is described here"?

definitions.md
```diff
@@ -233,21 +249,24 @@ In the context of:
...
 
 #### TipSet
 
-A collection of blocks mined by different miners, each an elected leader of a given epoch. All tip sets have the same parent-set and epoch number (height). TODO add picture.
+ A `TipSet` is a set of blocks that have the same parent set and same number of `tickets`, which implies they will have been mined at the same `height`. A `TipSet` can contain multiple blocks if more than one miner successfully mines a block at the same `height` as another miner. `TipSets` contain 1 or more blocks (i.e. null blocks are not included in `TipSets`).
```
dignifiedquire: This is a bit circular Above it says that height is the number of tipsets passed since genesis, and here it is tipset are the blocks with the same height.

dignifiedquire: > i.e. null blocks are not included in `TipSets`

then where are they included? aren't they part of the block technically?

definitions.md
```diff
@@ -233,21 +249,24 @@ In the context of:
...
 
 #### Verifiable
 
 Something that is verifiable can be checked for correctness by a third party.
 
+#### VDF
+
+A verifiable function that guarantees a time delay given some hardware assumptions and a small set of requirements. These requirements are efficient proof verification, random output, and strong sequentiality. Verifiable delay functions are formally defined [here](https://eprint.iacr.org/2018/601). A VDF will receive {set of public parameters (similar to [these](https://eprint.iacr.org/2018/712.pdf)), seed} and output {proof of correctness, output value}.
```
dignifiedquire: maybe add the title of these publications?

dignifiedquire: also the use of curly braces is very confusing

expected-consensus.md
```diff
@@ -1,70 +1,42 @@
-# Expected Consensus
+#  Expected Consensus
```
dignifiedquire: too much space ;)

definitions.md
anorth: FWIW the go-filecoin code, and everyones intuitions, say that "height" is the number of rounds, not the number of tipsets. We'll be swimming upstream to redefine that here. At present, the code doesn't have a use for a count of the non-null tips.

sternhenri:
Thank you!

sternhenri:
Noted. Will amend in a different PR to disambiguate rounds and heights!

definitions.md
sternhenri: to be handled in a different PR per https://github.com/filecoin-project/specs/pull/351

definitions.md
sternhenri: Disagree with circularity. The circular bit is just a reframing that helps cement what height is here. Re null blocks, removed. Confusing bit. A Tipset can never include just a null block, they are indeed included as part of the block, technically (through the ticket array).

Null blocks is a confusing concept overall we should get rid of.

data-structures.md
sternhenri: Holding off on specifying methods and the like for now, if that is what is meant...

PR #336, opened 2019-06-13
Fixing links

### LINKED ISSUE #291: Hugo-generated documentation site breaks all links
I think all the links in the hugo-generated site are broken. For example, see the links at the top of https://filecoin-project.github.io/specs/docs/expected-consensus/

A relative link like `faults.md` or `./faults.md` is broken because the page name appears as a path with a trailing `/`. Similarly an absolute ref like `/faults.md` is broken by the `/specs/docs/` path prefix. There may be a syntax that works, and is also compatible with raw markdown viewers like GitHub, but I haven't found it yet.

gnunicorn:
Attempt to make linking work again

- [x] fix #291 : links in hugo
   - [x] fix linking to `#`
   - [x] while still having links work on github 
- [x] use markdown links in menu
- [x] use menu to generate pandoc listing (single source of truth)
- [ ] fix links in PDFs

This also cleans up the file structure quite a bit by removing the symlinks in favor of just having everything under `docs`.


dignifiedquire:
@gnunicorn any update on this, would be great to get this in

gnunicorn:
Working on it, but the pdf links are a tricky problem. Also, broke my laptop and with that my setup, so all work got a bit delayed.

However we could review and merge as is, if you prefer.

warpfork:
May or may not be helpful -- if you don't have a tool for checking links yet, https://github.com/raviqqe/muffet is pretty sweet.  It even includes making sure that '#' links land somewhere.

dignifiedquire:
links on the site should be fixed in `master` now

dignifiedquire:
Closing for now. The main issues is fixed in `master`

PR #339, opened 2019-06-18
Add payment broker actor built-in ID address

anorth:


dignifiedquire:
@whyrusleeping @anorth with the changes to payment channels is this still needed?

whyrusleeping:
yeah, this is no longer needed. there isnt one single payment broker, but individual actors per channel

anorth:
I didn't realise payment channel spec had changed so dramatically!

Issue #308, opened 2019-05-29
Allow small grace period for timestamp checking


whyrusleeping:
We currently check that received blocks do not have a timestamp in the future. This could cause problems if nodes have out of sync clocks. We should probably check that the blocks arent more than ALLOWABLE_CLOCK_DRIFT in the future. This value could probably be as far as a minute, since we just want to know that miners arent submitting chains super far in the future.

However, doing this means that miner implementations who always use their local time for the time stamp could produce blocks that *before* their parents. (if one miner is a minute ahead of the next miner).

Bitcoin timestamps just need to ensure that the timestamp is greater than the median time of the last 11 blocks: https://en.bitcoin.it/wiki/Block_timestamp

https://medium.com/@lopp/bitcoin-timestamp-security-8dcfc3914da6

merced:
As miner, I would think the strategy would be setting blocktime = blocktime-ALLOWABLE_CLOCK_DRIFT while mining. And this would still cause validation fail for certain nodes and make their local chain out of sync with majorities of the net.

whyrusleeping:
Yeah, though if you make a block that a large portion of the network rejects, you're just wasting your time.

anorth:
The block timestamp is checked against two things:
- no less than BLOCKTIME past the parent tipset's timestamp
- no more than DRIFT past "now"

Is the grace period referring to the latter, or to both?

Our situation might be a bit different from bitcoin since, until VDFs, this delay is a critical part of maintaining block time (while bitcoin POW difficulty gives a reasonable expectation value of average block times). And when we have VDFs, my understanding is that there will be much less variance in the computation time than a traditional POW, so we might support tighter bounds.

whyrusleeping:
@anorth referring to the latter.

The problem that @merced points out is that if you allow block timestamps to be `X` less than the block time, then a malicious miner would always make a block `block_time - X` ahead, instead of the full `block_time`. If you *dont* allow this, then miners may receive a block that is `DRIFT` in the future, and when they go to make a block, they will either have to wait longer, or lie and put a future timestamp on their block. 

In any case, nodes in the network need to have fairly well synchronised time, bitcoin and ethereum (especially ethereum) require all participating nodes to run ntp (or some other time sync service) to ensure their clocks are as close as possible.

whyrusleeping:
Also agree that bitcoins model won't work for us, but its good background.

[...]

PR #340, opened 2019-06-18
added grace period and call out to network time

sternhenri:
Addresses https://github.com/filecoin-project/specs/issues/308. I'm holding off on the adjustable block-time idea from @whyrusleeping for now given I am not sure yet whether it doesn't cause a problem whereas attackers who can break the VDF can effectively prevent anyone else from mining by forcing the network to adopt their effective block-time (which is what the block delay sought to solve in the first place).

- Also chose to add some explicit mention of network time protocols to not have people just default to machine time.


dignifiedquire:
Missing an actual value for the drift

sternhenri:
I'd recommend a default on the order of 12*conf time or in FIL's case, 6 minutes... (BTC likewise uses 2 hours).

Though that's an impl choice imo

dignifiedquire:
@whyrusleeping please merge if this looks good to you

whyrusleeping:
@sternhenri i think that values > 1 block time are too much. In bitcoin, its okay if one blocks timestamp is before its parents. we are not okay with that.

allowing large drifts like this means that an attacker can 'rush', releasing many blocks at the same time with the bare minimum difference in timestamps, up to the drift limit. Granted, they have to be able to mine those blocks anyways, but it seems rather adversarial

sternhenri:
Why is it more ok with Bitcoin? Your point is well taken though and perhaps we should scale back to ~20 secs. Or 2/3 blocktime. It’s slightly on the short end for me given network prop times (ie 90th percentile slowness in btc gossip is closer to 40secs) but the downside is not awful... and easy to adjust for testnet. 

That said, we do have other means of limiting rushing, so to me this is more about relaxing clock times while limiting the potential this creates.

whyrusleeping:
Bitcoin doesnt care about the same sort of attacks we do. If you can generate bitcoin blocks faster than 10 minutes, you absolutely should do that, and the system wants you to.

> (ie 90th percentile slowness in btc gossip is closer to 40secs)

Thats for bitcoin, Ethereum is much less, its all a function of block size.

2/3 blocktime SGTM for now, we're assuming that all miners have fairly synchronized clocks (as ethereum does), and thats a pretty acceptable range.

sternhenri:
fair enough, changed reco. Your point is well taken, though I'd nuance it in that if you can generate in less than 10 mins, they can't stop you, but they'll update the difficulty.

I see this as more of a drawback of our VRC, just as with BTC this is also meant to mitigate potential rushing, however maybe our source of timing constraint (VDF vs hashing) is by nature a bit different...

PR #341, opened 2019-06-18
updated for weight punishment

sternhenri:
Solve for https://github.com/filecoin-project/consensus/issues/61.

Also patched the way in which the current block weight is changed to incentivize people including as many blocks as possible in their TipSet (downside is you can't know effective weight of your own block, only your parents').


sternhenri:
@ZenGround0 noticed rightfully so that the penalization can't apply to old weight, lest it break liveness (great catch, cc @bvohaska), so hold pending a quick refactor.

expected-consensus.md
```diff
@@ -333,9 +333,17 @@ TODO: ensure 'power' is properly and clearly defined
...
+weight is a term P times the sum of a constant `V` times the number of blocks included in that TipSet in the given round (can only be calculated in the next round), and `X` - a function of the total power in the network as reported in the Power Table. P will be equal to some term P_i^P_n and seeks to punish selfish mining. So we have:
+
+`Weight = ParentWeight + P(V*NumBlocksMinedInThisRound + X)`
+
+ The exact value for these parameters remain to be determined, but for now we set:
+
+- `V = 2 * number of blocks in the round`
+- `X = log(TotalPower in Power Table l blocks back)`
+- `P_i = .87`
+- `P_n = the number of tickets in this block's Ticket array` if that number is  >= 3, 0 otherwise.
```
ZenGround0: Something we need to keep in mind: keeping infinite precision in our weight is not feasible.  Right now we only track a delta weight >= 1/1000 [see here](https://github.com/filecoin-project/go-filecoin/blob/master/types/fixed_point.go#L5)

It turns out that we don't hit this fraction until after 50 null blocks ( (0.87)^50 < 1/1000) so this is not a big concern.  The implementation must round these events up to a delta of 1/1000 to have guaranteed liveness throughout time however.

expected-consensus.md
```diff
@@ -333,9 +333,17 @@ TODO: ensure 'power' is properly and clearly defined
...
-weight is a constant `V`, plus `X` - a function of the total power in the network as reported in the Power Table.  The exact value for `V` and the magnitude of the power ratio value are
-still to be determined, but for now `V = 10` and `X = log(TotalPower)`.
+The weight at each block is equal to its `ParentWeight` plus that block's delta weight. Delta
+weight is a term P times the sum of a constant `V` times the number of blocks included in that TipSet in the given round (can only be calculated in the next round), and `X` - a function of the total power in the network as reported in the Power Table. P will be equal to some term P_i^P_n and seeks to punish selfish mining. So we have:
+
+`Weight = ParentWeight + P(V*NumBlocksMinedInThisRound + X)`
+
+ The exact value for these parameters remain to be determined, but for now we set:
+
+- `V = 2 * number of blocks in the round`
```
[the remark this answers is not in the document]
sternhenri: It was, I adapted it based on my own understanding. Wouldn't we want to base this on the number of blocks produced in the given round: ie favor larger tipsets?

expected-consensus.md
sternhenri: I did not know that. I will work to adapt the functions accordingly. Thank you!

sternhenri:
@ZenGround0 sim work is proving these parameters imperfect. That + the requisite precision: let me know if you'd like me to refactor entirely or whether we can merge for now with the joint understanding that we will update the params in the coming weeks.

sternhenri:
@whyrusleeping would love you to opine on the blocksFactor (constant or factor of number of blocks mined in a round).

expected-consensus.md
```diff
@@ -333,9 +333,17 @@ TODO: ensure 'power' is properly and clearly defined
...
+- `X = log(TotalPower in Power Table l blocks back)`
+- `P_i = .87`
```
whyrusleeping: 0.87 is an oddly specific number. Where does that come from?

sternhenri: [The Weighting fn observable](https://observablehq.com/d/3812cd65c054082da).

Note 1: The number is most certainly going to change. Simulations shows it's not performing quite as I'd like yet. Regardless, should probably be removed from the spec I'd argue (in favor of our having a separate FIL-params guideline at Testnet that fills in variables).

Note 2: I'm cleaning this observable, the conf time one and the consensus repo this week, so I can make them all fully open and accessible to people farther away from the topic.

expected-consensus.md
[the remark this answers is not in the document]
sternhenri: @whyrusleeping any thoughts?

sternhenri:
Added https://github.com/filecoin-project/specs/pull/341/commits/6dc432b7f3dad15105f7c8a8c3a46dd7a031b5f1 @ZenGround0 per an old convo we had. Lmk if it does not works for you :)

expected-consensus.md
```diff
@@ -324,9 +324,23 @@ Due to network propagation delay, it is possible for a miner in round N+1 to omi
...
+The weight should be rounded down to its fourth decimal at each height (i.e. each time it is recalculated).
+
+```sh
+Note that if your implementation does not allow for rounding to the fourth decimal, miners should apply the [tie-breaker below](#selecting-between-tipsets-with-equal-weight). Weight changes will be on the order of single digit numbers on expectation, so this should not have an outsized impact on chain consensus across implementations.
+```
+
+ The exact value for these parameters remain to be determined, but for illustration, we can set:
+
+- `V = 2 * number of blocks in the round`
+- `X = log(TotalPower in Power Table l blocks back)`
```
dignifiedquire: just to be clear, is this "blocks" back or "tickets" back?

sternhenri: good catch!!!! Rounds (as in tickets) back but either way, L is gone

PR #344, opened 2019-06-20
actors: add some todos and details

dignifiedquire:


actors.md
```diff
@@ -79,16 +79,20 @@ func Exec(code Cid, params []byte) Address {
 
 	// This generates a unique address for this actor that is stable across message
 	// reordering
-	addr := VM.ComputeActorAddress()
+	// TODO: where do `creator` and `nonce` come from?
+	addr := VM.ComputeActorAddress(creator, nonce)
```
dignifiedquire: is `creator` the same as `message.from`?

whyrusleeping: creator is `msg.From`, and nonce is `msg.Nonce`.

actors.md
```diff
@@ -79,16 +79,20 @@ func Exec(code Cid, params []byte) Address {
...
 		Code:    code,
 		Balance: msg.Value,
+		Head:    nil,
+		Nonce:   0,
 	}
 
 	// The call to the actors constructor will set up the initial state
-	// from the given parameters
+	// from the given parameters, setting `actor.Head` to a new value when successfull.
+	// TODO: can constructors fail?
```
whyrusleeping: yes, they can

PR #348, opened 2019-06-26
limit number of sectors a storage miner actor to max number of sectors proved by a single PoSt

laser:
As per [elsewhere](https://github.com/filecoin-project/specs/blob/master/proofs.md#generatepost) in the spec, the `GeneratePoSt` operation generates a PoSt over `POST_SECTORS_COUNT` sealed sectors. We learned from @nicola yesterday that we must cap the number of sectors which a single storage miner actor may commit to `POST_SECTORS_COUNT`; if a single miner operator wishes to commit more than `POST_SECTORS_COUNT` sectors, they will need to provision additional storage miner actors.

A few changes fall out of this revelation:

- The `submitPoSt` message must include a single `PoStProof`.
- A single storage miner actor must be able to commit no more than `POST_SECTORS_COUNT` sectors.


actors.md
```diff
@@ -491,6 +491,12 @@ func CommitSector(sectorID SectorID, commD, commR, commRStar []byte, proof SealP
...
 
+	// ensure that the miner cannot commit more sectors than can be proved with
+	// a single PoSt
+	if miner.Sectors.Size() == MaxNumberSectorsPerMiner() {
+		Fatal("too many sectors")
+	}
+
 	miner.ActiveCollateral += coll
 
 	// Note: There must exist a unique index in the miner's sector set for each
```
dignifiedquire: shouldnt this be less or equal?

laser:
@dignifiedquire 

> where is the maximum defined?

The spec does not specify a maximum value, but rather refers to `POST_SECTORS_COUNT`. I agree that the protocol should disambiguate. CC @nicola

actors.md
laser: @dignifiedquire 

> shouldnt this be less or equal?

Sorry, I do not understand. Why would the guard be less or equal?

whyrusleeping: I'd put `>=`, just because the paranoid in me says "oh no what if..."

PR #349, opened 2019-06-26
Flesh out ZigZag Proof of Replication

schomatis:
Towards https://github.com/filecoin-project/research/issues/11.

Still incomplete, but without any review to guide this process I don't think adding more content will make any meaningful progress.

Diff (ignoring file rename) [here](https://github.com/filecoin-project/specs/compare/6aa0a38..fix/zigzag-porep/proof-generation).

Initial guideline:
> the DRG code is just Rust code for `parents`.
> - moving from straight Rust code to Go-like pseudocode. You can find the original source that was ported in `go-proofs`. The bucket sampling algorithm is described in https://web.stanford.edu/~bfisch/porep_short.pdf.
> - Need to describe the expansion parents, zigzagging, Feistel.


zigzag-porep.md
```diff
@@ -196,22 +197,23 @@ func DeriveChallenges(challenges LayerChallenges, layer uint, leaves uint,
 
 ### Layer Replication
 
+TODO: Define `Graph`. We need to decide if this is an object we'll explicitly define or if its properties (e.g., `GRAPH_SIZE`) are just part of the replication parameters and all the functions just refer to the _same_ graphs being manipulated across the entire replication process. (At the moment I've avoided defining a `Graph` structure as in other specs I didn't see any object methods, just standalone functions.)
+
```
porcuquine: I don't think `Graph` needs to be a first-class part of the spec. It's an implementation detail of how the problem is decomposed in `rust-fil-proofs` as a proof toolkit.

However, the cleanest way to specify based on what we have might still involve a small amount of state to track the direction of the current layer. If bundling that and any other convenience state into a `Graph` struct makes sense, that seems reasonable, at least for now.

zigzag-porep.md
```diff
@@ -0,0 +1,529 @@
...
+### Layer Challenge Counts
+
+TODO: define `Challenge` (or find existing definition)
+
+```go
+func ChallengesForLayer(challenge Challenge, layer uint) uint {
+
+    switch challenge.Type {
+        case Fixed:
+            return challenge.Count
```
nicola: We should remove ambiguity, there should be no "fixed" zigzag

zigzag-porep.md
```diff
@@ -0,0 +1,529 @@
...
+            r := 1.0 - TAPER;
+            t := min(l, TAPER_LAYERS)
+
+            totalTaper := pow(r, t)
+
+            calculated := ceil(totalTaper * challenge.count)
+
+            // Although implied by the call to `ceil()` above, be explicit
+            // that a layer cannot contain 0 challenges.
+            max(1, calculated)
```
nicola: this is wrong, the minimum number of challenges should be calculated from https://observablehq.com/d/bbabac1947b79011#gen_zigzag_taper

zigzag-porep.md
```diff
@@ -0,0 +1,529 @@
...
+}
+```
+
+### Challenge Derivation
+
+TODO: define `Domain` (for practical purposes a `uint`) and `LayerChallenges` (or find existing definition).
+
+```go
+func DeriveChallenges(challenges LayerChallenges, layer uint, leaves uint,
+    replicaId Domain, commitment Domain, k uint) []uint {
```
nicola: we should rename commitment with seed, this would make the spec interactive-porep friendly

nicola:
The current spec needs more review, I added a few comments in the code.
I am approving it for now and I will be opening an issue for deeper reviewing.

zigzag-porep.md
porcuquine: Specifically, we need to remove `TAPER_LAYERS` (and usage), and add `MIN_LAYER_CHALLENGES`. That constant should replace `1` in `max(1, calculated)` and should be set to 20.

The implementation should then be changed to match the spec.

PR #355, opened 2019-07-01
Switch to ipld schema

### LINKED ISSUE #352: Switch to use IPLD Schemas

dignifiedquire:
Closes #352

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -4,12 +4,32 @@ Any implementations of the Filecoin actors must be exactly byte for byte compati
 
 This spec describes a set of actors that operate within the [Filecoin State Machine](state-machine.md). All types are defined in [the basic type encoding spec](data-structures.md#basic-type-encodings).
 
-- [Init Actor](#init-actor)
-- [Storage Market Actor](#storage-market-actor)
-- [Storage Miner Actor](#storage-miner-actor)
-- [Payment Channel Broker Actor](#payment-channel-broker-actor)
+## Actor State
 
-## Built In Actors
+These below used `kindeded` representation, as the type can be inferred from the context, in which
+they are used (`Actor` or `UnsignedMessage`).
+
+```sh
+type ActorState union {
+    | InitActorState
+    | AccountActorState
+    | StorageMarketActorState
+    | StorageMinerActorState
+    | PaymentChannelBrokerActorState
+    | MultisigActorState
+} representation kinded
+
+type ActorMethod union {
+    | InitActorMethod
+    | AccountActorMethod
+    | StorageMarketActorMethod
+    | StorageMinerActorMethod
+    | PaymentChannelBrokerActorMethod
+    | MultisigActorMethod
+} representation kinded
+```
+
+## System Actors
 
 Some state machine actors are 'system' actors that get instantiated in the genesis block, and have their IDs allocated at that point.
 
@@ -18,50 +38,60 @@ Some state machine actors are 'system' actors that get instantiated in the genes
 | 0    | InitActor          | Network Init            |
 | 1    | AccountActor       | Network Treasury        |
 | 2    | StorageMarketActor | Filecoin Storage Market |
-|  99 | AccountActor | Burnt Funds |
+| 99   | AccountActor       | Burnt Funds             |
 
-## Init Actor
+## Built In Actors
+
+### Init Actor
+
+- **Code Cid**: `<codec:raw><mhType:identity><"init">`
 
 The init actor is responsible for creating new actors on the filecoin network. This is a built-in actor and cannot be replicated. In the future, this actor will be responsible for loading new code into the system (for user programmable actors). ID allocation for user instantiated actors starts at 100. This means that `NextID` will initially be set to 100.
 
-```go
-type InitActor struct {
-	// Mapping from Address to ID, for lookups.
-	AddressMap map[Address]BigInt
-
-	NextID BigInt
+```sh
+type InitActorState struct {
+    addressMap {Address:ID}<Hamt>
+    nextId UInt
 }
+
+type InitActorMethod union {
+    | InitConstructor 0
+    | Exec 1
+    | GetIdForAddress 2
+} representation keyed
 ```
 
-### Code Cid
+#### `Constructor`
 
-`<codec:raw><mhType:identity><"init">`
+**Parameters**
 
-| Index     | Method Name       |
-| -------- | ---------- |
-| 1  | `Exec`     |
-| 2 | `GetIdForAddress` |
+```sh
+type InitConstructor struct {
+}
 
-### `Exec(code Cid, params []Param) Address`
+```
 
->  This method is the core of the `Init Actor`. It handles instantiating new actors and assigning them their IDs.
+**Algorithm**
 
-### Parameters
+#### `Exec`
 
-| Name     | Type       | Description                                                  |
-| -------- | ---------- | ------------------------------------------------------------ |
-| `code`   | `Cid`      | A pointer to the location at which the code of the actor to create is stored. |
-| `params` | `[] Param` | The parameters passed to the constructor of the actor.       |
+This method is the core of the `Init Actor`. It handles instantiating new actors and assigning them their IDs.
 
+**Parameters**
 
+```sh
+type Exec struct {
+    ## Reference to the location at which the code of the actor to create is stored.
+    code &Code
+    ## Parameters passed to the constructor of the actor.
+    params ActorMethod
+} representation tuple
+```
 
-`Param` is the type representing any valid arugment that can be passed to a function.
-
-TODO: Find a better place for this definition.
-
+**Algorithm**
 
 ```go
-func Exec(code Cid, params []byte) Address {
+func Exec(code &Code, params ActorMethod) Address {
 	// Get the actor ID for this actor.
 	actorID = self.NextID
 	self.NextID++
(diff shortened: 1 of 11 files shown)


actors.md
```diff
@@ -4,10 +4,27 @@ Any implementations of the Filecoin actors must be exactly byte for byte compati
...
 This spec describes a set of actors that operate within the [Filecoin State Machine](state-machine.md). All types are defined in [the basic type encoding spec](data-structures.md#basic-type-encodings).
 
-- [Init Actor](#init-actor)
-- [Storage Market Actor](#storage-market-actor)
-- [Storage Miner Actor](#storage-miner-actor)
-- [Payment Channel Broker Actor](#payment-channel-broker-actor)
+## Actor State
+
+```sh
+type ActorState union {
```
whyrusleeping: There isnt a representation specified for these, and i'm not sure if we should actually put all the actor state types in a union. This will get tricky as we add more actors...

actors.md
```diff
@@ -18,50 +35,58 @@ Some state machine actors are 'system' actors that get instantiated in the genes
...
 
 The init actor is responsible for creating new actors on the filecoin network. This is a built-in actor and cannot be replicated. In the future, this actor will be responsible for loading new code into the system (for user programmable actors). ID allocation for user instantiated actors starts at 100. This means that `NextID` will initially be set to 100.
 
-```go
-type InitActor struct {
-	// Mapping from Address to ID, for lookups.
-	AddressMap map[Address]BigInt
+```sh
+type InitActorState struct {
+    addressMap {Address:ID}
```
whyrusleeping: this will need to be a HAMT, I think we write this as:
`addressMap &{Address:ID}<HAMT>`

cc @warpfork

retrieval-market.md
```diff
@@ -21,62 +21,68 @@ The main components are as follows:
 
 The v0 `retrieval market` will initially be implemented as two `libp2p` services. It will be request response based, where the client who is requesting a file sends a `retrieval deal proposal` to the miner. The miner chooses whether or not to accept it, sends their response which (if they accept the proposal) includes a `signed retrieval deal`, followed by the actual requested content, streamed as a series of bitswap block messages, using a pre-order traversal of the dag. Each block should use the [bitswap block message format](https://github.com/ipfs/go-ipfs/blob/master/exchange/bitswap/message/message.go#L216). This way, the client should be able to verify the data incrementally as it receives it. Once the client has received all the data, it should then send a payment channel SpendVoucher of the proposed amount to the miner. This protocol may be easily extended to include payments from the client to the miner every N blocks, but for now we omit that feature.
 
-```go
+```sh
 type RetDealProposal struct {
-	// Ref is the cid of the data to be retrieved
-	Ref Cid
+	## Reference to the data being retrieved.
+	ref Link
```
warpfork: Link to what?  `&Any`?

payments.md
```diff
@@ -34,63 +34,67 @@ Let's say that `A` wants to make such an arrangement with `B`. `A` should create
...
-	DataCommitment []byte
+	dataCommitment Bytes
 
-	MinCloseHeight uint64
+	minCloseHeight UInt
+}
 
-	Sig Signature
+type SignedVoucher struct {
+    voucher SpendVoucher
```
warpfork: Whitespace police.

payments.md
```diff
@@ -226,24 +228,28 @@ To accomplish this, we introduce the Payment Reconciliation Protocol.
...
 
 Bob receives this request, and checks that all the fields are correct, and then ensures that the difference between ReqVal and the vouchers sum is sufficient (this is a parameter that the client can set).  Then, he sends back a response which either contains the requested voucher, or an error status and message.
 
-```go
+```sh
 type ReconcileResponse struct {
-	Combined Voucher
+	combined Voucher
+	status  Status
+	message String optional
```
warpfork: Modifiers like 'optional' come before the TypeTerm.

retrieval-market.md
```diff
@@ -21,62 +21,68 @@ The main components are as follows:
...
+    | RejectedResponse 1
+    | ErrorResponse 2
+}
 
-type RetDealResponse struct {
-	Status  ResponseStatus
-	Message string
+type AcceptedResponse struct {}
+type RejectedResponse struct {
+    message String optional
```
warpfork: Modifiers like 'optional' come before the TypeTerm.

retrieval-market.md
```diff
@@ -21,62 +21,68 @@ The main components are as follows:
...
-	Data   []byte
+	## Cid prefix parameters for this block. It describes how to
+	## hash the block to verify it matches the expected value.
+	refix CidPrefix
+	data Bytes
+}
+
+## Represents all the metadata of a Cid.
+## It does not contains  any actual content information.
+type CidPrefix struct {
```
warpfork: Is this necessary to specify in this spec?  Maybe the answer is yes; I don't have deep context spooled up in my brain here.  But in most other places I just treat CIDs as fairly opaque.

retrieval-market.md
```diff
@@ -21,62 +21,68 @@ The main components are as follows:
...
 
-The query should include the CID of the piece that the client is interested in retrieving. The response contains whether or not the miner will serve that data, the price they will accept for it. 
+The query should include the CID of the piece that the client is interested in retrieving. The response contains whether or not the miner will serve that data, the price they will accept for it.
 
-```go
+```sh
 type RetQuery struct {
-	Piece Cid
+    ## TODO: what exactly does this link to?
+	piece Link
```
warpfork: +1 to that TODO :)

retrieval-market.md
```diff
@@ -21,62 +21,68 @@ The main components are as follows:
...
 }
 
-type RetQueryResponse struct {
-	Status   RetQueryStatus
-	MinPrice TokenAmount
+type RetQueryResponse union {
+    | AvailableResponse
+    | UnavailableResponse
```
warpfork: A `representation` clause on unions is required.  (See "Unions are a special case..." in [schema/Representations.md](https://github.com/ipld/specs/blob/afe04f38c8870044602e52eb14356f8a60c61d90/schema-layer/schemas/representations.md#kinds-and-their-representations).)

warpfork:
Coooool!

warpfork:
Oh good lord.  Github hid the entire diff to the `actors.md` file in the "large diff click to load" nonsense.  Review continues.

actors.md
warpfork: So I think I tried to talk @dignifiedquire into this the other day.

It's true it might get tricky, but it also might be a perfectly reasonable exercise of schema evolution/migration stuff.  Add more actors?  Add union members.  Parsing data that has an "actor" value the current program doesn't recognize?  Well, halting is probably what we want to do anyway, neh?

retrieval-market.md
whyrusleeping: the reason for this is that we're sending a raw array of bytes, and the cid prefix specifies how we should turn it into ipld.

actors.md
dignifiedquire: sounds reasonable to me

actors.md
dignifiedquire: Is there a way to say "all maps use hamts"? I would really like all of them in the actors to use hamts without having to remember this..

retrieval-market.md
dignifiedquire: unclear, @whyrusleeping ?

payments.md
dignifiedquire: :bomb:

payments.md
dignifiedquire: that is really confusing :(

retrieval-market.md
dignifiedquire: okkkkk

dignifiedquire: fixed

retrieval-market.md
dignifiedquire: fixed

payments.md
dignifiedquire: fixed

actors.md
dignifiedquire: fixed

actors.md
```diff
@@ -4,12 +4,32 @@ Any implementations of the Filecoin actors must be exactly byte for byte compati
...
+
+```sh
+type ActorState union {
+    | InitActorState
+    | AccountActorState
+    | StorageMarketActorState
+    | StorageMinerActorState
+    | PaymentChannelBrokerActorState
+    | MultisigActorState
+} representation kinded
```
warpfork: I don't think these can be kinded; and this is incomplete either way.

To use kinded unions, each member must have a distinct representation kind -- so for example, you can't use two different types which are both represented as map, nor two different types represented as lists, and so on.

If you have two structs, it's still possible to have them be discriminated in kinded union if one has `representation map` (reprkind==map) and the other has `representation tuple` (reprkind==list).  But those are just details to the above rule.

All members in a kinded union have to have a kind keyword after them, in order to make all this locally lexically clear.  More than one of these types would have the kind "map", and so something's fishy here.

(Long story short, you probably just mean to use `representation keyed` here, or something.)

actors.md
```diff
@@ -334,85 +423,89 @@ func IsMiner(addr Address) bool {
...
+    | UpdatePeerID 13
+    | ChangeWorker 14
+} representation keyed
 ```
 
-### Code Cid
+#### `Constructor`
 
-`<codec:raw><mhType:identity><"sminer">`
+Along with the call, the actor must be created with exactly enough filecoin for the collateral necessary for the pledge.
```
anorth: There is no pledge, that's an old concept.

PR #359, opened 2019-07-01
Cosmetic changes plus specification that null blocks are not included in tipsets

pooja:


expected-consensus.md
```diff
@@ -1,63 +1,64 @@
...
+A `ticket` is used as a source of randomness in EC leader election. Every block depends on an `ElectionProof` derived from a `ticket`. At least one new `ticket` is produced with every new block.
 {{% /definition %}}
 
 {{% definition %}}
- A `TipSet` is a set of blocks with the same parent set, and same number of tickets (which implies they will have been mined at the same height).
+ An `ElectionProof` is derived from a past `ticket` and is included in every block header. The `ElectionProof` proves that the miner was eligible to mine a block in that `round`.
  {{% /definition %}}
 
-Expected Consensus is a probabilistic Byzantine fault-tolerant consensus protocol. At a high
+## Basic algorithm
```
dignifiedquire: Maybe just "Algorithm"?

expected-consensus.md
```diff
@@ -1,63 +1,64 @@
...
-This spec describes how to implement the protocol in general, for Filecoin-specific processes, see:
+This spec describes how the expected consensus (EC) protocol works in general. To read more about Filecoin-specific processes, see:
 
-- [Mining Blocks](mining.md#mining-blocks) on how consensus is used.
+- [Mining blocks](mining.md#mining-blocks) on how consensus is used in block mining.
 - [Faults](./faults.md) on slashing.
-- [Storage Market](./storage-market.md#the-power-table) on how the power table is maintained.
-- [Block data structure](data-structures.md#block) for details on fields and encoding.
+- [Storage market](./storage-market.md#the-power-table) on how the power table is maintained.
+- [Block data structure](data-structures.md#block) for details on fields and encoding
```
dignifiedquire: are these cross references actually useful, one can just look at the table of contents

expected-consensus.md
pooja: Yup, can change that

expected-consensus.md
pooja: Yeah, I think they're useful so people know where exactly expected consensus can be seen in the protocol

expected-consensus.md
```diff
@@ -111,25 +112,23 @@ func CheckIfWinnerAtRound(key PrivateKey, n Integer, parentTipset Tipset) (bool,
...
 ```
 
-
 TODO: get accurate estimates for K and L, potentially merge both to a single param.
 
 Note: Validity of blocks beyond appropriate ticket generation (defined below) is defined by the specific protocol using EC. For the Filecoin definition of a valid block, see the [mining spec](mining.md).
 
-The basic algorithm can be broken down by looking in turn at its two major components:
-
+This basic algorithm can be broken down by looking in turn at its two major components:
```
dignifiedquire: "the algorithm"

PR #361, opened 2019-07-01
Verify miner is valid in blockVerif

sternhenri:
Should close https://github.com/filecoin-project/specs/issues/315

Most of it was already taken care of in https://github.com/filecoin-project/specs/commit/cf8bc2657135b49d96b41980a55f16c9090b0117#diff-7cd370c7e68a45a697aaa6f19a205df3

Added some cleanup here.

One detail to keep track of @whyrusleeping is the recommended implementation of `isMiner`: it simply checks whether the miner has **any** power. If the slashing Values (currently full power) were to change for either `slashConsensusFault` or `slashStorageFault`, a slashed miner may be able to participate.

My sense is that this is how it should be: ie if they don't get fully slashed then they are by definition still able to mine, but wanted to explicit that.


actors.md
```diff
@@ -152,17 +152,12 @@ type AccountActor struct {
 
 ## Storage Market Actor
 
-The storage market actor is the central point for the Filecoin storage market. It is responsible for registering new miners to the system, and maintaining the power table. The Filecoin storage market is a singleton that lives at a specific well-known address.
+The storage market actor is the central point for the Filecoin storage market. It is responsible for registering new miners to the system, and maintaining the power table. The Filecoin storage market is a singleton, defined in terms of a particular state in the Filecoin state tree and that lives at a specific well-known address.
 
-```go
-type StorageMarketActor struct {
```
whyrusleeping: the actors state should not be removed, thats a thing that needs to be clearly defined.

storage-market.md
```diff
@@ -14,7 +14,7 @@ TODO: This is a high level overview of how the storage market interacts with com
 
 ## The Market Interface
 
-This interface, written using Go type notation, defines the set of methods that are callable on the storage market actor. The storage market actor is a built-in network actor. For more information about Actors, see TODO.
+This interface, written using Go type notation, defines the set of methods that are callable on the storage market actor. The storage market actor is a built-in network actor, it can be retrieved using a call to `LoadStorageMarket(state CID)`. For more information about Actors, see [the actors spec](./actors.md).
```
whyrusleeping: saying `LoadStorageMarket(...)` here seems like a bit of an implementation detail to me. If its defined somewhere already, then i'm fine to keep it, otherwise lets just say "it's state can be read from the state tree" or something similar.

dignifiedquire:
I believe that with the move to ipld schemas we can define

```sh
type TipSet [&BlockHeader]
```

making it very easy

actors.md
sternhenri: Happy to add back in, but I thought the interface in storage-market spec did that well enough. What am I not understanding? It also seemed more fleshed out than what was in actors (which I interpreted as stale).

What can I do to fix?

storage-market.md
sternhenri: I get that, we just use it in a lot of pseudocode and it's not clearly defined anywhere. But point well taken.

sternhenri:
ok. worth adding in your opinion? May be a separate PR...

storage-market.md
```diff
@@ -44,6 +44,14 @@ type StorageMarket interface {
 	// accounting of the total storage in the storage market.
 	UpdateStorage(delta BytesAmount)
 
+  // PowerLookup returns a given miner's committed storage in the system.
+  // This number is used during leader election and block validation.
+  PowerLookup(miner Address) BytesAmount
+  
+  // Convenience method that simply returns whether a given miner's committed
+  // power is > 0.
+  IsMiner(miner Address) bool
```
anorth: The state value for a miner's power is currently on the miner actor, not here. It would be more straightforward to consult the miner directly, and consult the market only for the boolean value of whether a miner is participating. This method will just have to send a message to the miner actor to work it out.

I don't think a "convenience" method that composes both belongs in the spec'd interfaces. We should require only the bare primitives.

(Alternatively, it would also make sense for the power values to be stored in the market actor instead).

actors.md
whyrusleeping: the contents of actors.md is not stale, its the absolute truth of how the network works. If something seems off to you there, please point it out.

sternhenri: mmmh I was just going off of the fact that `totalStorage` was its only field. Therefore, thought we might be missing things (ie stale) given the interface for storage-market in storage-market.md..

Probably issue on my end.

whyrusleeping:
@sternhenri can you adjust the `GetPower` method on the `MinerActor` and the `GetTotalStoragePower` method on the `StorageMarketActor` to deal with the changes?

sternhenri:
Done. With that said, I concur with @anorth that it may be worth removing both!

sternhenri:
Sorry, which changes are you referring to exactly? Those suggested by @anorth?

storage-market.md
```diff
@@ -14,7 +14,7 @@ TODO: This is a high level overview of how the storage market interacts with com
 
 ## The Market Interface
 
-This interface, written using Go type notation, defines the set of methods that are callable on the storage market actor. The storage market actor is a built-in network actor. For more information about Actors, see TODO.
+This interface, written using Go type notation, defines the set of methods that are callable on the storage market actor. The storage market actor is a built-in network actor, it can be retrieved from the state tree. For more information about Actors, see [the actors spec](./actors.md).
```
dignifiedquire: pleas use `[actors.md]` as the syntax to not break links on the website

sternhenri:
Arr. You're right. The addition of `TipSets` to the data-structures.md file may be warranted. Nothin beyond that. 

Feel free to merge or close.

PR #335, opened 2019-06-12
Addressing comments from data-structures spec review

pooja:
Not yet ready for merge, but would appreciate comments regardless. If any merge conflicts as well, I can fix those, just lmk

PR #363, opened 2019-07-02
fix(data-structures): pull in additions from spec review

dignifiedquire:
Replaces #335

### DIFF
--- a/data-structures.md
+++ b/data-structures.md
@@ -1,19 +1,38 @@
 # Data Structures
 
-This document serves as an entry point for the data structures in filecoin.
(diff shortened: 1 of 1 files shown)


data-structures.md
```diff
@@ -1,19 +1,38 @@
 # Data Structures
 
-This document serves as an entry point for the data structures in filecoin.
+This document serves as an entry point for understanding the data structures in Filecoin. In this document, we give an introduction to each of the protocol data structures (linking to more detailed data structure specs in some places) and then explain how to encode these data structures for use in other parts of Filecoin such as the network protocols and the blockchain.
```
pooja: Can we simplify to:

This document serves as an entry point for understanding the data structures in Filecoin. In this document, we give an introduction to each of the protocol data structures and then explain how to encode these data structures for use in other parts of Filecoin (e.g. network protocols and the blockchain).

dignifiedquire: maybe avoid the double "this document" by dropping the first sentence?

PR #367, opened 2019-07-02
feat: define SectorSet

dignifiedquire:
Supersedes #333

### DIFF
--- a/data-structures.md
+++ b/data-structures.md
@@ -231,10 +231,15 @@ type Bitfield Bytes
 
 ### SectorSet
 
-{{% notice todo %}}
-Define me
-{{% /notice %}}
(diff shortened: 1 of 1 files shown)


whyrusleeping:
I implemented an AMT here: https://github.com/filecoin-project/go-amt-ipld

Its integrated into lotus here: https://github.com/filecoin-project/go-lotus/pull/163

mishmosh:
Pending a decision from #333.

ZenGround0:
It's probably roughly better to merge this than to drop it.  I'm on the fence.  SectorSet terminology appears to be deprecated, but if there are no links to amt yet that's useful to include.

PR #368, opened 2019-07-02
Clarify that block rewards are not locked up

pooja:
Closes https://github.com/filecoin-project/specs/issues/126


mining.md
```diff
@@ -319,7 +319,7 @@ An eligible miner can start by filling out `Parents`, `Tickets` and `ElectionPro
...
 
-Once the miner has the aggregate `ParentState`, they must apply the mining reward. This is done by adding the correct amount to the miner owner's account balance in the state tree. See [block reward](#block-rewards) for details.
+Once the miner has the aggregate `ParentState`, they must apply the block reward. This is done by adding the correct block reward amount to the miner owner's account balance in the state tree. The reward will be spendable immediately in the next block. See [block reward](#block-rewards) for details on how the block reward is structured. See [Notes on Block Reward Application](#notes-on-block-reward-application) for some of the nuances in applying block rewards.
```
whyrusleeping: technically, the mining reward is spendable in the same block (aka, the block the miner created) not the next block.

mining.md
```diff
@@ -367,11 +367,23 @@ Decay = e^(ln(0.5) / (HalvingPeriodBlocks / AdjustmentPeriod))
...
+### Notes on Block Reward Application
+
+As mentioned above, every round, a miner checks to see if they have been selected as the leader for that particular round (see [Secret Leader Election](expected-consensus.md#secret-leader-election) in the Expected Consensus spec for more detail). Thus, it is possible that multiple miners may be selected as winners in a given round, and thus, that there will be multiple blocks produced at the same block height (forming a TipSet). Each of the winning miners will apply the block reward directly to their actor's state in their state tree. 
+
+Once a TipSet is produced and propagated, other nodes on the network will validate the TipSet. The full procedure for how to verify a TipSet can be found above in [Block Validation](#block-validation). At a high level, in order to validate a set of blocks in a TipSet, the validating node will first apply the block reward value directly to each of the parent nodes' accounts and then apply the messages to the resulting state. Thus, each of the miners who produced a block in the TipSet will receive a block reward. There will be no lockup. These rewards can be spent immediately (in the next block and later).
```
whyrusleeping: tipsets arent really created and propagated, individual blocks are created, and nodes that receive them form tipsets out of other blocks at the same height with the same parents.

the general algorithm for computing the state of a tipset is:

for each block in the tipset:
  apply the block reward for the miner of the block
  apply the state transition for each message in that block

The way this paragraph is worded, it sounds like you would apply all the block rewards up front, and then apply all the messages.

mining.md
pooja: Got it, fixing

mining.md
pooja: Cool, will fix!

pooja:
@whyrusleeping Addressed the comments above! Can you take a look?

sternhenri:
Why did we remove any mention of conf time or others?

pooja:
Hey! I initially had it in but it was because our definition of
confirmation time has different implications than the more common
confirmation time definition of Bitcoin and other projects. We didn't want
people to worry that our stated time is long, even though our confirmation
time isn't just a heuristic but actually essentially finality.

It was more headache than it was worth to include it in the spec. But if
you feel really strongly about it, we can talk about adding it back (maybe
take it up with whyrusleeping)

sternhenri:
No I understand, that makes sense to me. Though in fact Bitcoin’s stated time is a misnomer that most wallets don’t use at face value.

Agree spec may not be the right place to do it but we should explicitly give guidance for it somewhere and maybe just say “immediately (or at client’s discretion)” in the spec, to indicate there’s some room for impl choice there

pooja:
I think this could be a PR if you'd like to amend the wording!

PR #370, opened 2019-07-02
Add dot

Kubuxu:


nicola:
Great contribution, thank you for improving the spec

Kubuxu:
My grammar might not be great but I can see a missing dot. :sweat_smile:

PR #381, opened 2019-07-08
Fix message type

whyrusleeping:
@dignifiedquire this snuck past my review. The message type will have a integer field to denote the message, and opaque bytes as the parameters. These bytes are passed into the VM, and the actor gets to decide how they are decoded. Any encoding we use for the built-in actors is merely convention (similar to the conventions defined by solidity around method calls).


dignifiedquire:
but that is not needed, the number gets encoded with the type, that was what our last discussion was about

PR #383, opened 2019-07-08
Slashing Mechanism

zixuanzh:
Not exactly sure for the following, 
- [x] add/remove collateral for a specific miner address 
- [x] function call for burning a specific amount of collateral
- [ ] parameter selection for how fast slasher share goes to 1

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -324,34 +324,30 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
 		Fatal("blocks do not prove a slashable offense")
 	}
 
-	if !self.Miners.Has(msg.From) {
-		Fatal("slash collateral must only be called by a miner actor")
-	}
-
 	miner := AuthorOf(block1)
 
 	// Check if miner has been slashed
+	// miner should be removed upon consensus fault slashing
 	if !self.Miners.Has(miner) {
 		Fatal("miner has been slashed")
 	}
 
-	// Burn all of the miners collateral
-	// const slashedCollateral = self.GetCollateral(miner)
-	const slashedCollateral = miner.Balance
-	// self.SetCollateral(miner, 0)
-	miner.Balance = 0.0
+	// Burn miner pledge collateral
+	// TODO determine amount to slash if not the entire balance
+	const minerPledgeCollateral = PledgeCollateralForPower(miner.Power)
+	const slashedCollateral = consensusFaultSlashAmount(minerPledgeCollateral)
 
 	// Some of the slashed collateral should be paid to the slasher
-	// growthRate determines how fast the slasher share of slashed collateral will increase as block elapses
-	// TODO parameter selection on growthRate and initialShare
-	// Current growthRate results in slasherShare reaches 1 after 30 blocks
+	// GROWTH_RATE determines how fast the slasher share of slashed collateral will increase as block elapses
+	// current GROWTH_RATE results in SLASHER_SHARE reaches 1 after 30 blocks
+	// TODO: parameter selection on GROWTH_RATE and INITIAL_SHARE
+	// TODO: define arithmetic precision and rounding for this operation
 	const blockElapsed = VM.CurrentBlockHeight() - block1.Height
-	const growthRate = 1.26
-	const initialShare = 0.001
-	const slasherShare = Min(Pow(initialShare, growthRate), 1.0)
-	// self.SetCollateral(msg.From, self.GetCollateral(msg.From) + slasherShare * slashedCollateral)
-	msg.From.Balance = msg.From.Balance + slasherShare * slashedCollateral
-	self.BurnCollateral((1.0 - slasherShare) * slashedCollateral)
+	const GROWTH_RATE = 1.26
+	const INITIAL_SHARE = 0.001
+	const SLASHER_SHARE = Min(Pow(INITIAL_SHARE, GROWTH_RATE), 1.0)
+	VM.TransferFunds(SLASHER_SHARE * slashedCollateral, msg.From)
+	self.BurnFunds((1.0 - SLASHER_SHARE) * slashedCollateral)
 	
 	// Remove the miner from the list of network miners
 	self.Miners.Remove(miner)
--- a/actors.md
+++ b/actors.md
@@ -334,8 +334,11 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
 
 	// Burn miner pledge collateral
 	// TODO determine amount to slash if not the entire balance
-	const minerPledgeCollateral = PledgeCollateralForPower(miner.Power)
-	const slashedCollateral = consensusFaultSlashAmount(minerPledgeCollateral)
+	// Tentatively, consensus fault should slash all pledge collateral
+	// only the balance is slashed if balance is less than collateral
+	// consensusFaultSlashAmount can return the Min for now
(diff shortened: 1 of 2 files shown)


actors.md
```diff
@@ -324,13 +324,35 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
 		Fatal("blocks do not prove a slashable offense")
 	}
 
+	if !self.Miners.Has(msg.From) {
+		Fatal("slash collateral must only be called by a miner actor")
```
decentralion: Why?

actors.md
```diff
@@ -324,13 +324,35 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
...
+	}
+
 	miner := AuthorOf(block1)
 
-	// TODO: Some of the slashed collateral should be paid to the slasher
+	// Check if miner has been slashed
+	if !self.Miners.Has(miner) {
```
decentralion: a miner is removed from the pool on slashing? this surprises me.

actors.md
```diff
@@ -324,13 +324,35 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
...
+		Fatal("miner has been slashed")
+	}
 
 	// Burn all of the miners collateral
-	miner.BurnCollateral()
-
+	// const slashedCollateral = self.GetCollateral(miner)
```
decentralion: remove commented code

actors.md
```diff
@@ -324,13 +324,35 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
...
+	const slashedCollateral = miner.Balance
```
decentralion: I thought we weren't slashing the entire balance on a fault? Can we leave a placeholder here for the amount slashed, e.g. something like:

`miner.Balance -= slashingAmountForFault(fault)`

actors.md
```diff
@@ -324,13 +324,35 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
...
+	// self.SetCollateral(miner, 0)
+	miner.Balance = 0.0
+
+	// Some of the slashed collateral should be paid to the slasher
+	// growthRate determines how fast the slasher share of slashed collateral will increase as block elapses
+	// TODO parameter selection on growthRate and initialShare
+	// Current growthRate results in slasherShare reaches 1 after 30 blocks
+	const blockElapsed = VM.CurrentBlockHeight() - block1.Height
+	const growthRate = 1.26
```
decentralion: can we factor these out as constants somewhere so it's easier for us to keep track of all the system parameters to tune? cc @anorth is there a consistent way we do this already

actors.md
zixuanzh: @decentralion My understanding was that you have to be a miner to have a collateral balance. Only those with a collateral balance (can be 0) can be rewarded when they call slash. I can totally be wrong on this.

actors.md
zixuanzh: Yes, this is what it currently is in the spec but likely to be changed. cc @whyrusleeping

actors.md
zixuanzh: will do after @anorth confirms that miner.Balance is the right way to get it value and we can directly set miner.Balance.

actors.md
whyrusleeping: For consensus slashing, yes. the miner is removed from the set of valid miners.

whyrusleeping: for storage fault slashing however, this should not be the case

actors.md
whyrusleeping: `self.Balance` should be fine for accessing it, but in order to remove funds, they need to be moved somewhere. If burning funds, use `BurnFunds`

actors.md
```diff
@@ -324,13 +324,35 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
...
+	const initialShare = 0.001
+	const slasherShare = Min(Pow(initialShare, growthRate), 1.0)
+	// self.SetCollateral(msg.From, self.GetCollateral(msg.From) + slasherShare * slashedCollateral)
+	msg.From.Balance = msg.From.Balance + slasherShare * slashedCollateral
```
whyrusleeping: use `Transfer` to move funds to other actors

actors.md
```diff
@@ -324,13 +324,35 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
...
+	miner.Balance = 0.0
+
+	// Some of the slashed collateral should be paid to the slasher
+	// growthRate determines how fast the slasher share of slashed collateral will increase as block elapses
+	// TODO parameter selection on growthRate and initialShare
+	// Current growthRate results in slasherShare reaches 1 after 30 blocks
+	const blockElapsed = VM.CurrentBlockHeight() - block1.Height
+	const growthRate = 1.26
+	const initialShare = 0.001
+	const slasherShare = Min(Pow(initialShare, growthRate), 1.0)
```
anorth: Not necessarily right now, but we need to define the arithmetic precision and rounding mode for this operation

actors.md
zixuanzh: @decentralion I think we might be slashing the entire amount here if it is consensus fault. Otherwise, we have to decide how much to slash. cc @whyrusleeping

actors.md
zixuanzh: @decentralion since we are slashing for consensus fault, a miner is removed from the pool in this case.

actors.md
anorth: A miner's balance probably be greater than their collateral requirements, but could be less. We should slash the min of those two. IMO we should not burn more than the miner's collateral requirement just because they happen to have that on balance.

zixuanzh: @anorth resolved in latest commit

actors.md
```diff
@@ -326,11 +326,32 @@ func SlashConsensusFault(block1, block2 BlockHeader) {
...
+
+	// Some of the slashed collateral should be paid to the slasher
+	// GROWTH_RATE determines how fast the slasher share of slashed collateral will increase as block elapses
+	// current GROWTH_RATE results in SLASHER_SHARE reaches 1 after 30 blocks
+	// TODO: parameter selection on GROWTH_RATE and INITIAL_SHARE
+	// TODO: define arithmetic precision and rounding for this operation
+	const blockElapsed = VM.CurrentBlockHeight() - block1.Height
+	const GROWTH_RATE = 1.26
+	const INITIAL_SHARE = 0.001
+	const SLASHER_SHARE = Min(Pow(INITIAL_SHARE, GROWTH_RATE), 1.0)
```
whyrusleeping: I think this should be:
```
slasherSlash = min( INITIAL_SHARE * Pow(growthRate, blockElapsed), 1.0)
```
right?

whyrusleeping: (blockElapsed should definitely be involved somehow)

whyrusleeping: We're also going to have to be careful here, floating point precision and rounding needs to be very explicitly defined.

zixuanzh: Not exactly sure on the right floating point precision and rounding here though

mishmosh:
@zixuanzh The last updates were in Sept, around the time of the big spec overhaul. Is this PR stale or still active?

zixuanzh:
I was under the impression that this has been added to the current spec but looks like it was left as a TODO. I will author the change here into the current spec. cc @sternhenri @jzimmerman

zixuanzh:
Added in #789

PR #384, opened 2019-07-09
add pseudocode for checking miner power

whyrusleeping:


actors.md
```diff
@@ -519,6 +519,8 @@ type StorageMinerActorMethod union {
     | GetSectorSize 11
     | UpdatePeerID 12
     | ChangeWorker 13
+    | IsSlashed 14
+    | IsLate 15
```
dignifiedquire: can you add a todo to define these?

storage-market.md
```diff
@@ -119,6 +119,27 @@ The `power table` is exported by the storage market for use by consensus. There
 
 TODO: rephrase the above to make it clear that power is updated only on PoSt submission and when slashed.
 
+To check the power of a given miner, use the following:
+
+```go
+func GetMinersPowerAt(ts TipSet, m Address) Integer {
```
dignifiedquire: can you clarify in which situations this would be actually called?

sternhenri: argh. I could have used this working on #403. Hard to read through PRs before starting one. :(

PR #394, opened 2019-07-16
restore POST_SECTORS_COUNT definition

laser:
`POST_SECTORS_COUNT` was defined previously, but this definition was (presumably inadvertently) deleted in #79f045362ebc07f4960b996110d95d3d32587a68.


laser:
cc @nicola

nicola:
I think this might have been related to friedel realization that we could construct a PoSt for unbounded number of sectors (he has that written up in his new PR).

I am happy otherwise to put this back until his PR gets merged.

PR #403, opened 2019-07-23
Powertable lookback 

sternhenri:
Change of how we handle power table lookbacks in leader election: fully removed, using safe power as of last post in order to simplify slashing etc.

A few notes on this:
- Case B presents the following advantages:
    - Overall simpler, and secure by default rather than thanks to mitigations
But:  
    - case B will be expensive since the verifier will have to look back up to 2PP heights in order to find when the storage market actor provided their last proof (may be easy optimizations there, I have some in mind).

TODO:
- [x] Align on approach (@whyrusleeping @anorth @dignifiedquire @sternhenri)
- [x] If B, spec exactly how we get the appropriate value for L (at an API level)
- [x] remove current duplication of checking for whether miner was slashed (both in general block validation and in leaderElection validation in mining.md), potentially refactor `isMiner` method to also check that provingPeriodEnd has not past yet (i.e. not a case of unreported but obvious fault).
- [x] Check for fenceposts.
- [x] Review post https://github.com/filecoin-project/specs/pull/384 merge
Wider context: 

This issue arose out of a conversation I was having with @dignifiedquire regarding how we handled faults (from a convo with @Whyrusleeping), specifically the necessity to slash pledge collateral (consensus) for storage faults (storage market).

In that conversation, Why and I had taken the power table lookback parameter (from which a miner chooses where to calculate their fraction of power from in leader election) to be a network-wide parameter. Thereby an effective sampling rate, whose value determines how "stale" the power data we use is (ie how many new proofs are left out of the determination). (CASE A)

Dig on the other hand, was under the belief that this power table lookback was a miner-specific parameter, equal to the height at which the miner got their before-last PoSt on-chain. In that sense, it was used to look back to the miner's last proven power and compare that to total power in the chain 1 round back (CASE B): they earn blocks up to 2 proving Periods-1 rounds late, but only on secure power.

In CASE A, we need to slash consensus collateral on storage faults lest the miner cut their storage right after submitting a proof and keep unduly earning block rewards for provingPeriod + power table lookback) rounds.

In CASE B, that is not an issue. Because the lookback is specific to the miner's earnings are more 'delayed' but the above attack would not work: the miner earns from "secure storage" that has been proven again.


sternhenri:
I will add that regardless of where we end up on L value, the extra validation steps checking for slashing here should be merged.

anorth:
I think that any construction that slashes pledge collateral for a storage fault(*) is untenable from a miner's operational risk perspective. I have made this case to @dignifiedquire @nicola and @protocolin  previously; please let me know if I should lay it out again here or elsewhere.

A construction that computes storage power from storage that *has in fact been proven* is much simpler and with smoother edges than a construction that computes power from an expectation of future proof and then heavily penalises (inevitable) failure to prove as indistinguishable from a malicious act.

(*) where storage fault means failure to get a PoSt on chain within a fairly generous time window (e.g. a datacentre-level network or power outage), or failure to include some individual sectors in a PoSt due to unpredictable unavailability.

whyrusleeping:
@anorth the faults in question are the miner cheating at consensus. Effectively lying about how much power they have. It doesnt matter that the fault is technically due to them having lost storage.

A miner could put up no storage collateral, seal a bunch of garbage data, drop it all immediately after their first post, and still have power until after their next PoSt submission. This is fine from a security standpoint IFF pledge collateral can be slashed for this (proportional to the amount of storage dropped).

anorth:
In the context of this issue, making:

> drop it all immediately after their first post, and still have power until after their next PoSt submission

...not true is an alternative that it seems like some lookback mechanism could address. But disclaimer I haven't thought deeply through that.

I will write up more about my risk objections.

sternhenri:
I agree with @anorth's last point @whyrusleeping, re 
> This is fine from a security standpoint IFF pledge collateral can be slashed for this

I think if you choose to use a farther back lookback such that only storage that has been book-ended (before after) with valid PoSTs is ever used for consensus, you should be ok.

That said, from convos with @dignifiedquire I still believe there is deeper-seated misalignment behind this issue. I would recommend using a forthcoming spec call to try and resolve it synchronously (saving us all time at the cost of fleshing out a written debate that will take longer to resolve -- cc @pooja)

sternhenri:
After a conversation with @dignifiedquire and @whyrusleeping today (interlude within a larger one). I believe we are all aligned with this, @anorth. A few specific asks for review outlined in code.

Will be cleaner to review once #290 is merged (sorry for those dependencies).

mining.md
```diff
@@ -132,7 +132,7 @@ func VerifyBlock(blk Block) {
 
 	// 3. Verify miner has not been slashed and is still valid miner
 	curStorageMarket := LoadStorageMarket(blk.State)
-	if !curStorageMarket.IsMiner(blk.Miner) {
+	if !curStorageMarket.IsMiner(blk.Miner) && !blk.Miner.ProvingPeriodEnd <= blk.height {
```
sternhenri: duplicate checks here to l. 260. Which should be kept? Both checks should happen together in any case. 

May be worth amending `isMiner` if it's not used elsewhere... @dignifiedquire ?

mining.md
```diff
@@ -210,75 +197,127 @@ func (state StateTree) LookupPublicKey(a Address) PubKey {
...
+		iii. # Check for an unreported storage fault
+			if M.ProvingPeriodEnd < N
+				return 0
+2. Determine the miner's power fraction
+		i. 	# Determine total storage 1 round back
+  			S_n <-- storageMarket(N)
+	  		p_n <-- S_n.GetTotalStorage()
+		ii. # Determine miner's power at last valid PoSt submission
+  			miner <-- block.miner
+  			S_m <-- storageMarket(miner.provingPeriodEnd - provingPeriodDuration(miner.SectorSize)
```
sternhenri: spot check needed above

expected-consensus.md
```diff
@@ -144,132 +145,158 @@ the prior block, or "losing" tickets generated in this epoch.
...
-  i. y <-- Map(pp,W) : {0,1}^n <-- pp x W
-  ii. If y < p_f then W is a winning lottery ticket
-    Return 1
-  iii. If y > p_f then W is not a winning lottery ticket
-    Return 0
+0. Prepare new election proof
+	i. newEP <-- New()
+1. Determine the miner's power fraction
+	i. 	# Determine total storage 1 round back
+  		S_n <-- storageMarket(N)
```
sternhenri: spot check: will new state have been computed by this point? Will total power be accounted for in this round? or is this a fencepost?

expected-consensus.md
```diff
@@ -144,132 +145,158 @@ the prior block, or "losing" tickets generated in this epoch.
...
+  		p_n <-- S_n.GetTotalStorage()
+    ii. # Determine own power as of last submitted PoSt
+        appropriateHeight <-- self.ProvingPeriodEnd - ProvingPeriodDuration(self.SectorSize)
+        S_m <-- storageMarket(appropriateHeight)
+  		p_m <-- S_m.PowerLookup(self)
```
sternhenri: likewise here... should we look at `A =  ProvingPeriodEnd - Duration` or at `A - 1`?

expected-consensus.md
```diff
@@ -144,132 +145,158 @@ the prior block, or "losing" tickets generated in this epoch.
...
+4. Return the new ticket
 ```
 
-#### Checking election results
+### Checking election results
+
+Now, a miner must also check whether they are eligible to mine a block in this round. For how Election Proofs are validated, see [election validation](mining.md#election-validation).
 
-Now, a miner must also check whether they have won in this round.
+To do so, the miner will use tickets from K blocks back as randomness to uniformly draw a value from 0 to 1. Comparing this value to their power, they determine whether they are eligible to mine. A user's `power` is defined as the ratio of the amount of storage they proved as of their last PoSt submission to the total storage in the network as of the current block.
```
anorth: With the appropriate bookkeeping in the miner actor, the storage proven at last PoSt should be the same value as their storage power in the current block, so both can be sampled from the same point. The devil is in "appropriate bookkeeping".

anorth:
I said a silly thing about sampling near the chain head. A delay of one proving period plus any grace periods is necessary in order to detect one or more sectors failing, thus missing the right book-end.

anorth:
An [illustrative diagram](https://docs.google.com/spreadsheets/d/1WT5auHKiPZ44HbsapopdPIUvQzJEQdPdecN9pPpMRuI/edit#gid=0). The miner does indeed enjoy power for one period after in fact dropping a sector, but balanced by the delay gaining the power in the first place.

With such a construction we might not even need grace periods and late fees - just lose power immediately (although refer to a question I have yet to post in an issue about whether a sealed sector can be re-posted after faulting, without re-sealing).

sternhenri:
So I misspoke in previous posts about bookend. Sorry for changing things so fast/often (cost of sync vs async). 

I no longer see why Bookend is better than simply using power from the last valid PoST. The problem of someone dropping power immediately after a PoST is true in either case (whether you use the two consecutive posts or just the last one). Your point about the fact that things even out because power kicks in late by the same amount, also stands in both cases...

I added what I think to your diagram (I'm assuming commiting a sector takes a PP, which is not nec true). I couldn't see disk usage in your section, but it struck me that potentially the miner isn't given power for all moments in which their disk is occupied.

On the last point, I agree with you. I would cut power to 0 immediately, not waiting for grace period, however I would only slash storage collateral after grace period. (higher cost). This is in this PR already.

decentralion:
I want to chime in from a cryptoecon perspective, and say that I agree with @anorth that slashing pledge collateral for storage faults is really unreliable. We're setting pledge collateral based on high-level system considerations like creating long-term incentive alignment, managing FIL liquidity, etc. This has led to high economic value for pledge collateral--which I think is fine, so long as miners know that the pledge collateral is safe; ie they will never lose it unless they deliberately misbehave. If there's a risk of losing pledge collateral by accident, it creates a really big new risk for miners and may change miner economics considerably.

So, glad to hear that it sounds like we'll be able to make pledge collateral independent from storage faults. It will also simplify the slashing/fault spec a lot.

sternhenri:
^ provably misbehave

anorth:
> I couldn't see disk usage in your section, but it struck me that potentially the miner isn't given power for all moments in which their disk is occupied.

That's an excellent addition to the diagram and I agree so long as we are convinced that `CommitSector` consumes the disk space for the whole period. In the case of a single miner, I agree, both in case the sector is being sealed, or already stored and being re-committed. I assume the PoRep construction prevents two miner actors from committing the same sector.

anorth:
> I would only slash storage collateral after grace period

What do you mean by this? Details are probably best left for a follow-up PR, but if you mean
a) deal clients (if any) become eligible to arbitrate their deal and claim storage collateral, 👍 
b) some nontrivial collateral is burned to the network, 😞 

"Slash" is a bit of a blurry word to use in the context of deal arbitration (but pretty clear for consensus-style burning of a pledge).

sternhenri:
> That's an excellent addition to the diagram and I agree so long as we are convinced that CommitSector consumes the disk space for the whole period. In the case of a single miner, I agree, both in case the sector is being sealed, or already stored and being re-committed. I assume the PoRep construction prevents two miner actors from committing the same sector.

Your assumption is right (to the best of my knowledge), your point on whole period well taken. In practice you can seal a bit faster, but it's a simple bound on power (ie limited advantage). I'll think on it, but the simplicity of separation of concerns may make this worthwhile even with no mitigation.

Also, I found the source of my own confusion re "bookends" and my flip flopping. We may actually be saying the same thing. The q is: "if a miner commits to 10 sectors, and PoSts then commits to another 10 within the Proving Period and PoSts again... So we have a PoSt committing 10, and then a second one committing 20. Between the second and third PoSts, does the miner have power for 10, or for 20?"

I had meant 10 initially, and then switched to 20. I believe this relates to our above discussion...

Re slash, fair enough. I meant broadly either mechanism, insofar as it would mean the miner would have to pay up again to regain their power without SEALing. @dignifiedquire and I are shipping this to spec tomorrow, and he will be working on what I mean by "slashing" specifically :)

anorth:
Great, let's defer discussion of "slashing" for the next PR.

definitions.md
```diff
@@ -161,20 +161,18 @@ A piece is a portion of a file that gets fitted into a sector.
...
 Security Parameter. Polling time is the time between two online PoReps in a PoSt proof.
 
+#### Power
+
+See `Power Fraction`.
+
 #### Power Fraction
 
-A miner's `Power Fraction` is the ratio of their committed storage over Filecoin's committed storage. It is used in leader election.
+A miner's `Power Fraction` or `Power` is the ratio of their committed storage as of their last PoST submission over Filecoin's total committed storage as of the current block. It is used in leader election.
```
dignifiedquire: maybe "as of their last **expected** PoSt submission", as missing a PoSt changes the power too

expected-consensus.md
```diff
@@ -65,22 +65,23 @@ func Mine(minerKey PrivateKey) {
...
-  minerPower := GetPower(tipset.state, key.Public())
-  totalPower := GetTotalPower(tipset.state)
+  minerPower := GetPower(miner)
+  totalPower := state.GetTotalPower()
 
-  return IsProofAWinner(eproof, minerPower, totalPower), eproof
+  if IsProofAWinner(eproof, minerPower, totalPower)
+    return eproof
+  else
+    return 0
```
dignifiedquire: the signature off the function expects `bool, ElectionProof`, but you return sth different here

expected-consensus.md
dignifiedquire: This one seems right to me, as this allows for detection of late PoSt

mining.md
dignifiedquire: - nit: `blk.Miner.ProvingPeriodEnd > blk.height` is easier to read
- I am not sure what exactly this protects against, could you elaborate?

mining.md
```diff
@@ -252,45 +252,61 @@ In order to determine that the mined block was generated by an elegible miner, o
...
+			if VRFState == "INVALID":
+				return 0
 5. Everything checks out, it's a valid election proof
 		return 1
 ```
 
+Note: For clarity we explicit what may happen when a miner fails to submit a PoSt:
+- If they are currently late to submit a PoSt, 1.iii above is triggered and they will not mine.
+- If they are late to submit a PoSt and are slashed/reported for it, then they will be taken out of the power table (at least temporarily) and 1.ii will be triggered.
+- If they are late to submit a PoSt but the fault was unreported before they finally submitted it, then they can resume mining normally (they got lucky).
```
dignifiedquire: Power should be lost for any period, even if they are not slashed explicitly. I am not sure how you end up in this situation.

definitions.md
sternhenri: I think it might be confusing to distinguish between "expected PoSt" and "actual PoSt" we should just say PoSt submission when it happened and "failed to PoSt" or something when it didn't. 

So your change is great as is.

expected-consensus.md
sternhenri: True. Fixed

sternhenri:
@anorth checked with @dignifiedquire. It's not that the disk is taken up for entire SEAL. Simply that power from committed sector doesn't kick in until you've posted your current PoSt (ie one PoSt later). So we're all set.

Issue #176, opened 2019-03-06
Modify spec so we can compute slashing conditions in state machine


ZenGround0:
Making updates in the consensus slashing rules PR it has occurred to me that we don't have the necessary things to actually compute consensus slashing from within the `SlashConsensusFault` state machine method.

(1) To slash when there are overlapping ticket rounds `SlashConsensusFault` needs some way to read ticket round numbers from the two proving block headers.  Because `Height` is moving to a measure of the number of *blocks* in a chain and no longer a measure of *tickets* the [block data structure](https://github.com/filecoin-project/specs/blob/master/data-structures.md#block) needs to include a round number to evaluate this slashing condition only from block headers.  Alternatively we could have nodes maintain round number locally and include a new method on the [state machine execution context](https://github.com/filecoin-project/specs/blob/master/data-structures.md#block) that looks up and returns the base round number of a block.

(2a) To slash when there is a dropped ancestor block we need proof that there is a chain of blocks from `b1` back to ancestor TipSet `A` such that another block `b2` is a valid extension of `A` but is not in `A`.  One way to do this is by adding in an execution context function that can return ancestors of `b1` up to the parent of `A`.  

(2b) It is harder to handle the case where the `SlashConsensusFault` caller's chain is built off of `b1` and the fork is built off of `A`.  In this case the execution context will not in general have access to the fork as it may not be stored locally and the methods on the execution context should not trap into the network.  It seems like the best way to handle this is to carry a list of block headers from `b1` to the parent of `A` in the slashing message.  The parent hashes and block signatures can be processed to prove the existence of the fork from `b1` to `A` within `SlashConsensusFault`, and again the relationship between `b2` and `A` can be proven using `A`'s parent.  This mechanism will handle the (2a) case and should be used there too to keep things simpler.

whyrusleeping:
@ZenGround0 thanks for pointing this out!

re: 1, I think either suggestion is fine, including an additional field in the block isnt *that* bad (annoying extra bytes, but not the end of the world), and also adding a method  to the context that 'does the right thing' under the hood also sounds good (more potentially unaccounted for work, also weird questions around data availability, but maybe it could work).

re 2: Yeah, data availability here is an issue, submitting all necessary blocks to the `SlashConsensusFault` method is the right thing to do here. This also makes the case that we should work to shrink the block headers as much as possible to make this not become abhorrently expensive.

pooja:
[...]

PR #412, opened 2019-07-26
Explicitly add parent tipset to slashing function argument

ZenGround0:
By requiring the parent tipset of `block1` to be passed explicitly as a parameter this PR closes out the concern of #176 as now all recipients of this message can apply it to their state regardless of which forks they are tracking.  

Taking this opportunity to clean up the method a little bit by requiring the arguments to be ordered during message construction for type (2) slashing conditions.

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -282,33 +282,29 @@ type SlashConsensusFault struct {
 **Algorithm**
 
 ```go
-func shouldSlash(block1, block2 BlockHeader) bool {
+func shouldSlash(block1, block2 BlockHeader, block1Parents TipSet) bool {
 	// First slashing condition, blocks have the same ticket round
 	if sameTicketRound(block1, block2) {
 		return true
 	}
 
-	// Second slashing condition, miner ignored own block when mining
-	// Case A: block2 could have been in block1's parent set but is not
-	block1ParentTipSet := parentOf(block1)
-	if !block1Parent.contains(block2) &&
-		block1ParentTipSet.Height == block2.Height &&
-		block1ParentTipSet.ParentCids == block2.ParentCids {
-		return true
-	}
-
-	// Case B: block1 could have been in block2's parent set but is not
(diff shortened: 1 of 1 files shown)


actors.md
```diff
@@ -282,33 +282,29 @@ type SlashConsensusFault struct {
...
-		block2ParentTipSet.Height == block1.Height &&
-		block2ParentTipSet.ParentCids == block1.ParentCids {
+	// Second slashing condition, miner ignored own block when mining.
+	// block2 could have been in block1's parent set but is not
+	// verify that block1's parents don't contain block2
+	if block1Parents.Cids == block1.ParentCids &&
+		!block1Parents.contains(block2) &&
+		block1Parents.Height == block2.Height &&
+		block1Parents.ParentCids == block2.ParentCids {
 		return true
```
sternhenri: This is to say: Block1Parent is indeed Block1's parent, block1parent omits block 2, even though block 2 should be in block1parent...

Aren't you forgetting a check which is that block2.miner == block1.miner? (otherwise could just be network failure)

you might also check that block2's is the min ticket (rational reason for omitting) but arguably that doesn't matter. Let's penalize either way.

sternhenri:
here's a quick idea for a third condition... What do you think?

- Miner provably switches from a heavier chain to a lighter chain. 

This may be silly, but I can't think of an honest reason to do this... I'm sure there is one. (I thought about this as I was trying to figure out why we had determined it had to be parent omission as opposed to ancestor omission, and it's because there might be a heavier chain omitting the ancestor (ie other user entropy accounting for the drop)). However, what they have in common is that this error is proof that a miner is knowingly abandoning a block/opting for a lighter chain, which must mean they are operating on private info -- i.e. grinding.

sternhenri:
Rendered obsolete by https://github.com/filecoin-project/specs/pull/789

PR #414, opened 2019-07-26
bump up actor IDs by one so 0 is free for sending

whyrusleeping:
If we reserve 0 for constructors, then it becomes rather difficult to just send money to an actor. So now method ID 1 is reserved for constructors, ID 0 is kept empty for actors to receive funds on, and the rest are open.


whyrusleeping:
note added

Issue #391, opened 2019-07-15
add treasury and vesting to spec


decentralion:
As drafted in #325, we need a way to calculate the total available FIL, e.g. via `availableFil := minedFil + vestedFil - burnedFil`. @anorth suggested using a single treasuryActor, which is the source of block rewards, vested FIL, and to whom burned FIL returns. in that case, checking available FIL would be as simple as subtracting the total token supply from the current balance of the treasury actor. 

We already have a treasury actor who starts with the full token balance, and is transfers out block rewards. So the main thing to spec is the vesting. Vesting is a little complicated because there are multiple different vest schedules (6 month, 1 year, etc).

There are a few ways we could encode this:
- The treasury actor directly contains the vesting logic, and effects transfers to every vestee directly
- There is a single vesting actor which receives funds from the treasury actor, and effects transfers to the vestees
- There is a vesting actor for each vest schedule, each of which receives funds from the treasury actor, and effects transfers to vestees

These are basically different tradeoffs of putting the complexity in the treasury actor vs. delegating to other actors. Assuming we go with the delegated approach, we can also decide when the treasury actor sends the funds: either all at once in the genesis block, or incrementally (e.g. monthly) as the funds vest. The first approach is to implement, but it makes calculating the total available FIL a little harder (because finding the available FIL now requires iterating over each vesting actor).

We should also give some thought to how vested funds are released. A simple approach is to have each vesting actor transfer the vested FIL to a receiving account every month, but this might spam the network with transactions, and it's not clear how the transaction fee should be chosen. A better approach would be to architect it so that each vestee chooses when to withdraw funds, and can choose an appropriate transaction fee at that time.

anorth:
I suggest that vesting means that funds become available to be withdrawn, but does not actually automatically send the funds anywhere. Things are much simpler to reason about and code if an explicit network message is required to move funds (or change any state in general). The receiver of funds clearly has the incentive to send that message in order to receive their funds.

Minor note: burnt funds go to the burnt funds actor (id 99), which I think is fine.

taoshengshi:
> I suggest that vesting means that funds become available to be withdrawn, but does not actually automatically send the funds anywhere. Things are much simpler to reason about and code if an explicit network message is required to move funds (or change any state in general). The receiver of funds clearly has the incentive to send that message in order to receive their funds.
> 
> Minor note: burnt funds go to the burnt funds actor (id 99), which I think is fine.

[...]

PR #416, opened 2019-07-30
Add linear unlocking functionality for funds held in multisig actors

jesseclay:
This PR adds the ability to put a linear unlock on funds held in multisig actors by passing in a particular unlock period (in blocks from time of account creation -- each block estimated to be 30 seconds). This will be used for Filecoin investor account unlocking, however this could also be used for other use-cases wishing to keep funds locked until a certain block height. With this function the account holder(s) are able to transfer up to and including all funds that have unlocked at that point. Any attempt to transfer more than that will fail.

If the user wishes to not use this functionality at all they can just pass `0` to the initialization function for `unlockDuration`, then no unlock will be used and all funds will be spendable immediately.

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -1522,10 +1522,10 @@ func Propose(to Address, value TokenAmount, method String, params Bytes) UInt {
 	self.Transactions.Append(tx)
 
 	if self.Required == 1 {
-		if !self.canSpend(value) {
+		if !self.canSpend(tx.value) {
 			Fatal("transaction amount exceeds available")
 		}
-		tx.RetCode = vm.Send(tx.To, tx.Value, tx.Method, tx.Params) 
+		tx.RetCode = vm.Send(tx.To, tx.Value, tx.Method, tx.Params)
 		tx.Complete = true
 	}
 
--- a/actors.md
+++ b/actors.md
@@ -1345,6 +1345,9 @@ type MultisigActorState struct {
     signers [Address]
     required UInt
     nextTxId UInt
+    initialBalance UInt
+    startingBlock UInt
+    unlockDuration UInt
     transactions {UInt:Transaction}
 }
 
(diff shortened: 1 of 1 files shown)


jesseclay:
and a big thanks to @whyrusleeping for the help on this PR.

anorth:
> This will be used for Filecoin investor account unlocking

How does this relate to the ideas expressed in https://github.com/filecoin-project/specs/issues/391? cc @decentralion

decentralion:
As an implementation of unlocking vested funds, this sounds good to me -- I like that the funds progressively unlock but it still requires a user action to transfer them out.

For the calculation of available FIL in #391 - it looks like we would have a separate multisig for every vestor's vest (one investor could have multiple multisigs if they bought with multiple vesting periods). So it wouldn't make sense to sum over all these contracts every time we are caulculating available FIL. However, the whole vesting schedule is known in advance, so we could just hardcode the vesting schedule for calculation purposes independently from the mechanism that actually vests the FIL. It would admit the possibility that these two desynchronize, but if we are careful it won't happen (and the complexity is not that high).

actors.md
```diff
@@ -1440,6 +1448,9 @@ func Propose(to Address, value TokenAmount, method String, params Bytes) UInt {
 	self.Transactions.Append(tx)
 
 	if self.Required == 1 {
+		if !self.canSpend(value) {
```
whyrusleeping: `tx.Value` here

dignifiedquire:
needs rebasing on master

actors.md
jesseclay: https://github.com/filecoin-project/specs/pull/416/commits/36a09c037300206aeef0cb9df415aa7489613ae8

actors.md
```diff
@@ -1560,6 +1571,9 @@ func Approve(txid UInt) {
 	tx.Approved.Append(msg.From)
 
 	if len(tx.Approved) >= self.Required {
+		if !self.canSpend(tx.Value) {
+			Fatal("transaction amount exceeds available")
+		}
 		tx.RetCode = Send(tx.To, tx.Value, tx.Method, tx.Params)
```
jesseclay: @whyrusleeping shall we change this to vm.Send to remain consistent with 1528?

jesseclay:
@dignifiedquire should be up to date now: https://github.com/filecoin-project/specs/pull/416/commits/ff27875c57d094d92c7862fd607c2e9f0e1ecc43

actors.md
whyrusleeping: Yeah, thats probably the right thing to do.

jesseclay: https://github.com/filecoin-project/specs/pull/416/commits/1428d0f1e46a0f78376ff30d78adec279465ceef

jesseclay:
@dignifiedquire @whyrusleeping all comments addressed now. LMK if you'd like me to merge or just feel free.

PR #423, opened 2019-07-31
Prevent MultiSig actor lockout and a takeover

Kubuxu:
When removing Singers and changing the Required number of approvals it was previously possible to lockout the MultiSig.

The introduction of `increaseReq` is to prevent a takeover when the number of Signers is increased from 1.
Without it, there is no way to safely add another signer to a MultiSig with just one Signer.

Also, retain return code of called methods.

Removed `created` as it wasn't being set.

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -1440,7 +1440,7 @@ func Propose(to Address, value TokenAmount, method String, params Bytes) UInt {
 	self.Transactions.Append(tx)
 
 	if self.Required == 1 {
-		tx.RetCode := vm.Send(tx.To, tx.Value, tx.Method, tx.Params)
+		tx.RetCode = vm.Send(tx.To, tx.Value, tx.Method, tx.Params)
 		tx.Complete = true
 	}
 
@@ -1566,7 +1566,7 @@ type AddSigner struct {
 **Algorithm**
 
 ```go
-func AddSigner(signer Address, incraseReq bool) {
+func AddSigner(signer Address, increaseReq bool) {
 	if msg.From != self.Address {
 		Fatal("add signer must be called by wallet itself")
 	}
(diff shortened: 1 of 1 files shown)


actors.md
```diff
@@ -1440,7 +1440,7 @@ func Propose(to Address, value TokenAmount, method String, params Bytes) UInt {
 	self.Transactions.Append(tx)
 
 	if self.Required == 1 {
-		vm.Send(tx.To, tx.Value, tx.Method, tx.Params)
+		tx.RetCode := vm.Send(tx.To, tx.Value, tx.Method, tx.Params)
```
whyrusleeping:
```suggestion
		tx.RetCode = vm.Send(tx.To, tx.Value, tx.Method, tx.Params)
```

actors.md
```diff
@@ -1559,19 +1559,23 @@ func ClearCompleted() {
...
     signer Address
+    increaseReq bool
 } representation tuple
 ```
 
 **Algorithm**
 
 ```go
-func AddSigner(signer Address) {
+func AddSigner(signer Address, incraseReq bool) {
```
whyrusleeping:
```suggestion
func AddSigner(signer Address, increaseReq bool) {
```

actors.md
```diff
@@ -1597,6 +1601,10 @@ func RemoveSigner(signer Address) {
 	if !self.isSigner(signer) {
 		Fatal("given address was not a signer")
 	}
+	if len(self.Signers)-1 < self.Required {
```
dignifiedquire: why not have an option for reducing required as well?

Kubuxu: Added.

actors.md
```diff
@@ -1652,6 +1661,9 @@ func ChangeRequirement(req UInt) {
 	if req < 1 {
 		Fatal("requirement must be at least 1")
 	}
+	if req > len(self.Signers) {
+		Fatal("requirement must be less than nuber of signers")
```
dignifiedquire: typo nuber

Kubuxu:
```suggestion
		Fatal("requirement must be less than number of signers")
```

PR #429, opened 2019-08-02
clean up unused fields in the payment info struct

whyrusleeping:
the channel is an outdated concept, and the payer is in the payment channel actors state.


anorth:
FYI @acruikshank

PR #439, opened 2019-08-15
actors.md: Split StorageMiner.PaymentVerify

magik6k:
Having `PaymentVerify` in a single method made things harder to interpret, and required slightly more on-chain bytes

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -519,7 +519,8 @@ type MinerInfo struct {
 | `ChangeWorker` | 14 |
 | `IsSlashed` |  15 |
 | `IsLate` | 16 |
-| `PaymentVerify` | 17 |
+| `PaymentVerifyInclusion` | 17 |
+| `PaymentVerifySector` | 18 |
 
 #### `Constructor`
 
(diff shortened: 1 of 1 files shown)


whyrusleeping:
LGTM, lets get a review in from @dignifiedquire too

anorth:
Is this primarily aimed at deal extension and other cases where a miner has already sealed the data? In the "normal" case, I understand a client has to produce all the vouchers before a miner will commit to sealing their data. Or is the expectation instead that a client might take this small risk because they can challenge and reclaim their funds if the miner doesn't produce a PIP for them?

cc @acruikshank

magik6k:
When starting a deal, client sends the miner a voucher with `PaymentVerifyInclusion`, and after the miner seals the sector, he can nicely ask the client off-chain (giving inclusion proof, possibly a small discount), to produce a new voucher on the same lane with `PaymentVerifySector`.

The reason for that is `PaymentVerifySector` is much smaller and much cheaper to post to chain.

actors.md
```diff
@@ -1122,51 +1123,71 @@ func IsSlashed() (bool) {
...
-    Fatal("miner does not have sector referenced by piece inclusion proof")
-  }
+Verifies a storage market payment channel voucher's 'Extra' data by checking for presence of a specified sector in miner's sector set.
+
+Miners should prefer payment vouchers with this method used for validation over `PaymentVerifyInclusion`, because posting them to the chain will be much cheaper.
 
-  return ValidatePIP(pip, dataCommitment)
+Clients should only create such vouchers after verifying that miners have related sectors in their sector set, and after checking piece inclusion proof.
+
+Miners can incentivize clients to produce such vouchers by applying small 'discount' to amount of token clients have to pay.
```
dignifiedquire: > by applying **a** small discount to **the** amount of token clients have to pay

dignifiedquire:
@magik6k @whyrusleeping I understand it is useful, but I am having trouble justifying the complication of yet another method, vs the actual savings this gives. Could give  a little more detail why this change is warranted?

magik6k:
PaymentVerifyInclusion / PaymentVerifySector are doing 2 different things:

* One is verifying piece inclusion proof, the other is only checking for sector in miners' sector set.
* One is expensive to run, the other is cheap (depending on how gas charging works it might be useful to have those methods split)
* Serialized params for PaymentVerifySector are tiny (probably <= 5 bytes in most cases)
* The 2 methods are actually simpler IMO (there is only one correct path for each, params can be interpreted in one way only)

whyrusleeping:
@dignifiedquire adding methods isnt complicated at all, but having larger VM methods with more error cases and conditionals is drastically more complex (think of the tests).

All this is doing is breaking the single method with nested conditionals into two separate methods. Makes it easier to reason about, and easier to test.

PR #440, opened 2019-08-16
docs(specify SectorID type as uint64)

### LINKED ISSUE #437: define SectorID type
The [data-structures](https://github.com/filecoin-project/specs/blob/master/data-structures.md#sectorset) section of the specification doesn't say anything about this type, and the `SectorSet` data structure (indexed by sector id) isn't specified either. I suggest an unsigned, 64-bit integer (which is what `go-filecoin` and the sector builder use today) for SectorID.

Quoting @whyrusleeping 

> [...] the definition of the SectorID is missing. It should just be an integer. Technically up to 31 bytes long, but I think for all practical purposes we can just use a uint64 as we will never have a single miner with 2^64 sectors

Relevant comment [here](https://github.com/filecoin-project/go-sectorbuilder/issues/5#issuecomment-515815855).

laser:
Fixes #437.


laser:
@dignifiedquire - Updated with expected transformation inputs/outputs and added some prose as per your request.

data-structures.md
```diff
@@ -319,6 +319,25 @@ A type to represent an amount of filecoin tokens.
...
 
+### SectorID
+
+Uniquely identifies a miner's sector.
+
+```sh
+type SectorID uint64
+```
+
+Note: The [`Seal`][seal-link] and [`VerifySeal`][verify-seal-link] proofs
```
whyrusleeping: This feels like a detail that shouldnt be exposed. The proofs code should just accept integers, and whatever weird serialization and padding it needs it should just do internally.

PR #452, opened 2019-08-22
clean up data structures for retrieval protocol

whyrusleeping:
working through retrieval implementation. Clarifying the actual format of the datastructures here.

One slightly out-of-the-blue thing here is that i added an 'unsealing' response status code. The idea here is that we're telling the client that we are going to respond to their request, but we need them to come back in a little bit

### DIFF
--- a/retrieval-market.md
+++ b/retrieval-market.md
@@ -24,7 +24,10 @@ The v0 `retrieval market` will initially be implemented as two `libp2p` services
 ```sh
 type RetDealProposal struct {
 	## Reference to the data being retrieved.
-	ref Link
+	ref Cid
+
+  # mode specific parameters saying which data to retrieve
+  params RetParams
 
 	## The total amount that the client is willing to pay for the retrieval of the data.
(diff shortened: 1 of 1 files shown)


retrieval-market.md
```diff
@@ -63,14 +77,16 @@ type CidPrefix struct {
...
+
 `Retrieval miners` should also support a query service that allows clients to request pricing information from a miner.
 
 The query should include the CID of the piece that the client is interested in retrieving. The response contains whether or not the miner will serve that data, the price they will accept for it.
 
 ```sh
 type RetQuery struct {
     ## TODO: what exactly does this link to?
-	piece Link
+	piece Cid
```
magik6k: This should also allow specifying `RetParams`. (In case of unixfs we should also define that if `size==0`, then the clients wants the entire thing)

sternhenri:
Thanks so much for this! I expect this will end up being useful to retrieval market impl, but underlying files have changed too much since. Closing accordingly.

PR #453, opened 2019-08-22
Patch/ticket in nulls

sternhenri:
Three big changes:
- Diagrams added
- Precision of what to do if multiple blocks in a TipSet have losing tickets
- spec'd ticket comparison steps.

### DIFF
--- a/expected-consensus.md
+++ b/expected-consensus.md
@@ -102,7 +102,35 @@ In cases in which no winning ticket is found by any miner in a given round (i.e.
 
 In order to pressure the network to converge on a single chain, each miner may only submit one block per round (see: [`Slashing`](#slashing)).
 
-TODO: pictures of ticket chain and block chain
+```
+                     ┌────────────────────────────────────────────────────────────────────────────────────┐                                                
+                      │                                                                                    │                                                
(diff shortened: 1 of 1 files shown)


expected-consensus.md
```diff
@@ -352,7 +433,7 @@ potentially long chain scans would be required to compute a given block's weight
 
 ### Selecting between TipSets with equal weight
 
-When selecting between TipSets of equal weight, a miner chooses the one with the smallest min ticket (by bytewise comparison).
+When selecting between TipSets of equal weight, a miner chooses the one with the smallest min ticket.
```
ZenGround0: My reading of the above discussion is that tickets are comparable against tickets but I don't know how to compare ticket arrays with each other.  Comparing ticket arrays is the thing we really want for 1) tie breaking, 2) consistent message application order.  It is the thing we should include in the spec.

Right now I'm imagining we just take the first ticket of the array and compare.

ZenGround0: Reading above I'm actually guessing its the final ticket, not the first.  But let's make sure to make that clear.

sternhenri: Understood. Yes! Thanks

sternhenri:
outdated

PR #465, opened 2019-08-29
Add ZigZag Commitments and support for generating spec Markdown from Org mode.

porcuquine:
This PR supersedes #463. Because this uses a forked branch, we can see the result of publication on the [fork's GitHub Pages site](https://porcuquine.github.io/specs/docs/zigzag-commitments/).

~Until actually rebuilding the spec, we won't be able to see the results of this presentation online.~

See the result of publishing at the .

You can see that if you run locally.

Content and presentation (aside from theme and page layout) should be in sync with this draft: https://filecoin-project.github.io/cryptolab/replication/zigzag-commitments/zigzag-commitments

This is not intended to be complete, but it has undergone review from @irengia and @lucaniz. I am creating this as a draft pending only final approval from them.

Please review for inclusion and presentation/infrastructure now.


pooja:
[Add ZigZag Commitments and support for generating spec Markdown from Org mode.](https://trello.com/c/WQMmD5n0/17-add-zigzag-commitments-and-support-for-generating-spec-markdown-from-org-mode)

jbenet:
hey @porcuquine -- what commands do you use to generate the `.md` output from `.org` ? and what orient files are needed? (modeling an example of the new tooling)

porcuquine:
It uses `ox-hugo` mode. I think the command is `org-hugo-export-to-md`. I usually invoke it interactively, but you should be able to run in batch mode as well.

Because there is code to evaluate, you will need to either configure org not to execute babel code blocks on export — or else ensure that the connection to Lisp has been established before you export.

PR #500, opened 2019-09-12
incomplete: storageCollateral is actually pledgeCollateral

whyrusleeping:
Was implementing `SlashConsensusFault` and it was calling this method, but actually meaning for it to be PledgeCollateral. 

The references to this need to be updated, and there is one place where we actually care about storage collateral, but the way to calculate storage collateral should be done on the miner (since miners set their own storage collateral prices)

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -217,7 +217,7 @@ type StorageMarketActorState struct {
 | `GetTotalStorage` | 5 |
 | `PowerLookup` | 6 |
 | `IsMiner` | 7 |
-| `StorageCollateralForSize` | 8 |
+| `PledgeCollateralForSize` | 8 |
 
 #### `Constructor`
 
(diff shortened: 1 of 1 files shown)


anorth:
I agree this should be about pledge collateral.

I think this method could also just be a free function of `nsectors` * `sectorsize` and chain height (and similar for a future storage collateral one). It's arbitrarily on the storage market actor without relying on any of its state, but involving the overhead of message passing. But if we want to use actors just as an organising namespace, I guess so. Maybe we'd want some notion of functions vs methods, e.g. so that if we switch to async message passing we can still get return values?

mishmosh:
@zixuanzh can you review & recommend to incorporate or close?

zixuanzh:
Closing because the current spec has distinguished between different collateral slashing for different faults. Please re-open if any new issues are found.

PR #514, opened 2019-09-13
schemas: remove usage of 'UInt' in IPLD schemas

rvagg:
_(one of a few PRs to attempt to get the IPLD Schemas conforming to current spec)_

Specific typing restrictions are specified as adjuncts to schemas, for now as as descriptive documentation text but we will also soon have a method to also supply such information for codegen purposes, probably language-specific. IPLD only has an `Int` to offer via plain schemas.

I think I've added documentation where it's needed, including the note at the bottom of data-structures.md and the note about `SectorID` needing to allow up to uint64 range. More could be added if anyone thinks it's needed. Keep in mind that there's also going to need to be some codegen hinting metadata provided as adjunct (in some form) so we can revisit these when we get to that (soon) and be more specific so these schemas produce proper Go types (and hopefully other languages too).

Ref: https://github.com/ipld/specs/blob/master/design/history/exploration-reports/2019.07-int-ranges-as-adjuncts.md
Ref: https://github.com/ipld/specs/blob/data-model-motivation/data-model-layer/data-model.md#motivation

### DIFF
--- a/actors.md
+++ b/actors.md
@@ -32,7 +32,7 @@ The init actor is responsible for creating new actors on the filecoin network. T
 ```sh
 type InitActorState struct {
     addressMap {Address:ID}<Hamt>
-    nextId UInt
+    nextId Int
 }
 ```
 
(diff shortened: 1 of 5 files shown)


whyrusleeping:
*cries in wasted space*

PR #517, opened 2019-09-13
schema: misc IPLD schema fixups and adjustments

rvagg:
This is a WIP. There's still a few things that prevent a full parse, notes below about that which we'll need to figure out. This pulls in #514 #515 #516 to get us closer to passing schema spec but this could be detached from them (with some minor rebase pain) if needed.

So, review the last commit in the list only, see those other PRs for those commits.

I've also taken the liberty of fixing some tab/space confusion in both the schema sections and the Go code blocks I found along the way. This isn't essential, IPLD Schemas are agnostic to this, but I think _mostly_ tabs are being used here and I imagine they're more comfortable for Go folks. 

The outstanding items to get a successful parse in all of these are:

1. Unions with integer representations, including `DealState` and `Options`, `Status` in network-protocols.md, are not supported
2. Type aliases, including `AcceptedParams`, `FailedParams`, `StagedParams` in network-protocols.md and `ErrorResponse` in retrieval-market.md are not supported

Also I have some changes to get pushed to the schema parsers to get some other things working like the advanced layouts from #515 but that'll happen shortly I hope.


payments.md
```diff
@@ -71,8 +71,8 @@ type ReconcileResponse struct {
 
 ## TODO: what are the possible status cases?
 type Status enum {
```
rvagg: unsure what the purpose of this is, by default enums are just strings but we are discussing getting an int enum in but it'll need ints next to each item

retrieval-market.md
```diff
@@ -69,13 +69,13 @@ The query should include the CID of the piece that the client is interested in r
 
 ```sh
 type RetQuery struct {
-    ## TODO: what exactly does this link to?
+	## TODO: what exactly does this link to?
 	piece Link
 }
 
 type RetQueryResponse union {
```
rvagg: unsure what the intent here was, keyed unions need a key, but perhaps there was something else going on? my keys might be too verbose or there might be an alternative approach appropriate here

PR #521, opened 2019-09-18
build: fix build on linux

Stebalien:
Modern Linux systems default to python3.


Stebalien:
Also fixes `make serve`.

mishmosh:
thanks for this @Stebalien and sorry for the loooong delay. @sternhenri is going to check if this still needed & valid given the many changes to build process since Sept.

PR #529, opened 2019-09-26
Very rough parent state variable refactor

whyrusleeping:
This is a change thats coming down the pipe from the spec team, sketching it out within the current spec to give others and idea of whats happening.


anorth:
This has now been captured in the new spec.

PR #530, opened 2019-09-26
Simplify logic of mining rewards

Kubuxu:
Below is an updated description.

----

It is the same model (exponential decay) but without `AdustmentPeriod` and with much simpler (and convergent) math.

Exponential distribution is defined as `dN/dt = -λN` but that formula is correct for continuos process.
In the case of a discrete process (our case), we have to work from the solution of the differential equation:
```
N(t) = N(0)*e^(-λ*t)
IV = N(0) - N(1)
IV = N(0) - N₀*e^(-λ)
IV = N(0) * (1-e^(-λ))
(1-e^(-λ)) = IV/N(0)  (eq. 1)
```
Then the reward is:
```
R(t) = N(t-1) - N(t)
R(t) = N(0) * e^(-λ*(t-1)) - N(0) * e^(-λ*(t))
R(t) = N(0) * e^(-λ*(t-1)) * (1 - e^(-λ)) // substitute eq. 1
R(t) = N(t-1) * IV / N(0)
N(t) = N(t-1) - R(t)
```
Where `N(t)` is NetworkActorBalance.

With the set IV, the schedule is within 3.94311e-12 FIL at 6 years.
The source of an observed error is limited precision of FIL unit (attoFIL) but is small enough that it should not matter.

### DIFF
--- a/src/systems/filecoin_blockchain/struct/block_producer/_index.md
+++ b/src/systems/filecoin_blockchain/struct/block_producer/_index.md
@@ -127,29 +127,26 @@ An eligible miner broadcasts the completed block to the network (via [block prop
 
 # Block Rewards
 
-Over the entire lifetime of the protocol, 1,400,000,000 FIL (`TotalIssuance`) will be given out to miners. The rate at which the funds are given out is set to halve every six years, smoothly (not a fixed jump like in Bitcoin). These funds are initially held by the network account actor, and are transferred to miners in blocks that they mine. The reward amount remains fixed for a period of 1 week (given our 30 second block time, this  is 20,160 blocks, the `AdjustmentPeriod`) and is then adjusted. Over time, the reward will eventually become zero as the fractional amount given out at each step shrinks the network account's balance to 0.
+Over the entire lifetime of the protocol, 1,400,000,000 FIL (`TotalIssuance`) will be given out to miners. The rate at which the funds are given out is set to halve every six years, smoothly (not a fixed jump like in Bitcoin). These funds are initially held by the network account actor, and are transferred to miners in blocks that they mine. Over time, the reward will eventually become close zero as the fractional amount given out at each step shrinks the network account's balance to 0.
 
 The equation for the current block reward is of the form:
 
 ```
-Reward = IV * (Decay ^ (BlockHeight / 20160))
+Reward = (IV * RemainingInNetworkActor) / TotalIssuance
 ```
 
-`IV` is the initial value, and is computed by taking:
+`IV` is the initial value, and is set to:
 
 ```
-IV = TotalIssuance * (1 - Decay)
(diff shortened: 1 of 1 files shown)


Kubuxu:
It is based on exponential decay equations https://en.wikipedia.org/wiki/Exponential_decay .

Where our parameters are half-life (derives lambda, in the spec), and `TotalIssuance` (N(0)). 
`IV` is `-dN/dt` at time 0.

so:
```
dN/dt = -λN
IV = λ*N(0)
λ = IV/N(0)
dN/dt = -IV*N/N(0)
```
Where `dN/dt` is how much we send out in each block. In this way we iteratively aproximate (very very closly) exponential decay.
If there is something else to explain, ask.

zixuanzh:
@Kubuxu this makes perfect sense, thanks for the clarification and good work. One quick note, we previously had the notion of `AdjustmentPeriod` and this PR is removing it. There are economic reasons why we may or may not want to keep the `AdjustmentPeriod`. Should we account for it by setting `λ = ln(2) / (HalvingPeriodBlocks / AdjustmentPeriod)`? `Reward` is only updated every `AdjustmentPeriod` and pro-rated for every block.

Kubuxu:
I've discussed it shortly with @whyrusleeping.

First off implementing the incremental model with `AdjustmentPeriod` is much harder than without it (as then you cannot base the reward on the current value of NetworkActor, you would have to store the state of what should be the reward over an AdjustmentPeriod).

In my opinion, unless we really need the AdjustmentPeriod, I would err on the side of simplicity.

Kubuxu:
I've derived a new better parameter that gets closer to 50% distributed after 6 years (with this one we would distribute 26FIL to much after 6 years). Updating the spec.

Kubuxu:
Copying commit description:

Exponential distribution is defined as `dN/dt = -λN` but that formula is correct for continuos process.
In the case of a discrete process (our case), we have to work from the solution of the differential equation:
```
N(t) = N₀*e^(-λ*t)
IV = N(0) - N(1)
IV = N₀ - N₀*e^(-λ)
IV = N₀ * (1-e^(-λ))
(1-e^(-λ)) = IV/N₀  (eq. 1)
```
Then the reward is:
```
R = C(0) - C(1)
R = C₀ - C₀*e^(-λ)
R = C₀ * (1-e^(-λ)) // substitute eq. 1
R = (C₀ * IV) / N₀
```
Where `C` is NetworkActorBalance.

Above means that the `Reward` equation is the same, but the IV equation is different.
The previous IV was causing the reward schedule to be 26FIL out at the six years mark (halving point).
With the new IV, the schedule is within 3.94311e-12 at the same point in time (as implemented in a Filecoin client).
The source of an observed error is limited precision of FIL unit (attoFIL) but is small enough that it should not matter.

Kubuxu:
Thanks for that, it is much cleaner. :+1:

PR #533, opened 2019-10-03
Implementation friendly weight calculations

Kubuxu:
Changes:
  - split `P` into a fraction
  - `Weight` is just an integer
  - Multiply the increase of weight by 1000 to have more significant digits.
  - Change TotalPower inclusion to be `floor(log2(TotalPower))` which corresponds to the binary length of the given value.

### DIFF
--- a/expected-consensus.md
+++ b/expected-consensus.md
@@ -327,22 +327,20 @@ Due to network propagation delay, it is possible for a miner in round N+1 to omi
 It is possible for forks to emerge naturally in Expected Consensus. EC relies on weighted chains in order to quickly converge on 'one true chain', with every block adding to the chain's weight. This means the heaviest chain should reflect the most amount of work performed, or in Filecoin's case, the most storage provided.
 
 The weight at each block is equal to its `ParentWeight` plus that block's delta weight. Delta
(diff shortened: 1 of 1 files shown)


expected-consensus.md
```diff
@@ -327,22 +327,20 @@ Due to network propagation delay, it is possible for a miner in round N+1 to omi
...
+The weight should be calculated using big integer arithmetic with order of operations defined above.
 
-```sh
-Note that if your implementation does not allow for rounding to the fourth decimal, miners should apply the [tie-breaker below](#selecting-between-tipsets-with-equal-weight). Weight changes will be on the order of single digit numbers on expectation, so this should not have an outsized impact on chain consensus across implementations.
-```
+The exact value for these parameters remain to be determined, but for illustration, we can set:
 
- The exact value for these parameters remain to be determined, but for illustration, we can set:
+- `V = 2`
+- `X = floor(log2(TotalPower in Power Table))` - length of TotalPower in bits
```
ZenGround0: Reiterating we want this for weight to be readily computable, at least in go.  Unclear if there is a downside to the lack of precision but should be evaluated.

Kubuxu: It would be as easy in other languages. It is much easier to compute than the log of an arbitrary precision integer.
Worst case scenario the algorithm is repeated division by 2 until you have 0 left.

Another reason is the removal of floats from these calculations.  Floats are hard to do consistently.
If we decide that it lacks precision, we should define an algorithm for computing that value using arbitrary precision integers.

sternhenri:
Thanks for your notes @Kubuxu. Closing in favor of https://github.com/filecoin-project/specs/pull/538. Would love your notes there please.

PR #538, opened 2019-10-07
updated ec spc and weight fn

sternhenri:
Updated ec with pseudocode to be moved into Go.

Three open questions:
- (@jbenet) happy with weighting function as is (spec. forkFactor) or tune ahead of testnet?
- (@Kubuxu) notes on making it easier to implement (sorry)
- (@whyrusleeping) (@jbenet) notes on new ticket rule to ensure |tickets(tipset)| == |blocks(tipset)|? Decision time to keep or drop...

Idea is to ensure 1 ticket per block always, with downside being ability for miners to grind through nonces in parallel. Though they would have been able to do that in old way through a batch of k at a time (with k the randomness lookback). I claim the difference is negligible without a working VDF.


src/algorithms/expected_consensus.md
```diff
@@ -292,10 +339,16 @@ w[r+1] = w[r] + floor(1000 * (wForkFactor(wPowerFactor[r+1] + wBlocksFactor[r+1]
```
sternhenri: @Kubuxu, section of interest for you is below.

Kubuxu: I would be for defining function: `log2b(x) = floor(log2(x))` and explaining that it is a binary length of `x`.

src/algorithms/expected_consensus.md
```diff
@@ -292,10 +339,16 @@ w[r+1] = w[r] + floor(1000 * (wForkFactor(wPowerFactor[r+1] + wBlocksFactor[r+1]
 
 with, for a given tipset ts in round r+1:
 
-- wBlocksFactor[r+1] = v * |blocksInTipset(ts)|
+- wBlocksFactor[r+1] = v * |blocksInTipset(ts)| = v * k
   - with v = vt * log2(totalPowerAtTipset(ts) * vs)/vs
```
Kubuxu:
```
v = vt * log2(totalPowerAtTipset(ts) * vs) / vs
v = vt * (log2(totalPowerAtTipset(ts) + log2(vs)) / vs
v = vt * log2(totalPowerAtTipset(ts) / vs + vt * log2(vs) / vs
v = vt * log2(totalPowerAtTipset(ts) / vs + sigma; where sigma = vt * log2(vs) / vs
```

Now what I don't understand is:
given parameters below:
```
vt = 350
vs = 1e-5
sigma = -581337416
```

Unless `log2` is already used as the proposed by me `log2b(x)`.

src/algorithms/expected_consensus.md
Kubuxu: In general, any division should be left for last (as it is the lossy operation in integer arithmetic). The moment division is performed should be defiend using the paranthesis so there is no ambiuity.

 The multiplication by 1000 should be done in every term before the division is performed to achieve greater precision. (another option could be multiplication by 2^16 as it maps well to bigint arithmetic).

src/algorithms/expected_consensus.md
Kubuxu: Proposed change:
redefine `vs`:  as `vS = 1/vs` - fractional `vs` hides divisions then:
`v = vt * vS * log2b(totalPowerAtTipset(ts)/vS)`

src/algorithms/expected_consensus.md
```diff
@@ -292,10 +339,16 @@ w[r+1] = w[r] + floor(1000 * (wForkFactor(wPowerFactor[r+1] + wBlocksFactor[r+1]
...
 - wPowerFactor[r+1]  = log2(totalPowerAtTipset(ts))
```
Kubuxu: Then
`wPowerFactor[r+1]  = log2b(totalPowerAtTipset(ts))`

src/algorithms/expected_consensus.md
```diff
@@ -292,10 +339,16 @@ w[r+1] = w[r] + floor(1000 * (wForkFactor(wPowerFactor[r+1] + wBlocksFactor[r+1]
...
-- wForkFactor[r+1]   = CDF(X, k) with X -> Bin(eNumberOfBlocksPerRound * numberOfMinersInPowerTable, 1/numberOfMinersInPowerTable)
+Take X -> Bin(eNumberOfBlocksPerRound * numberOfMinersInPowerTable, 1/numberOfMinersInPowerTable), for simplicity, we take:
+X -> Bin(eNumberOfBlocksPerRound * 10,000, 1/10,000). We have:
+- wForkFactor[r+1]   = 1 if |blocksInTipset(ts)| > E[X] - 2*stdDev[X], CDF(X, |blocksInTipset(ts)) otherwise.
```
Kubuxu: Define `E[X]` as a const for the network.

----

If I understand correctly `wForkFactor` is a fractional multiplier (`wForkFactor <= 1`) for the other factors (notation of `w[r+1]` is not clear in this regard.
Then the value should be:
1. Split into numerator and denominator.
2. Well defined for values of `|blocksInTipset(ts)|`.

Given its current spec, it is impossible for two different people to arrive at the same values.

src/algorithms/expected_consensus.md
Kubuxu: Read other comments first

---

Given everything else I would define `w[r+1]` as:
`w[r+1] = w[r] + (wForkFactor_num * (wPowerFactor[r+1] + wBlocksFactor[r+1) * 2^16) / wForkFactor_den`

`2^16` can be replaced by `2^8` depending on precision required.

Kubuxu: I would also move the note about integer arithmetic from the bottom to the top.

anorth:
FYI @ZenGround0

whyrusleeping:
@sternhenri how do i sample randomness from a null block round?

src/algorithms/expected_consensus.md
```diff
@@ -159,46 +125,48 @@ We use the ECVRF algorithm from [Goldberg et al. Section 5](https://tools.ietf.o
...
+    i. newTicket <-- New()
+1. Draw prior ticket
+    i. priorTicket <-- chain.TicketAtEpoch(N-1)
+2. Run it through VRF and get determinstic output
+    i. # take the VRFResult of that ticket as input, specifying the personalization (see data-structures)
+            input <-- VRFPersonalization.Ticket | priorTicket.Output
+    ii. # run it through the VRF and store the output in the new ticket
+            newTicket.vrfResult <-- self.VRFKeyPair.Generate(input)
+4. Return the new ticket
+```
```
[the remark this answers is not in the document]
jbenet: Something like this? (still not sure it's right)
```go
func prepareNewTicket(chain, epoch, minerElectionKey) {
  // 0. prepare new ticket
  newTicket := tickets.NewTicket()

  // 1. draw prior ticket
  priorTicket := TicketAtEpoch(chain, epoch - 1)
  
  // 2. run it through the VRF and get deterministic output
  // 2.i. take the VRFResult of that ticket as input, specifying the personalization (see data structures)
  input := VRFPersonalization.Ticket | priorTicket.Output // TODO: what is this operation?
  // 2.ii. run through VRF
  newTicket.VRFResult := vrf.KeyedVRF(minerElectionKey, input)

  return newTicket
}
```

src/algorithms/expected_consensus.md
```diff
@@ -159,46 +125,48 @@ We use the ECVRF algorithm from [Goldberg et al. Section 5](https://tools.ietf.o
...
 
 Succinctly, the process of verifying a block's tickets is as follows.
 ```text
 Input: received block, storage market actor S, miner's public key PK, a public VDF validation key vk
 Output: 0, 1
 
-0. Get the tickets
-    i. tickets <-- block.tickets
-For each ticket, idx: tickets
+0. Get the ticket
```
jbenet: code/pseudocode here would be good too

src/algorithms/expected_consensus.md
```diff
@@ -220,52 +188,96 @@ If the miner scratches a winning ticket in this round, it can use newEP, along w
...
+        normalized_power <-- p_f * 2^HashLen
+    ii. # Compare the miner's power fraction to the value of the ticket drawn: more power means win more often.
+        # winning ticket
+        if readLittleEndian(VRFOutput) <= normalized_power
+            return newEP
+        # otherwise parentTicket is not a winning lottery ticket
+        else 
+            Return 0
+Return 1
+```
```
jbenet: I can't verify this. @nicola, @whyrusleeping ?

src/algorithms/expected_consensus.md
```diff
@@ -220,52 +188,96 @@ If the miner scratches a winning ticket in this round, it can use newEP, along w
...
 
-For election proof generation, see [checking election results](expected-consensus.md#checking-election-results).
+If successful, the miner can craft a block, passing it to the block producer. If unsuccessful, it will wait to hear of another block mined this round to try again. In the case no other block was found in this round the miner can increment nonce and try again.
+While a miner could try to run through multiple nonces in parallel in order to quickly generate a block, this effort will be futile as the honest majority of miners will reject blocks crafted with ElectionProofs whose nonces prove too high (see below).
```
jbenet: there is no "honest majority of miners", the model is rational majority

src/algorithms/expected_consensus.md
```diff
@@ -220,52 +188,96 @@ If the miner scratches a winning ticket in this round, it can use newEP, along w
...
         ii. # Check that the election proof was correctly generated by the miner
             # using the appropriate ticket
-            VRFState <-- ECVRF_Verify(miner.PK, electionProof.VRFProof, input)
-            if VRFState == "INVALID":
+            VRFState <-- electionProof.Verify(input, miner.PK)
+            if VRFState == false:
+                return 0
+        iii. Ensure that the scratched ticket is a winner
+            # internally -> map p_f onto [0, 2^HashLen] and compare the miner's scratchValue to the miner's normalized power fraction
+            if electionProof.IsWinning(p_f) == false:
```
jbenet: cc @nicola, @whyrusleeping  pls verify

src/algorithms/expected_consensus.md
```diff
@@ -329,15 +348,15 @@ The probability that two Tipsets with different blocks would have all the same t
 
 {{<label finality>}}
 ## Finality in EC
-TODO
+EC enforces a version of soft finality whereby all miners at round N will reject all blocks coming in prior to round N-F. For illustrative purposes, we can take F to be 500. While strictly speaking EC is a probabilistically final protocol, choosing such an F simplifies miner implementations and enforces a socially-bound finality at no cost to liveness in the chain.
```
jbenet: - [ ] "... will reject all blocks that fork off prior to round N-F ...". ("coming in prior" is ill defined)
- [ ] s/"socially bound"/"macroeconomic consensus" or something like that?

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,12 +44,6 @@ Conversely, storage faults only lead to power loss once they are detected (up to
...
-
-Miner M1 has a provingPeriod of 30. M1 submits a PoST at height 39. Their next `provingPeriodEnd` will be 69, but M1 can submit a new PoST at any height X, for X in (39, 69]. Let's assume X is 67.
-
-At height Y in (39, 67], M1 will attempt to generate an `ElectionProof` using the storage market actor from height 39 for their own power (and an actor from Y for total network power); at height 68, M1 will use the storage market actor from height 67 for their own power, and the storage market actor from height 68 for total power and so on.
-
 ## Repeated leader election attempts
 
-In the case that no miner is eligible to produce a block in a given round of EC, the storage power consensus subsystem will be called by the block producer to attempt another leader election.
\ No newline at end of file
+In the case that no miner is eligible to produce a block in a given round of EC, the storage power consensus subsystem will be called by the block producer to attempt another leader election by incrementing the nonce appended to the ticket drawn from the past in order to attempt to craft a new valid `ElectionProof`
```
jbenet: This is not super clear. make the function explicit

jbenet:
I'm not able to review the weighing function in the next few days -- i'm not paged in enough. Can somebody else? (@nicola @nikkolasg @jzimmerman)

nicola:
Hey team, I am blocked on this being merged since I am doing changes which will require this part to compile and in order to do so, I need to either re-do some things in this PR or wait.

src/systems/filecoin_blockchain/blockchain/block.id
```diff
@@ -45,6 +46,8 @@ type Chain struct {
     HeadEpoch()                 Epoch
     FinalizedEpoch()            Epoch
     LatestCheckpoint()          Epoch
+    AddTipset(ts Tipset)        
 
+    TicketAtEpoch(epoch Epoch)  Ticket
```
nicola: this is also defined in Storage Power Consensus Subsystem https://github.com/filecoin-project/specs/blob/7e4fde83c7699f8ec33a4c8735768e4bc0bbc6fd/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.id#L27

We should remove one or the other.

src/algorithms/expected_consensus.md
sternhenri: This is great feedback. Thanks @kubuxu for the review. Super useful.

src/algorithms/expected_consensus.md
```diff
@@ -280,35 +295,41 @@ Just as there can be 0 miners win in a round, multiple miners can be elected in
...
-- A wBlocksFactor: which adds weight to the chain proportional to the number of blocks mined in a given round. Like wForkFactor, it rewards miner cooperation (which will yield more blocks per round on expectation).
+- wForkFactor: which seeks to cut the weight derived from rounds in which produced Tipsets do not correspond to what an honest chain is likely to have yielded (pointing to selfish mining or other non-collaborative miner behavior).
+- wPowerFactor: which adds weight to the chain proportional to the total power backing the chain, i.e. accounted for in the chain's power table. 
+- wBlocksFactor: which adds weight to the chain proportional to the number of blocks mined in a given round. Like wForkFactor, it rewards miner cooperation (which will yield more blocks per round on expectation).
 
-We have:
+The weight should be calculated using big integer arithmetic with order of operations defined above. We use brackets instead of parentheses below for legibility. We have:
 
-w[r+1] = w[r] + floor(1000 * (wForkFactor(wPowerFactor[r+1] + wBlocksFactor[r+1])))
+`w[r+1] = w[r] + wPowerFactor[r+1] + wBlocksFactor[r+1]`
```
Kubuxu:
```suggestion
`w[r+1] = w[r] + (wPowerFactor[r+1] + wBlocksFactor[r+1]) * 2^8`
```

As in L322, the result is multiplied by `2^8`, the top-level definition should probably be the same.

jbenet:
This needs to be rebased over master. there's a bunch of commits here that are already in master.

src/systems/filecoin_blockchain/blockchain/block.id
sternhenri: still working on that one but yes!

sternhenri:
Loads of progress. Still WIP. Left TBD:
- [ ] Review Block validation
- [ ] Consensus faults
- [ ] Power table 
- [ ] Fix randomness retrieval

jbenet:
I think you need to rebase on top of master, i think @nicola made a bunch of changes to get this area to compile

src/algorithms/expected_consensus.md
```diff
@@ -280,35 +133,41 @@ Just as there can be 0 miners win in a round, multiple miners can be elected in
...
+- wBlocksFactor[r+1] =  `wPowerFactor[r+1] * wRatio * b / e`
+  - with `b = |blocksInTipset(ts)|`
+  - `e = expected number of blocks per round in the protocol`
+  - and `wRatio in ]0, 1[`
+Thus, for stability of weight across implementations, we take: 
+- wBlocksFactor[r+1] =  `(wPowerFactor[r+1] * b * wRatio_num) / (e * wRatio_den)`
 
-The weight should be calculated using big integer arithmetic with order of operations defined above. The multiplication by 1,000 and flooring is meant to help generate uniform weights across implementations.
+We get: 
+- `w[r+1] = w[r] + log2b(totalPowerAtTipset(ts)) * 2^8 + (log2b(totalPowerAtTipset(ts)) * len(ts.blocks) * wRatio_num * 2^8) / (e * wRatio_den)`
```
magik6k: `log2b(totalPowerAtTipset(ts))` will essentially become a constant at higher network sizes (it may not catch a 2x drop in power), and it's quite painful to compute.

IMO it should be either dropped, or made more 'aggressive' (something between log and linear)

Kubuxu: @magik6k it is : `log2b(x big.Int) = x.BitLen() - 1`
AFAIK, it is there to prevent case where minority miner forks off and slashes other miners.

magik6k: You can still slash ~50% of power in the worst case without anyone noticing now

sternhenri: Hi @magik6k, this is super useful feedback.

Remember that this is added to weight at every round, and the other factor is going to be smaller (by wRatio_num/wRatio_den), so while it is indeed a constant, a weight drop by 20% over the space of finality (say ~500 blocks) will be noticeable.

On the flip-side, a power-rental attack will have to be maintained for a bit of time before it becomes irrecoverable. Say I can rent 80% of the network for a few hours, can an honest majority collude to ignore my fork and get the heaviest back in some reasonable amount of time? This is what the more 'aggressive' functions lack...

Thereafter, @Kubuxu is right in terms of what this seeks to prevent as a main reason.

What do you think?

In any case, I need to make clear that log2b is a parameter choice as is wRatio so will update to reflect that. Log2 is simply fave candidate rn.

sternhenri:
Done with main content switches @jbenet, now just need to get it compiling before it can be merged. This branch includes:
- Storage power actor changes
- all power table accounting
- all collateral explanations
- consensus faults explanations
- randomness sampling throughout the chain
- specific code for ticket and election proof generation and validation
- new ticket format for @whyrusleeping 
- weight function specc'd
and more.

Does not include:
- storage faults flow
- completed block validation flow

sternhenri:
@porcuquine @jbenet any thoughts on test failure here. Work compiles fine for me. Thought I'd ask before diving in in case it's obvious to either of you!

porcuquine:
@sternhenri It looks like this needs `make fmt-code` changes to be checked in. This would be clearer if we factor the formatting check into a separate CI job. I did the quickest, dirtiest thing for v0.

jbenet:
That's great to hear @sternhenri will review today

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -30,26 +30,114 @@ We must distinguish between both types of "miners" (storage and block miners). {
 
 However, given Filecoin's "useful Proof-of-Work" is achieved through file storage (PoRep and PoSt), there is little overhead cost for storage miners to participate in leader election. Such a {{<sref storage_miner_actor>}} need only register with the {{<sref storage_power_actor>}} in order to participate in Expected Consensus and mine blocks.
 
-{{<label power_table>}}
-## The Power Table
+## Repeated leader election attempts
 
-The portion of blocks a given miner generates through leader election in EC (and so the block rewards they earn) is proportional to their `Power Fraction` over time. That is, a miner whose storage represents 1% of total storage on the network should mine 1% of blocks on expectation.
+In the case that no miner is eligible to produce a block in a given round of EC, the storage power consensus subsystem will be called by the block producer to attempt another leader election by incrementing the nonce appended to the ticket drawn from the past in order to attempt to craft a new valid `ElectionProof` and trying again.
```
jbenet: add note about clock synchrony? (ie why cant i just do all them at once until i win? because people wont accept your blocks early)

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -30,26 +30,114 @@ We must distinguish between both types of "miners" (storage and block miners). {
...
-To illustrate this, an example:
+This ticket chain will track one-to-one with a block at each height in a given subchain, but omits certain details including other blocks mined at that height.
 
-Miner M1 has a provingPeriod of 30. M1 submits a PoST at height 39. Their next `provingPeriodEnd` will be 69, but M1 can submit a new PoST at any height X, for X in (39, 69]. Let's assume X is 67.
+It is composed inductively as follows. For a given chain:
 
-At height Y in (39, 67], M1 will attempt to generate an `ElectionProof` using the storage market actor from height 39 for their own power (and an actor from Y for total network power); at height 68, M1 will use the storage market actor from height 67 for their own power, and the storage market actor from height 68 for total power and so on.
+- At height 0, take the genesis block, return its ticket
+- At height n+1, take the heaviest tipset in our chain at height n.
+    - select the block in that tipset with the smallest final ticket, return its ticket
```
jbenet: what does "the heaviest tipset in our chain at height n" mean? either:
- (a) it means "oh there's exactly one tipset in our chain at heigh n, because "our chain" just means the links walking back `block.Parents`
- (b) or it means "oh there's many tipsets in our chat at height n, because "out chain" means all possible chains we have, with all possible tipsets, we have to pick the heaviest at each epoch.

if (a), just remove the "take the heaviest tipset in our chain at height n", there's only one.
if (b), then i think this is ambiguous and maybe wrong because - what happens if i have tipsetA and tipsetB each the heaviest tipset in epochA and epochB respectively, but NOT part of the same history (dont link to each other).

jbenet: - for any tipset:
  - select the block in that tipset with the smallest final ticket, return its ticket.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -30,26 +30,114 @@ We must distinguish between both types of "miners" (storage and block miners). {
...
+ - `T`-- estimated time for SEAL, dependent on sector size
+ - `G = T + variance`-- necessary flexibility to account for network delay and SEAL-time variance. 
+
+We expect Filecoin will be able to produce estimates for sector commitment time based on sector sizes, e.g.:
+`(estimate, variance) <--- SEALTime(sectors)`
+G and T will be selected using these.
+
+#### Picking a Ticket to Seal
+
+When starting to prepare a SEAL in round X, the miner should draw a ticket from X-F with which to compute the SEAL.
```
jbenet: why is this here? shouldn't we just say tickets are how we produce randomness in the chain at a given epoch, and then let other things (ie seal) say how they use that randomness?

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -30,26 +30,114 @@ We must distinguish between both types of "miners" (storage and block miners). {
...
+
+#### Verifying a Seal's ticket
+
+When verifying a SEAL in round Z, a verifier should ensure that the ticket used to generate the SEAL is found in the range of rounds [Z-T-F-G, Z-T-F+G].
```
jbenet: this should be defined in Sealing, not here in SPC? unless you want it here because it's relevant to verifiable resource commitment?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.id
```diff
@@ -28,16 +31,50 @@ type StoragePowerActor struct {
...
+    )
+    miners {addr.Address: StorageMiner}
+    GetMinerPower(addr addr.Address) StoragePower
+    GetTotalPower() StoragePower
+    GetMinerPublicKey(addr addr.Address) filcrypto.PubKey
+    IncrementPower(addr addr.Address, numSectors UVarint)
+    SuspendPower(addr addr.Address, numSectors UVarint)
+    UnsuspendPower(addr addr.Address, numSectors UVarint)
+    RemovePower(addr addr.Address, numSectors UVarint)
+    RemoveAllPower(addr addr.Address, numSectors UVarint)
```
jbenet: i dont think we need all these methods, let's only have the ones we absolutely need (smaller potential problem/audit surface)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.md
```diff
@@ -4,3 +4,40 @@ title: Storage Power Actor
...
+Conversely, storage faults only lead to power loss once they are detected (up to one proving period after the fault) so miners will mine with no more power than they have used to store data over time.
+
+Put another way, power accounting in the SPC is delayed between storage being proven or faulted, and power being updated in the power table (and so for leader election). This ensures fairness over time.
+Finally, we also account for "suspended" power in order to make accounting simple in the presence of faults (see {{<sref consensus_faults>}}) and subsequent recovery.
+
+The Miner lifecycle in the power table should be roughly as follows:
+- MinerRegistration: A new miner with an associated worker public key and address is registered on the power table by the storage mining subsystem, along with their associated sector size (there is only one per worker).
+- UpdatePower: These power increments and decrements (in multiples of the associated sector size) are called by various storage actor (and must thus be verified by every full node on the network). Specifically:
```
jbenet: > in multiples of the associated sector size

why not just raw Bytes at every point? or GB?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.md
```diff
@@ -4,3 +4,40 @@ title: Storage Power Actor
...
+    - Power is incremented to account for a new SectorCommitment at the first PoSt past the first ProvingPeriod.
+    - All Power is decremented immediately after a missed PoSt.
+    - Power is decremented immediately after faults are declared, proportional to the faulty sector size.
+    - Power is incremented after a PoSt recovering from a fault.
+    - Power is definitively removed from the Power Table past the sector failure timeout (see {{<sref storage_faults>}})
```
jbenet: see the sector state cycle that we're working on in #569

jbenet:
This needs a rebase unfort-- let's do it soon to avoid much greater divergence.

@zixuanzh wonder how much this conflicts with #569 -- should we merge this first and rebase #569 or vv.?

jbenet:
@sternhenri I rebased for you on top of master (#569 and more) o/

jbenet:
given this is a huge PR, i'm leaning towards merging it before all the comments are addressed (and avoid further merge conflicts

jbenet:
Ok i'm going to merge @sternhenri please address comments as a follow-on PR. Also reconcile w/ new deals/post work -- some things may have been duped/broken

jbenet:
cc @zixuanzh pls review this, and maybe help reconcile w/ new deals work if anything is off

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.md
zixuanzh: To summarize, only sectors in the Active state will command power. A Sector becomes Active after first PoSt from Committed and Recovering stages. Power is immediately decremented when an Active Sector enters the Failing state (through `DeclareFaults` or `Cron`) and when an Active Sector expires.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.id
```diff
@@ -1,52 +1,61 @@
...
+    IsValidConsensusFault(faults ConsensusFaultType, blocks [block.Block]) bool
 }
 
-type StorageMiner struct {
-    minerAddress  addr.Address
-    minerStorage
-    VRFKeyPair    filcrypto.VRFKeyPair
+type ConsensusFaultType union {
+    DoubleForkMiningFault  ConsensusFault
+    ParentGrindingFault    ConsensusFault
```
zixuanzh: should UncommittedPowerFault live here too?

src/systems/filecoin_blockchain/struct/block/block.go
```diff
@@ -12,50 +11,52 @@ func SmallerBytes(a, b util.Bytes) util.Bytes {
...
+	// parents := current.Parents
+	// for i := range dist {
+	// 	current = parents
+	// 	parent = current.Parents
+	// }
+	// return current
+}
 
-	return true
+func (bl *Block_I) TicketAtEpoch(epoch ChainEpoch) Ticket {
```
whyrusleeping: so does this mean that randomness drawn from a null round is just the randomness of the 'next' block after the series of nulls?

whyrusleeping:
Still unanswered:

> @sternhenri how do i sample randomness from a null block round?

src/algorithms/expected_consensus.md
```diff
@@ -161,115 +86,43 @@ We use the ECVRF algorithm from [Goldberg et al. Section 5](https://tools.ietf.o
...
-### Election Validation
+Succinctly, the process of crafting a new ElectionProof in round N is as follows. We use:
 
-For election proof generation, see [checking election results](expected-consensus.md#checking-election-results).
+    The ECVRF algorithm (must yield a pseudorandom, deterministic output) from Goldberg et al. Section 5, with:
+        Sha256 for our hashing function
+        Secp256k1 for our curve
 
-In order to determine that the mined block was generated by an eligible miner, one must check its `ElectionProof`.
+Note: We draw the miner power from the prior round. This means that if a miner wins a block on their ProvingPeriodEnd even if they have not yet resubmitted a PoSt, they retain their power (until the next round).
```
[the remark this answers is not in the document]
sternhenri: This is talking specifically about the edge case that is: what if I win a block in the round in which i am challenged (and don't submit). Do I still win or is it forfeited. So this is just meant to avoid fencepost errors. I'll clarify

src/algorithms/expected_consensus.md
```diff
@@ -329,18 +188,22 @@ The probability that two Tipsets with different blocks would have all the same t
...
+- (2) `parent grinding fault`: one block's parent is a Tipset that could have validly included the other block according to Tipset validity rules, however the parent of the first block does not include the other block.
 
   - While it cannot be proven that a miner omits known blocks from a Tipset in general (i.e. network latency could simply mean the miner did not receive a particular block) in this case it can be proven because a miner must be aware of a block they mined in a previous round.
 
-Any node that detects this occurring should take both block headers, and call [`storagemarket.SlashConsensusFault`](actors.md#slashconsensusfault). The network will then take all of that node's collateral, give a portion of it to
-the reporter, and keep the rest.
\ No newline at end of file
+Any node that detects either of the above events should submit both block headers to the `StoragePowerActor`'s `ReportConsensusFault` method. The "slasher" will receive a portion (TODO: define how much) of the offending miner's {{<sref pledge_collateral>}} as a reward for notifying the network of the fault.
```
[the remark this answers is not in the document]
sternhenri: I think this supercedes that. 

And yes, I expect the cronActor could call these though in the case of consensus faults it wouldn't be able to (too costly).

src/algorithms/expected_consensus.md
```diff
@@ -329,18 +188,22 @@ The probability that two Tipsets with different blocks would have all the same t
...
+(TODO: FIP of submitting commitments to block headers to prevent miners censoring slashers in order to gain rewards).
+
+It is important to note that there exists a third type of consensus fault directly reported by the `CronActor` on `StorageDeal` failures via the `ReportUncommittedPowerFault` method:
+- (3) `uncommitted power fault` which occurs when a miner fails to submit their `PostProof` and is thus participating in leader election with undue power (see {{<sref storage_faults>}}).
```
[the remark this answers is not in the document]
sternhenri: Agreed. I expect all of this is defined in the storage faults section.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.id
```diff
@@ -17,29 +17,63 @@ type StoragePowerActor struct {
...
+}
 
-    // RegisterStorageMiner(SealProof)				Address
-    // TryUpdateStorage(base.Address, newPower)			error
+type StorageMiner struct {
+    MinerAddress         addr.Address
+    MinerStoragePower    block.StoragePower
+    MinerSuspendedPower  block.StoragePower
```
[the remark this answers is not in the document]
sternhenri: Two things. I think passing address is better design since we would want to use the same methods for both the cronactor and self-reporting. So I chose to leave it there. Happy to change though.

My thought on suspended power is that it is necessary in order to account for collateral vs power. For instance if I declare a fault and my power drops, I should not be able to withdraw collateral. For that reason we need an intermediary state for power between just "present" and "absent", hence "suspended".

It may be worth our making a state machine for faults @zixuanzh

src/systems/filecoin_blockchain/struct/block/block.go
sternhenri: Sorry for lag. No, it means that it is the randomness from the prior "live" round. (null rounds just means nonce increment). I'll specify further

PR #541, opened 2019-10-09
Restore ZigZag Commitments and PoSt Parameters.

porcuquine:
During recent reorgs of the repo, the ZigZag Commitments and Proof of Spacetime Parameters sections were lost. This PR restores them by doing the following:

- Change relative paths used in .org files.
- Change Makefile to use `-R` instead of `-r` when copying org-generated content.
- Reverts to using CDN for MathJax. NOTE: config query parameter passed.
- Manually regenerated .md from .org (necessary for now).

As per discussion with @jbenet, CDN usage is temporary. You should now be able to see correctly rendered math in the ZigZag commitments section (along with graphs). This should suffice for testing alternate installation/configuration methods for MathJax, KaTeX, or whatever.


Makefile
```diff
@@ -245,7 +245,7 @@ watch-code: .PHONY
 
 DOTs=$(shell find src -name '*.dot')
 MMDs=$(shell find src -name '*.mmd')
-SVGs=$(DOTs:%=%.svg) $(MMDs:%=%.svg)
+SVGs=$(DOTs:%=%.svg) #$(MMDs:%=%.svg)
```
porcuquine: This was a hack/workaround because otherwise I get build errors.

src/algorithms/post/_index.md
```diff
@@ -2,7 +2,7 @@
 title: Proof-of-Spacetime
 entries:
 - rational_post
-- proof_of_spacetime_parameters
+- proof_of_spacetime_parameters.md
```
porcuquine: Needed to disambiguate `.md` and `.org` versions of the file.

PR #544, opened 2019-10-10
SDR

porcuquine:
NOTE: There is a lot more work to do here, but what remains will be more structural and begin touching other parts of the spec. I'd like to merge this now as a good synchronization point.

This branch contains updates from ZigZag to SDR.

I have closed #541 in favor of this and folded that work in here.

- [x] Replace ZigZag Commitments with SDR Commitments.
- [x] Replace vanilla ZigZag description with SDR description [at a surface level].
- [x] Replace ZigZag circuit spec with SDR circuit spec [at a cosmetic level].


src/algorithms/porep/porep_commitments/sdr_commitments/2
```diff
@@ -0,0 +1,22 @@
+There is no class named FILECOIN.THEORY::REPLICA-LAYER-GRAPH.
+   [Condition of type SB-PCL:CLASS-NOT-FOUND-ERROR]
```
jbenet: is this output supposed to be here? looks like a bug?

src/algorithms/porep/porep_commitments/sdr_commitments/sdr_commitments.org
```diff
@@ -0,0 +1,328 @@
+#+TITLE: Stacked DRG Commitments
```
jbenet: i won't review the content of this. leaving to you @porcuquine or @nicola

src/algorithms/porep/stacked_drg.md
```diff
@@ -1,10 +1,10 @@
 ---
-title: ZigZag-PoRep
```
jbenet: do we want to keep ZigZag for the future? i guess we can always bring it back.

src/algorithms/porep/stacked_drg_circuit.md
```diff
@@ -0,0 +1,161 @@
+## Stacked DRG: Offline PoRep Circuit Spec
```
jbenet: start these files with `#` -- the spec document should now shift headers correctly when grafting this header tree into the larger document.

jbenet:
@porcuquine couple minor comments. I'll merge this now to avoid further drift, feel free to resolve these comments w/ additional PR, or leave them if low prio

PR #549, opened 2019-10-10
Clarify that File abstracts the filesystem, doesn't require Unix

anorth:


src/systems/filecoin_files/file/file.id
```diff
@@ -1,8 +1,9 @@
-// Path is a unix file path, pointing to a file
+// Path is an opaque locator for a file (e.g. in a unix-style filesystem).
```
jbenet: it is meant to be unix style path. it is not completely opaque. slashes imply hierarchical structure (im ok with this language here, but slashes will be meaningful)

jbenet:
LGTM

PR #552, opened 2019-10-11
Make the VM section compile

jbenet:
This is a large stab at getting the VM runtime in place.

The following packages exist:

- The `vm/address` package is "the address of an actor in the vm"
  - arguably this should be moved into `vm/actor`
- The `vm/actor` package is "the definition of actors and key bits around them"
- The `vm/state_tree` package is "the thing that tracks all vm execution state"
- The `vm/message` package is "the external user invocation of a method"
- The `vm/sysactors` package is "the set of vm system actors (init, account, cron)"
- The `vm/runtime` package is "the access actors have from within"
  - this includes `InvocInput` and `InvocOutput` which are used to apply messages
- The `vm/runtime/exitcode` package is "the list of constant VM exit codes"
- The `vm/runtime/gascost` package is "the list of constant gas costs"
- the `vm/interpreter` package is "the thing that applies the messages"
  - runtime execution moved to functional application (not staging things)

- Much of this needs to be refined and reduced, many obvious improvements. this is a good next iteration, which compiles.
- Much of this needs to be fixed -- there are some erros. (all found marked with `//TODO`)


src/systems/filecoin_blockchain/vm/interpreter/vm_interpreter.go
```diff
@@ -39,131 +20,177 @@ func (vmi *VMInterpreter_I) ApplyMessageBatch(inTree st.StateTree, msgs []Messag
...
-//------------
+func applyError(tree st.StateTree, exitCode msg.ExitCode, gasUsed msg.GasAmount) (outTree st.StateTree, ret msg.MessageReceipt) {
+	return outTree, &msg.MessageReceipt_I{
+		ExitCode_:    exitCode,
+		ReturnValue_: nil,
+		GasUsed_:     gasUsed,
+	}
+}
+
+func gasToFIL(gas msg.GasAmount, price msg.GasPrice) actor.TokenAmount {
```
nicola: why are these function not methods of some struct?

src/systems/filecoin_blockchain/vm/message/message.id
```diff
@@ -14,30 +14,30 @@ type Message union {
...
-    value     actor.TokenAmount
-
-    gasPrice  actor.TokenAmount
-    gasLimit  actor.TokenAmount
+    CallSeqNum  actor.CallSeqNum
+    Value       actor.TokenAmount
 
-    method    actor.MethodNum
-    params    actor.MethodParams  // Serialized parameters to the method.
+    GasPrice
```
nicola: does this have no type?

jbenet: It does -- `GasPrice GasPrice`. the codeGen tool removes redundancy between name and type.

src/systems/filecoin_blockchain/vm/interpreter/vm_interpreter.go
jbenet: they don't need to be. we need some simple helper functions. they do not need to be functions on structs

nicola:
It compiles for me. Someone expert in the vm should review the technical parts.

src/systems/filecoin_blockchain/vm/runtime/runtime.id
```diff
@@ -0,0 +1,85 @@
...
+    Invocation     InvocInput
+    State          VMState
+
+    // Fatal is a fatal error, and halts VM execution.
+    // This is the normal error condition of actor execution.
+    // On Fatal errors, the VM simply does not apply the state transition.
+    // This is atomic across the entire contract execution.
+    //
+    // TODO: @why wants to get rid of this.
+    Fatal(string)
```
anorth: So does @anorth.

PR #554, opened 2019-10-11
blockhain/vm -> vm

jbenet:
This PR promotes the vm into its own first-class system. It is one of
the most complex areas of the code, and can be implemented somewhat
separately from everything else. This will make the whole thing much more
understandable and implementable, especially if parties want to implement
the VM as a separate, sandboxed process.

This part also rearranges the blockchain package, and cleans up a bunch of
imports into the blockchain. This gets rid of base_blockchain (yay).


src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -0,0 +1,56 @@
...
+    Parents          Tipset
+    Weight           ChainWeight
+    Epoch            ChainEpoch
+
+    // Miner Info
+    MinerAddress     addr.Address
+
+    // State
+    StateTree        st.StateTree
+    Messages         MessageRoot
```
anorth: Putting messages here confuses the parent state context of the state tree and receipts. The latter two are the state, the messages is a proposal for the next tipset.

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -0,0 +1,56 @@
...
+// TODO: remove this. header is signed
+type SignedBlock struct {
+    block  &Block
+    sig    filcrypto.Signature
+}
+
+type Block struct {
+    Header    BlockHeader
+    Messages  [msg.Message]
+    Receipts  [msg.MessageReceipt]
```
anorth: Receipts don't belong with the block, they belong with a tipset. I don't think they need to be propagated in blocksub, only recalculated when a tipset is evaluated.

src/systems/filecoin_blockchain/struct/block/block.id
jbenet: Agree! good catch thanks will reorganize

src/systems/filecoin_blockchain/struct/block/block.id
jbenet: Yes, I think that's right ^ -- i'll remove them from here.

I guess we only have MessageReceipts in the `BlockHeader` because the VM execution happens "outside of the state tree" as ETH1  (as opposed to tracking Messages, Receipts, etc inside, as ETH2)

PR #555, opened 2019-10-11
Re-organize code according to spec and implementations and add c…

nicola:


src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.id
```diff
@@ -1,26 +1,41 @@
...
 type StoragePowerActor struct {
+    Miners [addr.Address]  // TODO this should have a different type to account for power
+
     // call by StorageMiningSubsytem on miner creation
-    RegisterMiner(ownerAddr addr.Address, workerPubKey filcrypto.PubKey) addr.Address
+    CreateStorageMiner(
+        ownerAddr     addr.Address
+        workerPubKey  filcrypto.PubKey
+        sectorSize    util.UInt
+        peerId        libp2p.PeerID
```
jbenet: I understand this was here before. This is not a good registry of libp2p keys -- they may change frequently, and libp2p keys need to be associated with more than just storage miners (retrieval miners, clients, bootstrap nodes). I'm OK adding this to reflect status quo, but expect this to be moved elsewhere before mainnet.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
```diff
@@ -9,26 +9,51 @@ type SectorCommitment struct {}
...
-    SubmitPoSt(
-        postProof        base_mining.PoStProof
-        sectorStateSets  sector.SectorStateSets
-    )
-    computeProvingPeriodEndSectorState()
     CommitSector(
         seed            Seed
         sector          sector.SectorID
-        sealCommitment  base_mining.SealCommitment
+        sealCommitment  base_mining.SealCommitment  // TODO: it's more expensive to do a struct
```
jbenet: - This is almost certainly going to remain as a struct. I understand the additional cost.
- `base_mining.SealCommitment` looks wrong. I believe this needs to be `sector.OnChainSealVerifyInfo`

jbenet:
@nicola merge as you prefer. (merge sooner to avoid merge conflicts). but note comments above

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.id
nicola: ok, will add a TODO comment to it.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
nicola: 👍

PR #557, opened 2019-10-12
Data Transfer Spec, Fleshed Out

hannahhoward:
# Goals

Flesh out data transfer interface

# Implementation

- Abstract out elements that are role specific (storage client, retrieval client, storage miner, retrieval miner) from the actual operation of data transfers
- instead just express two operations: a push & a pull. a push means you will send data to the recipient, a pull means you are requesting data from a recipient
- concept of a "DataTransferVoucher" which expresses enough information to validate and initiate the transfer
- change the transfer request format to simply a Piece ID + an ipld selector (would be a recursive/** for the whole piece)
- concept of a validator -- basically when a push transfer is received on the other end, validates whether the node will receive that push, when a pull transfer is received on the other end, validates whether data will get sent back
- tighten up the notification interface -- for registering and listening to events for a transfer
- add terminology definitions
- add protocol specification
- add flows and diagrams
- add more explanatory text
- copy Selectors spec from ipld

# For Discussion

This proposal separates the negotiation of the data transfer into a separate network round trip than the data transfer itself. This means all data transfers have at minimum 2 network round trips.

There are several reasons for this:
1. By having a negotiation roundtrip, the negotiation is separated from the protocol used to move data. Presumably, lots and lots of other protocols to move data might one day exist. This would mean negotiation logic could remain the same, even if the transport protocol changed.
2. In general, negotiating and agreeing upon transfers before attempting transfer makes tracking and managing transfers a bit easier
3. If the transfers are done with Graphsync, significant changes to the protocol are required to do it in one trip.
4. There will likely be situations where the client is behind a NAT. Having two round trips means that that all requests can initiate on the client, which establishes the connection before the graphsync request occurs.

That said I'm open to suggestions for doing this in a single round trip.


src/systems/filecoin_files/data_transfer/data_transfer_subsystem.id
```diff
@@ -1,66 +1,134 @@
...
+	requestValidator RequestValidator
+	pieceStore    piece.PieceStore
+
+	// open a data transfer that will send data to the recipient peer and
+	// transfer the whole piece
+	OpenPushDataChannel(to libp2p.PeerID, voucher DataTransferVoucher, PieceRef CID) ChannelID
+
+	// open a data transfer that will send data to the recipient peer and
+	// transfer parts of the piece that match the selector
+	OpenPartialPushDataChannel(to libp2p.PeerID, voucher DataTransferVoucher, PieceRef CID, Selector ipld.Selector) ChannelID
```
anorth: The distinction between these two is working around a limitation of IPLD selectors, which IMO are too opinionated about how they can be used. However, even if an unbounded recursive selector is not made available, can we not always know a good upper bound anyway? When sending data to a miner, the client knows the complete data depths and provides the selector to the miner.

hannahhoward: yes, I am inclined to simplify these methods and will update

jbenet:
# Quick update

- @hannahhoward this is AWESOME! 👍 great work here, i think this is super solid.
  - great code
  - great prose
  - great diagrams
  - in @zixuanzh's words "this is what every spec PR should be ⭐️ 👍 "
- great to have selectors stuff here 👍 
- i have some comments/questions about some of the flows
- and, i think some of the prose may need updating after the recent (a) pieces discussion, and (b) markets "data must be received before deal is signed" decisions.
- I'll go through the markdown file and mark them as inline comments here in this PR.

## cleaning the pieces notion out

i think it would be good to see if we can write all the Data Transfer module _without making reference to Pieces_ -- ie, only using `ipld.CID` and `ipld.Selector` 
  - this is because Retrieval Market requests may be for _any_ data (eg payload data, blockchain itself, potentially even sealed sectors) and do not need to come from `Pieces` specifically.
  - looking at spec+prose, i think `PieceRef` is only here because it came from `deal.StorageDeal.PieceRef`. So i think it can just turn into `ipld.CID`.
 - this would also make Data Transfer much more usable/reusable. eg consider this thing plugged into other standalone filecoin tools (eg to move market orders in bulk), or making its way into go-ipfs some day, or plugged into other protocols. (it's a pretty powerful underlying component)

## rebasing branch + some commits

See  http://github.com/filecoin-project/specs/commits/data_transfer 
- rebased on top of master, and made some commits (see log below). 
- I think we may need to open a new PR to use that branch as base? or just update `hannahoward/specs/data_transfer` to `filecoin-project/specs/data_transfer`) ? you decide
- (oh, one note, i'd like to ship this as part of spec v1.1 (shipping monday), so i may go through and merge this in a different PR on monday)

---

commit f4cc3039f4d499152fc84843da8970c9779e4cb6 (HEAD -> data_transfer, origin/data_transfer)
Author: jbenet <juan@benet.ai>
Date:   Sat Oct 26 21:49:41 2019 -0700

    data_transfer: added modules diagram

    This diagram shows how the Data Transfer component fits in with
    other parts, like the markets.

    I avoided showing individual event arrows for the protocol flows,
    because the mermaid diagrams are much better (and more editable than a png).

commit 3f6ee855e347b9b111b0692f556a8f6b4c35f69b
Author: jbenet <juan@benet.ai>
Date:   Sat Oct 26 21:49:20 2019 -0700

    data_transfer: changed to mmd diagrams

    - Requester -> Responder fixes
    - added <br /> to module labels
    - expanded Note to cover "whole system"
      - mermaid doesn't provide grouping of modules
      - this helps understand what's part of "one thing"

commit 9ba08d39b62822f2b20f549551003daf97818296
Author: jbenet <juan@benet.ai>
Date:   Sat Oct 26 21:45:38 2019 -0700

    data_transfer: Updated diagram paths to be relative

    In recent commits, the `diagram` shortcode changed to pull
    diagrams using relative paths (like readfile).

commit 485aacf4318131d7e3c1441b57b21e5444ca1a4a
Author: jbenet <juan@benet.ai>
Date:   Sat Oct 26 20:32:36 2019 -0700

    data_transfer: rebased and formatted code

    since original PR, we now have formatting rules and CI checks for it.

    run `make fmt-code` to format your code locally.

jbenet:
And, sorry it took so long for me to review! 😿

src/systems/filecoin_files/data_transfer/_index.md
```diff
@@ -3,4 +3,92 @@ menuTitle: Data Transfer
...
+2. The requestors' data transfer module will send a push request to the responder along with the data transfer voucher. It also puts the data transfer in the scheduler queue, meaning it expects the responder to initiate a transfer once the request is verified
+3. The responder's data transfer module validates the data transfer request via the Validator provided as a dependency by the responder
+4. The responder's data transfer module schedules the transfer
+5. The responder makes a GraphSync request for the data
+6. The requestor receives the graphsync request, verifies it's in the scheduler and begins sending data
+7. The responder receives data and can produce an indication of progress
+8. The responder completes receiving data, and notifies any listeners
+
+The push flow is ideal for storage deals, where the client initiates the push
+once it verifies the the deal is signed and on chain
```
jbenet: i think this needs to change after last week's discussions -- the data transfer is made before the deal is signed and on chain

src/systems/filecoin_files/data_transfer/_index.md
```diff
@@ -3,4 +3,92 @@ menuTitle: Data Transfer
...
 
+_Data Transfer_ is a system for transferring all or part of a `Piece` across the network when a deal is made.
+
+# Terminology
+
+- **Push Request**: A request to send data to the other party
+- **Pull Request**: A request to have the other party send data
+- **Requestor**: The party that initiates the data transfer request (whether Push or Pull)
+- **Responder**: The party that receives the data transfer request
+- **Data Transfer Voucher**: A wrapper around storage or retrieval data that can identify and validate the transfer request to the other party
```
jbenet: at first i wasn't in love with the word voucher, but i think it's pretty great:

> **voucher (n.)** a small printed piece of paper that entitles the holder to a discount, or that may be exchanged for goods or services.

i think this is much better than the prior name (auth token)

src/systems/filecoin_files/data_transfer/_index.md
```diff
@@ -3,4 +3,92 @@ menuTitle: Data Transfer
...
+Note that the Negotiation and Transfer stages can occur in seperate round trips,
+or potentially the same round trip, where the requesting party implicitly agrees by sending the request, and the responding party can agree and immediately send or receive data.
+
+# Example Flows
+
+## Push Flow
+
+{{< diagram src="./docs/systems/filecoin_files/data_transfer/push-flow.mmd.svg" title="Data Transfer - Push Flow" >}}
+
+1. A requestor initiates a Push transfer when it wants to send data to another party.
```
jbenet: wonder if `requestor`, `responder` etc should all be `code` words

src/systems/filecoin_files/data_transfer/_index.md
```diff
@@ -3,4 +3,92 @@ menuTitle: Data Transfer
...
+## Push Flow
+
+{{< diagram src="./docs/systems/filecoin_files/data_transfer/push-flow.mmd.svg" title="Data Transfer - Push Flow" >}}
+
+1. A requestor initiates a Push transfer when it wants to send data to another party. 
+2. The requestors' data transfer module will send a push request to the responder along with the data transfer voucher. It also puts the data transfer in the scheduler queue, meaning it expects the responder to initiate a transfer once the request is verified
+3. The responder's data transfer module validates the data transfer request via the Validator provided as a dependency by the responder
+4. The responder's data transfer module schedules the transfer
+5. The responder makes a GraphSync request for the data
+6. The requestor receives the graphsync request, verifies it's in the scheduler and begins sending data
```
jbenet: is "graphsync request" a data structure or function call? if so may be good to have `graphsync.Request`. (potentially defined in `libraries/ipfs/graphsync`. 🤷‍♂

src/systems/filecoin_files/data_transfer/_index.md
```diff
@@ -3,4 +3,92 @@ menuTitle: Data Transfer
...
+
+1. Negotiation - the requestor and responder agree to the transfer by validating with the data transfer voucher
+2. Transfer - Once both parties have negotiated and agreed upon, the data is actually transferred. The default protocol used to do the transfer is Graphsync
+
+Note that the Negotiation and Transfer stages can occur in seperate round trips,
+or potentially the same round trip, where the requesting party implicitly agrees by sending the request, and the responding party can agree and immediately send or receive data.
+
+# Example Flows
+
+## Push Flow
```
jbenet: not sure but there may be an "alternate push flow" possible/useful too that starts sending along with the Data Transfer request, and provides all the data needed to start accepting. this would shave 1 RTT. (maybe just out of scope for now until it's needed)

src/systems/filecoin_files/data_transfer/_index.md
```diff
@@ -3,4 +3,92 @@ menuTitle: Data Transfer
 title: "Data Transfer in Filecoin"
 ---
 
+_Data Transfer_ is a system for transferring all or part of a `Piece` across the network when a deal is made.
```
jbenet: curious if you have opinions about the name @hannahhoward -- "Data Transfer" sounds somewhat bland and potentially more generic/less useful than it should be. Wonder if some name closer to Graphsync may be worth rebinding too. (i called it this because i didnt know what to call it at the time yet, and didnt want to introduce a superfluous name)

src/systems/filecoin_files/data_transfer/_index.md
jbenet: (this could also just come later, post mainnet)

src/systems/filecoin_files/data_transfer/_index.md
```diff
@@ -3,4 +3,92 @@ menuTitle: Data Transfer
...
+{{< diagram src="./docs/systems/filecoin_files/data_transfer/push-flow.mmd.svg" title="Data Transfer - Push Flow" >}}
+
+1. A requestor initiates a Push transfer when it wants to send data to another party. 
+2. The requestors' data transfer module will send a push request to the responder along with the data transfer voucher. It also puts the data transfer in the scheduler queue, meaning it expects the responder to initiate a transfer once the request is verified
+3. The responder's data transfer module validates the data transfer request via the Validator provided as a dependency by the responder
+4. The responder's data transfer module schedules the transfer
+5. The responder makes a GraphSync request for the data
+6. The requestor receives the graphsync request, verifies it's in the scheduler and begins sending data
+7. The responder receives data and can produce an indication of progress
+8. The responder completes receiving data, and notifies any listeners
```
jbenet: LGTM

src/systems/filecoin_files/data_transfer/_index.md
```diff
@@ -3,4 +3,92 @@ menuTitle: Data Transfer
...
+1. A requestor initiates a Pull transfer when it wants to receive data from another party. 
+2. The requestors' data transfer module will send a pull request to the responder along with the data transfer voucher. 
+3. The responder's data transfer module validates the data transfer request via a PullValidator provided as a dependency by the responder
+4. The responder's data transfer module schedules the transfer (meaning it is expecting the requestor to initiate the actual transfer)
+5. The responder's data transfer module sends a response to the requestor saying it has accepted the transfer and is waiting for the requestor to initiate the transfer
+6. The requestor schedules the data transfer
+7. The requestor makes a GraphSync request for the data
+8. The responder receives the graphsync request, verifies it's in the scheduler and begins sending data
+9. The requestor receives data and can produce an indication of progress
+10. The requestor completes receiving data, and notifies any listeners
```
jbenet: LGTM

src/systems/filecoin_files/data_transfer/_index.md
```diff
@@ -3,4 +3,92 @@ menuTitle: Data Transfer
...
+1. A requestor initiates a Pull transfer when it wants to receive data from another party. 
+2. The requestor’s DTM schedules the data transfer
+3. The requestor makes a Graphsync request to the responder with a data transfer request
+4. The responder receives the graphsync request, and forwards the data transfer request to the data transfer module
+5. The requestors' data transfer module will send a pull request to the responder along with the data transfer voucher. 
+6. The responder's data transfer module validates the data transfer request via a PullValidator provided as a dependency by the responder
+7. The responder's data transfer module schedules the transfer
+8. The responder sends a graphsync response along with a data transfer accepted response piggypacked
+9. The requestor receives data and can produce an indication of progress
+10. The requestor completes receiving data, and notifies any listeners
```
jbenet: this 1-RTT flow is great, LGTM 👍

jbenet: - this reminded me of a discussion at filecoin team week, that maybe the validators should be plugged into graphsync instead of DT? or that maybe DT is a "more advanced graphsync module w/ scheduler and validation awareness" (kind of like a "more advanced HTTP Client or HTTP Server with connection pools and pipelining or whatever else"). 
- curious if this connection seems better or worse to you + how you think about graphsync?  it may be worthwhile to just make this module be "part of graphsync"
- (eg when i think about go-ipfs using graphsync, i think of it trying to plug in some validators (eg accept pushes if i've friended this node, or part of the same ipfs-cluster, or if bitswap ratio is X, or ...)

src/systems/filecoin_files/data_transfer/data_transfer_subsystem.id
```diff
@@ -1,66 +1,127 @@
...
-    totalSize      UVarint
-    sent           UVarint
-    received       UVarint
-
-    channelType()  DataTransferType  @(cached)
-    transferNum()  Float             @(cached)
+	// an identifier for this channel shared by request and responder, set by requestor through protocol
+	transferID     TransferID
+	// base CID for the piece being transferred
+	PieceRef      CID
```
jbenet: can we call this just `RootCID` ? or `RequestCID` or just `CID`? (trying to remove "piece" concept from DT to make it more fundamental)

src/systems/filecoin_files/data_transfer/data_transfer_subsystem.id
```diff
@@ -1,66 +1,127 @@
...
+	RetrievalDealVoucher
+}
+
+type StorageDealVoucher struct {
+	deal StorageDeal
+}
+
+type RetrievalDealVoucher struct {
+	deal RetrievalDeal 
+}
```
jbenet: - i think these vouchers should probably be defined in the Storage Market area. (there may be an annoying import problem to make the `DataTransferVoucher` union --  not sure if `id` takes `interface{}` yet. 
- Could be defined in `data_transfer_subsystem.go` as an interface

src/systems/filecoin_files/data_transfer/data_transfer_subsystem.id
```diff
@@ -1,66 +1,127 @@
...
-// Assumes access to error-checked, ordered, reliable transmission protocol
-type Scheduler struct {
-    ScheduleTransfer(x UInt, y Piece)
-    requestData(x DataTransferChannel) union {Ret Bytes, Err error}
-    sendData(x DataTransferChannel, y Bytes) error?
+type TransferID UInt
+
+type ChannelID struct {
+	to libp2p.PeerID
+	id TransferID
```
jbenet: a bit confused:
- does that mean the requestor's `DataTransferSubsystem.dataTransfers` map has `ChannelIDs` where the `to` field is `DataTransferSubsystem.host.PeerID` ? (ie itself?).
  - if so, maybe `requesterPeer`
- or does `to` mean "a channel/connection to peerX"), and it contains "the other side's peerID" 
  - if so, maybe `peer` may be a clearer name, or `remote` or `remotePeer` (as libp2p calls it)
- or is `to` supposed to be "the responder" ("a requester sends a request **to** the responder")
  - if so maybe calling it `responder` or `responderPeer` makes sense?
- (my understanding is off -- please correct me. i'm trying to square the "identifier for this channel shared by request and responder, **set by requestor** through protocol" with one map on `ChannelID` only)
- the distinction above matters in cases where the same pair of peers can have two active `DataTransferChannels` with the same `TransferID` (can happen if both are requesting stuff from each other) - i think that's OK because the `TransferID` is relative to the `requester` peer, so it is always well defined. (even if both sides request stuff from each other, and collide in TransferID, it's still unambiguous. BUT im not sure if `ChannelID` is unambiguous, it depends on the above.
---
- i'm reminded that IP network stacks tend to keep 5-tuples `(proto, srcAddr, srcPort, dstAddr, dstPort)`  and that stream muxers like yamux/etc keep stream ids in each direction
  - maybe keeping a 3-tuple makes sense `type ChannelID { requesterPeer, responderPeer, requesterTransferID }` ? not sure
  - i think i agree that we dont need a 4-tuple: `type ChannelID { localPeer, remotePeer, localTransferID, remoteTransferID }` because one channel is not bidirectional, has a single direction.

jbenet:
@hannahhoward  I merged this all in #605  -- I'll leave this PR open for the discussion, feel free to close when relevant.

hannahhoward:
Thanks sorry @jbenet  - I will follow up with these comments soon.

ianconsolata:
@hannahhoward assuming this is irrelevant now, so I'm gonna close this. Feel free to ping me if that is a mistake.

PR #558, opened 2019-10-12
Improve proofs

porcuquine:
First steps toward implementing SealSector. Looking for feedback on whether this is the right way to do it.

This feels at once exceptionally verbose *and* incomplete/non-exhaustive. That might just be the state of the art, but if there is a different/better way, I'd like to course correct.


jbenet:
> This feels at once exceptionally verbose and incomplete/non-exhaustive. That might just be the state of the art, but if there is a different/better way, I'd like to course correct.

- I think this is right, yes
- We should be lazy-loading things. we dont need _everything_ to be written out, mostly critical functions.
- Yes, the `.id` outputs end up very, very verbose unfortunately. You can drop down to Go instead of using the `.id` types if you prefer. You might end up with more concise implementation functions, and more verbose interfaces/struct definitions.
- The important part is that by compiling, we use the type system to check for correctness, and that we have the details spelled out.
- Already having to do this caught bugs in our thinking and changed algorithms/definitions/types -- the value is that, at cost of verbosity.

cc @jzimmerman to add thoughts

porcuquine:
I am 💯 on the type-checking. I think I'm just trying to find my way through the border so there is a clean penetration into a simpler world. I'll keep pushing attempts, since I think I'm getting there.

To reiterate: I'm not complaining. Rather, I want to understand the current limits, eventual aspirations, and best practices so I can chart the right path for what we have right now.

jbenet:
@porcuquine SGTM. -- my comments mostly for observers reading through and confused about tradeoffs

We need to write up some docs re how `codeGen` turns `.id` types to `.go` types, and how to use them/best practices.

src/systems/filecoin_mining/storage_proving/sealer/sealer.go
```diff
@@ -0,0 +1,57 @@
...
+func (s *SectorSealer_I) SealSector(si SealInputs) *SectorSealer_SealSector_FunRet_I {
+	sid := si.SectorID()
+
+	commD := sector.UnsealedSectorCID(s.ComputeDataCommitment(si.UnsealedPath()).As_commD())
+
+	buf := make(util.Bytes, si.SealCfg().SectorSize())
+	f := file.FromPath(si.SealedPath())
+	length, _ := f.Read(buf)
+
+	// TODO: How do we meant to handle errors in implementation methods? This could get tedious fast.
```
jbenet: (responding to comment) i think we want to handle them to be clear that implementors will need to deal with these potential problems. I'd prefer error returns (go/rust way), but im OK with panics for now, especially for assertions

jzimmerman: +1 -- I definitely agree we should handle them. I think panics/asserts are best initially, pending a later move to error return values (or, ideally, Rust-style error-handling macros).

jzimmerman:
This looks like a good style for the hybrid `.id`/`.go` implementations as the system is currently structured. The verbosity is hopefully temporary -- I'm currently working on prototyping ways to write implementations in `.id` source directly, and I think that continuing to go through some of these examples will provide valuable insight into the best approach there.

src/systems/filecoin_mining/storage_proving/sealer/sealer.go
```diff
@@ -0,0 +1,57 @@
...
+	return &SectorSealer_VerifySeal_FunRet_I{}
+}
+
+func (s *SectorSealer_I) ComputeDataCommitment(unsealedPath file.Path) *SectorSealer_ComputeDataCommitment_FunRet_I {
+	return &SectorSealer_ComputeDataCommitment_FunRet_I{}
+}
+
+func ComputeReplicaID(sid sector.SectorID, commD sector.UnsealedSectorCID, seed sector.SealRandomSeed) *SectorSealer_ComputeReplicaID_FunRet_I {
+
+	_, _ = sid.MinerID(), (sid.Number())
```
sternhenri: nit, I'd just add some slight comments where as a spec we want to strike some balance of not having the go-like interface take over potentially more descriptive work...

src/systems/filecoin_mining/storage_proving/sealer/sealer.id
```diff
@@ -20,13 +18,12 @@ type SectorSealer struct {
...
-    //    PieceLength   UInt,
-    //  ) union { error, PieceInclusionProofs }
-    //
-    //  VerifyPieceInclusionProof() struct {}
+    ComputeDataCommitment(unsealedPath file.Path) union {commD sector.UnsealedSectorCID, err error}
+    ComputeReplicaID(
+        sid    sector.SectorID
+        commD  sector.UnsealedSectorCID
+        seed   sector.SealRandomSeed
+    ) union {replicaID Bytes, err error}
```
sternhenri: this looks great above

PR #566, opened 2019-10-15
SDR [WIP]

porcuquine:
Work in progress.

Question for @jzimmerman: What's the right way to add support for `[]UInt`? I did the obvious thing in `util.go` (after failing on other 'obvious things'), but I suspect I haven't done everything — because I have to prefix (`util.UInts`) whereas I don't have to do that when using `Bytes`.


jzimmerman:
The `Uint` issue is due to a temporary ad-hoc solution I've implemented for pulling in symbols from util: in order to reference a particular symbol, it currently must be added manually to the list in the `TranslateGoIdent` function in `codeGen.go`. Hopefully we should be able to eliminate this requirement fairly soon.

porcuquine:
Thanks, that worked.

src/systems/filecoin_mining/sector/sealing.id
```diff
@@ -33,14 +31,25 @@ type OnChainSealVerifyInfo struct {
...
-    CommRLast             SealedSectorCID
+    CommRLast             Commitment
     CommC                 Commitment
 
     // TODO: This may be a partially-cached tree.
     // this may be empty
     CachedMerkleTreePath  file.Path
 }
 
+type ProofAuxTmp struct {
```
jbenet: since we now have two of these, what about naming them for the use:
```
type SealVerifyAux struct{ ... } // aux info to verify SealOutputs with
type SealProofVerifyAux struct { ... } // aux info to verify a CreateSealProofOutputs with
```

tools/codeGen/util/util.go
```diff
@@ -22,11 +23,15 @@ type _UnusedCompileAssert2 = [8 - unsafe.Sizeof(int(0))]byte
 type UVarint = uint64
 type Varint = int64
 type UInt = uint64
+type UInts = []uint64 // Subject to change when array support comes.
```
jbenet: there is array support, did you try: `[UInt]` ?

jbenet:
LGTM! great to see this taking shape and connected into the code -- will get even better once generics land, but already this is forcing us to be super concrete + clear with all the artifacts-- awesome

.gitignore
```diff
@@ -10,3 +10,5 @@ hugo/content/codeGen/main
 .DS_Store
 *.gen.go
 hugo/data/version.yml
+bin/
```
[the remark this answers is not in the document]
porcuquine: I don't remember why I added that. It was for my own sanity, but I might have gotten into a funny build state due to my configuration. I'll remove that line for now and worry about it later.

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -0,0 +1,296 @@
+package filproofs
+
+import big "math/big"
+import "encoding/binary"
+import . "github.com/filecoin-project/specs/util"
```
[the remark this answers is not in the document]
porcuquine: Sure, no problem. I did it because `util` is a jarring prefix for basic types.

jbenet:
I'm going to merge this in the interest of keeping velocity high and rebasing other PRs -- the minor comments can be addressed in followup PR (if at all)

src/systems/filecoin_mining/sector/sealing.id
porcuquine: Mmmm, that's not quite what they are though. They're both produced by `SealSector`. One holds things which will be required for `CreateSealProof`. The other holds the subset of those things which will be needed later when generating `PoSt`.

tools/codeGen/util/util.go
porcuquine: Nope. I missed the memo.

Issue #478, opened 2019-09-09
Hello protocol and practical secure bootstrapping


anorth:
When a new node joins a network it must determine the head of the chain which it will download and validate, before it has a validated chain to test against (obviously). Downloading, and especially validating, are a significant amount of work so it's highly desirable to pick the "right" chain, at least up to a point fairly close to the consensus head. After syncing up to "pretty close", a node can sync all chains branching after that point.

The spec provides for the `Hello` protocol, in which nodes exchange a *single* head tipset key and ~height~weight once upon connection, and a chain syncing protocol which can traverse backwards from some head (in flux, see #461). [Edit: I think the change from height to weight must be recent, and makes the below even more difficult].

While most of the policy here should be left to implementations, they can only work with what the protocol provides. A simple approach (for illustration) might be to take the chain specified by a majority of a trusted set of peers. However, since the chain moves, these peers will disagree about the very tip of the chain quite often, despite good behaviour; blocks take some time to propagate across the network, may be withheld temporarily, and nodes may temporarily disagree on the contents of the head tipset. Discovering and querying a large set of peers will take some time, during which the chain will move on, so later peers will almost certainly disagree with earlier ones, despite in fact having the same chain.

This all makes a simple majority painful to determine. With the spec as-is, implementations must do something like attempt to fetch some head length of the chain from each peer (despite being unable to fully validate it) and seek a consensus at some earlier block height.

Some proposals:
- Quantize the tipset reported in `Hello`, e.g. to an exact multiple of 10 blocks. This would cause well-behaved nodes to agree exactly most of the time
- Report multiple tipsets in `Hello`, e.g. head, a quantized `head - 10`, and quantized `head - 100`
- Some explicit notion of finality and/or checkpoints past which nodes will not accept a re-org, and then sync (or state-fetch) that finality-horizon or checkpoint before then syncing all chains after that point.

I'm open to all of these being rejected in favour of the implementation-complex but spec-simple method of concurrently syncing unvalidated chain segments from many peers.

Note that this issues is (trying to be) only about determining which tipset to sync a full chain from, not actually fetching it. After using, say, some trusted bootstrap nodes to determine a head, a node should then fetch the blocks from random peers to avoid DOSing the bootstrap nodes.

For background see some discussion in #106 and https://github.com/filecoin-project/specs/pull/196/files (mostly since reverted as being too prescriptive on implementations)

pooja:
[...]

PR #568, opened 2019-10-15
Adding ChainSync to spec (WIP)

jbenet:
**Warning: this PR is controversial, and very WIP**. Please read with that in mind/context. 

Much of this can be reduced in scope, but i needed to put this out there as is to cover the extensive complexity and security requirement here, so that we're all on the same page, and that if we drop something, we do so with very open eyes together.

Please see the rendering here:
https://ipfs.io/ipfs/QmZ4CmrdLEiGXC2hfF7uFAj3qkHCfKgzqCPc3VnGcTKbQm/#systems__filecoin_blockchain__chainsync

7102105 (jbenet, 9 hours ago)
   chainsync fsm diagram

d9f2dc8 (jbenet, 15 hours ago)
   blockchain/ChainSync: wrote up ChainSync, w/ FSM

   Commit includes:
   - summary of ChainSync
   - in-depth protocol prose for ChainSync
   - ChainSync FSM states prose
   - ChainSync FSM diagram
   - Progressive Block Validation WIP
   - Progressive Block Propagation WIP
   - Graphsync recommendations
   - Gossipsub recommendations
   - lip2p recommendations (PeerDiscovery)

7a1272b (jbenet, 30 hours ago)
   mining: fsm diagrams


jbenet:
This is a take on everything we've discussed in last 2-3 weeks around:
- chain syncing security
- bandwidth efficient block/message propagation
- hard/trusted checkpoints
- statetree syncing via graphsync

This still does not have (still WIP):
- the _exact_ algorithm for how to advance in the chain, and how to prioritize blocks to validate (ie what checks to do, exactly when, and how to pick between all the potential blocks we might want to validate at these different levels. 
- exact algorithm will be security critical, as otherwise it will be easy to DOS lots of nodes with a single laptop)

jbenet:
This is unevenly described,  too much detail in some places, and not enough detail in others. I am seeking feedback on things unclear. I _know_ there's a lot of key details left unsaid, that i need to write out. Your questions will help me prioritize what to write out.

jbenet:
# graphsync notes/Qs

@hannahhoward I'm using a bunch of Grapsync functions in this PR that may not exist. They _don't need to_, but the functionality is probably needed. May be worth discussing what is needed. AFAIU, the ability to send a selector to a single peer and get back the resulting graph is there. That should be enough for the basics. 

The other features i'd like to be able to use (but may not be there, and up to us to decide whether to add/when). Marked `[x]` with the things _i think_ we have:

- [x] `graphsync.Pull(peer, sel)` <-- get everything from `peer` with given selector (if `sel` is supported)
- [x] `selectors.SelectCID(cid)` <--- select a single cid
- [x] `selectors.SelectAll(cid)` <--- select everything under a cid
- [ ] `graphsync.Pull(peer, sel, c.IpldStore)` <-- smartly look inside `c.IpldStore` to not request things we already have
- [ ] `selectors.LimitedSelect(otherSelector, limit)` <--- be able to limit the number of objects returned
- [ ] `graphsync.Pull(peers, ...)` <--- be able to ask multiple peers smartly. (ie ask for different branches of the tree. this can be done using `LimitedSelector(cid, limit)`
- [ ] `graphsync.CacheSerialization(sel)` <--- tell graphsync to compute and cache the output of pulling `sel`, so we dont have to deserialize, look through, and walk the whole graph. can precompute

jbenet:
# Qs for @whyrusleeping 

- Can you look through the state machine definition? i tried what you suggested re writing out the states, clear definitions of when you move out. also showed it visually (it's pretty simple, [comparatively](http://www.texample.net/media/tikz/examples/PNG/tcp-state-machine.png))
- Can you look through `CHAIN_CATCHUP` and `CHAIN_FOLLOW` and lmk how close that is to what you do already? The `TargetHeads` and `BestTargetHead` should be capturing our discussions on head selection and working towards it. 
- Note: the FSM is conceptual, its existence in code is not _required_, perhaps the biggest concern is making sure NOT to finalize blocks (advance finality) until we're sure we're on the right chain (which i captured w/ the two different `CHAIN_` states.
- Can you look through the **security conditions** marked for various states?
- Can you look through **Progressive Block Validation** and **Progressive Block Propagation**
  - I'm perfectly fine adding these during testnet, and pruning down what we need from each of these (eg drop `MessagePoolHints`), but we need _some_ for security and bandwidth constraints
- Can you look through the `Bootstrap` state and lmk if it's reasonable / makes sense, what doesn't? specially interested in security bugs
- What of this seems the hardest/most time consuming? 
- Look through the `libp2p.PeerDiscovery` parts -- which of these optional ones do you think we should have? i lean toward `libp2p.PersistentPeerstore` as the biggest bang for buck because most/all of it is already in go-libp2p

# Qs for @ZenGround0

- Can you look through the **ChainSync Summary** and see if it makes sense given our conversations about finality and checkpoints? 
- Can you look through `CHAIN_CATCHUP` and `CHAIN_FOLLOW` and lmk how close that is to what you do already? how different is this from what you do already?
- Can you look through/review the **security conditions** throughout the FSM?
- see also Qs below for @anorth 

# Qs for @anorth and @ZenGround0 

- Can you look through the  **ChainSync Summary** and see if it makes sense given our conversations about finality and checkpoints? 
- Note: the FSM is conceptual, its existence in code is not _required_, perhaps the biggest concern is making sure NOT to finalize blocks (advance finality) until we're sure we're on the right chain (which i captured w/ the two different `CHAIN_` states.
- Can you look through the FSM and lmk what here seems infeasible/too hard? We certainly don't need all of it at testnet/mainnet, and we can get by with weaker security. But we'll want to get to strong security conditions in this critical spot asap. i'd like to shoot for a reasonable but secure enough target
- Can you look through the `Bootstrap` state? i bundled a lot of our concerns about network bootstrapping into this, because this is _the key part_ where secure bootstrapping is needed. The `PeerDiscovery` protocols discussion is also relevant to this-- specifically the configurable bootstrap list is required. 
  - i wanted to make the `libp2p.PersistentPeerstore` thing required too, because that gives us a ton of bootstrap security potential, and i know most/all of this is already in go-libp2p, but not sure what the todos for gfc would be.
- This shouldn't be introducing many new concepts/things (aside from the checkpointing / sync statetree stuff). It may look new because of the arrangement/collection into this protocool, The rest is reshuffling where it appears. My impression is this is already much of what gfc achieves, with answers for some key open Qs

# Qs for @sternhenri and @nicola 

- can you look through the finality and **security considerations** parts?
- can you look through **progressive block propagation**?
- can you look through **progressive block validation**?
- specifically interested in suggestions for where to fit in the `ChainWeight` sorting. I tried to put it in terms of `TargetHeads` (a list kept sorted by chain quality -- `block.ChainWeight`) and `BestTargetHead`
- this is a rough write up of the protocol we discussed 3-4 weeks ago -- can you follow? i've started to write out the id/code for it (id in `chainsync.id`, code not there yet). some different abstractions. might abandon/defer to later. feedback useful.

# Qs for @jzimmerman 

- can you review the whole ChainSync section? this should have a ton of the concerns we've worked on in past months, and should address the key attack vectors we were concerned with in BlockSyncing / Propagation.
- specifically, are the FSM states roughly sound? i dont describe the use of the priority queue here yet, that's coming. i wanted to get it out in code first-- maybe we can do this later in the week?
- this writeup is still WIP and not complete

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+  - Step 1. Bootstrap to the network, and acquire a "secure enough" set of peers (more details below)
+  - Step 2. Bootstrap to the `BlockPubsub` channels
+  - Step 3. Listen and serve on Graphsync
+- **Part 3: Synchronize trusted checkpoint state (`SYNC_CHECKPOINT`)**
+  - Step 1. Start with a `TrustedCheckpoint` (defaults to `GenesisCheckpoint`).
+  - Step 2. Get the block it points to, and that block's parents
+  - Step 3. Graphsync the `StateTree`
+- **Part 4: Catch up to the chain  (`CHAIN_CATCHUP`)**
+  - Step 1. Maintain a set of `TargetHeads`, and select the `BestTargetHead` from it
+  - Step 2. Synchronize to the latest heads observed, validating blocks towards them
```
whyrusleeping: You want to make sure the head you selected actually links to your last checkpoint. Naively trying to sync forward towards a target that might not line up is hard (and DoSable)

jbenet: - Yes agree - and syncing back just from the head has big DOS attack surface too. [0] need to pick common intermediate targets (or use log skip lists...). haven't written the detail -- will do so.
  - anyway, In the common case, thankfully, people will download and load the chain manually along w/ the software, and wont have to worry about it. (which is why im Ok hitting testnet and even mainnet w/ this risk -- ie download headers all the way back, then validate fwd.)

[0] Downloading just headers for 1yr of a single chain is on the order of many GB. Downloading that only to find out that head is broken over and over -- reminds of old bittorrent DOS attacks of never giving the last pieces, or sending multi-GB bogus pieces that can only be detected after downloading the whole thing

jbenet: (there's a very simple + clean solution using an actor to track `TrustedCheckpoints` and a log skip list back to it, i'll write it up as a candidate for you to eval)

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+      - If the `sender` is propagating the block at all to `receiver`, it is in their interest to provide the full content to `receiver` when asked. Otherwise the block may not get included at all.
+      - Lots of security assumptions here -- this needs to be hyper verified, in both spec and code.
+  - **Step 1. (sender) `Send SignedBlock`**:
+      - Block propagation begins with the `sender` propagating the `block.SignedBlock` - `Gossipsub.Send(b block.SignedBlock)`
+      - This is a very light object (~200 bytes), which fits into one network packet.
+      - This has the BlockCID, the MinerAddress and the Signature. This enables parties to (a) learn that a block from a particular miner is propagating, (b) validate the `MinerAddress` and `Signature`, and decide whether to invest more resources pulling the `BlockHeader`, or the `Messages`.
+      - Note that this can propagate to ther rest of the network before the next steps complete.
+  - **Step 2. (receiver) `Pull BlockHeader`**:
+      - if `receiver` **DOES NOT** already have the `BlockHeader` for `b.BlockCID`, then:
+        - `receiver` requests `BlockHeader` from `sender`:
```
ZenGround0: Throughout this section the spec calls for fetching from the sender.  Who is sender, the latest gossipsub router or the pubsub originator?  In either case this could be overly prescriptive.  Particularly in the case of messages most other (non-sender) peers should be able to fulfill the Pull Messages request and the network might operate more efficiently if the protocol gives them the freedom to do so.

Guessing this isn't the case, but if the sender is the original publisher that would be problematic because the whole network would hit that node at once.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+        - This is a light-ish object (<4KB).
+        - This has many fields that can be validated before pulling the messages. (See **Progressive Block Validation**).
+  - **Step 3. (receiver) `Pull MessageHints` (TODO: validate/decide on this)**:
+      - if `receiver` **DOES NOT** already have the full block for `b.BlockCID`, then:
+        - `receiver` requests `MessagePoolHints` from `sender`: ...
+        - `MessagePoolHints` are TBD -- this is a compressed representation of messages that can expedite message propagation by leveraging prior `MessagePool` syncing.
+        - (This is an extension and not required. can just do Step 4.)
+  - **Step 4. (receiver) `Pull Messages`**:
+      - if `receiver` **DOES NOT** already have the full block for `b.BlockCID`, then:
+        - if `receiver` has _some_ of the messages:
```
ZenGround0: I want to clarify that I believe this _some_ messages condition requires that the receiver ran step 3 and has message hints.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+  - for better security, use many different `libp2p.PeerDiscovery` protocols
+- `BlockPubsub.Bootstrap()`: establish connections to `BlockPubsub` peers
+- `Graphsync.Serve()`: set up a Graphsync service, that responds to others' queries
+- **Chain State and Finality**:
+  - In this state, the **chain MUST NOT advance** beyond whatever the node already has.
+  - No new blocks are reported to consumers.
+  - The chain state provided is whatever was loaded from prior executions (worst case is `LastTrustedCheckpoint`).
+- **security conditions to transition out:**
+  - `Network` connectivity MUST have reached the security level acceptable for `ChainSync`
+  - `BlockPubsub` connectivity MUST have reached the security level acceptable for `ChainSync`
```
ZenGround0: Are these security levels specified by the protocol? Seems like it could have network wide implications if everyone chooses lax security here but maybe not so its ok to leave up to user?

hannahhoward:
@jbenet

In terms of what graphsync can do -- I can't seem to check stuff on your comment, but here's my point by point:
- [x] - graphsync.Pull(peer, sel) <-- get everything from peer with given selector (if sel is supported)
- [x] selectors.SelectCID(cid) <--- select a single cid
- [x ] selectors.SelectAll(cid) <--- select everything under a cid 
   ** THIS IS IMPLEMENTED BUT WITH A LIMIT ** -- you don't want to just recursive dag walk infinitely in an untrusted context (in service of responding to a peer you don't trust). Trusted contexts may get implemented as part of the data transfer module, but I don't know if this applies here.
- [ ] graphsync.Pull(peer, sel, c.IpldStore) <-- smartly look inside c.IpldStore to not request things we already have
    - Not implemented, but may be partially implemented w/ Data Transfer Module in Graphsync protocol through what I call "Exclusion sets" -- essentially you would add to a query "If you cross these CIDs in the selector, don't send the blocks back cause I have them". Converting ipld store + selector to an exclusion set would be trivial (just execute the selector locally) up until the point you missed a block. This wouldn't work for blocks you already had but where only connected to what the selector traverses through a block you didn't have. My hope is this isn't something you need.

- [x] selectors.LimitedSelect(otherSelector, limit) <--- be able to limit the number of objects returned
   - This is effectively, implemented through recursive selectors and depth limits, in fact, it's sort of "more implemented" than Selector.selectAll

 - [ ] graphsync.Pull(peers, ...) <--- be able to ask multiple peers smartly. (ie ask for different branches of the tree. this can be done using LimitedSelector(cid, limit)
    - Nope. I would at least for now not choose to naturally implement this inside of the graphsync library itself. I'd make something above that breaks up the selector and makes individual queries.
- [ ] graphsync.CacheSerialization(sel) <--- tell graphsync to compute and cache the output of pulling sel, so we dont have to deserialize, look through, and walk the whole graph. can precompute
    - Nope. Doesn't feel that hard to implement, if we're thinking in terms of a sel -> cid mapping. everything just becomes a raw block before it goes over the network.

You may want to look at where the selector spec landed:
https://github.com/ipld/specs/blob/master/selectors/selectors.md
It's a pretty powerful DSL, minus the Conditions which aren't implemented

It can give you a sense of what selectors do and don't support.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+
+## ChainSync FSM: `CHAIN_CATCHUP`
+
+- While in this state:
+  - `ChainSync` is well-bootstrapped, and has an initial **trusted** `StateTree` to start from.
+  - `ChainSync` is receiving latest `Blocks` from `BlockPubsub`
+  - `ChainSync` starts fetching and validating blocks (see _Block Fetching and Validation_ above).
+  - `ChainSync` has unvalidated blocks between `ChainSync.FinalityTipset` and `ChainSync.TargetHeads`
+- **Chain State and Finality**:
+  - In this state, the **chain MUST NOT advance** beyond whatever the node already has.
```
anorth: What does this mean? That `TargetHeads` does not change? That the best validated head does not change? The latter seems wrong since in this state the node will progressively validate blocks that it has fetched, pushing the frontier forward.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+- **security conditions to transition out:**
+  - `StateTree` for `LastTrustedCheckpoint` MUST be stored locally and verified (hashing is enough)
+- **transitions out:**
+  - once node receives and verifies complete `StateTree` for `LastTrustedCheckpoint`: move to `CHAIN_CATCHUP`
+
+## ChainSync FSM: `CHAIN_CATCHUP`
+
+- While in this state:
+  - `ChainSync` is well-bootstrapped, and has an initial **trusted** `StateTree` to start from.
+  - `ChainSync` is receiving latest `Blocks` from `BlockPubsub`
```
anorth: Why does it need this? This seems irrelevant to fetching and validating up to `TargetHeads`, or do you explicitly mean that `TargetHeads` should be changing? If the latter, completely closing the gap between `ChainSync.FinalityTipset ... ChainSync.BestTargetHead` could be very difficult as the target keeps moving, but also unnecessary before switching to "follow" mode so long as validation has reached FinalityTipset.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+- **Chain State and Finality**:
+  - In this state, the **chain MUST NOT advance** beyond whatever the node already has.
+  - No new blocks are reported to consumers.
+  - The chain state provided is the available `Blocks` and `StateTree` for all available epochs,
+    specially the `FinalityTipset`.
+- **security conditions to transition out:**
+  - Gaps between `ChainSync.FinalityTipset ... ChainSync.BestTargetHead` have been closed:
+    - All `Blocks` and their content MUST be fetched, stored, linked, and validated locally.
+      This includes `BlockHeaders`, `Messages`, etc.
+    - Bad heads have been expunged from `ChainSync.BestHeads`. Bad heads include heads that initially
```
anorth: What is `BestHeads`? The various names `BestTargetHead`, `BestHeads`, `TargetHeads` are a bit confusing. `BestHeads` doesn't appear in the code further down. Does it refer to fully validated heads? But then this expunging comment doesn't make sense.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+- **transitions out:**
+  - once gaps between `ChainSync.FinalityTipset ... ChainSync.BestHeads` are closed: move to `CHAIN_FOLLOW`
+
+## ChainSync FSM: `CHAIN_FOLLOW`
+
+- While in this state:
+  - `ChainSync` is well-bootstrapped, and has an initial **trusted** `StateTree` to start from.
+  - `ChainSync` fetches and validates blocks (see _Block Fetching and Validation_).
+  - `ChainSync` is receiving and validating latest `Blocks` from `BlockPubsub`
+  - `ChainSync` DOES NOT have unvalidated blocks between `ChainSync.FinalityTipset` and `ChainSync.TargetHeads`
```
anorth: As soon as a new block is received by gossip, does `TargetHeads` not change? So then that new block would be unvalidated.

src/systems/filecoin_blockchain/chainsync/blocksync/_index.md
```diff
@@ -0,0 +1,58 @@
...
+On reception of a new block over the appropriate libp2p channel (see gossib_sub), the _Block Syncer_'s _OnNewBlock_ method is invoked. Thereafter, the syncer must perform syntactic validation on the block to discard invalid blocks and forward the others for further validation by the {{<sref chain_manager>}}. At a high level, a syntactically valid block:
+
+- must include a well-formed miner address
+- must include at least one well-formed ticket
+- must include an election proof which is a valid signature by the miner address of the final ticket
+- must include at least one parent CID
+- must include a positive parent weight
+- must include a positive height
+- must include a well-formed state root
+- must include well-formed messages, and corresponding receipts CIDs
```
anorth: Messages and receipts no longer correspond, since receipts now refers to the parent state

src/systems/filecoin_blockchain/chainsync/blocksync/_index.md
```diff
@@ -0,0 +1,58 @@
...
+
+Blocks are propagated over the libp2p pubsub channel `/fil/blocks`. The following structure is filled out with the appropriate information, serialized (with IPLD), and sent over the wire:
+
+```sh
+type BlockMessage struct {
+  header BlockHeader
+  secpkMessages []&SignedMessage
+  blsMessages []&Message
+}
+```
```
anorth: This conflicts with "Progressive Steps of Block Propagation"

src/systems/filecoin_blockchain/chainsync/chainsync.id
```diff
@@ -0,0 +1,95 @@
...
+    // CacheChainStore is the ipld Datastore through which to load and store
+    // chain datastructures that may be temporary and may be garbage collected.
+    CacheChainStore  ipld.Store
+
+    // Graph is a datastructure that provides efficient access to all
+    // nodes needed for validation, and keeps both forward and back links.
+    // Graph graph.BlockGraph
+
+    // FinalityTipset contains the latest tipset declared to be final.
+    // This is what we use to check against.
```
anorth: To check what against? 
A 1-sentence recapitulation of what "final" means would help here too.

src/systems/filecoin_blockchain/chainsync/chainsync.id
```diff
@@ -0,0 +1,95 @@
...
+    FinalityTipset   PartialGraph
+
+    // ValidG contains all the blocks we have validated, but are not
+    // yet declared Final. This means all the blocks connected into the
+    // main blockchain, but which ValidationGraph has not yet picked.
```
anorth: What does "picked" mean?

src/systems/filecoin_blockchain/chainsync/chainsync.id
```diff
@@ -0,0 +1,95 @@
...
+    ValidG           PartialGraph
+
+    // ToValidateG is the set of unvalidated Blocks fully connected
+    // to the validated graph. These are ready to be validated, all the
+    // data is local, and are ordered by total ChainWeight.
+    ToValidateG      PartialGraph
```
anorth: ~I don't think the chain weight can be computed until after validation (execution of all messages to discover what the state, in particular the power table, looks like).~
EDIT: ah, the block headers declares its weight, though this isn't validated yet.

anorth: To follow the naming pattern of predicates describing the data, suggest calling this `FetchedG`.

src/systems/filecoin_blockchain/chainsync/chainsync.id
```diff
@@ -0,0 +1,95 @@
...
+    UnconnectedG     PartialGraph
+
+    // TargetHeads is a list of candidate heads sorted by likelihood of being the
+    // best chain. (eg sorted by block.ChainWeight, even if not validated yet)
+    // TargetHeads SHOULD be a small set (<10)
+    //
+    // Obviously bad heads SHOULD be pruned out:
+    // - Clearly invalid ChainWeight (lower than BestHead or too high)
+    // - Heads that are not advancing in BlockPubsub
+    TargetHeads      [block.BlockCID]
```
anorth: I have assumed all the way through here that "head" is a tipset, not a single block. This and `BestTargetHead` contradict that. I think we might need a new word to refer to blocks that area at the fringe (have no children), distinct from tipsets that can be composed from those blocks.

src/systems/filecoin_blockchain/chainsync/chainsync.id
```diff
@@ -0,0 +1,95 @@
...
+    fullGraph                 BlockGraph
+
+    // Blocks is the set of all blocks in this PartialGraph
+    Blocks                    {block.BlockCID: struct {}}
+
+    // Heads is the set of root/head blocks
+    Heads                     [block.BlockCID]
+
+    // Tails is the set of leaf/tail blocks
+    Tails                     [block.BlockCID]
```
anorth: I'm confused. Does head refer to the newest thing, as in the "head of the chain" being the latest state? The use of "root" is confusing if so, since one might usually identify that with the genesis. Similarly "leaf" also sounds like the expanding fringe.

src/systems/filecoin_blockchain/chainsync/chainsync.id
```diff
@@ -0,0 +1,95 @@
...
+
+    // Add adds given BlockCID to the PartialGraph, updating Blocks,
+    // Roots, and Leaves as needed.
+    Add(c block.BlockCID)
+
+    // Remove deletes given BlockCID from the PartialGraph, updating
+    // Blocks, Roots, and Leaves as needed.
+    Remove(c block.BlockCID)
```
anorth: Is a graph necessarily connected, or is this just a bag of blocks, with each disconnected subgraph being reported in heads/tails?

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -45,12 +45,41 @@ type BlockHeader struct {
...
 }
+
+// Checkpoint represents a particular block to use as a trust anchor
+// in Consensus and ChainSync
+//
+// Note: a Block uniquely identifies a tipset (the parents)
+// from here, we may consider many tipsets that _include_ Block
+// but we must indeed include t and not consider tipsets that
+// fork from Block.Parents, but do not include Block.
+type Checkpoint BlockCID
```
anorth: With tipsets, using a single block here confuses me a bit. Things that would make more transparent sense would be a state CID, or a tipset identifier (list of block CIDs).

anorth:
Is the hello protocol now irrelevant, to be removed?

Using gossipsub as a required peer discovery mechanism is new for GFC, which currently relies on DHT traversal. I'm interested in the motivation for this.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
-- must include a positive height
-- must include a well-formed state root
-- must include well-formed messages, and corresponding receipts CIDs
-- must include a valid timestamp
+- **Part 1: Verify internal state (`INIT` state below)**
+  - SHOULD verify data structures and validate local chain
+  - Resource expensive verification MAY be skipped at nodes' own risk
+- **Part 2: Bootstrap to the network (`BOOTSTRAP`)**
+  - Step 1. Bootstrap to the network, and acquire a "secure enough" set of peers (more details below)
+  - Step 2. Bootstrap to the `BlockPubsub` channels
```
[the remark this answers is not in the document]
whyrusleeping: We use block pubsub from the get-go to ensure the sync process is always adequately informed. Theres a lot that can be done concurrent to the primary sync process to speed up and/or increase the certainty in the selected chain

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+### NaiveChainsync by just Graphsyncing blocks from BlockPubsub (MAY)
+
+It is highly probable that many nodes will find success by running a protocol
+as simple as this:
+
+```go
+func (c *NaiveChainsync) Sync() {
+  // for each of the blocks we see in the gossip channel
+  for block := range c.BlockPubsub.NewBlocks() {
+    c.IpldStore.Add(block)
```
whyrusleeping: is this assuming that the BlockPubsub validator is checking the high level (stateless) validity of the block?

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+
+    go func() {
+      // fetch all the content with graphsync
+      selector := ipldselector.SelectAll(block.CID)
```
whyrusleeping: select all here is kinda scarily undefined, does this imply we sync every cid linked by this block, recursively? This would imply syncing the state trees of every single block in the entire chain, something very expensive to do

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+
+### Progressive Block Validation
+
+- Blocks can be validated in progressive stages, in order to minimize resource expenditure.
+- Validation computation is considerable, and a serious DOS attack vector.
+- Secure implementations must carefully schedule validation and minimize the work done by pruning blocks without validating them fully.
+- **Progressive Stages of Block Validation**
+  - _(TODO: move this to blockchain/Block section)_
+  - **BV0 - Syntactic Validation**: Validate data structure packing and ensure correct typing.
+  - **BV1 - Light Consensus State Checks**: Validate `b.ChainWeight`, `b.ChainEpoch`, `b.MinerAddress`, are plausible (some ranges of bad values can be detected easily, especially if we have the state of the chain at `b.ChainEpoch - consensus.LookbackParameter`. Eg Weight and Epoch have well defined valid ranges, and `b.MinerAddress`
```
whyrusleeping: does this or does this not depend on having some amount of chain history linking to this block?

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+  - **BV4 - Verify MessageSigs**: Verify the signatures on messages
+  - **BV5 - Verify StateTree**: Verify the application of `b.Parents.Messages()` correctly produces `b.StateTree` and `b.MessageReceipts`
+- These stages can be used partially across many blocks in a candidate chain, in order to prune out clearly bad blocks long before actually doing the expensive validation work.
+
+### Progressive Block Propagation
+
+- In order to make Block propagation more efficient, we trade off network round trips for bandwidth usage.
+- **Motivating observations:**
+  - `Block` propagation is one of the most security critical points of the whole protocol.
+  - Bandwidth usage during `Block` propagation is the biggest rate limiter for network scalability.
```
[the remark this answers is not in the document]
whyrusleeping: This should be header plus ordered message cids

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+      - If the `sender` is propagating the block at all to `receiver`, it is in their interest to provide the full content to `receiver` when asked. Otherwise the block may not get included at all.
+      - Lots of security assumptions here -- this needs to be hyper verified, in both spec and code.
+  - **Step 1. (sender) `Send SignedBlock`**:
+      - Block propagation begins with the `sender` propagating the `block.SignedBlock` - `Gossipsub.Send(b block.SignedBlock)`
+      - This is a very light object (~200 bytes), which fits into one network packet.
+      - This has the BlockCID, the MinerAddress and the Signature. This enables parties to (a) learn that a block from a particular miner is propagating, (b) validate the `MinerAddress` and `Signature`, and decide whether to invest more resources pulling the `BlockHeader`, or the `Messages`.
+      - Note that this can propagate to ther rest of the network before the next steps complete.
+  - **Step 2. (receiver) `Pull BlockHeader`**:
+      - if `receiver` **DOES NOT** already have the `BlockHeader` for `b.BlockCID`, then:
```
whyrusleeping: So this implies that you want to *not* send out the block header? I'm confused, that seems like unnecessary round trips.

And if its not the block header, what are the 200 bytes you mention?

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+        - `receiver` requests `BlockHeader` from `sender`:
+            - `bh := Graphsync.Pull(sender, SelectCID(b.BlockCID))`
+        - This is a light-ish object (<4KB).
+        - This has many fields that can be validated before pulling the messages. (See **Progressive Block Validation**).
+  - **Step 3. (receiver) `Pull MessageHints` (TODO: validate/decide on this)**:
```
whyrusleeping: This makes propagation of blocks a multi-round trip process, and this call alone will make the bandwidth concerns worse than just sending {header, msgscids}. Really not sure what is being gained here.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+            // can be implemented on top of a graphsync that only pulls from a single
+            // peer and does not check local store first.
+        }
+        c.ChainCatchup() // on to CHAIN_CATCHUP
+    }
+    ```
+- **security conditions to transition out:**
+  - `StateTree` for `LastTrustedCheckpoint` MUST be stored locally and verified (hashing is enough)
+- **transitions out:**
+  - once node receives and verifies complete `StateTree` for `LastTrustedCheckpoint`: move to `CHAIN_CATCHUP`
```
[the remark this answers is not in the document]
whyrusleeping: That lines up with my understanding

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+  - New blocks are finalized as they cross the finality threshold (`ValidG.Heads[0].ChainEpoch - FinalityLookback`)
+  - New finalized blocks are reported to consumers.
+  - The chain state provided includes the `Blocks` and `StateTree` for the `Finality` epoch, as well as
+    candidate `Blocks` and `StateTrees` for unfinalized epochs.
+- **security conditions to transition out:**
+  - Temporary network partitions (see _Detecting Network Partitions_).
+  - Encounter gaps of >2 epochs between Validated set and a new `ChainSync.BestTargetHead`
+- **transitions out:**
+  - if a temporary network partition is detected: move to `CHAIN_CATCHUP`
+  - if gaps of >2 epochs form between the Validated set and `ChainSync.BestTargetHead`: move to `CHAIN_CATCHUP`
```
whyrusleeping: >2 ? that will happen frequently. Null blocks will happen quite often that cause gaps like this

src/systems/filecoin_blockchain/chainsync/blocksync/_index.md
jbenet: ah yes! great catch, thanks

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+  - Gaps between `ChainSync.FinalityTipset ... ChainSync.BestTargetHead` have been closed:
+    - All `Blocks` and their content MUST be fetched, stored, linked, and validated locally.
+      This includes `BlockHeaders`, `Messages`, etc.
+    - Bad heads have been expunged from `ChainSync.BestHeads`. Bad heads include heads that initially
+      seemed good but turned out invalid, or heads that `ChainSync` has failed to connect (ie. cannot
+      fetch ancestors connecting back to `ChainSync.FinalityTipset` within a reasonable amount of time).
+    - All blocks between `ChainSync.FinalityTipset ... ChainSync.BestHeads` have been validated
+      This means all blocks _before_ the best heads.
+  - Not under a temporary network partition
+- **transitions out:**
```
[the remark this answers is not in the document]
ZenGround0: Hmm I think I'm talking about a less dangerous but harder to fix problem.  The situation I'm flagging is when a node's finalized tipset as set during FOLLOW is already on a fork off of the main chain (due to undetected network partition for example).  The observation is that the existing spec of CATCHUP honors the finality condition set in follow, even though this condition was potentially based on bad assumptions.

I think this is not an issue if we trust that the network partition detector doesn't have false negatives.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+- **Progressive Steps of Block Propagation**
+  - **IMPORTANT NOTES**:
+      - these can be effectively pipelined. The `receiver` is in control of what to pull, and when. It is up them to decide when to trade-off RTTs for Bandwidth.
+      - If the `sender` is propagating the block at all to `receiver`, it is in their interest to provide the full content to `receiver` when asked. Otherwise the block may not get included at all.
+      - Lots of security assumptions here -- this needs to be hyper verified, in both spec and code.
+  - **Step 1. (sender) `Send SignedBlock`**:
+      - Block propagation begins with the `sender` propagating the `block.SignedBlock` - `Gossipsub.Send(b block.SignedBlock)`
+      - This is a very light object (~200 bytes), which fits into one network packet.
+      - This has the BlockCID, the MinerAddress and the Signature. This enables parties to (a) learn that a block from a particular miner is propagating, (b) validate the `MinerAddress` and `Signature`, and decide whether to invest more resources pulling the `BlockHeader`, or the `Messages`.
+      - Note that this can propagate to ther rest of the network before the next steps complete.
```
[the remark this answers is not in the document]
anorth: This has fairly significant implementation ramifications, so a clear decision one way or the other would be appreciated (you can change it later, but it'll be expensive).

One significant implication: if the BlockBrief is propagated before a node has fetched the header and messages, then the receiver of a BlockBrief has no idea where to fetch that data from: the sender is not particularly likely to have it. This seems to make the fetching described below not work very well.

On the other hand, if a node does assure that it has the header/messages before propagating, then the BlockBriefs will not zip across the network: they'll move at the speed allowed by the round-trips and transfer of headers and messages.

src/systems/filecoin_blockchain/chainsync/chainsync.id
anorth: I'm ok with most of this sync spec being phrased in terms of blocks. It does simplify a few things. But "head" in particular is ambiguous, so qualifying it always would help.

src/systems/filecoin_blockchain/struct/block/block.id
anorth: So how about a state CID instead of a Block CID?

Note that the genesis state is not constructable as a sequence of messages applied to an empty state (it contains system actors already installed). We actually don't need a genesis block so much as a genesis state. What would break if multiple miners produced blocks in round 1, all sharing the same parent state root?

anorth:
Thanks, the responses clear up a few things, one being the mechanics of finality. I think this could benefit from a brief exposition at the start explaining the intentions behind the finality tipset, best heads, various graph bits etc.

ZenGround0:
@jbenet here is a question that came up from discussion with @anorth:

 Imagine we go with the minimal `BlockStamp` proposed here for the pubsub propagated message and we keep the validation conditions as described.  

This pubsub message will be propagated before fetching the header per the validation conditions.  This means that peers can not be sure that their last-hop gossipsub peer actually has the header data that they are going to ask to fetch.

In fact they should *expect* that this peer *won't* have that data available on graphsync.  The `BlockStamp` message is designed to propagate in one RTT which it can only achieve by *omitting the header*.  This implies that the previous hop's graphsync fetch of the previous header is going to take > 1 RTT (otherwise just keep it in the gossipsub message and achieve the same goals).  Going off of the Progressive Steps of Block Propagation pubsub propagation happens at about the same time as header pull.  Assuming `BlockStamp` validation is negligible and uniform hop RTTs, the request for a header from the next hop will hit a node at 1 RTT while the complete header data will not arrive from the previous hop until the time it takes to transfer the block, > 1 RTT.  It seems peer-retry logic or backoff timeouts are critical for implementations to work in the common case, not just when something goes wrong.

The question is: should we fetch the header for a `BlockStamp` before propagating it to the next hop to avoid this need for backoff?  It seems like this will reduce the network propagation to the same speed as just transmitting the header in the pubsub message in the first place.  This leads to the follow up question: why not transfer the header directly in the pubsub message?

src/systems/filecoin_blockchain/struct/block/block.id
jbenet: > So how about a state CID instead of a Block CID? 

I'm not seeing how this is better? maybe im missing it? 

I understand some of the motivation -- the genesis block seems superfluous, in light that it points back to a parent `StateTree` anyway. And I see that the algorithms here can be changed to reason also about `StateTrees` as checkpoints, and to deal with cases of syncing blocks back to a `StateTree` instead of syncing blocks back to another block (this would add a lot `StateTree` oriented checks in addition to all the block oriented checks we already need).

this would change the logic of tools and processes/expectations around checkpoints that already exist in broader community w/ other chains -- this can be changed, but we would have to make sure a lot of tools supported looking up `StateTree` by cid (tools which already _would_ support lookup of blocks by cid but not necessarily a `StateTree`, eg simple block explorers, wallets, chain exporters, etc). this is different and would need more work, but doable. 

Im not seeing lots of benefit for these changes and additional tooling work -- but maybe im missing them? is there more beyond the conceptual re-writing and cleaning up of checkpoints? (namely, that they always point to a "thing" at a particular round of which there is only one valid instance-- ie only one `StateTree` per round)

I would be OK doing this (changing to use `StateTree` as the checkpoint) if `StateTree` also became the "root object" and included the blocks within it (as in ETH2, and some other proposed chains, which treat consensus as a VM-internal process, not VM-surrounding process). Making `StateTree` the root object would cement its standing as a key, root data structure, that would need to get support in a ton of tools. (but would still have a genesis block, see below.)

> Note that the genesis state is not constructable as a sequence of messages applied to an empty state (it contains system actors already installed). 

Yes

> We actually don't need a genesis block so much as a genesis state. 

I believe we do need a genesis block -- we derive randomness from the ticket there. Rather than special casing, it is easier/less complex to ensure a well-formed chain always, including at the beginning

> What would break if multiple miners produced blocks in round 1, all sharing the same parent state root?

invariants/tools expecting the chain to end at one block (this reminds of git's empty repo and the edge cases it introduces, instead of starting always with the same empty/first commit). 

In any case, the mainnet filecoin chain will need to start with a small chain stub of blocks.
- product reason, we actually have to include some data in different blocks. (dont worry about this at all till Q1 -- we'll produce the relevant chain stub)
- A lot of code expects lookbacks, especially actor code.  Rather than introducing a bunch of special case logic for what happens ostensibly once in network history (special case logic which adds complexity and likelihood of problems), it is easiest to assume the chain is always at least X blocks long, and the system lookback parameters are all fine and dont need to be scaled in the beginning of network's history.

jbenet:
> Is the hello protocol now irrelevant, to be removed?

hello can stay -- still useful to query for this info, like `libp2p.Identify` or https://gateway.ipfs.io/version -- I'll resurface it (removal was a bug)

> Using gossipsub as a required peer discovery mechanism is new for GFC, which currently relies on DHT traversal. I'm interested in the motivation for this.

- Feel very free to still use DHT traversal, but we should be aware the DHT is susceptible to eclipse and other attacks, so may not satisfy security requirements on its own
  - I set this as optional to make it _easier_ to implement filecoin, and _a bit more secure_. (implementing DHT on top of gossipsub is harder -- not all libp2p impls have it, but they do have gossipsub)
  - Certainly if you have it already, fine to run it
- Discovery through Gossipsub helps tree forming, and security of block broadcast.
  - not 100% sure, but it may be sufficient to consider with very low probability establishing new connections to peers we hear from as broadcasters of messages.
  - bitcoin uses a random connect/disconnects to ensure its gossip channel remains hard to control/eclipse)
- gossipsub may end up doing it underneath the hood (spoke about this to raul-- but this may end up elsewhere -- as "a simple peer exchange" protocol)
- if we do "a simple peer exchange" outside of gossipsub, then maybe gfc can get this for free. will connect w/ @raulk

jbenet:
## Answers to @ZenGround0 

> It is fairly different but the spec is in the right direction.

good to hear :)

> It all looks doable once we converge on details.  I think everyone agrees we can and should skip writing the SYNC_CHECKPOINT phase.

yep

> The spec has lots of great detail on how to sync blocks in CHAIN_FOLLOW and has almost nothing for how to sync in CHAIN_CATCHUP.  

Ok, will flesh out. It should be fetching block headers first, connect the graph, then fetch messages to do full validation. (but will spell this out in doc)

> The successive propagation section doesn't help with fetching blocks from the past that don't show up on pubsub.  

Right

> Confusingly there are lots of partial graph datastructures and other hints lying around indicating that there *is* an expectation of doing something specific to handle CHAIN_CATCHUP. 

Yes, those data structures are for a `CHAIN_CATCHUP` algorithm impl. Will get to it, but want to land some of the deals stuff first. if i can describe the algorithm in prose to satisfaction is it ok for code to wait?

> - Does syncing large sections backwards with graphsync as gofc does now still work or should we be syncing forward somehow?

**I think this can work in practice, given that:**
- we are listening in gossipsub for new heads, so we're exposed to what the chain is doing (and likely to consider the real best head as part of our targets)
- we sync headers first, to connect back to `FinalityTipset`
  - headers are relatively small, and could make it back quickly
  - but note headers for a whole year could be tens of GB
- we can run bootstrap nodes to serve chain data very quickly.
- many parties will download large swaths of the chain ahead of time (w/ the software) and load it in manually.

**It is safest, though, to advance in steps towards the targets.**
- I was striving for a protocol here that advanced using intermediate targets between valid fringe and head (eg `ValidG <- IntermediateBlock <- BestTargetHead`), but was dissuaded/redirected to higher prios
- to do this, we need a nice/cheap way of getting the intermediate targets:
  - the quantized heads @anorth mentioned in https://github.com/filecoin-project/specs/issues/478#issuecomment-544954498
  - the log skip list i mentioned above too
- 4 ways of doing this:
  - right in the hello protocol (as @anorth suggested) -- see also my comments https://github.com/filecoin-project/specs/issues/478#issuecomment-544962977
  - by just querying peers for blocks at specific round heights
  - using an actor to compute them w/ Cron, so we would just need to peek into the latest `StateTree` (ie having to fetch that part 🙄). can be amortized by calculating every ~1000 blocks.
  - putting a log skip-list like thing in the `BlockHeader` (oof, header gets larger-- 😢 -- could be amortized by putting it only every ~1000 blocks.
- but this may be more trouble than this is worth atm -- it is fine to hit testnet without this, and that gives me space to land the higher prio things before.
- what do you think? should we finish this and aim for the more secure version w/ intermediate?

> - Are messages and headers supposed to arrive separately as in CHAIN_FOLLOW or can we keep fetching full blocks during CHAIN_CATCHUP. 

- Would recommend fetching headers first -- connect the graph, verify blocks as far as you can, before fetching messages (the heaviest part)

> - How do we interleave syncing from target heads and updating those heads? 

- if only syncing one at a time: When querying (say, using graphsync) query for a limit of headers, so that if the head changes significantly, to a totally different fork, you catch it the next time you sync.
- if syncing many in parallel: sync to k target heads at once, where k < 5, and where chains are expected to be highly overlapping

> - Do nodes propagate over gossipsub during CHAIN_CATCHUP?

No-- should only listen. (unless this is required from gossipsub assumptions)

> Along the same lines separating target head management into its own concept is a great idea.  But currently it is also underspecified. 

I can write out the algorithm in mind in spec code, and that will force it all out. but i need to land some other higher prio stuff first. Can you ask more questions of what is underspecified to make sure i write it out in prose here

> Throughout the comments there is talk about what we can leave out until mainnet.  I would like us to come to a well understood agreement on that once the spec details are agreed upon.

Sounds good 👍

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,388 @@
...
+- **transitions out:**
+  - once node receives and verifies complete `StateTree` for `LastTrustedCheckpoint`: move to `CHAIN_CATCHUP`
+
+## ChainSync FSM: `CHAIN_CATCHUP`
+
+- While in this state:
+  - `ChainSync` is well-bootstrapped, and has an initial **trusted** `StateTree` to start from.
+  - `ChainSync` is receiving latest `Blocks` from `BlockPubsub`
+  - `ChainSync` starts fetching and validating blocks (see _Block Fetching and Validation_ above).
+  - `ChainSync` has unvalidated blocks between `ChainSync.FinalityTipset` and `ChainSync.TargetHeads`
```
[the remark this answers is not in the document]
jbenet: Yes, correct. if `LastTrustedCheckpoint` is our _latest_ finalized tipsed, use that as `ChainSync.FinalityTipset`. will write it up

src/systems/filecoin_blockchain/chainsync/_index.md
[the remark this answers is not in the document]
jbenet: > The observation is that the existing spec of CATCHUP honors the finality condition set in follow, even though this condition was potentially based on bad assumptions.

ah yes. indeed.

> I think this is not an issue if we trust that the network partition detector doesn't have false negatives.

Yes, or rather: unlikely and should be fixed manually.

src/systems/filecoin_blockchain/chainsync/chainsync.id
jbenet: ok will do

jbenet:
@ZenGround0 

> Imagine we go with the minimal BlockStamp proposed here for the pubsub propagated message and we keep the validation conditions as described.
>
> This pubsub message will be propagated before fetching the header per the validation conditions. This means that peers can not be sure that their last-hop gossipsub peer actually has the header data that they are going to ask to fetch.
>
> In fact they should expect that this peer won't have that data available on graphsync. The BlockStamp message is designed to propagate in one RTT which it can only achieve by omitting the header. This implies that the previous hop's graphsync fetch of the previous header is going to take > 1 RTT (otherwise just keep it in the gossipsub message and achieve the same goals). Going off of the Progressive Steps of Block Propagation pubsub propagation happens at about the same time as header pull. Assuming BlockStamp validation is negligible and uniform hop RTTs, the request for a header from the next hop will hit a node at 1 RTT while the complete header data will not arrive from the previous hop until the time it takes to transfer the block, > 1 RTT. It seems peer-retry logic or backoff timeouts are critical for implementations to work in the common case, not just when something goes wrong.

I believe all of the above is correct

> The question is: should we fetch the header for a BlockStamp before propagating it to the next hop to avoid this need for backoff? It seems like this will reduce the network propagation to the same speed as just transmitting the header in the pubsub message in the first place. This leads to the follow up question: why not transfer the header directly in the pubsub message?

The main reason is not speed, but wasted bandwidth. Getting it down to something that fits in 1 packet is nice for reducing duplicate bandwidth use. 

If you imagine that you will receive the header once per gossipsub peer (or if lucky, half of them), and that there is `EC.E_LEADERS=10` blocks per round, then we're talking the difference between:

```
16 peers, 1 pkt  -- 1 * 16 * 10 = 160 dup pkts (256KB) in <5s
16 peers, 4 pkts -- 4 * 16 * 10 = 640 dup pkts (1MB)   in <5s

32 peers, 1 pkt  -- 1 * 32 * 10 =   320 dup pkts (512KB) in <5s
32 peers, 4 pkts -- 4 * 32 * 10 = 1,280 dup pkts (2MB)   in <5s
```

2MB may not be worth saving-- and maybe gossipsub can be _much_ better about supressing dups.


**Important:** I am OK to not do this until we find it to be a problem -- ie skip step 1 and step 2 for testnet/mainnet, and propagate `BlockHeader` instead (as an alternate Step 1). If we find we need to reduce the bw usage at block propagation time, we adjust then.

src/systems/filecoin_blockchain/chainsync/_index.md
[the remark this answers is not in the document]
jbenet: Yes, understood. https://github.com/filecoin-project/specs/pull/568#issuecomment-544975833

jbenet:
@ZenGround0 to Q in #fil-specs:

- `BlockStamp` and `MessageHints` may not be that useful yet, so in terms of time/simplicity, may be best to skip for testnet, and see if it's a problem then. (we can crank up # of nodes, `EC.E_LEADERS`, and crank down block time to see where it breaks).
- MessageHints may save a lot of bandwidth -- we may need to do this more so than BlockStamp`, but let's defer.

I think though the right order of implementation is:

- Step 1. change to: Step 1b: w/`BlockHeader` instead of `BlockStamp` directly (still no messages, `MessageHints` or message `CIDs`)
- Step 2. (skip, already have `BlockHeader`)
- Step 3. change to Step 3b: pull `message CIDs` instead of `MessageHints`
- Step 4. same
- Step 5. same

if we see bandwidth being a problem in testnet, we can diagnose and decide to bring back either. I will (in the meantime) spec out and test `MessageHints` and get better understanding of what we can do with gossipsub duplicate message suppression.

cc @anorth @whyrusleeping

src/systems/filecoin_blockchain/struct/block/block.id
anorth: Ok, thanks. You understand my motivations and this proposal is reasonable.

anorth:
> I am OK to not do this until we find it to be a problem -- ie skip step 1 and step 2 for testnet/mainnet, and propagate BlockHeader instead (as an alternate Step 1). [etc]

Thanks. I see great utility in speccing and implementing something simple until we demonstrate (research or implementation) a problem that needs a more complex solution.

In this case, I would request that the spec today describe only what we currently expect implementations to align on for/during testnet. It's fine to note that operational experience may require a change before (or after some time during) mainnet. I would prefer the other options be demoted to an appendix, if not a gist or issue, just so that there's no confusion about what the immediately-expected behaviour is.

jbenet:
> In this case, I would request that the spec today describe only what we currently expect implementations to align on for/during testnet. It's fine to note that operational experience may require a change before (or after some time during) mainnet. I would prefer the other options be demoted to an appendix, if not a gist or issue, just so that there's no confusion about what the immediately-expected behaviour is.

Ok SGTM. will make sure it shows right. 

(i may find a way to comment it out in the markdown so it stays close to where it's supposed to be)

src/systems/filecoin_blockchain/chainsync/_index.md
jbenet: addressed

src/systems/filecoin_blockchain/chainsync/blocksync/_index.md
jbenet: addressed

raulk:
> if we do "a simple peer exchange" outside of gossipsub, then maybe gfc can get this for free. will connect w/ @raulk

We can totally do this. The simplest form would be an RPC protocol with two operations:

1. Advertise a set of hashes/CIDs for services we advertise, where the hash digests a contextualised string, e.g. `gossipsub:topic_name`.
2. Lookup peers for a given hash/CID.

src/systems/filecoin_blockchain/chainsync/chainsync.id
jbenet: addressed

src/systems/filecoin_blockchain/chainsync/chainsync.id
jbenet: addressd

src/systems/filecoin_blockchain/chainsync/chainsync.id
jbenet: addressed

src/systems/filecoin_blockchain/chainsync/chainsync.id
jbenet: addressed

jbenet: (for now keeping word Head but clearly noting it is a block. may still switch it before PR lands.

jbenet:
Can i get a review + approval so i can merge the PR?

raulk:
re: general-use peer exchange protocol, would something like this work for you, @jbenet? https://github.com/libp2p/specs/issues/222

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,490 @@
...
+specific to Filecoin's choices in state representation and consensus rules,
+but is general enough that it can serve other blockchains. `ChainSync` is a
+group of smaller protocols, which handle different parts of the sync process.
 
-# Block reception and Syntactic Validation
+# Terms and Concepts
 
-- Called by: libp2p {{<sref chainsync>}}
-- Calls on: {{<sref chain_manager>}}
+- `LastCheckpoint` the last hard social-consensus oriented checkpoint that `ChainSync` is aware of.
```
sternhenri: a link here to concept in consensus would be nice to disambiguate what could be meant by checkpoint

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,490 @@
...
+  - `chainsync.BlockCache` a temporary cache of blocks, to constrain resource expended
+  - `chainsync.AncestryGraph` a datastructure to efficiently link `Blocks`, `Tipsets`, and `PartialChains`
+  - `chainsync.ValidationGraph` a datastructure for efficient and secure validation of `Blocks` and `Tipsets`
+
+## Graphsync in ChainSync
+
+`ChainSync` is written in terms of `Graphsync`. `ChainSync` adds blockchain and filecoin-specific
+synchronization functionality that is critical for Filecoin security.
+
+### Rate Limiting Graphsync responses (SHOULD)
```
sternhenri: nice use of keyword SHOULD

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,490 @@
...
+  - `ChainSync` is well-bootstrapped, and has an initial **trusted** `StateTree` to start from.
+  - `ChainSync` is receiving latest `Blocks` from `BlockPubsub`
+  - `ChainSync` starts fetching and validating blocks (see _Block Fetching and Validation_ above).
+  - `ChainSync` has unvalidated blocks between `ChainSync.FinalityTipset` and `ChainSync.TargetHeads`
+- **Chain State and Finality**:
+  - In this state, the **chain MUST NOT advance** beyond whatever the node already has:
+    - `FinalityTipset` does not change.
+    - No new blocks are reported to consumers/users of `ChainSync` yet.
+    - The chain state provided is the available `Blocks` and `StateTree` for all available epochs,
+      specially the `FinalityTipset`.
```
sternhenri: not sure what the "specially" here means

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,490 @@
...
+    - finality must not move forward here because there are serious attack vectors where a node can be forced to end up on the wrong fork if finality advances before validation is complete up to the block production fringe.
+  - Validation must advance, all the way to the block production fringe:
+    - Validate the whole chain, from `FinalityTipset` to `BestTargetHead`
+    - The node can reach `BestTargetHead` only to find out it was invalid, then has to update `BestTargetHead` with next best one, and sync to it
+      (without having advanced `FinalityTipset` yet, as otherwise we may end up on the wrong fork)
+- **security conditions to transition out:**
+  - Gaps between `ChainSync.FinalityTipset ... ChainSync.BestTargetHead` have been closed:
+    - All `Blocks` and their content MUST be fetched, stored, linked, and validated locally.
```
sternhenri: I think this condition may be too strict. That is we need only have those blocks in the heaviest chain have been f/s/l/v locally. We are going depth first i believe to drive this (though obviously some breadth will be covered so doing). As soon as the heaviest "heard about" chain is validated, you can transition out. You needn't validate all other chains in order to move forward i believe.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,490 @@
...
+
+- While in theis state:
+  - `ChainSync` is well-bootstrapped, and has an initial **trusted** `StateTree` to start from.
+  - `ChainSync` fetches and validates blocks (see _Block Fetching and Validation_).
+  - `ChainSync` is receiving and validating latest `Blocks` from `BlockPubsub`
+  - `ChainSync` DOES NOT have unvalidated blocks between `ChainSync.FinalityTipset` and `ChainSync.TargetHeads`
+  - `ChainSync` MUST drop back to another state if security conditions change.
+  - Keep a set of gap measures:
+    - `BlockGap` is the number of remaining blocks to validate between the Validated blocks and `BestTargetHead`.
+      - (ie how many epochs do we need to validate to have validated `BestTargetHead`. does not include null blocks)
```
sternhenri: I'm not sure I understand the BlockGap purpose here. The idea being stop doing any progress work and just do validation if you fall behind? If so makes sense. In which case it should be a multiple of expectedNumberOfBlocksPerRound (and in that sense a proxy for epochGap in some way)

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,490 @@
...
+- `ChainSync` gathers a list of `TargetHeads` from `BlockPubsub`, sorted by likelihood of being the best chain (see below).
+- `ChainSync` makes requests for chains of `BlockHeaders` to close gaps between  `TargetHeads`
+- `ChainSync` forms partial unvalidated chains of `BlockHeaders`, from those received via `BlockPubsub`, and those requested via `Graphsync`.
+- `ChainSync` attempts to form fully connected chains of `BlockHeaders`, parting from `StateTree`, toward observed `Heads`
+- `ChainSync` minimizes resource expenditures to fetch and validate blocks, to protect against DOS attack vectors.
+  `ChainSync` employs **Progressive Block Validation**, validating different facets at different stages of syncing.
+- `ChainSync` delays syncing `Messages` until they are needed. Much of the structure of the partial chains can
+  be checked and used to make syncing decisions without fetching the `Messages`.
+
+## Progressive Block Validation
```
sternhenri: I think this section should be split out into its own md. We've had multiple issues in the past with validation being defined in a few spots. Given how key a topic it is, relevance to consensus, etc. I think it probably warrants a standalone block val section that everything else links to.

I should be touching it up wrt consensus as well.

sternhenri: Specifically we should have a canonical way of specifying (ie code) how you run various validations (eg ChainEpoch) and having a standalone file in which to link to the code would be great.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,490 @@
...
+
+Notes:
+- in `CHAIN_CATCHUP`, if a node is receiving/fetching hundreds/thousands of `BlockHeaders`, validating signatures can be very expensive, and can be deferred in favor of other validation. (ie lots of BlockHeaders coming in through network pipe, dont want to bound on sig verification, other checks can help dump blocks on the floor faster (BV0, BV1)
+- in `CHAIN_FOLLOW`, we're not receiving thousands, we're receiving maybe a dozen or 2 dozen packets in a few seconds. We receive cid w/ Sig and addr first (ideally fits in 1 packet), and can afford to (a) check if we already have the cid (if so done, cheap), or (b) if not, check if sig is correct before fetching header (expensive computation, but checking 1 sig is way faster than checking a ton). In practice likely that which one to do is dependent on miner tradeoffs. we'll recommend something but let miners decide, because one strat or the other may be much more effective depending on their hardware, on their bandwidth limitations, or their propensity to getting DOSed
+
```
sternhenri: We need a note on GC in chainSync. Specifically with the fact that top-X heaviest heads' blocks and graphs should be kept around until past the finality threshold. Others can be discarded, all can be discarded at finality (SHOULD)

sternhenri: Also, having a set of diagrams here would be super helpful to illustrate how this all works. We should have a toy example at the top or bottom of this index file to disambiguate intent and show off a lot of things (with code remaining canon)

src/systems/filecoin_blockchain/chainsync/_index.md
sternhenri: spotted below in `block.id` file. Should link to there!

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,58 +1,490 @@
...
+# Terms and Concepts
 
-- Called by: libp2p {{<sref chainsync>}}
-- Calls on: {{<sref chain_manager>}}
+- `LastCheckpoint` the last hard social-consensus oriented checkpoint that `ChainSync` is aware of.
+  This consensus checkpoint defines the minimum finality, and a minimum of history to build on.
+  `ChainSync` takes `LastCheckpoint` on faith, and builds on it, never switching away from its history.
+- `TargetHeads` a list of `BlockCIDs` that represent blocks at the fringe of block production.
+  These are the newest and best blocks `ChainSync` knows about. They are "target" heads because
+  `ChainSync` will try to sync to them. This list is sorted by "likelihood of being the best chain".
```
sternhenri: Explicitly state that this is in terms of chainWeight, as calculated thanks to the StoragePowerConsensus subsystem

src/systems/filecoin_blockchain/chainsync/_index.md
ZenGround0: As an implementer I think it is really critical to keep this the sequencing description in the syncing section.  Sequencing is a key thing that up to now has been woefully underspecified. 
 Validation order matters and is non-trivially sequenced along with fetching, propagating and state updating.

I'm not sure how much of the above suggestion is about spec organization so apologies if I'm misunderstanding.  For example I'm totally happy if Progressive Block Valdation continues to exist with a full description of the sequence but points into those functions in the section you're proposing.

jbenet:
@sternhenri 

> let's make sure to split validation out of graphsync altogether. It's one of these parts of the spec that is bound to get defined 8 times in slightly different ways so we need a canonical "block validation" spec (I'll merge that from consensus into it thereafter).

Maybe -- i'll keep it here for now and reconsider later. most of block validity only makes sense in terms of a chain. some functions can be defined in `block` package, some functions can be defined in `blockchain`, but most of it makes sense to be split into stages in chainsync.

> - I think your requirement on block sync that all chains be validated before we move from catch up to follow is overly strong. We can weaken it at no loss in correctness.

No, cannot. without validating all the way to the best head, can fall into the wrong fork. consider two `TargetHeads` which fork `FinalityTipset + 10,000` epochs back. One appears clearly better than the other. Suppose the "better-looking" one is also invalid, and the product of an attacker, which can only be detected upon validation of intermediate nodes. If the node syncs towards that one, and moves to `CHAIN_FOLLOW` (ie starts advancing `FinalityTipset`) before validating the whole chain and finding the error, the node will get stuck once the error is found, as it cannot switch to the other, correct head.

src/systems/filecoin_blockchain/chainsync/_index.md
jbenet: > "Gaps between `ChainSync.FinalityTipset ... ChainSync.BestTargetHead` have been closed:"

this means only the chain to `BestTargetHead`, so i think it covers your concern.

note that if that one turns out to be invalid, need to keep validating towards the next best one, before moving to `CHAIN_FOLLOW`

src/systems/filecoin_blockchain/chainsync/_index.md
jbenet: > -  The idea being stop doing any progress work and just do validation if you fall behind? 

yes.

- and: the difference between `BlockGap` and `EpochGap` is null blocks. we can tolerate more of those than real blocks gap.

src/systems/filecoin_blockchain/chainsync/_index.md
jbenet: > I think this section should be split out into its own md. We've had multiple issues in the past with validation being defined in a few spots. Given how key a topic it is, relevance to consensus, etc. I think it probably warrants a standalone block val section that everything else links to.

I'll probably put the functions themselves into chain package (block validation only makes sense in terms of a chain), but this section will remain here, as these stages are key to understand in chainsync, and _only_ needed split like this here. I wont put all the specifics of how to validate each thing here -- that belongs in the canonical section. This section here can point to the code + functions in the canonical section.

> My bad confused with previous section

with which section? i agreed with your (now crossed out) comment

src/systems/filecoin_blockchain/chainsync/_index.md
jbenet: > We need a note on GC in chainSync. Specifically with the fact that top-X heaviest heads' blocks and graphs should be kept around until past the finality threshold. Others can be discarded, all can be discarded at finality (SHOULD)

yeah great point

> Also, having a set of diagrams here would be super helpful to illustrate how this all works. We should have a toy example at the top or bottom of this index file to disambiguate intent and show off a lot of things (with code remaining canon)

what kind of diagrams would you like to see? (rough sketch for me, i'll make them later)

jbenet: addressed (cache)

src/systems/filecoin_blockchain/chainsync/_index.md
jbenet: fine to use `ChainWeight` now, but I think this will end up being more than `ChainWeight` -- for example there's some calculation about "likelihood of being correct" -- eg if the `ChainWeight` looks outlandish, or ChainEpoch seems off by a small factor (ie, may be right, but seems a bit off), or `MinerAddress` belongs to a miner which has made a number of mistakes/attacks

jbenet: addressed

jbenet:
Rebased

jbenet:
@raulk 

> re: general-use peer exchange protocol, would something like this work for you, @jbenet? libp2p/specs#222

Yep, it does! that would be great

src/systems/filecoin_blockchain/chainsync/_index.md
sternhenri: Yes. You're completely right. I focused on the "all blocks" but it is appropriately scoped. Thank you

src/systems/filecoin_blockchain/chainsync/_index.md
sternhenri: Fair point

src/systems/filecoin_blockchain/chainsync/_index.md
sternhenri: As long as there is a single canonical detailed section dedicated to how block Val is done, I am happy.

src/systems/filecoin_blockchain/chainsync/_index.md
sternhenri: The diagrams that we made when first discussing this in LA. Showing a toy example with say 10 blocks all coming in, then being worked through the pipeline into fringe and then into longest target head. 

So some set of colored blocks being validated and classified over multiple steps allowing us to see the sequence of how "sync" should work.

sternhenri:
We're on the same page re chain sync. 

On the block validation section split. I remain a bit worried that the meat of how validation happens and what it should cover is useful enough across various parts of the protocol that I'd want it split out and have chain sync point to that... But let's merge and we'll amend if it indeed becomes a problem.

PR #569, opened 2019-10-15
Deals onchain v2 with new state transitions

zixuanzh:


zixuanzh:
This PR specifies on chain storage deal flow and behavior. There are five states when a `Sector` is on chain.
- `Committed`: The Sector has just been committed through CommitSector and it does not account for power at this point.
- `Recovering`: After a Sector is marked as Faulted, miners can call StorageMinerActor.RecoverFaults to recover previously Faulted sectors. However, power is only gained when a Sector becomes Active
- `Active`: After completing the first PoSt successfully, Active sectors can come from both Committed and Recovering sectors. 
- `Faulted`: Faulted sectors are marked by either DeclareFaults or CronActor for missing a PoSt. Power will be immediately suspended and some StorageDealCollateral will be lost.
- `Cleared`: Cleared sectors no longer exist in the network, they are either expired or removed because of an extended fault. A sector being at fault for three consecutive proving period will be cleared by the protocol. In this version, a sector expires when the last deal in the sector has expired.

Some of the graphics that @jbenet made will help explain the deal flow better. It is likely that there will be more changes and improvement. The following items are not included in this PR apart from the inline comments but the overall flow and logic are there.
- [ ] Actions triggered by CronActor
- [ ] Proper on chain calls and actor code
- [ ] Complete flow description in markdown

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -1,21 +1,309 @@
...
+		// TODO: proper failure
+		panic("TODO")
+	}
+
+	// The proof is verified, process ProvingSet.GetOnes():
+	// ProvingSet.GetOnes() contains SectorCommitted, SectorActive, SectorRecovering
+	// ProvingSet itself does not store states, states are all stored in SectorStates
+	var powerUpdate uint64 = 0
+
+	for _, sectorNo := range sm.ProvingSet_.GetOnes() {
```
nicola: this sounds very expensive, the previous code did not need to loop through every sector of a miner, but only through the expired ones and the faulty ones..

src/systems/filecoin_mining/sector/sectorset.go
```diff
@@ -0,0 +1,21 @@
+package sector
+
+func (css *CompactSectorSet) GetOnes() []SectorNumber {
```
jbenet: call it `SectorsInSet` or `SectorsOn`

src/systems/filecoin_mining/sector/sectorset.go
```diff
@@ -0,0 +1,21 @@
...
+	var sectorNo []SectorNumber
+	return sectorNo
+}
+
+func (css *CompactSectorSet) GetZeros() []SectorNumber {
```
jbenet: call it `SectorsNotInSet` or `SectorsOff`

src/systems/filecoin_mining/storage_mining/sector_state_enum.go
```diff
@@ -0,0 +1,62 @@
...
+// SectorState is an enum of
+// SectorCommitted, SectorRecovering, SectorActive, SectorFailing, SectorCleared
+// FaultCount is only relevant for SectorRecovering and SectorFailing.
+// FaultCount is relevant in accounting for the number
+// of consecutive proving periods that a sector is Failing.
+// Sectors that are in Failing for more than CONSECUTIVE_FAULT_COUNT_LIMIT
+// in a row will result in Sectors getting cleared and miners penalized.
+// The enum is written in this awkward way because of golang limitation.
+
+const MAX_CONSECUTIVE_FAULT_COUNT uint8 = 3
```
jbenet:
```go
type FaultCount uint8
const MAX_CONSECUTIVE_FAULTS = FaultCount(3)
```

src/systems/filecoin_mining/storage_mining/sector_state_enum.go
```diff
@@ -0,0 +1,62 @@
...
+
+const (
+	SectorCommittedStateNo  uint8 = 0
+	SectorRecoveringStateNo uint8 = 1
+	SectorActiveStateNo     uint8 = 2
+	SectorFailingStateNo    uint8 = 3
+	// SectorClearedStateNo    uint8 = 4 // deleted from sm
```
jbenet: Still have it as a state, and make this state 0 (the null state). 

```go
const (
	SectorClearedStateNo    uint8 = 0
	SectorCommittedStateNo  uint8 = 1
	SectorActiveStateNo     uint8 = 2
	SectorRecoveringStateNo uint8 = 3
	SectorFailingStateNo    uint8 = 4
)
```

(also reordered a couple)

src/systems/filecoin_mining/storage_mining/sector_state_enum.go
```diff
@@ -0,0 +1,62 @@
...
+
+type SectorState struct {
+	StateNumber uint8
+	FaultCount  uint8
+}
+
+func SectorCommitted() SectorState {
+	return SectorState{
+		StateNumber: SectorCommittedStateNo,
+		FaultCount:  0,
```
jbenet: add comment: 
```go
		FaultCount:  0, // always zero for SectorCommitted
```

src/systems/filecoin_mining/storage_mining/sector_state_enum.go
```diff
@@ -0,0 +1,62 @@
...
+	return SectorState{
+		StateNumber: SectorRecoveringStateNo,
+		FaultCount:  count,
+	}
+}
+
+func SectorActive() SectorState {
+	return SectorState{
+		StateNumber: SectorActiveStateNo,
+		FaultCount:  0,
```
jbenet: add comment: 
```go
		FaultCount:  0, // always zero for SectorActive
```

src/systems/filecoin_mining/storage_mining/sector_state_enum.go
```diff
@@ -0,0 +1,62 @@
...
+func SectorFailing(count uint8) SectorState {
+	return SectorState{
+		StateNumber: SectorFailingStateNo,
+		FaultCount:  count,
+	}
+}
+
+// SectorCleared is not directly represented in the spec
+// when a Sector is cleared, it is deleted from sm
+// func SectorCleared() SectorState {
```
jbenet: bring back. useful to have defined, even if not used in the map

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -1,21 +1,355 @@
...
-import base_mining "github.com/filecoin-project/specs/systems/filecoin_mining"
 import sector "github.com/filecoin-project/specs/systems/filecoin_mining/sector"
-import sealing "github.com/filecoin-project/specs/systems/filecoin_mining/sector"
+import block "github.com/filecoin-project/specs/systems/filecoin_blockchain/struct/block"
+import poster "github.com/filecoin-project/specs/systems/filecoin_mining/storage_proving/poster"
+
+// If a Post is missed (either due to faults being not declared on time or
+// because the miner run out of time, every sector is reported as failing
+// for the current proving period.
+func (sm *StorageMinerActor_I) CronAction() {
```
jbenet: `StorageMinerActor` is not called directly by `CronActor`. it is called by `StoragePowerActor`:

```
CronActor.EpochTick(vmr.Runtime)
|__ StoragePowerActor.EpochTick(vmr.Runtime)
    |__ StorageMnerActor.CheckPostSubmission()
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -1,21 +1,355 @@
...
+	// TODO: detemine how much pledge collateral to slash
+	// TODO: SendMessage(spa.SlashPledgeCollateral)
+	var powerUpdate uint64 = 0
+
+	for sectorNo, _ := range sm.Sectors() {
+		state := sm.SectorStates()[sectorNo]
+		switch state.StateNumber {
+		case SectorCommittedStateNo:
```
jbenet: this switch is great 👍

jbenet: i think this switch should move into a function `sm.failSectors([]sectorNo)` and this function should just do this:

```go
func (sm *StorageMinerActor_I) CheckPostSubmission() {
  if (post was not submitted) {
    nums := getSectorNums(sm.Sectors())
    sm.failSectors(nums)
  }
}
```

the other case (DeclareFaults) would just do this:

```go
func (sm *StorageMinerActor_I) DeclareFaults(faultSet sector.CompactSectorSet) {
  ...
  sectors := faultSet.SectorsInSet()
  sm.failSectors(sectors)
  ...
}
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -1,21 +1,355 @@
...
+
+// move Sector from Committed/Recovering into Active State
+// reset FaultCount to zero
+func (sm *StorageMinerActor_I) activateSector(sectorNo sector.SectorNumber) {
+	sm.SectorStates()[sectorNo] = SectorActive()
+}
+
+// move Sector from Active/Committed/Recovering into Failing State
+// and increment FaultCount
+func (sm *StorageMinerActor_I) failSector(sectorNo sector.SectorNumber) {
```
jbenet: i think the big fault switch should be here, and this should take a list:

```go
func (sm *StorageMinerActor_I) failSector(sectors []sector.SectorNumber) {
    var powerUpdate = StoragePower(0)
    for sectorNo, sector := range sectors {
        switch state.StateNumber {
        case SectorCommittedStateNo:
        ... // big switch here
        }
    }
    sma.UpdatePower(minerID, power)
}
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -1,21 +1,355 @@
...
+	newFaultCount := sm.SectorStates()[sectorNo].FaultCount + 1
+	sm.ProvingSet_.Remove(sectorNo)
+	sm.SectorStates()[sectorNo] = SectorFailing(newFaultCount)
+}
+
+// increment FaultCount and check if greater than MaxFaultCount
+// move from Failing to Cleared State if so
+func (sm *StorageMinerActor_I) checkMFC(sectorNo sector.SectorNumber) {
```
jbenet: `MFC` in this function call should be explicit `checkMaxFaultCount`.

Also, `checkX` should not mutate state (here it does). Maybe call it `incrementFaultCounter(n sector.SectorNumber)`

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -1,21 +1,355 @@
...
+//     - State Transition
+//       - Failing / Recovering / Active / Committed -> Cleared
+//     - Remove SectorNumber from Sectors, SectorStates, ProvingSet
+func (sm *StorageMinerActor_I) SubmitPoSt(postSubmission poster.PoStSubmission) {
+	// Verify correct PoSt Submission
+	isPoStVerified := sm.verifyPoStSubmission(postSubmission)
+	if !isPoStVerified {
+		// no state transition, just error out and miner should submitPoSt again
+		// TODO: proper failure
+		panic("TODO")
```
jbenet: yep, correct. should note here (as a comment):

```go
// miners get to try again. we do not need to address the failure case here
// it will be addressed via a `CronActor` calling `StoragePowerActor`
```
or something like this

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jbenet: yeah, let us think about it some more

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -1,21 +1,355 @@
...
+	// otherwise increment FaultCount in SectorStates()
+	for _, sectorNo := range sm.ProvingSet_.GetZeros() {
+		state := sm.SectorStates()[sectorNo]
+		switch state.StateNumber {
+		case SectorFailingStateNo:
+			sm.checkMFC(sectorNo)
+		default:
+			// TODO: proper failure
+			panic("Invalid sector state in ProvingSet.GetZeros")
+		}
```
jbenet: this switch needs to not fail in the other cases. 

I think either this
```go
switch state.StateNumber {
case SectorFailingStateNo:
    sm.checkMFC(sectorNo)
}
```

or this

```go
switch state.StateNumber {
case SectorFailingStateNo:
    sm.checkMFC(sectorNo)
case SectorClearedStateNo:     // nothing
case SectorCommittedStateNo:   // nothing
case SectorActiveStateNo:      // nothing
case SectorRecoveringStateNo:  // nothing
default: 
    panic("Invalid sector state in ProvingSet.GetZero")
}
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -1,21 +1,355 @@
...
+	}
+
```
jbenet: I would add a note here (as a comment) that expiration needs to happen in the same flow, _after_ states change to Failing or Active (ie Committed and Recovering pass through those before expiring)

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -1,21 +1,355 @@
...
+			// SendMessage(sma.SettleExpiredDeals(sc.DealIDs()))
+			sm.clearSector(expiredSectorNo)
+		default:
+			// TODO: proper failure
+			panic("Invalid sector state in SectorExpirationQueue")
+		}
+
+		expirationPeek = sm.SectorExpirationQueue().Peek().Expiration()
+
+	}
```
jbenet: this pattern for the loop is a bit error prone:
```go
foo := expr
for foo <= val {
    // do a lot of stuff
    foo = expr
}
```
instead, use:

```go
queue := expirationPeek := sm.SectorExpirationQueue()
for peek := queue.Peek(); peek.Expiration() <= currEpoch; peek = queue.Peek() {
    // do lots of stuff
}
```

src/systems/filecoin_markets/deal/_index.md
```diff
@@ -3,6 +3,39 @@ menuTitle: Deals
...
+There are two types of deals in Filecoin markets, storage deals and retrieval deals. Storage deals are recorded on the blockchain and enforced by the protocol. Retrieval deals are off chain and enabled by micropayment channel by transacting parties. All deal negotiation happen off chain and a request-response style storage deal protocol is in place to submit agreed-upon storage deals onto the network with `CommitSector` to gain storage power on chain. Hence, there is a `StorageDealProposal` and a `RetrievalDealProposal` that are half-signed contracts submitted by clients to be counter-signed and posted on-chain by the miners.
+
+Filecoin Storage Market Deal Flow
+
+====== Add Storage Deal and Power ======
```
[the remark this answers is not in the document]
jbenet: addressed in 435ab9f

src/systems/filecoin_markets/deal/_index.md
```diff
@@ -3,6 +3,39 @@ menuTitle: Deals
...
+
+1.`StorageClient` and `StorageProvider` call `StorageMarketActor.AddBalance` to deposit funds into Storage Market. There are two fund states in the Storage Market, `Locked` and `Available`.
+a. `StorageClient` and `StorageProvider` can call `WithdrawBalance` before any deal is made. (move to state X)
+2. `StorageClient` and `StorageProvider` negotiate a deal off chain. `StorageClient` sends a `StorageDealProposal` to a `StorageProvider`.
+a. `StorageProvider` verifies the `StorageDeal` by checking address and signature of `StorageClient`, checking the proposal has not expired, checking `StorageClient` did not call withdraw in the last X Epoch, checking both `StorageProvider` and `StorageClient` have sufficient available balances in `StorageMarketActor`.
```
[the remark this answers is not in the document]
jbenet: addressed in 435ab9f

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jbenet: addressed in 435ab9f

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jbenet: addressed in 435ab9f

jbenet:
Rebased on top of master after #566 

@porcuquine could you look at d38df8fc911b707cebe366e4c0deb4095e232980 ? -- it is based on changes @nicola and I made in a9ab8d7 (using epoch instead of randomess to verify on chain)

Questions
- [ ] is epoch to verify ok or do we need to put the randomness?
- [ ] w/ UnsealedCID gone from `OnChainVerifyInfo`, it can also get removed from `SectorCommitment` right?

jbenet:
# SectorState FSM

image

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -264,24 +267,18 @@ func (sm *StorageMinerActor_I) DeclareFaults(faultSet sector.CompactSectorSet) {
...
+	for _, sectorNo := range faultSet.SectorsOn() {
 		state := sm.SectorStates()[sectorNo]
-		// sc := sm.Sectors()[sectorNo]
 
 		switch state.StateNumber {
-		case SectorActiveStateNo:
-			// SlashStorageDealCollateral
-			// SendMessage(sma.slashStorageDealCollateral(sc.DealIDs()))
-			sm.failSector(sectorNo)
+		case SectorActiveSN:
```
zixuanzh: Some storage deal collateral should be slashed here or no?

src/systems/filecoin_markets/deal/_index.md
```diff
@@ -3,6 +3,52 @@ menuTitle: Deals
...
+There are two types of deals in Filecoin markets, storage deals and retrieval deals. Storage deals are recorded on the blockchain and enforced by the protocol. Retrieval deals are off chain and enabled by micropayment channel by transacting parties. All deal negotiation happen off chain and a request-response style storage deal protocol is in place to submit agreed-upon storage deals onto the network with `CommitSector` to gain storage power on chain. Hence, there is a `StorageDealProposal` and a `RetrievalDealProposal` that are half-signed contracts submitted by clients to be counter-signed and posted on-chain by the miners.
+
+Filecoin Storage Market Deal Flow
+
+# Add Storage Deal and Power
+
+- 1. `StorageClient` and `StorageProvider` call `StorageMarketActor.AddBalance` to deposit funds into Storage Market. There are two fund states in the Storage Market, `Locked` and `Available`.
```
anorth: The funds serve different purposes for the two parties. Could you clarify that those purposes are escrow and deal collateral? `StorageDealCollateral` is used below without definition. Can single actor be both roles at once (for different deals, for the same deal even)? Do the two reasons for locking funds need to be tracked separately?

src/systems/filecoin_markets/deal/_index.md
```diff
@@ -3,6 +3,52 @@ menuTitle: Deals
...
+    - `StorageClient` and `StorageProvider` can call `WithdrawBalance` before any deal is made. (move to state X)
```
anorth: The should be able to call `WithdrawBalance` at any time, regardless of deals, and withdraw any unlocked funds.

src/systems/filecoin_markets/deal/_index.md
```diff
@@ -3,6 +3,52 @@ menuTitle: Deals
...
+- 7. Sectors that are declared faulty for `storagemining.MaxFaults` consecutive ChainEpochs will result in `StoragePowerActor.SlashPledgeCollateral`.
+  - TODO: set `X` parameter
+
+# Submit PoSt
+
+(TODO: move into Storage Mining)
+
+On every PoSt Submission, the following steps happen.
+
+- 8. `StorageMinerActor` first verifies the PoSt Submission. All Sectors will be considered in `SpottedFaults` if PoSt submission has failed (move to State 14).
```
anorth: I think this is confused:
- if an attempted PoSt submission is rejected, nothing should happen - the miner still has time to submit a correct one
- "failure" occurs when there is no submission after deadline, so you can't talk about checking it in response to the submission that didn't happen. that has to be in terms of cron.

src/systems/filecoin_markets/deal/_index.md
```diff
@@ -3,6 +3,52 @@ menuTitle: Deals
...
+- 9. If `CommittedSectors` are proven in `PoStSubmission.SectorSet`, Storage Miner gains power for these newly committed sectors.
+- 10. If there are `DeclaredFaultySectors` , `Sector` in that set will not be challenged.
+- 11. For all other sectors, payment will be processed by invoking `StorageMarketActor.ProcessStorageDealsPayment` and miner available balances will be updated.
+- 12. Decide which Sectors have expired by looking at the `SectorExpirationQueue`. Sectors expire when all deals in that Sector have expired. `StorageDealCollateral` for both miners and users will only be returned when all deals in the Sector have expired. This is done by calling `StorageMarketActor.SettleExpiredDeals` and the Sector will be deleted from `StorageMinerActor.Sectors`.
```
anorth: > Sectors expire when all deals in that Sector have expired.

This is an unexpected change to me. IMO miners should explicitly remove sectors when they no longer want to claim power for them, not have them automatically lose power. Collateral can remain pinned until that explicit action if necessary. I don't think we need a concept of sector expiration at all, unless it refers to the explicit action taken by a miner (but then "done" would be a better word).

If there's an explicit protocol decision to have storage power expire automatically, that should be called out loudly. It implies a lot more re-sealing work for miners who want to maintain power but don't have new, paying client deals in enough quantity to fill their storage. And a difficult decision when self-dealing to determine how long to set the deal before paying the cost of re-sealing. One might hope for *short* self-deals so that storage can be quickly repurposed to client deals, but this mechanism incentivises very long self deals, which might exclude unknown future client deals. @jbenet

src/systems/filecoin_markets/deal/_index.md
```diff
@@ -3,6 +3,52 @@ menuTitle: Deals
...
+
+# Spot Faults
+
+(TODO: move into Storage Mining)
+
+- 13. `CronActor` calls `StoragePowerActor.EpochTick` at every block. This calls `StorageMinerActor.CheckPoSt` on all the miners whose `ProvingPeriod` is up.
+  - If no PoSt is submitted by the end of the `ProvingPeriod`, `StorageMinerActor` spots the missing PoSt, and sets all sectors to `Failing`.
+    - TODO: reword in terms of a conditional in the mining cycle
+  - When there are sector faults are spotted, both `StorageDealCollateral` and `PledgeCollateral` are slashed, and power is lost.
```
anorth: As stated, this sounds like a draconian penalty for a single unavoidable failure (e.g. of network or power). I hope there are some missing qualifiers about how much is penalised.

src/systems/filecoin_markets/storage_market/storage_market_actor.id
```diff
@@ -48,15 +49,17 @@ type StorageMarketActor struct {
...
     // call by CronActor / onPoStSubmission on ActiveSet to process deal payment
     // go through StorageDealIDs, if IDs are active in MarketActor
     // payment will be processed
-    handleStorageDealPayment(storageDealIDs [deal.DealID])
+    ProcessStorageDealsPayment(dealIDs [deal.DealID])
 
-    // call by CronActor / onPoStSubmission on FaultSet to slash deal collateral
-    slashStorageDealCollateral(storageDealIDs [deal.DealID])
+    // call by CronActor / on DeclareFault on FaultSet to slash deal collateral
+    // Deals should be slashed for a single proving period
```
anorth: > Deals should be slashed for a single proving period

What does this mean?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
anorth: Should it not return an error code, pay gas fees etc? I don't think actor code should have a panic anywhere.

PR #571, opened 2019-10-17
Remove CommD from SealCommitment

zixuanzh:
As discussed, CommD can be computed on the fly and they won't be stored on chain. PieceRef and CommP will be needed to generate CommD and verify seal proof. StorageClient generates a CommP for data in a StorageDealProposal and StorageProvider needs to verify that CommP before accepting and signing the proposal. We still need alignment on the following:

- [x] Decide if we need to store PieceRef on chain @nicola 
- [ ] Decide if we need to store (sector) Expiration in SealCommitment, Expiration is already stored on chain once in `sm.SectorExpirationQueue` and we are storing it a second time in `sm.Sectors`.
- [x] Define input type and interfaces for `sm.verifySeal` @porcuquine


src/systems/filecoin_markets/deal/deal.id
```diff
@@ -8,17 +8,25 @@ type DealCID ipld.CID
...
-    Provider            addr.Address
-    ProposalExpiration  block.ChainEpoch
-    DealExpiration      block.ChainEpoch
-    StoragePrice        actor.TokenAmount
-    StorageCollateral   actor.TokenAmount
-    ProposerSignature   Signature
+    // TODO: Decide if PieceRef needs to go on chain
+    // PieceRef is the hash of the data in native structure. This will be used for
+    // certifying the data transfer. Reference for transit.
+    PieceRef                DealCID
```
anorth: I *think* the outcome of https://gist.github.com/jbenet/8af91b75091f05a1c92a774f9d9a56c8 is that only CommP needs to go on-chain. Words are tricky here (I don't know what "native" means), but I don't think the CID of the IPLD graph closest to the raw data the client wants to store needs to go on chain. CommP is implicitly coupled to it anyway, since the CommP is calculated by serializing the graph (into a CAR file) then creating the merkle tree from those byptes.

anorth: A root CID of the not-a-merkle-tree IPLD graph may be transmitted to the miner as part of deal negotiation, but not inside this structure.

anorth: cc @hannahhoward @jbenet

porcuquine: Agreed — those decisions are more recent and supplant what appears here. @jbenet will be cleaning these data structures up soon.

jbenet: This needs to be:
```go
    PieceCID piece.PieceCID // CommP`
```

jbenet: This one needs to be removed -- not needed. (we cannot fit it on chain)

src/systems/filecoin_markets/deal/deal.id
```diff
@@ -8,17 +8,25 @@ type DealCID ipld.CID
...
+    // The data hashed in a form that is compatible with the proofs system.
+    // Reference for actual storage in a sector.
+    CommP                   ipld.CID  // 35 bytes
```
jbenet: Rename to `PieceCID`:
```go
    PieceCID piece.PieceCID // CommP`
```

Rename also: `piece.PieceID -> piece.PieceCID`

zixuanzh:
Merging for velocity since CommD is now removed. Recognized that there are discrepancies in `StorageMarket` because I was previously unclear on the data types. Will go through this part in the next day or two and reach convergence with a separate PR.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -293,7 +294,27 @@ func (sm *StorageMinerActor_I) verifySeal(onChainInfo sector.OnChainSealVerifyIn
...
+	var pieceInfos []sector.PieceInfo // = make([]sector.PieceInfo, 0)
+
+	for dealId := range onChainInfo.DealIDs() {
+		// FIXME: Actually get the deal info from the storage market actor and use it to create a sector.PieceInfo.
+		_ = dealId
+
+		pieceInfos = append(pieceInfos, nil)
+	}
+
+	new(proving.StorageProvingSubsystem_I).VerifySeal(&sector.SealVerifyInfo_I{
```
jbenet: The actor wont have access to `proving.StorageProvingSubsystem_I` -- we should be calling something deeper, maybe `FilecoinProofsSubsystem.VerifySeal` (btw if that is a pure function, maybe it should be a static function (not bound to `FilecoinProofsSubsystem` type)

jbenet: since `filproofs` is a library, it doesn't need the `FilecoinProofsSubsystem` type. those functions can just be defined on the package directly.  (if `.id` doesn't allow, it can do it in the go file directly)

anorth: I am expecting proofs methods to be exposed by the VMContext or similar, even if that just proxies directly to proofs implementations. I'm hoping actor code won't import *anything* outside of their own packages.

porcuquine: Just a note that I simply needed to plumb this somehow and knew that VM details were not nailed down yet. Don't read anything into it, and I agree that there should be a context object serving as the single-point-of-contact between actors and the external world, including `filproofs`. If I missed the point, please clarify so I can best understand how to do these things in the future.

PR #585, opened 2019-10-22
Add power accounting for collateral and slashing

zixuanzh:
This PR implements and connects StorageMinerActor and StoragePowerActor. Concepts of PowerReport and PowerTable are introduced to account for miner power and collateral requirement based on changes in their sector states.

There are three possible states for a unit of Power. It can be Active, Inactive, or None. Active power is derived from having sectors in the Active state. Inactive power corresponds to sectors in Committed, Recovering and Failing states. Sectors enter the Cleared state either through exceeding maximum consecutive fault count (Terminated) or expiration. StorageMinerActor keeps track of a SectorTable that records the number of Sectors in each state. At DeclareFaults, SubmitPoSt, and CheckPoStSubmissionHappened, a PowerReport is sent to StoragePowerActor to update power and perform slashing. A PowerReport contains up-to-date counts of sectors in each Sector state (with one exception) and new slash count for each of the three fault scenarios (DeclaredFaults, DetectedFaults, and TerminatedFaults).

Note that consensus fault slashing is not yet included and exact punishment for each of the above. This PR sets up the structure to perform power and collateral accounting and facilitate slashing of collateral.


jbenet:
@zixuanzh -- spoke w/ @whyrusleeping and went through the sector state diagram. We talked through it all, and arrived at some more observations:
- we need to not apply state changes to proving set during the challenge period of the post -- this is because any of `CommitSector`, `RecoverSector` and `DeclareFaults` during the challenge period would muddle/mess with the PoSt computation/submission (a bit complex).
- we *could* disallow or stage sector state changes (error prone, complex) during the challenge period.
- or simply best fix is to have a CronActor driven action that freezes the proving set at the beginning of the challenge period (ie at the epoch when the PoSt challenge is issued).
- this has the side effect of also giving us the counts we need for power amount updates (on successful submit post, the (frozen) proving set is now the power -- on unsuccessful, easy, everything failed) 
- + one more (big simplification for us) should only report/update power w/ SPA at end of proving period (not in between). this reduces a lot of complexity for us. (sorry, we may not need the counts after all!)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -1,43 +1,285 @@
...
-func (spa *StoragePowerActor_I) ReportConsensusFault(slasherAddr addr.Address, faultType ConsensusFaultType, proof []block.Block) {
-	panic("TODO")
+	// TODO: anything to check here?
+	newMiner := &PowerTableEntry_I{
+		ActivePower_:            block.StoragePower(0),
+		InactivePower_:          block.StoragePower(0),
+		AvailableBalance_:       actor.TokenAmount(0),
+		LockedPledgeCollateral_: actor.TokenAmount(0),
+	}
+	spa.PowerTable()[msgSender] = newMiner
```
anorth: `msgSender` doesn't seem like the right key here, it should be the create actor's address.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -1,43 +1,285 @@
...
+
+	powerEntry := spa.PowerTable()[msgSender]
+	pledgeCollateralRequired := spa.GetPledgeCollateralReq(powerEntry.ActivePower() + powerEntry.InactivePower())
+
+	if pledgeCollateralRequired < powerEntry.LockedPledgeCollateral() {
+		return true
+	}
+
+	if pledgeCollateralRequired < (powerEntry.LockedPledgeCollateral() + powerEntry.AvailableBalance()) {
+		spa.lockPledgeCollateral(msgSender, (pledgeCollateralRequired - powerEntry.LockedPledgeCollateral()))
```
anorth: `IsPledgeCollateralSatisfied` implied a pure function: name the method `LockPledgeCollateral` or `EnsurePledgeCollateral` or similar if it changes state.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -1,43 +1,285 @@
...
+	currEntry, found := spa.PowerTable()[address]
+	if !found {
+		// TODO: proper throw
+		panic("TODO")
+	}
+
+	if currEntry.Impl().LockedPledgeCollateral() < amount {
+		// TODO: proper throw
+		panic("TODO")
+	}
```
anorth: Probably best to slash it anyway!

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -56,21 +304,21 @@ func (spa *StoragePowerActor_I) Surprise(ticket block.Ticket) []addr.Address {
...
+	numSurprised := uint(len(spa.PowerTable())) / provingPeriod
 
-	// TODO: seem inefficient but spa.Miners() is now a map from address to power
-	minerAddresses := make([]addr.Address, len(spa.Miners()))
+	// TODO: seem inefficient but spa.PowerTable() is now a map from address to power
+	minerAddresses := make([]addr.Address, len(spa.PowerTable()))
 
 	index := 0
-	for address := range spa.Miners() {
+	for address, _ := range spa.PowerTable() {
```
jbenet: i think this is going to pick all of them -- iterating over all of the power table. you want:

```go
table := spa.PowerTable()
numMiners := len(table)
random := 4 // TODO: get randomness from chain
offset := random % numMiners
for i := 0; i < newSurprised; i++ {
  j := (offset + i) % numMiners 
  minerAddresses[i] = table[j]
}
```
Note somewhere:
- this picks `newSurprised` consecutive miners, starting from a random `offset`
- good for sampling w/o replacement, but may be bad as it would be the same cohort always challenged at the same time
- if we want to pick properly random set, can: 
  - make a list of indices first, shuffle it, then pick consecutive from those 
  - or pick at random every time, but skip ones we've seen already (sample w/ replacement, but skip doubles)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.id
```diff
@@ -1,86 +1,73 @@
...
+type PowerReport struct {
+    ActivePower    block.StoragePower  // set value
+    InactivePower  block.StoragePower  // set value
+}
+
+type FaultReport struct {
+    NewDeclaredFaults    util.UVarint  // diff value
+    NewDetectedFaults    util.UVarint  // diff value
+    NewTerminatedFaults  util.UVarint  // diff value
+}
```
jbenet: this turned out very well

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.id
```diff
@@ -1,86 +1,73 @@
...
-    miners {addr.Address: StorageMiner}
-    GetMinerPower(addr addr.Address) block.StoragePower
-    GetTotalPower() block.StoragePower
-    GetMinerPublicKey(addr addr.Address) filcrypto.PubKey
-    IncrementPower(addr addr.Address, numSectors UVarint)
-    SuspendPower(addr addr.Address, numSectors UVarint)
-    UnsuspendPower(addr addr.Address, numSectors UVarint)
-    RemovePower(addr addr.Address, numSectors UVarint)
-    RemoveAllPower(addr addr.Address, numSectors UVarint)
-}
```
jbenet: love removing code

src/systems/filecoin_mining/sector/sectorset.go
```diff
@@ -19,3 +19,8 @@ func (css *CompactSectorSet) Remove(sectorNo SectorNumber) CompactSectorSet {
 	var newCompactSectorSet CompactSectorSet
 	return newCompactSectorSet
 }
+
+func (css *CompactSectorSet) DeepCopy() CompactSectorSet {
+	var newCompactSectorSet CompactSectorSet
+	return newCompactSectorSet
```
jbenet: does this need a todo?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -3,19 +3,104 @@ package storage_mining
...
+
+func (st *SectorTable_I) InactivePower() block.StoragePower {
+	return block.StoragePower((st.CommittedSectors_ + st.RecoveringSectors_ + st.FailingSectors_) * util.UVarint(st.SectorSize_))
+}
+
+// called by CronActor to notify StorageMiner of PoSt Challenge
+func (sm *StorageMinerActor_I) NotifyOfPoStChallenge() {
+	if sm.isChallenged() {
+		// TODO: proper throw
+		panic("should not be challenged when isChallenged")
```
jbenet: yeah we need to address this case. it can happen

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -3,19 +3,104 @@ package storage_mining
...
+	}
+
+	sm.isChallenged_ = true
```
jbenet: i think better than a bool would be to say the epoch:

```go
sm.challengeEpoch_ = rt.CurrentEpoch()
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jbenet:
```go
if sm.inChallengePeriod() {
  return // silent return, dont re-challenge
}
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -3,19 +3,104 @@ package storage_mining
...
+
+	// TODO: commit state change
+
+}
 
 // If a Post is missed (either due to faults being not declared on time or
 // because the miner run out of time, every sector is reported as failing
 // for the current proving period.
 func (sm *StorageMinerActor_I) CheckPoStSubmissionHappened() {
+	if !sm.isChallenged() {
```
jbenet:
```go
if !sm.inChallengePeriod() {
  return // probably got challenged twice
}
```
and

```go
func (sm *StorageMinerActor_I) inChallengePeriod() bool {
  sinceLastChallenge := runtime.CurrentEpoch() - sm.challengeEpoch_
  return sinceLastChallenge < EPOCHS_CHALLENGE_PERIOD
}
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -40,67 +125,77 @@ func (sm *StorageMinerActor_I) verifyPoStSubmission(postSubmission poster.PoStSu
...
+	case SectorRecoveringSN:
+		sm.SectorTable().Impl().RecoveringSectors_ -= 1
+	case SectorFailingSN:
+
+	default:
+		// TODO: proper failure
+		panic("Invalid sector state in CronAction")
+	}
+
+	sm.SectorTable().Impl().FailingSectors_ += 1
```
jbenet: i believe this is double counting if it was already failing. best to make sure numbers are conserved by editing in the exact same spots:

```go
switch state.StateNumber {
case SectorActiveSN:
    sm.SectorTable().Impl().ActiveSectors_ -= 1
    sm.SectorTable().Impl().FailingSectors_ += 1
case SectorCommittedSN:
    sm.SectorTable().Impl().CommittedSectors_ -= 1
    sm.SectorTable().Impl().FailingSectors_ += 1
case SectorRecoveringSN:
    sm.SectorTable().Impl().RecovveringSectors_ -= 1
    sm.SectorTable().Impl().FailingSectors_ += 1
case SectorFailingSN:
    // no change to numbers
default:
    panic("Invalid sector state in failSector")
}
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
```diff
@@ -21,51 +21,76 @@ type SectorExpirationQueue struct {
     Remove(n sector.SectorNumber)
 }
 
+type SectorTable struct {
+    SectorSize         sector.SectorSize
+    ActiveSectors      util.UVarint
+    CommittedSectors   util.UVarint
+    RecoveringSectors  util.UVarint
+    FailingSectors     util.UVarint
+    TerminatedSectors  util.UVarint  // transient State that get reset on every constructPowerReport
```
jbenet: this is a bit annoying because it is too similar to an additional state, and that creates more complexity. maybe change it to:

```go
TerminationFaults util.UVarint
// or
TerminationFaultCount util.UVarint
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -40,67 +125,77 @@ func (sm *StorageMinerActor_I) verifyPoStSubmission(postSubmission poster.PoStSu
...
 
-// increment FaultCount and check if greater than MaxFaultCount
-// move from Failing to Cleared State if so
-func (sm *StorageMinerActor_I) incrementFaultCount(sectorNo sector.SectorNumber) {
-	newFaultCount := sm.SectorStates()[sectorNo].FaultCount + 1
 	if newFaultCount > MAX_CONSECUTIVE_FAULTS {
 		// TODO: heavy penalization: slash pledge collateral and delete sector
 		// TODO: SendMessage(SPA.SlashPledgeCollateral)
+		sm.SectorTable().Impl().TerminatedSectors_ += 1
```
jbenet: seems like a state. `.TerminationFaults_ += 1` much nicer

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -40,67 +125,77 @@ func (sm *StorageMinerActor_I) verifyPoStSubmission(postSubmission poster.PoStSu
...
+		sm.SectorTable().Impl().FailingSectors_ -= 1
```
jbenet: why not call this within `clearSector` in a state switch? keeps it in the same place. `sm.clearSector` is the state transition. same applies to all similar spots that change counts right before calling `sm.clearSector`

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -40,67 +125,77 @@ func (sm *StorageMinerActor_I) verifyPoStSubmission(postSubmission poster.PoStSu
...
 		sm.clearSector(sectorNo)
-	} else {
-		// increment FaultCount
-		// TODO: SendMessage(sma.SlashStorageDealCollateral)
-		sm.ProvingSet_.Remove(sectorNo)
-		sm.SectorStates()[sectorNo] = SectorFailing(newFaultCount)
 	}
+
+	sm.ProvingSet_.Remove(sectorNo)
+	sm.Sectors()[sectorNo].Impl().State_ = SectorFailing(newFaultCount)
```
jbenet: bug? should not be setting state to failing if we just cleared it. the `} else {` was probably right?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -119,13 +214,19 @@ func (sm *StorageMinerActor_I) incrementFaultCount(sectorNo sector.SectorNumber)
...
 //     - State Transition
 //       - Failing / Recovering / Active / Committed -> Cleared
-//     - Remove SectorNumber from Sectors, SectorStates, ProvingSet
+//     - Remove SectorNumber from Sectors, ProvingSet
 func (sm *StorageMinerActor_I) SubmitPoSt(postSubmission poster.PoStSubmission) {
+
+	if !sm.isChallenged() {
+		// TODO: proper throw
+		panic("cannot SubmitPoSt when sm not isChallenged")
+	}
```
jbenet: nice

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -162,43 +256,56 @@ func (sm *StorageMinerActor_I) SubmitPoSt(postSubmission poster.PoStSubmission)
...
 	// ProvingSet.SectorsOff() contains SectorFailing
 	// SectorRecovering is Proving and hence will not be in GetZeros()
 	// heavy penalty if Failing for more than or equal to MAX_CONSECUTIVE_FAULTS
-	// otherwise increment FaultCount in SectorStates()
+	// otherwise increment FaultCount in Sectors().State
 	for _, sectorNo := range sm.ProvingSet_.SectorsOff() {
-		state := sm.SectorStates()[sectorNo]
-		switch state.StateNumber {
+		state, found := sm.Sectors()[sectorNo]
+		if !found {
```
jbenet: note why -- `// sm.ProvingSet_.SectorsOff() includes both failing and some cleared sectors -- only want failing`

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -207,12 +314,15 @@ func (sm *StorageMinerActor_I) expireSectors() {
 
 			// Settle deals
 			// SendMessage(sma.SettleExpiredDeals(sc.DealIDs()))
+
+			sm.SectorTable().Impl().ActiveSectors_ -= 1
```
jbenet: yeah move into `sm.clearSector`?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -235,58 +348,68 @@ func (sm *StorageMinerActor_I) expireSectors() {
...
-// - Update SectorStates
+// - Update State in Sectors()
 // - Remove Active / Commited / Recovering from ProvingSet
 func (sm *StorageMinerActor_I) DeclareFaults(faultSet sector.CompactSectorSet) {
 
-	var powerUpdate uint64 = 0
+	if sm.isChallenged() {
+		// TODO: proper throw
+		panic("cannot DeclareFaults when sm isChallenged")
+	}
```
jbenet: nice!

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -308,12 +431,21 @@ func (sm *StorageMinerActor_I) checkIfSectorExists(sectorNo sector.SectorNumber)
 // Currently deals must be posted on chain via sma.PublishStorageDeals before CommitSector
 // TODO(optimization): CommitSector could contain a list of deals that are not published yet.
 func (sm *StorageMinerActor_I) CommitSector(onChainInfo sector.OnChainSealVerifyInfo) {
+
+	if sm.isChallenged() {
+		// TODO: proper throw
+		panic("cannot CommitSector when sm isChallenged")
+	}
```
jbenet: we should be able to commit sectors while in challenge period -- this will be very annoying operationally if not. does the frozen proving set work well, or no?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
zixuanzh: Will use `EnsurePledgeCollateralSatisfied` here because there is another function called `lockPledgeCollateral`

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -1,43 +1,291 @@
...
+}
+
+func (spa *StoragePowerActor_I) IsPledgeCollateralSatisfied() bool {
+	var msgSender addr.Address // TODO replace this
+
+	powerEntry := spa.PowerTable()[msgSender]
+	pledgeCollateralRequired := spa.GetPledgeCollateralReq(powerEntry.ActivePower() + powerEntry.InactivePower())
+
+	if pledgeCollateralRequired < powerEntry.LockedPledgeCollateral() {
+		return true
```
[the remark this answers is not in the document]
zixuanzh: leaning towards no, this function just checks to make sure that miner has the collateral required. To enable unlocking, perhaps we can have `UnlockPledgeCollateral` instead of `unlockPledgeCollateral`?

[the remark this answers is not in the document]
zixuanzh: Alternatively, this function does not change state and miner needs to call lock and unlock to change state for it

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -1,43 +1,291 @@
...
+
+	currEntry.Impl().LockedPledgeCollateral_ = currEntry.LockedPledgeCollateral() - amount
+	currEntry.Impl().AvailableBalance_ = currEntry.AvailableBalance() + amount
+	spa.PowerTable_[address] = currEntry
+
+	// TODO: commit state change
+}
+
+func (spa *StoragePowerActor_I) getDeclaredFaultSlash(util.UVarint) actor.TokenAmount {
+	return actor.TokenAmount(0)
```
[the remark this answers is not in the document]
zixuanzh: Potentially, using functions as placeholders for now.

src/systems/filecoin_mining/sector/sectorset.go
zixuanzh: All CompactSectorSet operations are TODOs for now, adding to all

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
zixuanzh: adding these as comments to Surprise, it should be its own PR. cc @nicola

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
zixuanzh: If a miner gets challenged and submits a successful PoSt, they will be out of the ChallengePeriod and sm.inChallengePeriod() should return false. If a miner gets challenged and did not submit PoSt successfully, they should still be in the challenge period. Hence, correct, if !sm.inChallengePeriod(), we should return here 👍.

zixuanzh: However, just using EpochTime is not enough, need to mark miner successfully submitted a PoSt.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
zixuanzh: It will be changed to everything is moved to Failing state before clearing.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
zixuanzh: cc @whyrusleeping

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
zixuanzh: Given the current set up, freezing ProvingSet with out freezing SectorTable and Sectors wont work because we need to know the SectorTable breakdown within the ProvingSet to account for power and collateral. We are freezing all three of them with isChallenged(). I can add a separate commit that allows for CommitSector while isChallenged() and keep those sectors as a buffer. That buffer will be added to the ProvingSet and SectorTable upon successful SubmitPoSt or MissedPoSt. We can decide if we want to include this later.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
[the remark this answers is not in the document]
jbenet: Fine for now, but it is weird to leave more than enough locked. lets not block PR and revisit later

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -14,14 +14,46 @@ func (st *SectorTable_I) InactivePower() block.StoragePower {
...
+// Call by either SubmitPoSt or OnMissedPoSt
+// TODO: verify this is correct and if we need to distinguish SubmitPoSt vs OnMissedPoSt
+func (cs *ChallengeStatus_I) OnChallengeResponse(currEpoch block.ChainEpoch) ChallengeStatus {
+	cs.LastResponseEpoch_ = currEpoch
+	return cs
+}
+
+func (cs *ChallengeStatus_I) isChallenged() bool {
+	// true (isChallenged) when LastChallengeEpoch is later than LastResponseEpoch
+	return cs.LastChallengeEpoch() > cs.LastResponseEpoch()
```
jbenet: small naming note, i'd expect from naming alone that `LastResponseEpoch` would only be set by SubitPost, but this needs to be set by the `onMissedPost()` flow as well. Maybe `LastChallengeEndEpoch` (sounds worse, but more clear?) or something like it

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -170,29 +220,35 @@ func (sm *StorageMinerActor_I) failSector(sectorNo sector.SectorNumber, incremen
...
 		// TODO: heavy penalization: slash pledge collateral and delete sector
 		// TODO: SendMessage(SPA.SlashPledgeCollateral)
-		sm.SectorTable().Impl().TerminatedSectors_ += 1
-		sm.SectorTable().Impl().FailingSectors_ -= 1
-		sm.clearSector(sectorNo)
-	}
 
-	sm.ProvingSet_.Remove(sectorNo)
-	sm.Sectors()[sectorNo].Impl().State_ = SectorFailing(newFaultCount)
+		sm.clearSector(sectorNo, true)
```
jbenet: given that this only happens _once_ (in one code point, here) and that `sm.SectorTable().Impl().TerminationFaultCount_` is not a sector state, i think we should:

- move back to `clearSector(sectorNum)` (not taking in `isTerminated`) and just add the fault count here:

```
sm.clearSector(sectorNo)
sm.SectorTable().Impl().TerminationFaultCount_++
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -14,14 +14,46 @@ func (st *SectorTable_I) InactivePower() block.StoragePower {
...
+	}
+
+	// TODO: get currEpoch from runtime
+	var currEpoch block.ChainEpoch
+	sm.ChallengeStatus().Impl().OnNewChallenge(currEpoch)
+
+	// TODO: commit state change
+}
+
+func (sm *StorageMinerActor_I) loadStagedCommittedSectors() {
```
jbenet: `load` is an odd name for this (it's fine -- bigger fish to fry), but if we change other things, probably something like `processStagedCommittedSectors`

jbenet: this turned out pretty well, good solution 👍

jbenet:
I made the changes in 5fc3c95 so i can merge now (@zixuanzh feel free to revert if disagree)

jbenet:
Ahh-- looks like there's a merge conflict -- i'll rebase on master and resolve the conflict

jbenet:
great work @zixuanzh !

PR #589, opened 2019-10-24
make the website rendering link more prominent

mishmosh:


zixuanzh:
Approving the change but note that website rendering is not automatically updated. Still need to build from source to view the latest spec.

jbenet:
I rebased and added the disclaimer @zixuanzh asked for

PR #591, opened 2019-10-25
Move StorageMinerActor to new runtime/state paradigm to mitigate reen…

jzimmerman:
…trancy issues

### DIFF
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
@@ -72,7 +72,7 @@ func (a *StorageMinerActorCode_I) NotifyOfPoStChallenge(rt Runtime) {
 	h.Commit(st)
 }
 
-func (st *StorageMinerActorState_I) _updateProcessStagedCommittedSectors(rt Runtime) {
+func (st *StorageMinerActorState_I) _updateCommittedSectors(rt Runtime) {
 	for sectorNo, sealOnChainInfo := range st.StagedCommittedSectors() {
 		st.Sectors()[sectorNo] = sealOnChainInfo
 		st.Impl().ProvingSet_.Add(sectorNo)
@@ -97,8 +97,8 @@ func (a *StorageMinerActorCode_I) _submitFaultReport(
 		NewTerminatedFaults_: newTerminatedFaults,
 	}
 
+	rt.Fatal("TODO") // TODO: Send(SPA, ProcessFaultReport(faultReport))
 	panic(faultReport)
-	// TODO: Send(SPA, ProcessFaultReport(faultReport))
 
 	h, st := a.State(rt)
 	st.SectorTable().Impl().TerminationFaultCount_ = util.UVarint(0)
@@ -114,8 +114,8 @@ func (a *StorageMinerActorCode_I) _submitPowerReport(rt Runtime) {
 	}
 	h.Release(st)
 
+	rt.Fatal("TODO") // TODO: Send(SPA, ProcessPowerReport(powerReport))
 	panic(powerReport)
-	// TODO: Send(SPA, ProcessPowerReport(powerReport))
 }
 
 func (a *StorageMinerActorCode_I) _onMissedPoSt(rt Runtime) {
@@ -148,7 +148,7 @@ func (a *StorageMinerActorCode_I) _onMissedPoSt(rt Runtime) {
 	// end of challenge
 	h, st = a.State(rt)
 	st.ChallengeStatus().Impl().OnChallengeResponse(rt.CurrEpoch())
-	st._updateProcessStagedCommittedSectors(rt)
+	st._updateCommittedSectors(rt)
 	h.Commit(st)
 }
 
@@ -170,7 +170,7 @@ func (a *StorageMinerActorCode_I) _verifyPoStSubmission(rt Runtime, postSubmissi
 	// 1. A proof must be submitted after the postRandomness for this proving
 	// period is on chain
 	// if rt.ChainEpoch < sm.ProvingPeriodEnd - challengeTime {
-	//   panic("too early")
+	//   rt.Fatal("too early")
 	// }
 
 	// 2. A proof must be a valid snark proof with the correct public inputs
@@ -181,7 +181,7 @@ func (a *StorageMinerActorCode_I) _verifyPoStSubmission(rt Runtime, postSubmissi
 	// 2.3 Verify the PoSt Proof
 	// verifyPoSt(challenges, TODO)
 
-	panic("TODO")
+	rt.Fatal("TODO") // TODO: finish
 	return false
 }
 
@@ -200,7 +200,8 @@ func (st *StorageMinerActorState_I) _updateClearSector(rt Runtime, sectorNo sect
 		st.SectorTable().Impl().FailingSectors_ -= 1
 	default:
 		// Committed and Recovering should not go to Cleared directly
-		panic("invalid state in clearSector")
+		rt.Fatal("invalid state in clearSector")
+		// TODO: determine proper error here and error-handling machinery
 	}
 
 	delete(st.Sectors(), sectorNo)
@@ -219,8 +220,8 @@ func (st *StorageMinerActorState_I) _updateActivateSector(rt Runtime, sectorNo s
 	case SectorRecoveringSN:
 		st.SectorTable().Impl().RecoveringSectors_ -= 1
 	default:
-		// TODO: proper throw
-		panic("invalid state in activateSector")
+		// TODO: determine proper error here and error-handling machinery
+		rt.Fatal("invalid state in activateSector")
 	}
 
 	st.Sectors()[sectorNo].Impl().State_ = SectorActive()
@@ -261,11 +262,9 @@ func (st *StorageMinerActorState_I) _updateFailSector(rt Runtime, sectorNo secto
 		// no change to SectorTable but increase in FaultCount
 		st.Sectors()[sectorNo].Impl().State_ = SectorFailing(newFaultCount)
 	default:
-		// TODO: proper failure
-		panic("Invalid sector state in CronAction")
+		// TODO: determine proper error here and error-handling machinery
+		rt.Fatal("Invalid sector state in CronAction")
 	}
-
-	// TODO commit state change
 
 	if newFaultCount > MAX_CONSECUTIVE_FAULTS {
 		// TODO: heavy penalization: slash pledge collateral and delete sector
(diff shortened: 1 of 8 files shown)


src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
```diff
@@ -42,63 +42,77 @@ type SectorOnChainInfo struct {
...
     ChallengeStatus
 
     // contains mostly static info about this miner
-    Info                     &MinerInfo
+    Info                       &MinerInfo
 
     // TODO ProvingPeriodEnd   Epoch
 
-    NotifyOfPoStChallenge()
+    _isChallenged(rt Runtime)  bool
```
jbenet: i like the underscores, i think they're very clear

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -27,41 +43,50 @@ func (cs *ChallengeStatus_I) OnChallengeResponse(currEpoch block.ChainEpoch) Cha
...
+	st.ChallengeStatus().Impl().LastChallengeEpoch_ = rt.CurrEpoch()
+	h.Commit(st)
 }
 
-func (sm *StorageMinerActor_I) processStagedCommittedSectors() {
-	for sectorNo, sealOnChainInfo := range sm.StagedCommittedSectors() {
-		sm.Sectors()[sectorNo] = sealOnChainInfo
-		sm.ProvingSet_.Add(sectorNo)
-		sm.SectorTable().Impl().CommittedSectors_ += 1
+func (st *StorageMinerActorState_I) _updateProcessStagedCommittedSectors(rt Runtime) {
```
jbenet: maybe just `_updateStagedCommittedSectors` ? or `_updateCommittedSectors` ? method name is a mouthful now

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -27,41 +43,50 @@ func (cs *ChallengeStatus_I) OnChallengeResponse(currEpoch block.ChainEpoch) Cha
...
+	for sectorNo, sealOnChainInfo := range st.StagedCommittedSectors() {
+		st.Sectors()[sectorNo] = sealOnChainInfo
+		st.Impl().ProvingSet_.Add(sectorNo)
```
jbenet: can this just be `st.ProvingSet().Add(sectorNo)` ? or no?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -72,71 +97,76 @@ func (sm *StorageMinerActor_I) submitFaultReport(
 		NewTerminatedFaults_: newTerminatedFaults,
 	}
 
-	sm.SectorTable().Impl().TerminationFaultCount_ = util.UVarint(0)
-
 	panic(faultReport)
```
jbenet: - i think `panic` makes the go compiler ignore some errors below the `panic(.)` statement
- maybe this should be just a `// TODO` ?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -72,71 +97,76 @@ func (sm *StorageMinerActor_I) submitFaultReport(
...
 	// TODO: Send(SPA, ProcessFaultReport(faultReport))
+
+	h, st := a.State(rt)
+	st.SectorTable().Impl().TerminationFaultCount_ = util.UVarint(0)
```
jbenet: perhaps `st.AddTerminationCount(util.UVarint(0))` ?

jbenet: (keep the actor _code_ as algorithm-oriented as we can?)

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -72,71 +97,76 @@ func (sm *StorageMinerActor_I) submitFaultReport(
...
+		st._updateFailSector(rt, sectorNo, true)
 	}
+	st._updateExpireSectors(rt)
+	h.Commit(st)
 
-	sm.expireSectors()
+	h, st = a.State(rt)
+	newDetectedFaults := st.SectorTable().FailingSectors()
+	newTerminatedFaults := st.SectorTable().TerminationFaultCount()
+	h.Release(st)
```
jbenet: nice -- is the idea that calling `st.SectorTable()` after `h.Release()` would panic? (i'd like that...)

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -314,51 +344,57 @@ func (sm *StorageMinerActor_I) SubmitPoSt(postSubmission poster.PoStSubmission)
...
 			// TODO: proper failure
 			panic("Invalid sector state in ProvingSet.SectorsOff")
 		}
 	}
 
 	// Process Expiration.
-	sm.expireSectors()
+	st._updateExpireSectors(rt)
+
+	h.Commit(st)
```
jbenet: - this is probably the only case i've seen so far where it was hard to immediately see + remember that `h` is still being held, and where so much is going on that _could_ be calling some other actor.
- AFAIK, even calling `rt.Send(.)` in between `h, st := a.State(rt)` and `h.Commit(st)` is an error, correct?  (ie it doesn't need to be, in that it's only a problem if state is changed, but for secure programming, it would be nice if it indeed is an error

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -479,41 +520,37 @@ func (sm *StorageMinerActor_I) verifySeal(onChainInfo sector.OnChainSealVerifyIn
...
+func (a *StorageMinerActorCode_I) CommitSector(rt Runtime, onChainInfo sector.OnChainSealVerifyInfo) {
+	h, st := a.State(rt)
 
-	isSealVerified := sm.verifySeal(onChainInfo)
-	if !isSealVerified {
-		// TODO: proper failure
-		panic("Seal is not verified")
+	isSealVerificationCorrect := st._isSealVerificationCorrect(rt, onChainInfo)
+	if !isSealVerificationCorrect {
+		rt.Fatal("Seal verification failed")
```
jbenet: let's flag errors as something to discuss later

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -27,41 +43,50 @@ func (cs *ChallengeStatus_I) OnChallengeResponse(currEpoch block.ChainEpoch) Cha
...
 }
 
+func (st *StorageMinerActorState_I) _isChallenged(rt Runtime) bool {
+	return st.ChallengeStatus().IsChallenged()
+}
+
+func (a *StorageMinerActorCode_I) _isChallenged(rt Runtime) bool {
+	h, st := a.State(rt)
+	ret := st._isChallenged(rt)
+	h.Release(st)
```
jbenet: I think that `h.Release` and `h.Commit` should mangle `st` and `h` such that all further invocations on them panic. something like this:

```
func (h *ActorStateHandle) Release(st util.Any) {
  // whatever
  h.destroy() // all further invocations to h functions panic
  st.destroy() // all further invocations to st functions panic
}
```
It may be easy to achieve by making `h` and `st` be set to nil (if they're holding on to pointers, could work -- all function invocations would try to dereference nil and panic). But not sure

jbenet:
```go
// can passing in address let `Release` set `st = nil` and achieve the intent?
h.Release(&st)


func (h *ActorStateHandle) Release(st &util.Any) {
  // whatever
  h.Runtime = nil // all further invocations to h functions panic
  st = nil // all further invocations to h functions panic
}
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
zixuanzh: calling it `_update...` might be misleading tho, and we are missing commit state change in this function.

jzimmerman: Changing to `_updateCommittedSectors`.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jzimmerman: Unfortunately, no:

```
systems/filecoin_mining/storage_mining/storage_miner_actor.go:78:18: cannot call pointer method on st.ProvingSet()
systems/filecoin_mining/storage_mining/storage_miner_actor.go:78:18: cannot take the address of st.ProvingSet()
```

I believe the problem is that `ProvingSet()` returns the set by value. We could fix this by implementing functional-style "With" methods in the id compilation, e.g., `With_ProvingSet(newProvingSet)`, but this seems lower priority.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jzimmerman: Modifying these to use `rt.Fatal(.)` instead of `panic(.)`.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jzimmerman: Makes sense (but should this be `SetTerminationCount` rather than `AddTerminationCount`?)

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jzimmerman: Yes (or more precisely, would have the same effect as `rt.Fatal(.)`)

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jzimmerman: Correct, I think we should implement the runtime methods so that this is an error.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jzimmerman: OK -- adding comments in those locations.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jzimmerman: That would cause the functions to panic with a nil dereference, which we might not want -- if this is done by a runtime check, they can instead exit via `rt.Fatal`, causing the transaction to abort but not necessarily causing the interpreter to crash. Not totally clear; which would be preferable?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -314,51 +344,57 @@ func (sm *StorageMinerActor_I) SubmitPoSt(postSubmission poster.PoStSubmission)
...
 	// Reset Proving Period and report power updates
 	// sm.ProvingPeriodEnd_ = PROVING_PERIOD_TIME
 
-	// TODO: get currEpoch from runtime
-	var currEpoch block.ChainEpoch
-	sm.ChallengeStatus().Impl().OnChallengeResponse(currEpoch)
-	sm.processStagedCommittedSectors()
+	h, st = a.State(rt)
+	st.ChallengeStatus().Impl().OnChallengeResponse(rt.CurrEpoch())
+	st._updateCommittedSectors(rt)
```
zixuanzh: are we missing `h.Commit(st)` here?

jzimmerman: Yes, good catch -- just fixed in latest commit.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jbenet: dont think so, i think when we notice termination counts, we want to increment a count?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jbenet: ok sounds good -- looks like rt.Fatal is now there, but the panic is still there. should remove the panics, as they are hiding compiler errors in this function

jbenet:
LGTM, merging

PR #592, opened 2019-10-28
Runtime send functionality, error handling, interpreter

jzimmerman:
(Note: includes contents of previous PR on same branch)


src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+	prevState := rt._globalStatePending().GetActor(rt._actorAddress()).State()
+	prevStateCID := prevState.State()
+	if !ActorSubstateCID_Equals(prevStateCID, checkStateCID) {
+		rt.Abort("State CID differs upon release call")
+	}
+
+	rt._actorStateAcquired_ = false
+}
+
+func (rt *Runtime_I) Check(cond bool) Runtime_Check_FunRet {
```
jbenet: `Assert`?  -- `Check` does not imply abort.

jbenet: why is a return value needed in this function? why can't this be an empty return?  (harms readability / adds unnecessary complexity)

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+	h._initValue_ = ActorStateHandle__initValue_Make_None(&ActorStateHandle__initValue_None_I{})
+	return ret
+}
+
+func (rt *Runtime_I) _updateReleaseActorState(newStateCID ActorSubstateCID) {
+	rt._checkRunning()
+	rt._checkActorStateAcquired()
+
+	rt._globalStatePending_ = rt._globalStatePending().WithActorState(rt._actorAddress(), newStateCID)
+	rt._actorStateAcquired_ = false
```
jbenet: All this state handling shouldn't be in the runtime object itself. likely a subobject (`VMState`?), or the invocation

jbenet: The runtime has a lot of things and doing this here makes it harder to reason about.

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+	rt._checkGasRemaining()
+	rt._gasRemaining_ = rt._gasRemaining().Subtract(x)
+	rt._checkGasRemaining()
+}
+
+func (rt *Runtime_I) _refundGasRemaining(x msg.GasAmount) {
+	// TODO: check x >= 0
+	rt._checkGasRemaining()
+	rt._gasRemaining_ = rt._gasRemaining().Add(x)
+	rt._checkGasRemaining()
```
jbenet: great to have this! 👍

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+
+	rtOuter._transferFunds(rtOuter._actorAddress(), to, input.Value())
+	rtOuter._incrementCallSeqNum(to)
+
+	rtInner := Runtime_Make(
+		rtOuter._globalStatePending(),
+		to,
+		input.Value(),
+		rtOuter._gasRemaining(),
+	)
```
jbenet: - why are we making another VM runtime -- the intention is for the VM to only have one vm.Runtime.
- The runtime is meant to be a single syscalls-like layer, hopefully holding minimal state (or simply proxying calls to sub objects)
- I believe you're now using the runtime as the invocations pair (`InvocInput/InvocOutput`) were meant to be used. If you don't want the pair, maybe use a third object `Invocation`, or `ActorRuntime` to hold the invocation state changes? (i understand the intention of having each actor be isolated/sandboxed -- so  maybe `ActorRuntime`? `InvocationRuntime`? `InvocRuntime`? `MethodRuntime` ?

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+
+func (rt *Runtime_I) _deferredSendInternal(to addr.Address, input InvocInput, ignoreErrors bool) {
+	rt._checkRunning()
+	rt._checkStateLock(false)
+	rt._messageQueue_ = append(rt._messageQueue(), MessageQueueItem_Make(to, input, ignoreErrors))
+}
+
+func (rt *Runtime_I) DeferredSend(to addr.Address, input InvocInput) Runtime_DeferredSend_FunRet {
+	rt._deferredSendInternal(to, input, false)
+	return &Runtime_DeferredSend_FunRet_I{}
```
jbenet: what actor calls need `DeferredSend`?  do we strictly need this now? (i dont think we do)

if not, we should keep this optimization out-- it's a source of complexity (requires the message queue, which itself complicates the runtime). we may find we need it back, and can bring it back then.

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+	rt._running_ = true
+	catchRetCode := exitcode.CatchRuntimeErrors(func() {
+		actorCode.InvokeMethod(rt, method, params)
+		rt._checkStateLock(false)
+		rt._checkNumValidateCalls(1)
+		rt._checkNumReturnCalls(1)
+		for _, item := range rt._messageQueue() {
+			rt._sendInternal(item.to, item.input, item.ignoreErrors)
+		}
+		rt._messageQueue_ = []MessageQueueItem{}
```
jbenet: this got significantly more complicated, do we actually need this? (eg the message queue)

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+	})
+	rt._running_ = false
+
+	if catchRetCode.IsSuccess() {
+		if rt._output() == nil {
+			panic("Internal error: return call recorded but no output object returned")
```
jbenet: i believe most actor methods won't return objects even in success

src/systems/filecoin_vm/runtime/runtime.id
```diff
@@ -1,85 +1,81 @@
...
 
-    // Storage provides access to the VM storage layer
-    Storage    VMStorage
+    AcquireState() ActorStateHandle
 
-    // BlockHeight returns the height of the block this message was added to the chain in
-    Epoch()    st.Epoch
+    ReturnSuccess()
+    ReturnValue(Bytes)
+    ReturnError(exitCode exitcode.ExitCode)
```
jbenet: - Why are we using these and not using return values in `ActorCode_I` functions as we used to?
- this is harming readability, and defeating/not using the go compiler's return value checks

jbenet: If anything, we can use helpers to make the types:

```go
func (a *CronActorCode_I) Constructor(rt vmr.Runtime) vmr.IOutput {
  ...
  return vmr.ReturnSuccess() // returns a &vmr.IOutput_I{ ... }
}
```

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+func (rt *Runtime_I) _checkStateLock(expected bool) {
+	if rt._actorStateAcquired() != expected {
+		rt._apiError("State update and message send blocks must be disjoint")
+	}
+}
+
+func (rt *Runtime_I) _checkGasRemaining() {
+	if rt._gasRemaining().LessThan(msg.GasAmount_Zero()) {
+		rt._throwError(exitcode.SystemError(exitcode.OutOfGas))
+	}
```
jbenet: This turned out well i think! 👍

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+func (rt *Runtime_I) DeferredSendAllowingErrors(to addr.Address, input InvocInput) Runtime_DeferredSendAllowingErrors_FunRet {
+	rt._deferredSendInternal(to, input, true)
+	return &Runtime_DeferredSendAllowingErrors_FunRet_I{}
+}
+
+func (rt *Runtime_I) ValueSupplied() actor.TokenAmount {
+	return rt._valueSupplied()
+}
+
+func (rt *Runtime_I) Randomness(offset uint64) Randomness {
```
jbenet: - Keep `Randomness(e block.ChainEpoch, offset uint64)`
- Actors need to be able to access the randomness of past epochs (up to some K lookback). this is critical for SPC, PoSt, sealing, and more. 
- maybe `e` must be `K <= e <= now()` (or panics/errors)

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -20,177 +21,118 @@ func (vmi *VMInterpreter_I) ApplyMessageBatch(inTree st.StateTree, msgs []Messag
...
+		),
+		false,
+	)
+
+	if sendRet.ExitCode().AllowsStateUpdate() {
 		// success -- refund unused gas
-		outTree = out.OutTree // take the state from the invocation output
-		refundGas := message.GasLimit() - out.GasUsed
-		outTree = treeDepositFunds(outTree, fromActor, gasToFIL(refundGas, message.GasPrice()))
-		outTree = treeIncrementActorSeqNo(outTree, fromActor)
```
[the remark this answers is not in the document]
jbenet: - I do think there is utility in keeping a counter for message invocations, so that (for example) an actor would be able to get the number: `(epoch, callNum)` where `callNum` increments with each internal send in an epoch 
  - this would be useful for randomness generation (eg method can ask for `rt.Randomness(rt.Epoch(), rt.CallNum())`
- potentially even `(epoch, extCallNum, intCallNum)` where `extCallNum` is the external call number (eg the root invocation -- the message), and `intCallNum` is the internal call number in that stack.

src/systems/filecoin_vm/sysactors/register.go
```diff
@@ -0,0 +1,26 @@
...
+	_registerBuiltinActors()
+}
+
+func _registerBuiltinActors() {
+	// TODO
+
+	cron := &CronActorCode_I{}
+
+	vmr.RegisterActor(vmr.InitActorCodeCID, &InitActorCode_I{})
+	vmr.RegisterActor(vmr.CronActorCodeCID, cron)
```
anorth: Do the storage power and market actors need to be registered too?

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -20,177 +21,118 @@ func (vmi *VMInterpreter_I) ApplyMessageBatch(inTree st.StateTree, msgs []Messag
...
+		message.To(),
+		msg.InvocInput_Make(
+			message.Method(),
+			message.Params(),
+			message.Value(),
+		),
+		false,
+	)
+
+	if sendRet.ExitCode().AllowsStateUpdate() {
```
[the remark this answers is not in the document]
jzimmerman: Done

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -20,177 +21,118 @@ func (vmi *VMInterpreter_I) ApplyMessageBatch(inTree st.StateTree, msgs []Messag
...
 
-	// may return a different tree on succeess.
-	// this MUST get rolled back if the invocation fails.
-	var toActor actor.Actor
-	var err error
-	compTree, toActor, err = treeGetOrCreateAccountActor(compTree, message.To())
-	if err != nil {
-		return applyError(inTree, exitcode.ActorNotFound, gascost.ApplyMessageFail)
+	// TODO: Should we allow messages to implicitly create actors?
+	// - If so, need to specify how the actor code is determined
```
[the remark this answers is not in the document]
jzimmerman: Done

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -20,177 +21,118 @@ func (vmi *VMInterpreter_I) ApplyMessageBatch(inTree st.StateTree, msgs []Messag
...
+	// - We should be able to disallow this and still preserve functionality by
+	//   requiring two separate messages (create actor, transfer initial funds)
+	createRet := compTree.WithNewActorIfMissing(message.To())
+	if createRet.Which() == st.StateTree_WithNewActorIfMissing_FunRet_Case_error {
+		return inTree, _applyError(exitcode.ActorNotFound)
+	} else {
+		compTree = createRet.As_StateTree()
```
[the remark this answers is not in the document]
jzimmerman: Done.

-- Though, it's worth noting that the return tuple idiom can also be error-prone: it uses a product type where a sum is what's intended, and there's no static checking for the implicit property that "either ret is sensible and err is nil, or ret is nil and err is non-nil". General question: if we had a more readable way to check the tag of a sum type, would that be preferable to the tuple return? i.e.:

```
retTree, err := F(inTree)
if err != nil {
    // abort
}
```

versus:

```
retVal := F(inTree)
if retVal.IsError() {
    // abort
}
retTree := retVal.As_StateTree()
```

Not as Go-idiomatic, but, I think much less error-prone.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -20,177 +21,118 @@ func (vmi *VMInterpreter_I) ApplyMessageBatch(inTree st.StateTree, msgs []Messag
...
 
-	// deduct maximum expenditure gas funds first
-	// TODO: use a single "transfer"
-	compTree = treeDeductFunds(compTree, fromActor, maxGasCost)
-
-	// transfer funds fromActor -> toActor
-	// (yes deductions can be combined, spelled out here for clarity)
-	// TODO: use a single "transfer"
-	compTree = treeDeductFunds(compTree, fromActor, message.Value())
-	compTree = treeDepositFunds(compTree, toActor, message.Value())
```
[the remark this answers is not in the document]
jzimmerman: I think the following code already covers this check, doesn't it?

```
	maxGasCost := gasToFIL(message.GasLimit(), message.GasPrice())
	totalCost := message.Value() + actor.TokenAmount(maxGasCost)
	if fromActor.State().Balance() < totalCost {
		return inTree, _applyError(exitcode.InsufficientFunds)
	}
```

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -20,177 +21,118 @@ func (vmi *VMInterpreter_I) ApplyMessageBatch(inTree st.StateTree, msgs []Messag
...
+		message.GasLimit(),
+	)
+
+	sendRet, sendRetStateTree := rt.Impl().SendToplevelFromInterpreter(
+		message.To(),
+		msg.InvocInput_Make(
+			message.Method(),
+			message.Params(),
+			message.Value(),
+		),
```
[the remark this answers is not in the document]
jzimmerman: Done

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+	}
+
+	var toActorMethodGasBound msg.GasAmount
+	TODO() // TODO: obtain from actor registry
+	rtOuter._deductGasRemaining(toActorMethodGasBound)
+	// TODO: gasUsed may be larger than toActorMethodGasBound if toActor itself makes sub-calls.
+	// To prevent this, we would need to calculate the gas bounds recursively.
+
+	rtOuter._transferFunds(rtOuter._actorAddress(), to, input.Value())
+	rtOuter._incrementCallSeqNum(to)
```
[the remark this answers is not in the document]
jzimmerman: (see below)

src/systems/filecoin_vm/interpreter/vm_interpreter.go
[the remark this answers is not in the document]
jzimmerman: Sounds good -- I've modified the code to use the existing `CallSeqNum` counter only for external calls. Should we add the internal call counter also, or postpone that change for now?

src/systems/filecoin_vm/interpreter/vm_registry.go
```diff
@@ -1,78 +0,0 @@
-package interpreter
-
-import "errors"
-import actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
-import addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
-import vmr "github.com/filecoin-project/specs/systems/filecoin_vm/runtime"
-import sysactors "github.com/filecoin-project/specs/systems/filecoin_vm/sysactors"
```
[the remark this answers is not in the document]
jzimmerman: The issue here is that `runtime` needs the addresses of the system actors (though it doesn't need the actor code itself). I was able to work around this in the new commits by moving those addresses to the `addr` package, but I think the circularity may still cause problems further down the road.

src/systems/filecoin_vm/message/message.id
```diff
@@ -34,10 +35,19 @@ type SignedMessage struct {
     Signature  filcrypto.Signature
 }  // representation tuple
 
-type ExitCode UVarint
+type InvocInput struct {
+    Method  actor.MethodNum
+    Params  actor.MethodParams
+    Value   actor.TokenAmount
+}
```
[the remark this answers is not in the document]
jzimmerman: (moved state into separate VMContext struct as per discussion)

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+		_actorAddress_:           actorAddress,
+		_actorStateAcquired_:     false,
+		_actorStateAcquiredInit_: actorStateInit.State(),
+
+		_valueSupplied_:    valueSupplied,
+		_gasRemaining_:     gasRemaining,
+		_numValidateCalls_: 0,
+		_numReturnCalls_:   0,
+		_messageQueue_:     []MessageQueueItem{},
+		_output_:           nil,
```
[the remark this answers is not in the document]
jzimmerman: (moved state into separate VMContext struct as per discussion)

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,424 @@
...
+	globalState st.StateTree,
+	actorAddress addr.Address,
+	valueSupplied actor.TokenAmount,
+	gasRemaining msg.GasAmount) Runtime {
+
+	actorStateInit := globalState.GetActor(actorAddress).State()
+
+	return &Runtime_I{
+		_globalStateInit_:        globalState,
+		_globalStatePending_:     globalState,
```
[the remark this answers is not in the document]
jzimmerman: Unfortunately the trailing underscore is an artifact of the `.id` compilation, which I implemented before knowing about the lowercase naming convention. This and a few other `.id` issues have been pending for a while -- I can eval whether the changes required in codegen are reasonably straightforward, and if not, move the relevant structures to pure Go.

src/systems/filecoin_vm/runtime/runtime.go
jzimmerman: Renamed to `Assert`. Re: empty return, this is another `.id` codegen task that is hopefully straightforward (as above), but if not, will move the functions to pure Go.

src/systems/filecoin_vm/runtime/runtime.go
jzimmerman: (moved state into separate VMContext struct as per discussion)

src/systems/filecoin_vm/runtime/runtime.go
jzimmerman: (moved state into separate VMContext struct as per discussion)

src/systems/filecoin_vm/runtime/runtime.go
jzimmerman: (removed deferred send for now as per discussion)

src/systems/filecoin_vm/runtime/runtime.go
jzimmerman: (removed deferred send for now as per discussion)

src/systems/filecoin_vm/runtime/runtime.go
jzimmerman: I think this is a nomenclature issue -- `_output()` here isn't the actual output value of the method call, but rather is the `InvocOutput` instance (which may be non-nil, but still have nil as its method output value). Reworded and modified logic to use Go-native `return` instead (see below).

src/systems/filecoin_vm/runtime/runtime.id
jzimmerman: Fixed. (I believe this was leftover from a draft in which each method was run in its own goroutine, and we needed a way to signal completion -- but this should no longer be relevant.)

src/systems/filecoin_vm/runtime/runtime.go
jzimmerman: Done.

src/systems/filecoin_vm/sysactors/register.go
jzimmerman: Yes -- this function was moved from elsewhere so I didn't edit the content in the initial PR, but I've just added those entries.

src/systems/filecoin_vm/actor/address/address.go
```diff
@@ -4,6 +4,16 @@ package address
...
 
+// Addresses for singleton system actors
+var (
+	InitActorAddr           = &Address_I{} // TODO
+	CronActorAddr           = &Address_I{} // TODO
+	StoragePowerActorAddr   = &Address_I{} // TODO
+	StorageMarketActorAddr  = &Address_I{} // TODO
+	PaymentChannelActorAddr = &Address_I{} // TODO
+	BurntFundsActorAddr     = &Address_I{} // TODO
+)
```
jbenet: this stuff shouldn't leak into this package. `address` is fundamental -- it shouldn't know about any of these actors.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -83,18 +88,18 @@ func (a *StorageMinerActorCode_I) _isChallenged(rt Runtime) bool {
 }
 
 // called by CronActor to notify StorageMiner of PoSt Challenge
-func (a *StorageMinerActorCode_I) NotifyOfPoStChallenge(rt Runtime) {
-	rt.ValidateCallerIs(vmr.CronActorAddr)
+func (a *StorageMinerActorCode_I) NotifyOfPoStChallenge(rt Runtime) InvocOutput {
+	rt.ValidateCallerIs(addr.CronActorAddr)
```
jbenet: this shouldn't be coming from `addr`. runtime is a much better place to do it -- we can use a file like `exitcodes`

src/systems/filecoin_vm/interpreter/vm_interpreter.go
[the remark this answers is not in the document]
jbenet: > Should we add the internal call counter also, or postpone that change for now?

maybe wait until we need it? (one place i thought we needed it was the `ComputeActorAddress`, but i think `CallSeqNum` is enough)

src/systems/filecoin_vm/interpreter/vm_registry.go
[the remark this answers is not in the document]
jbenet: > The issue here is that runtime needs the addresses of the system actors (though it doesn't need the actor code itself). I was able to work around this in the new commits by moving those addresses to the addr package, but I think the circularity may still cause problems further down the road.

ah got it. (we could bring them into the runtime or their own package). probably fine to leave in `addr` for now, and clean them up after everything stabilizes more

src/systems/filecoin_vm/message/message.id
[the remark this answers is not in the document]
jbenet: LGTM! 👍

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,81 +1,405 @@
...
+	h._initValue = nil
+	return ret
 }
 
-func (r *Runtime_I) Fatal(string) Runtime_Fatal_FunRet {
-	panic("TODO")
+// Concrete instantiation of the Runtime interface. This should be instantiated by the
+// interpreter once per actor method invocation, and responds to that method's Runtime
+// API calls.
+type VMContext struct {
```
jbenet: yeah i think this worked out very well (runtime interface, context struct)

jbenet:
I fixed the merge conflicts

PR #593, opened 2019-10-28
Add vmr to StorageMarketActor and StoragePowerActor

zixuanzh:
per changes in #591


src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -347,10 +344,10 @@ func (spa *StoragePowerActor_I) Surprise(ticket block.Ticket) []addr.Address {
 	// var currBlockHeight util.UInt
 
 	// The number of miners that are challenged at this block
-	challengeCount := uint(len(spa.PowerTable())) / provingPeriod
+	challengeCount := uint(len(st.PowerTable())) / provingPeriod
```
jbenet: move the `h, st := a.State(rt)` line to right above this. (wrap the state look-ups tightly)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -392,6 +389,8 @@ func (spa *StoragePowerActor_I) Surprise(ticket block.Ticket) []addr.Address {
 		// SendMessage(sm.ExtendProvingPeriod)
 	}
 
+	h.Commit(st)
```
jbenet: - i think this is only reading (not committing), so this should be `h.Release(st)` ?
- given the state handle, the `Send` function will fail inside, will want to move it to a loop like this:

```go
h, st := a.State(rt)
// make all the variables

index := 0
for address, _ := range st.PowerTable() {
    minerAddresses[index] = address
    index++
}

for i := uint(0); i < challengeCount; i++ {
  var randomNumber uint
  minerIndex := randomNumber % uint(len(st.PowerTable()))
  minerAddress := minerAddresses[minerIndex]
  surprisedMiners = append(surprisedMiners, minerAddress)
}

h.Release()


// now send the messages (after state is released)
for _, addr := range surprisedMiners {
  rt.SendMessage(addr, ...)
}
```

jbenet: oh and potentially also:

```go
func (a *StoragePowerActorCode_I) Surprise(rt Runtime, ticket block.Ticket) []addr.Address {

  var provingPeriod int // TODO

  // sample the actor addresses
  h, st := a.State(rt)
  challengeCount := len(st.PowerTable()) / provingPeriod
  surprisedMiners := st.SampleMinersToSurprise(rt, challengeCount)
  st.Release() 

  // now send the messages
  for _, addr := range surprisedMiners {
    rt.SendMessage(addr, ...)
  }
}


func (a *StoragePowerActorState_I) SampleMinersToSurprise(rt Runtime, challengeCount int) []addr.Address {
  // this wont quite work -- a.PowerTable() is a HAMT by actor address, doesn't
  // support enumerating by int index. maybe we need that as an interface too, 
  // or something similar to an iterator (or iterator over the keys)
  // or even a seeded random call directly in the HAMT: myhamt.GetRandomElement(seed []byte, idx int)
  m := a.PowerTable() 

  return postSurpriseSample(rt, m, challengeCount)
}

// postSurpriseIndices implements the PoSt-Surprise sampling algorithm
func postSurpriseSample(rt Runtime, allMiners []addr.Address, provingPeriod int) []addr.Address {

  sm := make([]addr.Address, challengeCount)
  for i := 0; i < challengeCount; i++ {
    // rInt := rt.NextRandomInt() // we need something like this in the runtime
    rInt := 4 // xkcd prng for now.
    miner := allMiners[rInt % len(allMiners)]
    sm = append(sm, miner)
  }
  return sm
}
```

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -5,320 +5,315 @@ import (
...
+	// TODO: convert msgSender to MinerActorID
 	var minerID addr.Address
 
-	currEntry, found := spa.PowerTable()[minerID]
+	powerEntry, found := st.PowerTable()[minerID]
+
 	if !found {
-		// TODO: proper throw
-		panic("TODO")
+		rt.Fatal("spa.EnsurePledgeCollateralSatisfied: miner not found.")
```
jbenet: @jzimmerman the `h.Commit/Release` structure has one issue -- control flow returns:

```go
h, st := a.State(rt)

err := tryFoo()
if err != nil {
    return rt.ReturnError(err) // did not h.Release or h.Commit
}

h.Commit(st)
```

We should not use `defer` so we should enforce strict locking/unlocking of state, similar to `sync.Mutex`


```go
h, st := a.State(rt)

err := tryFoo()
if err != nil {
    h.Release()
    return rt.ReturnError(err)
}

h.Commit(st)
```

jbenet: - I imagine errors and `rt.Fatal` _can_ automatically release w/o side effect, BUT is automatic undo what we want? or do we want to force spec code writers to make sure they release properly? 
- i lean towards the latter, and calling `rt.ReturnX` to be a panic-worthy error if state has not been released

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -5,320 +5,315 @@ import (
...
+	// store ownerAddr and workerAddr there
+	// and return StorageMinerActor address
 
-	// TODO: send amountToSlash to TreasuryActor
-	panic(amountToSlash)
-	spa.PowerTable_[minerID] = currEntry
+	// TODO: minerID should be a MinerActorID
+	// which is smaller than MinerAddress
+	var minerID addr.Address
+	st.PowerTable()[minerID] = newMiner
```
jbenet: move the `h, st := a.State(rt)` call to just above this line. (tight call)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -5,320 +5,315 @@ import (
...
+	st.Impl().PowerTable_[minerID] = currEntry
+
+	h.Commit(st)
+
+	h, st = a.State(rt)
 
 	// TODO: send funds to msgSender
-	// TODO: commit state change
+
+	h.Commit(st)
```
jbenet: dont need this one

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -5,320 +5,315 @@ import (
...
+		rt.Fatal("spa.WithdrawBalance: insufficient balance.")
 	}
 
 	currEntry.Impl().AvailableBalance_ = currEntry.AvailableBalance() - amount
-	spa.PowerTable_[minerID] = currEntry
+	st.Impl().PowerTable_[minerID] = currEntry
+
+	h.Commit(st)
+
+	h, st = a.State(rt)
```
jbenet: dont need this one -- send is probably going to be done with `rt.Send(..)`

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -5,320 +5,315 @@ import (
...
+////////////////////////////////////////////////////////////////////////////////
 
-	isMinerVerified := spa.verifyMiner(msgSender)
-	if !isMinerVerified {
-		// TODO: proper throw
-		panic("TODO")
-	}
+func (a *StoragePowerActorCode_I) AddBalance(rt Runtime) {
+
+	h, st := a.State(rt)
```
jbenet: acquire the state in the last possible line (limit the scope / least authority)

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -4,97 +4,211 @@ import addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address
...
+		p.ClientCollateralPerEpoch() < deal.MIN_CLIENT_DEAL_COLLATERAL_PER_EPOCH {
+		return false
+	}
+
+	// verify client and provider has sufficient balance
+	clientBalanceA := st.Balances()[p.Client()].Available()
+	providerBalanceA := st.Balances()[p.Provider()].Available()
+
+	if clientBalanceA < (p.ClientBalanceRequirement()) ||
+		providerBalanceA < p.ProviderBalanceRequirement() {
```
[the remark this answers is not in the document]
zixuanzh: changing to `bal.Available() >= amt`

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -4,97 +4,211 @@ import addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address
...
+
+// Call by PublishStorageDeals and GetLastDealExpirationFromDealIDs (consider remove this)
+// This is the check before a StorageDeal appears onchain
+// It checks the following:
+//   - verify deal did not expire when it is signed
+//   - verify deal hits the chain before StartEpoch
+//   - verify client and provider address and signature are correct (TODO may not be needed)
+//   - verify StorageDealCollateral match requirements for MinimumStorageDealCollateral
+//   - verify client and provider has sufficient balance
+func (st *StorageMarketActorState_I) _verifyStorageDeal(rt Runtime, d deal.StorageDeal) bool {
```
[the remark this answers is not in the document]
zixuanzh: removing `_validateNewStorageDeal` from `GetLastDealExpirationFromDealIDs`. We might want to do more checks there but unclear what they will be.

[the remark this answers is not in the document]
zixuanzh: Let's also chat about this, there might be a simplification that can get rid of `GetLastDealExpirationFromDealIDs` without changing the system/product requirement.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
zixuanzh: `rt.Fatal` should undo all the state changes and return state to the initial state at state acquisition. Good to make explicit that that is the case but probably dont want to give code writers the option of choosing if state will be released or unrolled?

zixuanzh: `rt.Fatal` should undo all the state changes and return state to the initial state at state acquisition. Good to make explicit that that is the case but probably dont want to give code writers the option of choosing if state will be released or unrolled?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
zixuanzh: This is way cleaner than before 👍, will add this as placeholder and revisit in a later PR.

jbenet:
Fixed merge conflicts after new VM code landed

jbenet:
@zixuanzh see the new `UpdateRelease(.)` form

though, @jzimmerman i think the `h.Commit(st)` thing was better. think it feels more intuitive to take a lease/handle on something, and then call commit on it (than calling the `UpdateRelease` helper).

PR #595, opened 2019-10-29
Modify orient target to be all-orient, and add generic orient target.

porcuquine:
- Rename `orient` make target to `all-orient`.
- Add new `orient` target which will solves all `.orient` files with a corresponding `.json` file.
  - Output goes to `build/….orient.json`

### DIFF
--- a/bin/solve-orient.sh
+++ b/bin/solve-orient.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+system_file=$1
+input_file=$2
+output_file=$3
+
+mkdir -p $output_file && rmdir $output_file # Dirty hack to ensure output_file directory exists.
+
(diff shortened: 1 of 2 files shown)


bin/solve-orient.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+system_file=$1
+input_file=$2
+output_file=$3
+
+mkdir -p $output_file && rmdir $output_file # Dirty hack to ensure output_file directory exists.
+
+orient/bin/orient solve --system=$system_file --in=$input_file | jq > $output_file
```
jbenet: `jq` is not a dep-- should probably add it to `bin/install-deps-orient.sh` and `bin/install-deps-orient-user.sh`?

Issue #588, opened 2019-10-24
EC section inclusion bug


jbenet:
@sternhenri  i think there's some messup in how the EC section is getting added (messes w/ section numbers). i think it's embedding a file into `systems/filecoin_blockchain/storage_power_consensus` section from `algorithms/`

PR #601, opened 2019-10-29
change section numbering and remove ec duplication

sternhenri:
Removed EC from SPC section to leave in algorithms section and added link to EC section from SPC separately. 

Does this solve your aforementioned issue @jbenet ? (https://github.com/filecoin-project/specs/issues/588)


jbenet:
rebased on master

PR #604, opened 2019-10-30
Split CommitSector for Interactive PoRep

zixuanzh:


src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -543,44 +543,97 @@ func (st *StorageMinerActorState_I) _isSealVerificationCorrect(rt Runtime, onCha
...
 }
 
-// Currently deals must be posted on chain via sma.PublishStorageDeals before CommitSector
-// TODO(optimization): CommitSector could contain a list of deals that are not published yet.
-func (a *StorageMinerActorCode_I) CommitSector(rt Runtime, onChainInfo sector.OnChainSealVerifyInfo) {
+// Deals must be posted on chain via sma.PublishStorageDeals before PreCommitSector
+// TODO(optimization): PreCommitSector could contain a list of deals that are not published yet.
+func (a *StorageMinerActorCode_I) PreCommitSector(rt Runtime, info sector.SectorPreCommitInfo) {
+
+	// no checks needed
```
jbenet: do we need to validate any of `SectorPreCommitInfo` ? 
- ~~do we need to check dealIDs map to existing deals? probably worth checking here~~ there's a TODO below.
- is `SealRandomness` correct? do we need to check it now, or wait till `CommitSector`

src/systems/filecoin_mining/sector/sealing.id
```diff
@@ -50,6 +50,19 @@ type SealCommitment struct {
     Expiration  block.ChainEpoch
 }
 
+type SectorPreCommitInfo struct {
+    SectorNumber
+    SealedCID       SealedSectorCID  // CommR
+    SealRandomness
```
jbenet: do we need `SealRandomness`? or should it be the `epoch` the randomness comes from?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
zixuanzh: We might do some syntactic validation here but unclear how much we need to check in the PreCommit step. If that is the case, make sense to incur a cost for calling PreCommit

src/systems/filecoin_mining/sector/sealing.id
zixuanzh: I think we can get the Randomness deterministically from the chain with epoch. Will change this to Epoch for now. cc @porcuquine

porcuquine: Yes, I think the `Epoch` and `InteractiveEpoch` need to go on chain and are cheaper there than the randomness itself, which can be looked up again when needed. Once looked up for verification, the randomness itself can be passed to `VerifySeal`.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -543,44 +543,97 @@ func (st *StorageMinerActorState_I) _isSealVerificationCorrect(rt Runtime, onCha
...
+func (st *StorageMinerActorState_I) _sectorExists(sectorNo sector.SectorNumber) bool {
 	_, found := st.Sectors()[sectorNo]
 	return found
 }
 
-// Currently deals must be posted on chain via sma.PublishStorageDeals before CommitSector
-// TODO(optimization): CommitSector could contain a list of deals that are not published yet.
-func (a *StorageMinerActorCode_I) CommitSector(rt Runtime, onChainInfo sector.OnChainSealVerifyInfo) {
+// Deals must be posted on chain via sma.PublishStorageDeals before PreCommitSector
+// TODO(optimization): PreCommitSector could contain a list of deals that are not published yet.
```
porcuquine: As far as I can tell, the deal IDs aren't being passed to `PreCommitSector` at all, but they should be.

porcuquine: Nevermind, I see it hiding below the comment above.

zixuanzh: see `sealing.SectorPreCommitInfo` which contains `DealIDs` as an input to `PreCommitSector`

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -600,72 +634,79 @@ func (a *StorageMinerActorCode_I) ProveCommitSector(rt Runtime, info sector.Sect
...
-		DealIDs_:                   preCommitInfo.DealIDs(),
-		SectorNumber_:              preCommitInfo.SectorNumber(),
-	}
-
-	isSealVerificationCorrect := st._isSealVerificationCorrect(rt, onChainInfo)
-	if !isSealVerificationCorrect {
-		rt.Fatal("sm.ProveCommitSector: seal verification failed.")
-	}
+	// if fewer than MAX_PROVE_COMMIT_SECTOR_EPOCH has elapsed
+	if elapsedEpoch <= MAX_PROVE_COMMIT_SECTOR_EPOCH {
```
jbenet: we should check the error case first, handle it, return. otherwise proceed ([go error return convention](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow))

```go
if elapsedEpoch > MAX_PROVE_COMMIT_SECTOR_EPOCH {
    h.Commit(st)
    return // error
}

// ok we're good here.

... rest
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -23,6 +23,9 @@ func (a *StorageMinerActorCode_I) State(rt Runtime) (vmr.ActorStateHandle, State
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// TODO: placeholder
+const MAX_PROVE_COMMIT_SECTOR_EPOCH = block.ChainEpoch(3)
```
jbenet: maybe make it clearer this is a placeholder _value_ and that we'll pick it later, someone may think 3 is actually the number.

PR #606, opened 2019-10-30
sys actors -- Init, Cron, Account

jbenet:
Improved implementation of Init, Cron, Account.


src/systems/filecoin_vm/sysactors/init_actor.go
```diff
@@ -2,92 +2,134 @@ package sysactors
...
+	return h, state
+}
+func Release(rt Runtime, h vmr.ActorStateHandle, st InitActorState) {
+	checkCID := actor.ActorSubstateCID(rt.IpldPut(st.Impl()))
+	h.Release(checkCID)
+}
+func UpdateRelease(rt Runtime, h vmr.ActorStateHandle, st InitActorState) {
+	newCID := actor.ActorSubstateCID(rt.IpldPut(st.Impl()))
+	h.UpdateRelease(newCID)
+}
```
jbenet: @jzimmerman this boilerplate is tedious/error prone in every actor. (though 6 is probably fine :) )

src/systems/filecoin_vm/sysactors/init_actor.go
```diff
@@ -2,92 +2,134 @@ package sysactors
...
-		// reordering
-		// TODO: where do `creator` and `nonce` come from?
-		// TODO: CallSeqNum is not related to From -- it's related to Origin
-		// addr := rt.ComputeActorAddress(rt.Invocation().FromActor(), rt.Invocation().CallSeqNum())
-		var addr addr.Address // TODO
+func (a *InitActorCode_I) Constructor(rt Runtime) InvocOutput {
+	panic("TODO")
+}
 
+func (a *InitActorCode_I) Exec(rt Runtime, codeCID actor.CodeCID, constructorParams actor.MethodParams) InvocOutput {
```
jbenet: trying to keep as much as possible within `InitActor`

src/systems/filecoin_vm/sysactors/init_actor.go
```diff
@@ -2,92 +2,134 @@ package sysactors
...
+	actorID := s.NextID_
+	s.NextID_++
 	return actorID
 }
 
-func (_ *InitActorCode_I) GetActorIDForAddress(state InitActorState, address addr.Address) actor.ActorID {
-	return state.AddressMap()[address]
+func (_ *InitActorCode_I) _computeNewAddress(rt Runtime, id actor.ActorID) addr.Address {
+	// assign an address based on some randomness
+	// we use the current epoch, and the actor id. this should be a unique identifier,
```
anorth: A message can easily end up in a different epoch during a re-org. I don't think this will produce a stable address.

PR #608, opened 2019-10-30
Rational Calculations

nicola:


jbenet:
(this says draft -- do you want me to review it yet or wait?)

nicola:
nop, not needed, it's mostly old calculations that I am keeping around, they were meant to be useful at the beginning.

Closing this for now.

Issue #603, opened 2019-10-30
EC with multiple leaders


sternhenri:
How to do multiple leader election in EC? @ZenGround0 recently pointed out an issue that has been vastly overlooked in doing multiple leaders in EC. 

Currently election validation is specced [here](https://github.com/filecoin-project/specs/blob/master/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go#L126) but this code needs to be updated (along with the md file) to explicit how multiple leaders is done.

*(In this issue, we use  expected_number_of_leaders_per_round = e.)*

The straw-man solution would have us do something simple, namely:
- check that `power_fraction * e < electionProof.Output`

However, as @ZenGround0 points out this has unintended consequences, namely:
- There is no incentive to build power beyond 1/e of the network without creating another sybil (that may be fine)
- can get guaranteed wins (rather than probabilistic) with enough power
- Returns on investment are no longer linear
e.g. a 25% miner will deterministically win at every round for e >= 4.

Perhaps the **intended** system should be closer to running e independent leader elections, thereby making a 25% miner 68% likely to win at least once (`1-(1-.25)^4)`). But this also has issues:
- More complex to implement (basically implement @jzimmerman's virtual sybils idea)
- Harder to reason about miner behavior accordingly
- Defeats purpose (in terms of bw) of having multiple winners if same miner wins multiple times (per @nikkolasg)

Two questions follow:
- What is the exact intended behavior we should seek
- How do we achieve it specifically for EC.

I am in favor of retaining the straw man at this time but this is a core question that has a wide impact on occurrence of null rounds as well as incentives and winning distribution, so I wanted to bring you in to opine before moving this to spec (by this Friday Nov 1).

ZenGround0:
Proposed behaviors:

- Fairness -- miners win in proportion to storage power
- Tipset size -- we want to be able to fix the expected number of blocks
- Election process doesn't unduly pressure miners to sybil -- the 25% example above is maybe ok because its such a huge portion of network power but if all miner's have a clear incentive to never grow a miner actor above 0.001% or something that's a problem.

Places to look that have solutions we could rework
- Algorand 
- Spacemint 3.5 proof quality sampling

Throwaway idea
- Elections could output some number of wins easy to derive from election proof (I think algorand works more like this).  TIpset reward is then split proportionally by dividing one miner's number of wins / total.

nikkolasg:
I would propose something simple:
- one block per miner per round

[...]

PR #602, opened 2019-10-30
updates to clean up EC and SPC [WIP]

sternhenri:
Includes:

- LE nonce revamp (now done with block height appended to ticket rather than an explicit nonce)
- better wording in slashing
- bug fixing in ticket drawing code
- moving some code around

Missing:
- a bit more on timestamp + round number validation and cutoff
- generalization of EC to include multiple winners.

PR #610, opened 2019-10-31
Patch/spc work [WIP]

sternhenri:
Continuing https://github.com/filecoin-project/specs/pull/602.

- [ ] Code compiles
- [x] Multiple EC winners speccd (https://github.com/filecoin-project/specs/issues/603)
- [x] Clock system specc'd
- [x] Timestamp Validation specc'd (rest to come later)
- [x] BLS linked
- [ ] wider validation spec (left TODO)
- [x] worker key change spec (TODO: interface between a given miner and worker keys is not yet in spec)

Requests for review:

**@jbenet**
- Some of the code snippet inclusions aren't rendering (i.e. serve and look for "something's not right"). I think it's a directory structure issue in the /algorithms section but would love your thoughts
- validation of `clock_subsystem`, in particular for triggering election retries, and rejecting blocks coming in after epoch cutoff
- `block producer` seems to be pretty stale. Could you take a look?
- `key_store` first draft

**@zenground0**
- validation of consensus fault rules

**@kubuxu**
- validation of weight function
- validation of leader election as laid out in code and explanation

**@whyrusleeping**
- That keystore is consistent with your model

**@ZX**
- That Power table fits your model


nikkolasg:
Which solution are you speccing out then (from the issue) ?

src/algorithms/expected_consensus.md
```diff
@@ -197,11 +220,12 @@ Due to the existence of potential forks in EC, a miner can try to unduly influen
...
+- (1) `double-fork mining fault`: two blocks mined at the same epoch.
+- (2) `same-fork double-mining fault`: two blocks mined off of the same Tipset at different epochs (i.e. with different `ElectionProof`s generated from the same input ticket).
+- (3) `parent grinding fault`: one block's parent is a Tipset that provably should have included a given block but does not. While it cannot be proven that a missing block was willfully omitted in general (i.e. network latency could simply mean the miner did not receive a particular block), it can when a miner has successfully mined a block two epochs in a row and omitted one. That is, this condition should be evoked when a miner omits their own prior block. When a miner's block at epoch e + 1 references a Tipset that does not include the block they mined at e both blocks can be submitted to prove this fault.
```
sternhenri: @ZenGround0 would love thoughts on above stated conditions if you have!

src/systems/filecoin_blockchain/storage_power_consensus/expected_consensus.go
```diff
@@ -0,0 +1,49 @@
...
+	// totalPower := spc.PowerTable.GetTotalPower()
+
+	// // Conceptually we are mapping the pseudorandom, deterministic VRFOutput onto [0,1]
+	// // by dividing by 2^HashLen (64 Bytes using Sha256) and comparing that to the miner's
+	// // power (portion of network storage).
+	// return (minerPower*2^(len(electionProof.Output)*8) < electionProof.Output*totalPower)
+	return true
+}
+func (self *ExpectedConsensus_I) GetBlockRewards(electionProof ElectionProof, workerAddr addr.Address) UVarint {
+	panic("")
```
sternhenri: cc @Kubuxu still working on it, but it'll go here

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -3,31 +3,41 @@ title: Clock
...
-- explain synchrony requirements
-  - small clock drift -- <2s
-  - very important to have accurate time
-- explain how we can resync
-  - Network
-    - recommend various NTP servers
-  - Cesium clocks
-- Future work:
-  - VDF Clocks
+{{< readfile file="clock_subsystem.id" code="true" lang="go" >}}
```
sternhenri: cc @ZenGround0 per your question last week. Please let me know what else you need here.

src/algorithms/expected_consensus.md
steven004: What exactly means by 
*(1) `double-fork mining fault`: two blocks mined at the same epoch.*  ?

For example, in the following case, are [n+2] and [n+2'] at the same epoch? 
[n] <--- [n+1] <--- [n+2]
 &nbsp;&nbsp; \\
&nbsp;&nbsp;&nbsp;&nbsp;\\---- [n+1'] <---[n+2']

sternhenri: yes!

jbenet: a dot diagram may be useful there -- here:

```dot
digraph G {
  rankdir=RL;
  node [shape=box, style="filled", color="lightgreen"]

  B1 [label="B1\nMinerAny"]
  B2 [label="B2\nMinerA",   color=pink]
  B3 [label="B3\nMinerA",   color=pink]
  B4 [label="B4\nMinerB",   color=pink]
  B5 [label="B5\nMinerB",   color=pink]
  B6 [label="B6\nMinerC"]
  B7 [label="B7\nMinerC"]

  B6 -> B3
  B7 -> B4

  B4 -> B2 -> B1
  B5 -> B3 -> B1
}
```

looks like

image

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -1,17 +1,29 @@
...
-import power "github.com/filecoin-project/specs/systems/filecoin_blockchain/storage_power_consensus"
-import proving "github.com/filecoin-project/specs/systems/filecoin_mining/storage_proving"
-import sector "github.com/filecoin-project/specs/systems/filecoin_mining/sector"
-import util "github.com/filecoin-project/specs/util"
-import vmr "github.com/filecoin-project/specs/systems/filecoin_vm/runtime"
-import exitcode "github.com/filecoin-project/specs/systems/filecoin_vm/runtime/exitcode"
+import (
+	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
+	addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
+
```
jbenet: what happened here? doesnt seem formatted correctly

jbenet: did you run `make fmt-code` ?

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -3,31 +3,41 @@ title: Clock
...
+
+## Clock uses
+Specifically, the Filecoin system clock is used:
+- to validate incoming blocks and ensure they were mined in the appropriate round, looking at the wall clock time in conjunction with the block's `ElectionProof` (which contains the epoch number) (see {{<sref leader_election>} and {{<sref block_validation>}}).
+- to help protocol convergence by giving miners a specific cutoff after which to reject incoming blocks in this round (see {{<sref chain_sync>}}).
+- to maintain protocol liveness by allowing participants to try leader election in the next round if no one has produced a block in this round (see {{<sref storage_power_consensus>}}).
+
+In order to allow miners to do the above, the system clock must:
+1. have low clock drift: at most on the order of 2s (i.e. markedly lower than epoch time) at any given time.
```
jbenet: on the order of `1s` or `300ms`

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -3,31 +3,41 @@ title: Clock
...
+2. maintain accurate network time over many epochs: resyncing and enforcing accurate network time.
+3. set epoch number on client initialization equal to `epoch ~= (current_time - genesis_time) / epoch_time`
+
+It is expected that other subsystems will register to a NewRound() event from the clock subsystem.
+
+## Clock Requirements
+
+Computer-grade clock crystals can be expected to have drift rates on the order of [1ppm](https://www.hindawi.com/journals/jcnc/2008/583162/) (i.e. 1 microsecond every second or .6 seconds a week), therefore, in order to respect the first above-requirement,
+- clients SHOULD query an NTP server (`pool.ntp.org` is recommended) on an hourly basis to adjust clock skew.
```
jbenet: - maybe recommend a few in a listing:

```
pool.ntp.org
something else.org
etc.org
```

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -3,31 +3,41 @@ title: Clock
...
+  - We further recommend making 3 measurements in order to drop by using the network to drop outliers
+  - See how [go-ethereum does this](https://github.com/ethereum/go-ethereum/blob/master/p2p/discv5/ntp.go) for inspiration
+- clients CAN consider using cesium clocks instead for accurate synchrony within larger mining operations
```
jbenet: 👍

src/systems/filecoin_nodes/clock/clock_subsystem.go
```diff
@@ -2,10 +2,27 @@ package clock
 
 import "time"
 
-func (_ *WallClock_I) NowUTC() Time {
+func (_ *Clock_I) UTCTime() Time {
```
jbenet: Um, why changing the `Now` interface? 

this is to make it completely obvious. it's a very useful addition in Golang to distinguish "get the time right now" from time conversions of other time values.

jbenet: maybe `NowUTCTime()` if you want

src/systems/filecoin_nodes/clock/clock_subsystem.go
```diff
@@ -2,10 +2,27 @@ package clock
...
 	return Time(time.Now().Unix())
 }
 
-func (_ *WallClock_I) NowUTCNano() Time {
+func (_ *Clock_I) UnixNano() Time {
```
jbenet: same (keep `Now`)

src/systems/filecoin_nodes/clock/clock_subsystem.go
```diff
@@ -2,10 +2,27 @@ package clock
...
 	return Time(time.Now().UnixNano())
 }
+
+func (self *Clock_I) CurrentEpoch() ChainEpoch {
+	return self.currentEpoch_
+}
```
jbenet: - This should not live here. the `Clock` does not know the epoch, the Blockchain does, and that's related to finality in chainsync, bootstrap state and more. `Clock` here should not have (nor report) the epoch -- defer to the blockchain.

- there was a previous idea of a `Round/EpochClock` but that was going to be a smalll abstraction on top of (having a pointer to) the blockchain system, and deferring to it for calculation (including failing to return because system is not in a state able to return an epoch). we ended up not need it so i didnt write it.

jbenet: Oh maybe you want to calculate what epoch _it should be_ based on a prior epoch-to-time mapping? I think functions like this are appropriate for this package:

```go
func EpochAfterTime(start block.ChainEpoch, t Time) block.ChainEpoch {
   // returns start + EpochsInSeconds(t)
}

// genesis being hard coded
func EpochAfterGenesis() block.ChainEpoch {
   // returns EpochAfterTime(genesis, Now() - timeOfGenesis)
}
```

then caller can take responsibility of passing in the right start epoch, etc.

src/systems/filecoin_nodes/clock/clock_subsystem.go
```diff
@@ -2,10 +2,27 @@ package clock
...
+	// if self.UnixNano()-self.currentEpochStart > self.cutoffAt_ {
+	// 	return ChainEpochState.PastCutoff
+	// }
+	// return ChainEpochState.Active
+}
+
+func (self *Clock_I) ResetEpoch(genesisTime Time) struct{} {
+	panic("")
+	// self.currentEpoch = (self.UnixNano() - genesisTime) / epochDuration
+}
```
jbenet: remove

src/systems/filecoin_nodes/clock/clock_subsystem.go
```diff
@@ -2,10 +2,27 @@ package clock
...
+	return self.currentEpoch_
+}
+
+func (self *Clock_I) CurrentEpochState() ChainEpoch {
+	panic("")
+	// if self.UnixNano()-self.currentEpochStart > self.cutoffAt_ {
+	// 	return ChainEpochState.PastCutoff
+	// }
+	// return ChainEpochState.Active
+}
```
jbenet: remove -- dont hold state

src/systems/filecoin_nodes/clock/_index.md
sternhenri: I think given a 15s block time, 2 would be fine. But will constrain to 1.

src/systems/filecoin_nodes/clock/clock_subsystem.go
sternhenri: Was thinking this could be useful when you are syncing with network again and want to remove skew

src/systems/filecoin_nodes/clock/clock_subsystem.go
sternhenri: ok, however we do need a way to enforce "when" in the epoch we are as part of rejecting blocks received past cutoff in block validation... Will change so it's stateless.

src/systems/filecoin_nodes/clock/clock_subsystem.go
sternhenri: epochDuration (or EpochsInSeconds) should be pulled from blockchain?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.md
```diff
@@ -20,11 +20,26 @@ SPC provides a power table abstraction which tracks miner power (i.e. miner stor
 
 An invariant of the storage power consensus subsystem is that all storage in the power table must be verified. That is, miners can only derive power from storage they have already proven to the network.
 
-In order to achieve this, Filecoin delays updating power for new sector commitments until the first valid PoSt in the next proving period corresponding to that sector.
+In order to achieve this, Filecoin delays updating power for new sector commitments until the first valid PoSt in the next proving period corresponding to that sector. (TODO: potential delay this further in order to ensure that any power cut goes undetected at most as long as the shortest power delay on new sector commitments).
```
sternhenri: @zx just want to make sure this is what you had understood before. 

cc @nikkolasg

src/systems/filecoin_nodes/key_store/key_store.id
```diff
@@ -0,0 +1,15 @@
+import filcrypto "github.com/filecoin-project/specs/algorithms/crypto"
```
sternhenri: see below @whyrusleeping @jbenet 

I would argue we may want to disable the ability to have many workers per owner in v1.

We do need this abstraction if we want to enable key changes though (I can spec in forthcoming PR)

anorth:
I know Election PoSt might have interfered with some bits here, but could we converge and land the rest of it soon?

src/algorithms/expected_consensus/_index.md
```diff
@@ -197,11 +231,17 @@ Due to the existence of potential forks in EC, a miner can try to unduly influen
...
+Any node that detects any of the above events should submit both block headers to the `StoragePowerActor`'s `ReportConsensusFault` method. The "slasher" will receive a portion (TODO: define how much) of the offending miner's {{<sref pledge_collateral>}} as a reward for notifying the network of the fault.
 (TODO: FIP of submitting commitments to block headers to prevent miners censoring slashers in order to gain rewards).
 
 It is important to note that there exists a third type of consensus fault directly reported by the `CronActor` on `StorageDeal` failures via the `ReportUncommittedPowerFault` method:
-- (3) `uncommitted power fault` which occurs when a miner fails to submit their `PostProof` and is thus participating in leader election with undue power (see {{<sref storage_faults>}}).
+- (4) `uncommitted power fault` which occurs when a miner fails to submit their `PostProof` and is thus participating in leader election with undue power (see {{<sref storage_faults>}}).
```
jbenet: - PoSt (and PoRep) should not be mentioned in EC. no storage specific stuff should be mentioned here (this is very close, it only appears a little). Doing so breaks through / spaghettifies abstractions, which is painful. 
- Define the things EC needs from other modules (ie the power accounting, the power fault, etc.) and then satisfy them in another module that wires the two together (eg SPC). SPC should be the one mentioning PoSt, and how failure to submit post triggers the uncommitted power fault. etc.

src/systems/filecoin_mining/storage_proving/post_generator.md
```diff
@@ -1,92 +0,0 @@
----
-title: PoSt Generator
```
jbenet: - not sure it makes sense to remove this component. this is meant to be a clear explanation of how the `StorageProving` subsystem creates posts, submits them etc. 
- The EC work should not talk about PoSt Submission. SPC can mention what's needed. We need a component (part of storage proving) that describes how (in detail) an implementation needs to produce a post.
- i would expect most of the Election-PoSt prose to land here

jbenet:
Thanks @sternhenri ! 👏

src/algorithms/expected_consensus/_index.md
sternhenri: Fair enough. We just mentioned this storage fault as a "consensus fault" of sorts so for the sake of keeping things in the same place decided to include it here but will break out.

src/systems/filecoin_mining/storage_proving/post_generator.md
sternhenri: Agreed, the large issue was mostly that this md was a full dup of the `storage_mining/mining_cycle.md` and content was better suited there. Will recreate a `post_generator` for the `Election-PoSt`

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.md
nikkolasg: If i understand correctly, that means a miner does a sealing, then has to submit a post after that (when he's elected of after the "period" or when he's "surprised"). 
If so:
We can do this, but I don't think it's useful to do so, since a sealing already contains a "post" somehow, so it's already sure the miner have spend that much amount of storage. The next post proves that the miner *kept* that storage for that duration. So after a sealing, the power is already there and proven. Not sure if it is needed to require this additional delay. In short, tt definitely doesn't hurt, but i'm not sure if this is needed.

src/systems/filecoin_nodes/key_store/key_store.id
nikkolasg: >  I would argue we may want to disable the ability to have many workers per owner in v1.

yeah me too, I remember that the rationals were not clearly expressed last time we discussed this and were leaning towards being opinions rather than fact-based decisions

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.md
anorth: Only keeping storage for a duration should count for power, so has to be in arrears. This is also needed for symmetry with the possible lag in the network detecting that a sector has been dropped.

src/systems/filecoin_blockchain/struct/block_producer/_index.md
```diff
@@ -92,20 +76,18 @@ To create a block, the eligible miner must compute a few fields:
...
 
-They then apply the messages state transition, and generate a receipt for it containing the total gas actually used by the execution, the executions exit code, and the return value (see [receipt](data-structures.md#message-receipt) for more details). Then, they refund the sender in the amount of `(msg.GasLimit - GasUsed) * msg.GasPrice`. In the event of a message processing error, the remaining gas is refunded to the user, and all other state changes are reverted. (Note: this is a divergence from the way things are done in Ethereum)
+They then apply the messages state transition, and generate a receipt for it containing the total gas actually used by the execution, the executions exit code, and the return value . Then, they refund the sender in the amount of `(msg.GasLimit - GasUsed) * msg.GasPrice`. In the event of a message processing error, the remaining gas is refunded to the user, and all other state changes are reverted. (Note: this is a divergence from the way things are done in Ethereum)
```
anorth: The above paragraphs are a bit confused wrt the parent state calculation. The only time messages must be evaluated is when computing the parent state. The block reward must be paid for those parent blocks, not the block being proposed. Receipts are first generated when evaluating the parent state.
No evaluation is necessary for the messages being proposed in the new block (tho miners may speculatively do so to optimize what they include).

FYI @icorderi @acruikshank  note the divergence from ethereum here. The spec is asking for exact Gas accounting in the case of failures.

sternhenri: Yes. you are completely right. Sorry for the huge lag here, i take it you are updating but let me know if not and I will do so!

Also note that I took a stab at cleaning up what I clearly saw to be stale in this spec. You have a lot of latitude to drive here imho.

PR #617, opened 2019-10-31
Proofs interfaces

porcuquine:
- [x] Provide `filproofs` method to compute commD from `PieceInfo`s.
- [x] Call from `storage_miner_actor` into `storage_market_actor` to get commD from deal IDs.
- [x] Call from `storage_market_actor` into `filproofs` via `StorageProvingSubsystem` to get `commD` from PieceInfos.
- [x] Get randomness and interactive randomness from epochs in `storage_mining_actor`.
- [x] Pass `comm_d`, randomness, etc. to verify seal proof in `storage_mining_actor`.
- [x] Disallow sectors with to much padding (by refusing to compute commD for pieces that total insufficient bytes).

### DIFF
--- a/src/libraries/filcrypto/filproofs/algorithms.go
+++ b/src/libraries/filcrypto/filproofs/algorithms.go
@@ -662,7 +662,7 @@ func ComputeUnsealedSectorCIDFromPieceInfos(sectorSize UInt, pieceInfos []PieceI
 	if rootSize != sectorSize {
 		return unsealedCID, errors.New("Wrong sector size.")
 	}
-	// TODO: Enforce maximum padding allowable padding.
+
 	return UnsealedSectorCID(AsBytes_PieceCID(rootPieceInfo.PieceCID())), nil
 }
 
--- a/src/libraries/filcrypto/filproofs/algorithms.go
+++ b/src/libraries/filcrypto/filproofs/algorithms.go
@@ -9,6 +9,7 @@ import "encoding/binary"
 
 import util "github.com/filecoin-project/specs/util"
 import file "github.com/filecoin-project/specs/systems/filecoin_files/file"
+import piece "github.com/filecoin-project/specs/systems/filecoin_files/piece"
 import sector "github.com/filecoin-project/specs/systems/filecoin_mining/sector"
 import sectorIndex "github.com/filecoin-project/specs/systems/filecoin_mining/sector_index"
 
@@ -16,7 +17,7 @@ type SHA256Hash Bytes32
 type PedersenHash Bytes32
 type Bytes32 []byte
 type UInt = util.UInt
-type PieceInfo = sector.PieceInfo
+type PieceInfo = *sector.PieceInfo_I
 type Label Bytes32
 type Commitment = sector.Commitment
 
@@ -655,15 +656,13 @@ func (sdr *StackedDRG_I) VerifySeal(sv sector.SealVerifyInfo) bool {
 }
 
 func ComputeUnsealedSectorCIDFromPieceInfos(sectorSize UInt, pieceInfos []PieceInfo) (unsealedCID sector.UnsealedSectorCID, err error) {
-
 	rootPieceInfo := computeRootPieceInfo(pieceInfos)
 	rootSize := rootPieceInfo.Size()
 	if rootSize != sectorSize {
 		return unsealedCID, errors.New("Wrong sector size.")
 	}
 	// TODO: Enforce maximum padding allowable padding.
-	return rootPieceInfo.CommP(), nil
-
+	return UnsealedSectorCID(AsBytes_PieceCID(rootPieceInfo.PieceCID())), nil
 }
 
 // commD := rootPieceInfo.CommP()
@@ -743,8 +742,8 @@ func zeroPadding(size UInt) PieceInfo {
 func joinPieceInfos(left PieceInfo, right PieceInfo) PieceInfo {
 	util.Assert(left.Size() == right.Size())
 	return &sector.PieceInfo_I{
-		Size_:  left.Size() + right.Size(),
-		CommP_: UnsealedSectorCID(BinaryHash_SHA256Hash(AsBytes_UnsealedSectorCID(left.CommP()), AsBytes_UnsealedSectorCID(right.CommP()))), // FIXME: make this whole function generic?
+		Size_:     left.Size() + right.Size(),
+		PieceCID_: piece.PieceCID(BinaryHash_SHA256Hash(AsBytes_PieceCID(left.PieceCID()), AsBytes_PieceCID(right.PieceCID()))), // FIXME: make this whole function generic?
 	}
 }
 
@@ -963,13 +962,19 @@ func AsBytes_T(t util.T) []byte {
 }
 
 func AsBytes_UnsealedSectorCID(cid sector.UnsealedSectorCID) []byte {
-	panic("Unimplemented for T")
+	panic("Unimplemented for UnsealedSectorCID")
 
 	return []byte{}
 }
 
 func AsBytes_SealedSectorCID(CID sector.SealedSectorCID) []byte {
-	panic("Unimplemented for T")
+	panic("Unimplemented for SealedSectorCID")
+
+	return []byte{}
+}
+
+func AsBytes_PieceCID(CID piece.PieceCID) []byte {
+	panic("Unimplemented for PieceCID")
 
 	return []byte{}
 }
--- a/src/libraries/filcrypto/filproofs/algorithms.go
+++ b/src/libraries/filcrypto/filproofs/algorithms.go
@@ -658,6 +658,7 @@ func (sdr *StackedDRG_I) VerifySeal(sv sector.SealVerifyInfo) bool {
 func ComputeUnsealedSectorCIDFromPieceInfos(sectorSize UInt, pieceInfos []PieceInfo) (unsealedCID sector.UnsealedSectorCID, err error) {
 	rootPieceInfo := computeRootPieceInfo(pieceInfos)
 	rootSize := rootPieceInfo.Size()
+
 	if rootSize != sectorSize {
 		return unsealedCID, errors.New("Wrong sector size.")
 	}
--- a/src/libraries/filcrypto/filproofs/algorithms.go
+++ b/src/libraries/filcrypto/filproofs/algorithms.go
@@ -167,7 +167,7 @@ func (sdr *StackedDRG_I) Seal(sid sector.SectorID, subsectorData [][]byte, rando
 	var allKeyLayers [][]byte
 	var replicas [][]byte
 	var sealSeeds []sector.SealSeed
-	var subsectorCommDs []SHA256Hash
+	var subsectorCommDs []sector.UnsealedSectorCID
 	var subsectorCommDTreePaths []file.Path
 
 	for i, data := range subsectorData {
@@ -209,7 +209,7 @@ func (sdr *StackedDRG_I) Seal(sid sector.SectorID, subsectorData [][]byte, rando
 	return &result
 }
 
-func (sdr *StackedDRG_I) SealSubsector(subsectorIndex int, sid sector.SectorID, data []byte, randomness sector.SealRandomness) ([][]byte, []byte, sector.SealSeed, SHA256Hash, file.Path) {
+func (sdr *StackedDRG_I) SealSubsector(subsectorIndex int, sid sector.SectorID, data []byte, randomness sector.SealRandomness) ([][]byte, []byte, sector.SealSeed, sector.UnsealedSectorCID, file.Path) {
 	commD, commDTreePath := ComputeDataCommitment(data)
 	sealSeed := computeSealSeed(sid, subsectorIndex, randomness, commD)
 	nodeSize := int(sdr.NodeSize())
@@ -232,7 +232,7 @@ func (sdr *StackedDRG_I) GenerateCommitments(replica []byte, keyLayers [][]byte)
 	return commC, commRLast, commR, commCTreePath, commRLastTreePath
 }
 
-func computeSealSeed(sid sector.SectorID, subsectorIndex int, randomness sector.SealRandomness, commD Commitment) sector.SealSeed {
+func computeSealSeed(sid sector.SectorID, subsectorIndex int, randomness sector.SealRandomness, commD sector.UnsealedSectorCID) sector.SealSeed {
 	var proverId []byte // TODO: Derive this from sid.MinerID()
 	sectorNumber := sid.Number()
 
@@ -241,7 +241,7 @@ func computeSealSeed(sid sector.SectorID, subsectorIndex int, randomness sector.
 	preimage = append(preimage, littleEndianBytesFromUInt(UInt(sectorNumber), 8)...)
 	preimage = append(preimage, littleEndianBytesFromInt(subsectorIndex, 8)...)
 	preimage = append(preimage, randomness...)
-	preimage = append(preimage, commD...)
+	preimage = append(preimage, Commitment_UnsealedSectorCID(commD)...)
 
 	// FIXME: Implement
 	return sector.SealSeed{}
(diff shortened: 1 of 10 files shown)


src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -574,19 +574,26 @@ func (a *StorageMinerActorCode_I) DeclareFaults(rt Runtime, faultSet sector.Comp
...
+
+	receipt := rt.Send(&msg.InvocInput_I{
+		To_:     addr.StorageMarketActorAddr,
+		Method_: actor.MethodGetUnsealedCIDForDealIDs,
+		Params_: params,
+		Value_:  rt.ValueSupplied(), // TODO: figure out where tx fee should come from
+	})
 
-	var pieceInfos []sector.PieceInfo // = make([]sector.PieceInfo, 0)
+	if receipt.ExitCode() == exitcode.InvalidSectorPacking {
```
jzimmerman: By default, a failure exit code in a `Send` call will cascade, causing the current method to also fail immediately with a `MethodSubcallError`. To prevent this and check the error code manually, one must call `SendAllowingErrors` instead of `Send`.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -597,16 +604,14 @@ func (st *StorageMinerActorState_I) _isSealVerificationCorrect(rt Runtime, onCha
...
 			Partitions_:     st.Info().Partitions(),
 		},
 
-		// TODO: get Randomness sector.SealRandomness using onChainInfo.Epoch
-		//Randomness_:
-		// TODO: get InteractiveRandomness sector.SealRandomness using onChainInfo.InteractiveEpoch
-		//InteractiveRandomness_:
-		PieceInfos_: pieceInfos,
+		Randomness_:            sector.SealRandomness(rt.Randomness(onChainInfo.SealEpoch(), 0)),
+		InteractiveRandomness_: sector.InteractiveSealRandomness(rt.Randomness(onChainInfo.InteractiveEpoch(), 0)),
```
jzimmerman: It might be problematic to hard-code `0` as the offset here, in case other code accidentally uses `0` also. This likely merits further discussion, but for the moment it seems safest to me to treat these offsets in a similar fashion to MethodNums in the VM (in the sense that we should take special care to enforce their uniqueness).

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -312,6 +317,44 @@ func (a *StorageMarketActorCode_I) GetLastDealExpirationFromDealIDs(rt Runtime,
...
+	ret := SPS.ComputeUnsealedSectorCID(sectorSize, pieceInfos)
+
+	switch ret.Which() {
+	case proving.StorageProvingSubsystem_ComputeUnsealedSectorCID_FunRet_Case_unsealedSectorCID:
+		return rt.ValueReturn(util.Bytes(sector.UnsealedSectorCID(ret.As_unsealedSectorCID())))
+	case proving.StorageProvingSubsystem_ComputeUnsealedSectorCID_FunRet_Case_err:
+		return rt.ErrorReturn(exitcode.InvalidSectorPacking)
+	}
+
+	// Unreachable -- just make the compiler happy.
```
jbenet: adding the `default` case to the switch is probably more idiomatic

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -312,6 +317,44 @@ func (a *StorageMarketActorCode_I) GetLastDealExpirationFromDealIDs(rt Runtime,
 	return lastDealExpiration
 }
 
+func (a *StorageMarketActorCode_I) GetUnsealedCIDForDealIDs(rt Runtime, sectorSize util.UVarint, dealIDs []deal.DealID) vmr.InvocOutput {
```
jbenet: this souldn't be here in the `StorageMarketActor` -- this is chain code in the VM. this should be in the `SectorBuilder` i think. See the `SectorBuilder.NextSector()` function call (doesnt have code yet), i think this belongs there

src/systems/filecoin_mining/sector/sealing.id
```diff
@@ -26,7 +27,7 @@ type SealVerifyInfo struct {
     SealCfg
     Randomness             SealRandomness
     InteractiveRandomness  InteractiveSealRandomness
-    PieceInfos             [PieceInfo]
+    UnsealedCID            UnsealedSectorCID  // CommD
```
jbenet: aren't we using the `PieceInfos` to verify the sector on chain? how are we going to make sure `UnsealedSectorCID` (`CommR`) relates to relevant `PieceCIDs` (`CommPs`)?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -574,19 +574,26 @@ func (a *StorageMinerActorCode_I) DeclareFaults(rt Runtime, faultSet sector.Comp
...
-	// TODO: verify seal @nicola
-	// TODO: st.verifySeal(sectorID SectorID, comm sector.OnChainSealVerifyInfo, proof SealProof)
-
-	// verifySeal will also generate CommD on the fly from CommP and PieceSize
+	sectorSize := st.Info().SectorSize()
+	dealIDs := onChainInfo.DealIDs()
+	params := make([]actor.MethodParam, 1+len(dealIDs))
+	// TODO: serialize method param as {sectorSize,  DealIDs...}.
+
+	receipt := rt.Send(&msg.InvocInput_I{
```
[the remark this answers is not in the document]
jbenet: meaning the following?

```go
func (st *StorageMinerActorCode_I) _isSealVerificationCorrect(rt Runtime, onChainInfo sector.OnChainSealVerifyInfo) bool
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -574,19 +574,26 @@ func (a *StorageMinerActorCode_I) DeclareFaults(rt Runtime, faultSet sector.Comp
...
+		To_:     addr.StorageMarketActorAddr,
+		Method_: actor.MethodGetUnsealedCIDForDealIDs,
+		Params_: params,
+		Value_:  rt.ValueSupplied(), // TODO: figure out where tx fee should come from
```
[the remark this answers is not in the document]
jbenet: yeah we dont need to send value here

src/systems/filecoin_vm/actor/actor.go
```diff
@@ -6,6 +6,8 @@ const (
 	MethodSend        = MethodNum(0)
 	MethodConstructor = MethodNum(1)
 	MethodCron        = MethodNum(2)
+
+	MethodGetUnsealedCIDForDealIDs = MethodNum(99)
```
jbenet: we prob dont need this here, this should prob live in the same file as the `StorageMarketActor` code?

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -312,6 +317,44 @@ func (a *StorageMarketActorCode_I) GetLastDealExpirationFromDealIDs(rt Runtime,
...
+
+	switch ret.Which() {
+	case proving.StorageProvingSubsystem_ComputeUnsealedSectorCID_FunRet_Case_unsealedSectorCID:
+		return rt.ValueReturn(util.Bytes(sector.UnsealedSectorCID(ret.As_unsealedSectorCID())))
+	case proving.StorageProvingSubsystem_ComputeUnsealedSectorCID_FunRet_Case_err:
+		return rt.ErrorReturn(exitcode.InvalidSectorPacking)
+	}
+
+	// Unreachable -- just make the compiler happy.
+	return rt.ErrorReturn(exitcode.SystemError(exitcode.MethodSubcallError))
```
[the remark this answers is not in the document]
porcuquine: Yeah, but that doesn't prevent the `missing return at end of function`.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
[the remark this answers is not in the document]
porcuquine: How would I then get the `MinerInfo`, currently access with `st.Info()` — where `st` is a `StorageMinerActorState_I`?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
porcuquine: I think I must not understand the purpose/meaning of the offset here. Can you clarify?

src/systems/filecoin_markets/storage_market/storage_market_actor.go
porcuquine: Unfortunately, that also does not seem to save me from `missing return at end of function`.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
porcuquine: I'm not sure I follow. At the point that we are verifying a Seal proof, we need access to the deals which were actually put on chain — in order to determine whether or not the submitted proof actually corresponds to the claimed deals.

src/systems/filecoin_mining/sector/sealing.id
porcuquine: This is exactly what is computed by `StorageMarketActorCode_I.GetUnsealedCIDForDealIDs` (which relies indirectly on `filproofs` to do so). We populate the `SealVerifyInfo` only with the correct `UnsealedCID`. If the `PieceInfos` themselves were invalid (say, too large for the sector), then we will fail earlier without an `UnsealedSectorCID`.

src/systems/filecoin_vm/actor/actor.go
porcuquine: Won't it be hard to track `MethodNum` duplication if we don't have a centralized registry lexically in the code?

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -735,12 +742,12 @@ func zeroPadding(size UInt) PieceInfo {
 func joinPieceInfos(left PieceInfo, right PieceInfo) PieceInfo {
 	util.Assert(left.Size() == right.Size())
 	return &sector.PieceInfo_I{
-		Size_:  left.Size() + right.Size(),
-		CommP_: UnsealedSectorCID(BinaryHash_Blake2sHash(AsBytes_UnsealedSectorCID(left.CommP()), AsBytes_UnsealedSectorCID(right.CommP()))), // FIXME: make this whole function generic?
+		Size_:     left.Size() + right.Size(),
+		PieceCID_: piece.PieceCID(BinaryHash_Blake2sHash(AsBytes_PieceCID(left.PieceCID()), AsBytes_PieceCID(right.PieceCID()))), // FIXME: make this whole function generic?
```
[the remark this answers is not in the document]
porcuquine: This is an old comment, but the general answer is that until generics land in detail, the easiest way to use them involves a little hand-waving. We can pay off that debt when we have something concrete and compilable to problem-solve with. In other words, can we just ignore the issue for now?

src/systems/filecoin_vm/actor/actor.go
anorth: They don't need to be globally unique, only scoped to each actor.

porcuquine: I see, thanks. I'm not up to speed so appreciate the tutorial.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jbenet: - the offset is used to derive randomness. think of `rt.Randomness(onChainInfo.InteractiveEpoch(), 0)` like `H(epoch || <randomness from vrf @ epoch> || 0)` <-- the 0 is an additional input that lets the caller derive different randomness values for the same epoch.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -312,6 +319,37 @@ func (a *StorageMarketActorCode_I) GetLastDealExpirationFromDealIDs(rt Runtime,
...
+	}
+
+	Release(rt, h, st)
+
+	SPS := new(proving.StorageProvingSubsystem_I)
+	ret := SPS.ComputeUnsealedSectorCID(sectorSize, pieceInfos)
+
+	switch ret.Which() {
+	case proving.StorageProvingSubsystem_ComputeUnsealedSectorCID_FunRet_Case_unsealedSectorCID:
+		return rt.ValueReturn(util.Bytes(sector.UnsealedSectorCID(ret.As_unsealedSectorCID())))
```
jbenet: - we dont have access to `StorageProvingSubsystem` from within the VM, only `filproofs`. if there's helpers in there that we need, maybe we should move them out to filproofs, or make another library accessible.
- the reason `StorageProvingSubsystem` doesn't make sense here is that that whole subsystem is an instantiated object/system that only lives in a subset of filecoin nodes, the storge miners.
- in the VM, we need something that is callable in all filecoin nodes, without running any of the storage mining stuff. eg libraries -- calling `filproofs` directly

src/systems/filecoin_mining/sector/sealing.id
jbenet: got it

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -312,6 +319,37 @@ func (a *StorageMarketActorCode_I) GetLastDealExpirationFromDealIDs(rt Runtime,
...
+	for index, deal := range st.Deals() {
+		proposal := deal.Proposal()
+		pieceSize := util.UInt(proposal.PieceSize().Total())
+		totalPieceSize += pieceSize
+		pieceInfos[index].PieceCID_ = proposal.PieceCID()
+		pieceInfos[index].Size_ = pieceSize
+	}
+
+	Release(rt, h, st)
+
```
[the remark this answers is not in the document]
porcuquine: @jbenet Got it: this is actually how I had it originally (more or less), but @whyrusleeping noted that passing the `PieceInfo`s through the VM would be difficult and/or expensive from an implementation perspective. I made this change and more convenient proofs API to unblock him. I can change back, but this might be a useful place for you and he to discuss VM pragmatics/philosophy (which will also be valuable for me, since I'm still coming up to speed on the issues). cc: @jzimmerman @zixuanzh

src/systems/filecoin_markets/storage_market/storage_market_actor.go
porcuquine: Got it. I wasn't sure what the VM handle on verification should be. Will change.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
whyrusleeping: moving potentially large objects across vm boundaries is costly (serialization cost and allocations for that must be paid, you can't pass in-memory objects between actors). Pulling all of these out of the store only to shove them in an array that gets serialized, thrown off to another actor, and then unmarshaled back into that array, only to be reduced to a single hash immediately, is unnecessary.

We should work hard to make everything in the deal flow as low cost as possible, its already going to be tight, don't need to make things worse on ourselves.

We should strive to not have any methods that require potentially large complex structures to move between actors.

Additionally, I can see how this method is very relevant to the storage market, this can be framed as a helper that lets you easily check if your deal is part of a bigger hunk of data.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
porcuquine: Done.

porcuquine:
@zixuanzh This is rebased, and I've addressed requests apart from the outstanding issue which, I think, requires some further discussion between @jbenet and @whyrusleeping (and/or others). This decision touches on important design issues around actor/VM usage, so is worth having.

Link since GH is burying the discussion and hiding from the conversation page (for me at least): https://github.com/filecoin-project/specs/pull/617#discussion_r342503601

src/libraries/filcrypto/filproofs/algorithms.go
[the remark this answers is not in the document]
jzimmerman: Yes, I think so -- in the sense that we can just simulate generics when we need them, by writing a one-off type `F_A` where we would like to write `F<A>`.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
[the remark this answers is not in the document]
jzimmerman: This is a frustratingly missing static analysis -- I usually get around this by writing something like:

```go
function_statically_known_to_error_out(); panic("unreachable")
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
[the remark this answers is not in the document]
jzimmerman: You would need to acquire the state within the method:

```go
h, st := a.State(rt)
// ... st.Info() ...
Release(rt, h, st)        // if no modifications made; or
UpdateRelease(rt, h, st)  // if modifications made
```

src/systems/filecoin_vm/actor/actor.go
jzimmerman: (@porcuquine Although, they do need to avoid colliding with the reserved constants 0, 1, 2.)

src/systems/filecoin_markets/storage_market/storage_market_actor.go
jzimmerman: @porcuquine @jbenet @zixuanzh See also https://github.com/filecoin-project/go-filecoin/pull/3636#discussion_r344450703

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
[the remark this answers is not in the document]
porcuquine: Okay, I think I've done that now.

src/systems/filecoin_vm/actor/actor.go
porcuquine: Yeah, I chose 3 for that reason: `	MethodGetUnsealedCIDForDealIDs = actor.MethodNum(3)`.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
porcuquine: I'll be happy to change this again once we decide on a mechanism for exposing only the intended surface of extra-VM calls to actors.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
jbenet: > (serialization cost and allocations for that must be paid, you can't pass in-memory objects between actors).

i think this is an implementation problem? The requirement is a safe memory model, that ensures no mutable memory is shared across actor calls to ensure no changes are made. but that doesn't require serialization on the spot. (probably don't want to do that, will be really expensive, especially for cases where calls fail or run out of gas, which will need to be reverted as a whole. (dont want to hit disk every time)

For example, an implementation with functional, immutable data structures and good caching would do this very cheaply and **_safely_**, by passing references to immutable in-memory data structures with lazy serialization, or to CIDs that load quickly because the object is kept in cache until the VM invocation is finished

> We should work hard to make everything in the deal flow as low cost as possible, its already going to be tight, don't need to make things worse on ourselves.

Very much agree. And we need to not make things worse on ourselves in terms of protocol correctness and verification. Implementing and verifying all of the code is drastically easier when modules are well scoped and concerns don't leak through

src/systems/filecoin_markets/storage_market/storage_market_actor.go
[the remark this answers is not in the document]
jbenet: - >  more convenient proofs API to unblock him
   
    sgtm. this change says nothing about filproofs API. I think that API should definitely be optimized for callers from anywhere about the specific thing they need.
- This doesn't need to block testnet in implementation. I'm OK with whatever @whyrusleeping wants to do there
- But in the spec, (spec tests, and impl for mainnet) all sector & proof things in the VM need to be scoped to `StorageMinerActor`.  This is an important scope isolation thing.
- We'll need to sort out the larger performance optimization issue. @whyrusleeping is very right to be concerned about this
- This call (returning pieceInfos) is small compared to others, and all of those are small compared to the wasted costs of serializing and writing through during failures, including gas failures. Before mainnet, we'll end up running into these problems and likely optimize things in the VM itself

PR #620, opened 2019-11-04
ElectionPoSt

zixuanzh:
EDIT: 20 Nov
** In this PR:**
- [x] ElectionPoSt is implemented in Storage Miner
- [ ] ElectionPoSt is implemented in Expected Consensus
- [x] - Both SubmitElectionPoSt and SubmitCleanupPoSt call _onSuccessfulPoSt upon PoSt verification.
- [x] SubmitElectionPoSt assumes successful PoSt verification (to be checked by StoragePowerConsensusSubsystem as part of block verif). Otherwise, the block will not be included.
- [x] Add single-message ElectionPoSt, where miners are required to submit a PoSt proof whenever they mine a block.
- [x] Cleanup works as follows: storageMinerActor only responds to CleanupPoSts in appropriate challenge period
 - [x] Faults must be recovered before cleanup takes effect on power

**Remains for this PR:**
- [x] @porcuquine can you help massage interfaces so they fit with what Proofs provides. Specifically looking at any instances of `GenerateCandidates` and `GeneratePoSt`/`GeneratePoStProof` across the StorageProvingSubsystem and PoStGenerator :D
- [x] @porcuquine any markdown you wish to add on proofs side, esp data struct definitions
- [x] Settle cleanup question with @jbenet 
- [x] @sternhenri to add markdown
- [x] @sternhenri final pass through all pieces both in sectors, proofs, ec code, to ensure all is there

**Remains for future PR:**
- [ ] Flesh out both PoSt validation steps using proofs library


src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -447,6 +457,42 @@ func (a *StorageMinerActorCode_I) SubmitPoSt(rt Runtime, postSubmission poster.P
...
+
+// Called by StoragePowerConsensus subsystem after verifying the Election proof
+// and verifying the PoSt proof within the block. (this happens outside the VM)
+// Assume ElectionPoSt has already been successfully verified when the function gets called.
+func (a *StorageMinerActorCode_I) SubmitElectionPoSt(rt Runtime, postSubmission poster.PoStSubmission) InvocOutput {
+
+	h, st := a.State(rt)
+	st.ChallengeStatus().Impl().OnNewChallenge(rt.CurrEpoch())
+	UpdateRelease(rt, h, st)
+
```
jbenet: I would add the following comment in here:

```go
// we do not need to verify post submission here, as this should have already been done
// outside of the VM, in StoragePowerConsensus Subsystem. Doing so again would waste
// significant resources, as proofs are expensive to verify.
//
// notneeded := a._verifyPoStSubmission(rt, postSubmission)
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -447,6 +457,42 @@ func (a *StorageMinerActorCode_I) SubmitPoSt(rt Runtime, postSubmission poster.P
...
 
 	return rt.SuccessReturn()
+
+}
+
+// Called by StoragePowerConsensus subsystem after verifying the Election proof
+// and verifying the PoSt proof within the block. (this happens outside the VM)
+// Assume ElectionPoSt has already been successfully verified when the function gets called.
+func (a *StorageMinerActorCode_I) SubmitElectionPoSt(rt Runtime, postSubmission poster.PoStSubmission) InvocOutput {
+
```
jbenet: add a `//TODO() // TODO: validate caller` in here too. 
also state that:
- the caller MUST be the miner who won the block (who won the block should be callable as a a VM runtime call)
---
- we also need to enforce that this call happens only once per block, OR make it not callable by special privileged messages

src/systems/filecoin_mining/storage_mining/_index.md
```diff
@@ -27,3 +27,11 @@ Filecoin Storage Mining Subsystem
 # Sector in StorageMiner State Machine (both)
 
 {{< diagram src="diagrams/sector_fsm.dot.svg" title="Sector State Machine (both)" >}}
+
+# ElectionPoSt
+
+To enable short PoSt response time, miners are required submit a PoSt when they are elected to mine a block, hence PoSt on election or ElectionPoSt. When miners win a block, they need to immediately generate a PoStProof and submit that along with the ElectionProof. Both the ElectionProof and PoStProof are checked at Block Validation by `StoragePowerConsenusSubsystem`. When a block is included, a special message is added that calls `SubmitElectionPoSt` which will process sector updates in the same way as successful `SubmitSurprisePoSt` do. Note that it will be a single PoStProof in a single `SubmitElectionPoSt` message instead of a Commit-then-Prove scheme.
```
jbenet: This should discuss:
- the construction of the post itself (what goes into it, how we sample challenges, how we prove it, etc -- maybe wait till that finalizes)
- how we create the message to put into the VM

src/systems/filecoin_mining/storage_mining/_index.md
```diff
@@ -27,3 +27,11 @@ Filecoin Storage Mining Subsystem
...
+
+# SurprisePoSt
+
+The chain keeps track of the last time that a miner received and submitted a Challenge (Election or Surprise). It randomly samples X miners to surprise at every Epoch. For every miner sampled, a `NotifyOfPoStSurpriseChallenge` is issued and sent as an on-chain message to surprised `StorageMinerActor`. However, if the `StorageMinerActor` is already proving a SurprisePoStChallenge (`IsChallenged` is True) or the `StorageMinerActor` has received a challenge (by Election or Surprise) within the `MIN_CHALLENGE_PERIOD` (`ShouldChallenge` is False), then the PoSt surprise notification will be ignored and return success.
```
jbenet: note/address why we need both `ElectionPoSt` and `SurprisePoSt`

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jbenet: cc @jzimmerman see above ^ -- re the need for VM runtime to give knowledge on who the miner on the _block_ was (not just tipset) in the invocation call. (not sure this is the only place, if only place we could deal with it another way)

jzimmerman: Sounds good. Re: enforcing that it is only called once per block, I think we can just set a field in the state tree to record the epoch of the actor's most recent PoSt submission (analogous to `LastChallengeEpoch_`), and invalidate any `SubmitElectionPoSt` messages that don't advance the counter.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -53,7 +53,10 @@ func DeserializeState(x Bytes) State {
 ////////////////////////////////////////////////////////////////////////////////
 
 // TODO: placeholder epoch value -- this will be set later
-const MAX_PROVE_COMMIT_SECTOR_EPOCH = block.ChainEpoch(3)
+const MAX_PROVE_COMMIT_SECTOR_PERIOD = block.ChainEpoch(3)
+const MAX_SURPRISE_POST_RESPONSE_PERIOD = block.ChainEpoch(4)
```
jbenet: - this should probably be defined elsewhere -- it's used in a lot of parts of the codebase.
- Make sure to update or remove `POST_PROVING_PERIOD` in `post_generator.go`

src/systems/filecoin_mining/storage_mining/_index.md
zixuanzh: will update readme when more of ElectionPoSt finalizes

src/systems/filecoin_mining/storage_mining/_index.md
```diff
@@ -27,3 +27,11 @@ Filecoin Storage Mining Subsystem
 # Sector in StorageMiner State Machine (both)
 
 {{< diagram src="diagrams/sector_fsm.dot.svg" title="Sector State Machine (both)" >}}
+
+# ElectionPoSt
+
+To enable short PoSt response time, miners are required submit a PoSt when they are elected to mine a block, hence PoSt on election or ElectionPoSt. When miners win a block, they need to immediately generate a PoStProof and submit that along with the ElectionProof. Both the ElectionProof and PoStProof are checked at Block Validation by `StoragePowerConsenusSubsystem`. When a block is included, a special message is added that calls `SubmitElectionPoSt` which will process sector updates in the same way as successful `SubmitSurprisePoSt` do. Note that it will be a single PoStProof in a single `SubmitElectionPoSt` message instead of a Commit-then-Prove scheme.
+
+# SurprisePoSt
```
nikkolasg: I want to make the point that we actually don't need the full surprise post and can cut on the blockchain's memory footprint:

**Context**: One can see SurprisePost as two inner components: the "surprise" component and the "deterministic" component. The "surprise" part is what we understand (randomly sample at every block). The "deterministic" part is when a miner haven't been challenged in a given period (`MIN_CHALLENGE_PERIOD`) then that miner *has* to submit its post.

**Election Post**: In that setting, the way I understand it, *every* miner has to check whether he is eligible for producing a block or not. In a sense, that is the *same* behavior as surprise post already: every miner has to check at every block if they have to submit their post! The only difference is that in ElectionPost this check is private, and in SurprisePost this check is public. The point here is that in both cases, miner **are already on the hook every round**. They have to perform this check every round. 

Hence, I believe we **don't need the surprise component** of SurprisePost, but only the deterministic one. In other words, I would recommend to only spec the fact that if a miner hasn't submitted its post (therefore a block) after `MIN_CHALLENGE_PERIOD`), it has to submit its post, that's it.

jbenet: @nikkolasg yes, for security of power sure, but for the miner's storage & state on chain, no -- we need to clean up the state in the chain automatically, and we need to deal with cases where miner gets faulted but needs to recover. Look at the actor code, sector state machine, etc

jbenet:
We need the consensus parts in this PR too -- otherwise the change isn't near to complete. We need to see what miners need to compute (the sequential challenges, making the snark proofs, etc)

src/systems/filecoin_mining/storage_mining/_index.md
nikkolasg: What you telling about here sounds (1) more about engineering issues rather than security issues (clean up state?) or (2) issues that surprise post is not meant to solve at all (recover faults?). 
- What I am proposing if anything should simplify things rather than adding complexity. 
- We should address engineering issues explicitely and not as a by-product of another (security) feature, as this is a dangerous road that leads to more interconnected components without clear boundaries and with fuzzy links (which I believe we're trying to avoid)

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
anorth: We need to define when this implicit message is called. In a tipset with N blocks, are the N `SubmitElectionPoSt` calls the first three calls? Before or after cron? Or does each block's call precede only that block's (tipset-deduplicated) messages? Or all at the end?

jbenet:
- rebase over master, as @sternhenri's PR landed
- add in the new election-post stuff

cerasusland:
will surprisePoSt and electionPoSt replaced rational post now in used?

sternhenri:
yes @cerasusland

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
sternhenri: @anorth this will have to get done when we do the message passing on block verif in VM (scheduled in prio list)

src/systems/filecoin_mining/storage_mining/_index.md
```diff
@@ -27,3 +27,15 @@ Filecoin Storage Mining Subsystem
 # Sector in StorageMiner State Machine (both)
 
 {{< diagram src="diagrams/sector_fsm.dot.svg" title="Sector State Machine (both)" >}}
+
```
zixuanzh: We can remove this section altogether.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -73,8 +72,8 @@ func (cs *ChallengeStatus_I) OnNewChallenge(currEpoch block.ChainEpoch) Challeng
 	return cs
 }
 
-// Call by either SubmitPoSt or OnMissedPoSt
-// TODO: verify this is correct and if we need to distinguish SubmitPoSt vs OnMissedPoSt
+// Call by _onSuccessfulPoSt or _onMissedSurprisePoSt
+// TODO: verify this is correct and if we need to distinguish _onSuccessfulPoSt vs _onMissedSurprisePoSt
```
zixuanzh: I believe this is correct, we can remove the TODO here.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -85,17 +84,24 @@ func (cs *ChallengeStatus_I) IsChallenged() bool {
...
 	return ret
 }
 
+func (cs *ChallengeStatus_I) ShouldChallenge(currEpoch block.ChainEpoch, minChallengePeriod block.ChainEpoch) bool {
+	// true when currEpoch > LastChallengeEpoch + minChallengePeriod
+	return currEpoch > (cs.LastChallengeEpoch() + minChallengePeriod)
+}
+
+func (st *StorageMinerActorState_I) _shouldChallenge(rt Runtime, networkPower block.StoragePower) bool {
+	// pulled in from constants
```
zixuanzh: Let's define this in constant then and we can define this in ChallengeStatus.ShouldChallenge and this method will just be a pass-through.

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
```diff
@@ -62,21 +60,54 @@ func (sms *StorageMiningSubsystem_I) OnNewRound() {
...
-			sms.blockProducer().GenerateBlock(newEP, newTicket, sms.blockchain().BestChain().HeadTipset(), worker.Address())
+		var input []byte
+		input = append(input, spc.VRFPersonalizationPoSt)
+		input = append(input, randomnessK...)
+
+		postRandomness := worker.VRFKeyPair().Impl().Generate(input).Output()
+		// TODO: add how sectors are actually stored in the SMS
+		allSectors := make([]sector.SectorID, 0)
+
+		challengeTickets := sms.StorageProving().Impl().GeneratePoStCandidates(postRandomness, allSectors)
```
zixuanzh: we can and probably should get this from `ProvingSet` but we can update this in a separate PR.

src/algorithms/post/election_post.md
```diff
@@ -0,0 +1,153 @@
...
+challNumber = NumMiners / ProvingPeriod
+
+// Using the ticket to seed randomness, a miner is picked from the power table for each challenge
+sampledMiners = []
+For i in challNumber:
+    ranHash = H(ticket, i)
+    ranIndex = HashToInt(ranHash) mod len(PowerTable)
+    sampledMiners.append(PowerTable[ranIndex].address)
+```
+Leading to a miner getting randomly sampled once per proving period on expectation.
```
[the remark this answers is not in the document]
sternhenri: I added an explanation. It would be useful for you to read it over and convince yourself it is right. A few design decisions that matter are made here.

src/algorithms/post/election_post.md
```diff
@@ -0,0 +1,153 @@
...
+Only faults which have been reported at challenge time, will be accounted for. If any other faults have occured the miner will likely fail to submit a valid PoSt for this proving period.
+
+Any faulted sectors will not count toward miner power in {{<sref expected_consensus>}}.
+
+## Fault Recovery
+
+In order to recover from faults (and make the faulted sectors active once more), a miner must mark the faults as `recovered` and then submit a PoSt proving the recovering sectors.
+When such a PoSt proof is successfully submitted all faults are reset and assumed to be recovered. A miner must either (1) resolve a faulty sector and accept challenges against it in the next proof submission, (2) report a sector faulty again if it persists but is eventually recoverable, (3) report a sector faulty *and done* if the fault cannot be recovered.
+
+If the miner knows that the sectors are permanently lost, they can submit them as part of the `doneSet`, to ensure they are removed from the proving set. In this case, the minwer would pay a portion of their pledge and deal collaterals accordingly.
```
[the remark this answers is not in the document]
sternhenri: mmh this last part? it feels it should be added... can you add? If not, yes let's remove from readme.

src/systems/filecoin_mining/storage_mining/_index.md
sternhenri: moved it to the election-post.md. Thought it useful

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
sternhenri: agreed. if you know how to pull it from the actor, please do!

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.go
```diff
@@ -171,13 +172,20 @@ func (st *StoragePowerActorState_I) _sampleMinersToSurprise(rt Runtime, challeng
...
+	sampledMiners := make([]addr.Address, 0)
+
 	for chall := 0; chall < challengeCount; chall++ {
 		minerIndex := filproofs.TicketToRandomInt(ticket.Output(), chall, ptSize)
 		panic(minerIndex)
-		// hack
+		// hack to turn bigint into int
 		minerIndexInt := 0
-		sampledMiners = append(sampledMiners, allMiners[minerIndexInt])
+		potentialChallengee := allMiners[minerIndexInt]
```
zixuanzh: Let's send all the challenges and handle this at StorageMinerActor to avoid the round trip. Will make the change.

src/algorithms/post/election_post.md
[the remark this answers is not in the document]
zixuanzh: removing this from readme, unless we bring this back later.

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
zixuanzh: will do this in a separate PR

zixuanzh:
@sternhenri LGTM, please approve and merge since I opened the PR. Great work!

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -90,12 +90,12 @@ func (a *StorageMinerActorCode_I) _isChallenged(rt Runtime) bool {
 	return ret
 }
 
-func (cs *ChallengeStatus_I) ShouldChallenge(currEpoch block.ChainEpoch, minChallengePeriod block.ChainEpoch) bool {
-	return currEpoch > (cs.LastChallengeEpoch()+minChallengePeriod) && !cs.IsChallenged()
+func (cs *ChallengeStatus_I) ShouldChallenge(currEpoch block.ChainEpoch) bool {
+	return currEpoch > (cs.LastChallengeEpoch()+PROVING_PERIOD/2) && !cs.IsChallenged()
 }
 
-func (st *StorageMinerActorState_I) _shouldChallenge(rt Runtime, minChallengePeriod block.ChainEpoch) bool {
```
sternhenri: why undo this? it makes more sense to pass it in as a param so the logic can be run entirely from sampleMiners

PR #623, opened 2019-11-06
DRG/Rational calculations

nicola:


jbenet:
Is this ready for review or not yet?

nicola:
This unfortunately is again out of date (well, most part is still correct), since it is modeling the old version.

A new model is coming up soon.

PR #624, opened 2019-11-06
ActiveDeal for Storage Power

zixuanzh:


jbenet:
This LGTM in general -- i havent traced it carefully yet to check correctness. will review tmrw

zixuanzh:
Changes:

zixuanzh:
Changes:
- Only data stored in active deals are counting towards power.
-  `ActiveSectors`, `CommittedSectors`, `RecoveringSectors`, and `FailingSectors` are now `CompactSectorSet` instead of a numeric counter.
- `SectorUtilization` keeps track of utilization and active deal/power for each sector.
- `_getActivePower` and `_getInactivePower` are now defined on `StorageMinerActorState` instead of `PowerTable`.
-  `_updateSectorUtilization` needs to be called before calling `_getActivePower` and `_getInactivePower`.
- A sector is added to `SectorUtilization` when it becomes `Committed` and deleted from `SectorUtilization` when it is cleared.

anorth:
> Only data stored in active deals are counting towards power

My read of the code is that "active deals" means deals in active sectors, whether or not the deal has expired. Is that correct (the term is a bit misleading)? So a miner gains power for all the deals in their sector, and retains that same power regardless of some deals expiring, until the _last_ deal in a sector expires, at which point the sector becomes inactive they lose it all.

If that were not true, and deal expiry caused power expiry, the stakes and difficulty of deal sector packing would go way up, probably resulting in more self-dealing to align expiries.

sprwig:
Can self-dealing be considered as effective storage?

zixuanzh:
@anorth 

> My read of the code is that "active deals" means deals in active sectors, whether or not the deal has expired. Is that correct (the term is a bit misleading)? So a miner gains power for all the deals in their sector, and retains that same power regardless of some deals expiring, until the last deal in a sector expires, at which point the sector becomes inactive they lose it all.

This will no longer be the case after this PR or #629 has landed. Active deals refer to deals that are in an Active Sector and have not yet expired or terminated. 

> If that were not true, and deal expiry caused power expiry, the stakes and difficulty of deal sector packing would go way up, probably resulting in more self-dealing to align expiries.

I think part of the reason why we switched to this model is so that sector packing can be simpler (removal of the maximum padding requirement). Self dealing will still be in the allowable action space for agent in the network and the goal is NOT to make that impossible.

PR #628, opened 2019-11-08
Move from Blake2s to SHA256 hashing in PoRep.

porcuquine:
Just as it seems.


src/libraries/filcrypto/filproofs/feistel.go
```diff
@@ -85,7 +85,7 @@ func feistel(right uint32, key uint32, rightMask uint32) uint32 {
 	data[6] = byte(key >> 8)
 	data[7] = byte(key)
 
-	hash := blake2s.Sum256(data[:])
+	hash := blake2b.Sum256(data[:])
```
zixuanzh: not sure if this is right but just calling it out for another look, everything else LGTM.

PR #629, opened 2019-11-08
Storage Deal Payment

zixuanzh:
EDIT: 23 Nov
**In this PR:**
- [x] specify StorageMarketActor interactions
    - [x] only data in active deals count towards power (update sector power as deal expires through `DealExpirationAMT`
    - [x] keep track of onchain deal state and LastPaymentEpoch through `OnchainDeal` struct
    - [x] specify transitions between deal states, readme in #638 
    - [x] specify deal collateral slashing
    - [x] specify deal payment
    - [x] specify deal expiration
- [x] specify interactions between StorageMinerActor and StorageMarketActor and validation steps
- [x] split StorageMarketActor into StorageMarketActorState and StorageMarketActorCode
    
**Remains for future PRs**
- [ ] power slashing and accounting
- [ ] garbage collection for expired published deals (did not get activated)
- [ ] implement on chain message calls


src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+
+		duration := lastChallengeEndEpoch - activeDeal.LastPaymentEpoch()
+
+		if duration <= 0 {
+			// return for this particular deal instead of aborting the whole operation
+			// rt.Abort("sma._creditStorageDeals: no new payment to be credited.")
+			return
+		}
+
+		fee := actor.TokenAmount(uint64(duration) * uint64(activeDeal.Deal().Proposal().StoragePricePerEpoch()))
```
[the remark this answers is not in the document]
jbenet: readability and reducing errors with money math is nice 👌

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+
+		if fee > activeDeal.LockedStorageFee() {
+			// rt.Abort("sma._creditStorageDeals: cannot unlock more than already locked.")
+			return
```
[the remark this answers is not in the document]
jbenet: actually, is this an error condition? im not sure -- the prose version of what is supposed to happen will help here.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+}
+
+func (st *StorageMarketActorState_I) _slashDeclaredFaults(rt Runtime, dealIDs []deal.DealID) {
+
+	for _, dealID := range dealIDs {
+
+		deal, found := st.ActiveDeals()[dealID]
+
+		if !found {
+			// move on to the next deal and keep slashing
```
jbenet: this might be an error condition in some design versions. im ok with no error, but make it clearer -- say why, eg "the deal may no longer be active, and we are OK with a call to this function referencing inactive deals" or something like that.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+			return
+		}
+
+		// TODO: the exact slash amount is up for change
+		// TODO: check if provider runs out of collateral to slash here
+		deal.Impl().ProviderCollateralRemaining_ -= deal.Deal().Proposal().ProviderCollateralPerEpoch()
```
jbenet: - ideally no direct money arithmetic. Only `transfer`, `lock`, `unlock` functions that make symmetric transfers in exactly the same spot.
- sounds like this is not a transfer/lock/unlock, but merely marking collateral as spent/slashed.
- here, i would declare a function on `StorageMarketActor` 

```go
func (st *StorageMarketActorState_I) _slashDealCollateral(d *ActiveDeal) {
    st.CollateralSlashed += amt
    d.ProviderCollateralRemaining_ -= amt
}
``` 
- the money needs to be tallied up somewhere (added to some running count), exactly symmetrically.
- for easier correctness, accounting money should always be symmetric. that way everything is consistent, money numbers only "move", and are never created or destroyed. (even secondary accounts/tallies of money should be symmetric)

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+		if !found {
+			// move on to the next deal and keep slashing
+			return
+		}
+
+		// TODO: the exact slash amount is up for change
+		// TODO: check if provider runs out of collateral to slash here
+		// TODO: more sever slashing for detected faults vs declared faults
+		// maybe for the duration since the last Epoch that the sector was proven
+		deal.Impl().ProviderCollateralRemaining_ -= deal.Deal().Proposal().ProviderCollateralPerEpoch()
```
jbenet: same as above

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+
+func (st *StorageMarketActorState_I) _slashDetectedFaults(rt Runtime, dealIDs []deal.DealID) {
+
+	for _, dealID := range dealIDs {
+
+		deal, found := st.ActiveDeals()[dealID]
+
+		if !found {
+			// move on to the next deal and keep slashing
+			return
```
jbenet: continue

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+}
+
+func (st *StorageMarketActorState_I) _slashDetectedFaults(rt Runtime, dealIDs []deal.DealID) {
+
+	for _, dealID := range dealIDs {
+
+		deal, found := st.ActiveDeals()[dealID]
+
+		if !found {
+			// move on to the next deal and keep slashing
```
jbenet: also mark w/ error info

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+// TODO: decide what to do with unlocked storage fee here
+func (st *StorageMarketActorState_I) _slashTerminatedFaults(rt Runtime, dealIDs []deal.DealID) {
+
+	for _, dealID := range dealIDs {
+
+		deal, found := st.ActiveDeals()[dealID]
+
+		if !found {
+			// move on to the next deal and keep slashing
+			return
```
jbenet: same as above

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+		}
+
+		delete(st.ActiveDeals(), dealID)
```
jbenet: - this should be in a function like `st._terminateDeal(dealID)` or `st._clearActiveDeal(dealID)` or something. stray deletes are asking for state-keeping trouble. make the transitions very explicit and clear, so state accounting is easy. 
- ie, here in this function we want to call a higher level logical construct ("a SlashTerminatedFaults messages also terminates the deal accounting -- therefore call deal termination, or deal cleanup")

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+		clientAddr := deal.Deal().Proposal().Client()
+		clientCollateral := deal.Deal().Proposal().TotalClientCollateral()
+		st._unlockBalance(rt, clientAddr, clientCollateral+deal.LockedStorageFee())
+
+		// burn all deal.ProviderCollateralRemaining by sending them to TreasuryActor
+		// TODO: Send(deal.ProviderCollateralRemaining)
+	}
+}
+
+func (a *StorageMarketActorCode_I) ProcessSectorDealSlash(rt Runtime, info deal.BatchDealSlashInfo) {
```
jbenet: probably just `ProcessDealSlash`.

jbenet: if we need to enforce the caller is `StorageMinerActor`, then mark a todo for now

src/systems/filecoin_markets/storage_market/storage_market_actor.id
```diff
@@ -2,18 +2,21 @@ import actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
...
-type DealsAMT {deal.DealID: deal.StorageDeal}
+type PublishedDealsAMT {deal.DealID: deal.StorageDeal}
+type ActiveDealsAMT {deal.DealID: deal.ActiveStorageDeal}
 
 type StorageMarketActorState struct {
-    Balances  BalancesHAMT
-    Deals     DealsAMT
+    Balances        BalancesHAMT
+    PublishedDeals  PublishedDealsAMT
+    ActiveDeals     ActiveDealsAMT
```
jbenet: distinction between Published and Active deals not super clear, are these just two different states of the same thing? maybe say so. it looks though than an "ActiveDeal" includes other stuff (some accounting/tallies)

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -109,25 +99,50 @@ func (a *StorageMinerActorCode_I) NotifyOfPoStChallenge(rt Runtime) InvocOutput
 }
 
 func (st *StorageMinerActorState_I) _updateCommittedSectors(rt Runtime) {
-	for sectorNo, sealOnChainInfo := range st.StagedCommittedSectors() {
-		st.Sectors()[sectorNo] = sealOnChainInfo
+	for sectorNo, stagedInfo := range st.StagedCommittedSectors() {
```
jbenet: the name `_updateCommittedSectors` is not clear enough. maybe `_unstage` or `_moveCommittedSectorsToProvingSet` or something clear like this but shorter. maybe `_processStagedCommittedSectors`

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -136,22 +151,66 @@ func (a *StorageMinerActorCode_I) _submitFaultReport(
...
 	panic(faultReport)
 
+	if len(newDeclaredFaults) > 0 {
+		h, st := a.State(rt)
+		slashInfo := st._getDealSlashInfoFromSectors(rt, newDeclaredFaults.SectorsOn(), deal.SlashDeclaredFaults)
+		Release(rt, h, st)
+
+		// TODO: Send(StorageMarketActor, ProcessSectorDealSlash)
+		panic(slashInfo)
+	}
```
jbenet: there's a lot of code duplication here, which to me is usually a sign of error-prone-ness and need for decomposition, or similar

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -243,6 +302,80 @@ func (a *StorageMinerActorCode_I) _expirePreCommittedSectors(rt Runtime) {
...
+		newExpiredDealIDs := deal.CompactDealSet(make([]byte, totalDealCount))
+
+		if firstExpirationEpoch < currEpoch {
+			// this deal has expired
+			newExpiredDeal := utilizationInfo.DealExpirationQueue().Pop()
+			newUtilization -= newExpiredDeal.PayloadPower()
+			st.SectorUtilization()[sectorNo].Impl().ActiveDealIDs_.Remove(newExpiredDeal.DealID())
+			newExpiredDealIDs.Add(newExpiredDeal.DealID())
+
+			// this is a go while loop
```
jbenet: this is weird, and wrong, what are you trying to do?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -243,6 +302,80 @@ func (a *StorageMinerActorCode_I) _expirePreCommittedSectors(rt Runtime) {
...
+			firstExpirationEpoch = utilizationInfo.DealExpirationQueue().Peek().Expiration()
```
jbenet: i think you can just remove this line

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -243,6 +302,80 @@ func (a *StorageMinerActorCode_I) _expirePreCommittedSectors(rt Runtime) {
...
+
+		}
+
+		st.SectorUtilization()[sectorNo].Impl().CurrUtilization_ = newUtilization
+		ret = ret.Extend(newExpiredDealIDs)
+	}
+
+	return ret.DealsOn()
+
+}
```
jbenet: this function is complicated, hard to read, and doing a lot. maybe break it apart w/ a helper function or two that make the logic and intent super clear, so we can read it very simply.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
```diff
@@ -557,22 +690,53 @@ func (a *StorageMinerActorCode_I) DeclareFaults(rt Runtime, faultSet sector.Comp
...
+
+	h, st := a.State(rt)
+
+	utilization, found := st.SectorUtilization()[sectorNo]
+
+	if !found {
+		rt.Abort("sm.GetSectorPaymentInfo: sector number not found.")
+	}
+
+	isSectorActive := st.SectorTable().Impl().ActiveSectors_.Contain(sectorNo)
```
jbenet: - `sectorIsActive` 
- so that below it reads "if sector is active { ... }"

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
-	// TODO provider lose power for the FaultSet but not PledgeCollateral
-	// }
-	panic("TODO")
+	return ret
 }
 
 // Call by StorageMinerActor at CommitSector
-func (a *StorageMarketActorCode_I) GetLastDealExpirationFromDealIDs(rt Runtime, dealIDs []deal.DealID) block.ChainEpoch {
-
+func (a *StorageMarketActorCode_I) GetInitialUtilizationInfo(rt Runtime, dealIDs []deal.DealID) sector.SectorUtilizationInfo {
```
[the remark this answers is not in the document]
zixuanzh: this is because `StorageMinerActor` does not have access to details in a StorageDeal

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
zixuanzh: go through all expiring deals and settle payments. trying to piggyback off the heavy lifting in updateSectorUtilization

src/systems/filecoin_markets/storage_market/storage_market_actor.id
zixuanzh: Yes, they are two states of a deal, PublishedDeals are created when published but they only become active after `ProveCommit`. `ActiveDeal` also keeps track of all the accounting information related to payment and collateral that are not needed for PublishedDeals. Also not all PublishedDeals will make it into ActiveDeals (hence we saw delete and abort - will fix that one too)

src/systems/filecoin_markets/deal/deal.id
```diff
@@ -51,6 +53,32 @@ type StorageDeal struct {
...
+    LockedStorageFee                         actor.TokenAmount
+    UnlockedStorageFee                       actor.TokenAmount
+    LastPaymentEpoch                         block.ChainEpoch
+
+    UnlockStorageFee(fee actor.TokenAmount)  bool
+}
+
+type DealExpirationQueueItem struct {
+    DealID
+    PayloadPower  block.StoragePower
```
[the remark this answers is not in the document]
zixuanzh: want to be more specific here until we align on what exactly counts toward power in a Piece. There is PayloadSize and OverheadSize in PieceSize.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -220,17 +214,16 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+	// miner should add more balance and try again
 	for i, newDeal := range newStorageDeals {
 		if st._validateNewStorageDeal(rt, newDeal) {
 			st._lockFundsForStorageDeal(rt, newDeal)
 			id := st._generateStorageDealID(rt, newDeal)
-			st.Deals()[id] = newDeal
+			st.PublishedDeals()[id] = newDeal
 			response[i] = PublishStorageDealSuccess(id)
 		} else {
 			response[i] = PublishStorageDealError()
```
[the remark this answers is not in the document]
zixuanzh: think we should fail the call entirely, will update this

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+		publishedDeal, publishedFound := st.PublishedDeals()[dealID]
+		if !publishedFound {
+			rt.Abort("sma.VerifyPublishedDealIDs: unpublished deal ID.")
+		}
+
+		currEpoch := rt.CurrEpoch()
+
+		if publishedDeal.Proposal().StartEpoch() < currEpoch {
+			delete(st.PublishedDeals(), dealID)
+			rt.Abort("sma.VerifyPublishedDealIDs: deal has already started.")
```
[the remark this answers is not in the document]
zixuanzh: agree, will move garbage collect to somewhere else

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+		publishedDeal, publishedFound := st.PublishedDeals()[dealID]
+		if !publishedFound {
+			delete(st.PublishedDeals(), dealID)
+			rt.Abort("sma.ActivateSectorDealIDs: unpublished deal ID.")
+		}
+
+		if publishedDeal.Proposal().EndEpoch() <= rt.CurrEpoch() {
+			delete(st.PublishedDeals(), dealID)
+			rt.Abort("sma.ActivateSectorDealIDs: storage deal has expired.")
+		}
```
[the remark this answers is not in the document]
zixuanzh: same as above, need to some form of garbage collection when a deal is published but not proven for a while

src/systems/filecoin_markets/storage_market/storage_market_actor.go
```diff
@@ -242,74 +235,299 @@ func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDea
...
+		maxUtilization += dealPayloadPower
+
+	}
+
+	initialUtilizationInfo := &sector.SectorUtilizationInfo_I{
+		DealExpirationQueue_: dealExpirationQueue,
+		MaxUtilization_:      maxUtilization,
+		CurrUtilization_:     maxUtilization,
+		ActiveDealIDs_:       activeDealIDs,
+		LastDealExpiration_:  lastExpiration,
```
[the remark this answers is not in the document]
zixuanzh: Looks like we need both SectorUtilizationInfo (a tally of the total) and DealExpirationQueue. The latter keeps track of deal expiration in a sector and update the total tally. We can consider moving ActiveDealIDs from SectorUtilizationInfo into DealExpirationQueue.

src/systems/filecoin_markets/storage_market/storage_market_actor.go
zixuanzh: agree

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
zixuanzh: will do, this function is also where the heavy-lifting of iterating all active deals and update sector utilization happens

src/systems/filecoin_mining/storage_mining/storage_miner_actor.go
jbenet: i think this works:
```go
queue := utilizationInfo.DealExpirationQueue()
for queue.Peek().Expiration() < currEpoch {
    expiredDeal := queue.Pop()
    newUtilization -= expiredDeal.PayloadPower()

    // state changes
    st.SectorUtilization()[sectorNo].Impl().ActiveDealIDs_.Remove(newExpiredDeal.DealID())
    newExpiredDealIDs.Add(newExpiredDeal.DealID())
}
```

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
```diff
@@ -0,0 +1,288 @@
...
+func (st *StorageMarketActorState_I) _activateDeal(rt Runtime, deal deal.OnchainDeal) deal.OnchainDeal {
+
+	dealP := deal.Deal().Proposal()
+	deal.Impl().LastPaymentEpoch_ = dealP.StartEpoch()
+	st.Deals()[deal.ID()] = deal
+
+	return deal
+}
+
+// TODO: consider returning a boolean
```
sternhenri: otherwise the abort is fine, but may want to call such methods "safe+" as in "safeLockBalance" so it is immediately obvious that the checks happen within the method in code calling it.

src/systems/filecoin_markets/deal/deal.id
```diff
@@ -34,6 +33,8 @@ type StorageDealProposal struct {
     ClientCollateralPerEpoch      actor.TokenAmount  // potentially collapse into one with ProviderDealCollateral
 
     ClientBalanceRequirement()    actor.TokenAmount  // (ClientCollateralPerEpoch + StoragePricePerEpoch) * (EndEpoch - StartEpoch)
+    TotalStorageFee()             actor.TokenAmount  // StoragePricePerEpoch * (EndEpoch - StartEpoch)
+    TotalClientCollateral()       actor.TokenAmount  // ClientCollateralPerEpoch * Duration
     ProviderBalanceRequirement()  actor.TokenAmount  // ProviderCollateralPerEpoch * (EndEpoch - StartEpoch)
     Duration()                    block.ChainEpoch
```
sternhenri: Should we actually store duration? It seems misguided given start and end

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+}
+
+func (st *StorageMarketActorState_I) _assertDealExpireAfterMaxProveCommitWindow(rt Runtime, dealP deal.StorageDealProposal) {
+
+	currEpoch := rt.CurrEpoch()
+	dealExpiration := dealP.EndEpoch()
+
+	if dealExpiration <= (currEpoch + sector.MAX_PROVE_COMMIT_SECTOR_EPOCH) {
```
sternhenri: should this be <?

at least X sounds inclusive of X

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
```diff
@@ -0,0 +1,288 @@
...
+	if !found {
+		rt.Abort("sm._getOnchainDeal: dealID not found in Deals.")
+	}
+
+	return deal
+}
+
+func (st *StorageMarketActorState_I) _assertPublishedDealState(rt Runtime, dealID deal.DealID) {
+
+	deal := st._getOnchainDeal(rt, dealID)
```
sternhenri: abort message will be unclear in such cases. consider changing _getOnChainDeal to return both a deal and status (bool) so you can abort in caller method for better logging

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
```diff
@@ -0,0 +1,288 @@
...
+	amountToSlash := dealP.ProviderBalanceRequirement()
+
+	providerBal, found := st.Balances()[dealP.Provider()]
+	if !found {
+		rt.Abort("sma._slashDealCollateral: provider not found in balances.")
+	}
+
+	if providerBal.Locked() < amountToSlash {
+		amountToSlash = providerBal.Locked()
+		// TODO: decide on error handling here
```
sternhenri: slash it all anyways and log error?

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
```diff
@@ -0,0 +1,288 @@
...
+
+	deal := st._getOnchainDeal(rt, dealID)
+	st._assertActiveDealState(rt, dealID)
+
+	dealP := deal.Deal().Proposal()
+	delete(st.Deals(), dealID)
+
+	// return client collateral and locked storage fee
+	clientCollateral := dealP.TotalClientCollateral()
+	lockedFee := st._getStorageFeeSinceLastPayment(rt, deal, dealP.EndEpoch())
```
sternhenri: no. you want to return full fee.

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -0,0 +1,278 @@
...
+func (a *StorageMarketActorCode_I) ProcessDealExpiration(rt Runtime, dealIDs []deal.DealID) {
+	h, st := a.State(rt)
+
+	for _, dealID := range dealIDs {
+
+		expiredDeal := st._getOnchainDeal(rt, dealID)
+		st._assertActiveDealState(rt, dealID)
+
+		dealP := expiredDeal.Deal().Proposal()
+		fee := st._getStorageFeeSinceLastPayment(rt, expiredDeal, dealP.EndEpoch())
```
sternhenri: should this be all remaining fee? If so, may be worth checking below, lest the _unlockBalance inadvertently leave fees stuck in the deal

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -0,0 +1,278 @@
...
+
+		delete(st.Deals(), dealID)
+		st._transferBalance(rt, dealP.Client(), dealP.Provider(), fee)
+
+		// return storage deal collaterals to both miners and client
+		st._unlockBalance(rt, dealP.Provider(), dealP.ProviderBalanceRequirement())
+		st._unlockBalance(rt, dealP.Client(), dealP.TotalClientCollateral())
```
sternhenri: any remaining fee should be returned to client

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
sternhenri: as with balances

src/systems/filecoin_markets/deal/deal.id
zixuanzh: This is a helper function, we can call it FullDuration to be clear and precise.

src/systems/filecoin_markets/storage_market/storage_market_actor.id
```diff
@@ -38,44 +41,77 @@ type StorageMarketActorState struct {
...
+    // Verify the following:
+    //     - dealIDs have been registered in PublishedDeals
+    //     - deals will not expire by the time ProveCommit hits
+    //     - deals have not started
+    // Fail PreCommit if any of above is not true and abort
+    VerifyPublishedDealIDs(rt Runtime, dealIDs [deal.DealID])
+
+    // Call by sm.ProveCommitSector
+    // Verify the following:
+    //     - dealIDs are in PublishedDeals
```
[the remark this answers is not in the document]
zixuanzh: Ack, this is explicitly left as an optimization for later.

src/systems/filecoin_markets/storage_market/storage_market_actor.id
```diff
@@ -38,44 +41,77 @@ type StorageMarketActorState struct {
...
+    ProcessDealSlash(
+        rt           Runtime
+        dealIDs      [deal.DealID]
+        slashAction  deal.StorageDealSlashAction
+    )
 
-    GetLastExpirationFromDealIDs(rt Runtime, dealIDs [deal.DealID]) block.ChainEpoch
+    // batch process storage deal payment
+    // Note: if a sector enters into a Failing state but later recovers, client still pays miner as usual
+    // client only stops paying when a sector gets cleared (expired or faulted)
```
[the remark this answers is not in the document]
zixuanzh: Will update comment here. Clients effectively stop paying after entering into Failing state, they will pay again at the next _onSuccessfulPoSt or stop paying if the sector gets terminated.

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -0,0 +1,278 @@
...
+		ret = append(ret, onchainDeal)
+	}
+
+	UpdateRelease(rt, h, st)
+
+	return ret
+
+}
+
+func (a *StorageMarketActorCode_I) ProcessDealSlash(rt Runtime, dealIDs []deal.DealID, slashAction deal.StorageDealSlashAction) {
```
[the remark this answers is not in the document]
zixuanzh: Not at the moment for StorageMarketActor, happy to change it to only SlashTerminatedFault and get rid of SlashAction here but there will be other SlashActions in StoragePowerActor

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
zixuanzh: now that we removed deal tally to reduce chain foot print, it might be difficult to keep track of remaining fees if there is any. cc @magik6k

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -0,0 +1,278 @@
...
+	// this operation will be unrolled if any of the above triggers a throw
+	for i, newDeal := range newStorageDeals {
+		if st._validateNewStorageDeal(rt, newDeal) {
+			st._lockFundsForStorageDeal(rt, newDeal)
+			id := st._generateStorageDealID(rt, newDeal)
+
+			onchainDeal := &deal.OnchainDeal_I{
+				ID_:               id,
+				Deal_:             newDeal,
+				LastPaymentEpoch_: block.ChainEpoch(0), // 0 = inactive
```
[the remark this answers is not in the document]
zixuanzh: cc @magik6k

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
zixuanzh: I think it should still be <= since it will throw if this is true.

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
zixuanzh: yes, that is the intention. Will fix in future PR with better error handling.

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
zixuanzh: this locked fee is the full fee that has not been unlocked. Not sure if it is possible to pay out full fee, unless we are talking about the same thing here.

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
```diff
@@ -0,0 +1,288 @@
...
+	if !found {
+		rt.Abort("sma._unlockBalance: addr not found.")
+	}
+
+	currBalance.Impl().Locked_ -= amount
+	currBalance.Impl().Available_ += amount
+}
+
+// move funds from locked in client to available in provider
+func (st *StorageMarketActorState_I) _transferBalance(rt Runtime, fromLocked addr.Address, toAvailable addr.Address, amount actor.TokenAmount) {
```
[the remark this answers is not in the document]
zixuanzh: dont think TreasuryAddr is relevant here, removing it. This is transferring balance within BalancesHAMT that miners/clients have locked up.

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -72,7 +85,7 @@ func (a *StorageMarketActorCode_I) WithdrawBalance(rt Runtime, balance actor.Tok
 	senderBalance.Impl().Available_ = senderBalance.Available() - balance
 	st.Balances()[msgSender] = senderBalance
 
-	// TODO send funds to msgSender with `transferBalance` in VM runtime
```
zixuanzh: we are using a different transferBalance here in rt.Send(). This is left as a TODO for future PRs.

zixuanzh:
@sternhenri thanks for the reviewing and addressed them in comments and code. 
- We removed `dealTally` which keeps track of current deal payout to reduce onchain footprint cc @magik6k on how to do this better.
- money methods are intended to be atomic, please let me know how to make that more explicit
- I have removed the treasury actor and added explanation, please let me know if they are unclear.

merging soon for velocity. Thanks!

src/systems/filecoin_markets/storage_market/storage_market_actor.id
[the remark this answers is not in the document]
sternhenri: where is it explicited? may be worth adding as a comment somewhere

src/systems/filecoin_markets/storage_market/storage_market_actor.id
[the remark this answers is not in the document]
sternhenri: I understand but my point is the determination of how much to pay since last successful PoSt should deduct any epochs spent in the Failing state

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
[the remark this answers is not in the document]
sternhenri: let's keep as is

src/systems/filecoin_markets/storage_market/storage_market_actor.id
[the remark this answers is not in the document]
sternhenri: I understand but my point is the determination of how much to pay since last successful PoSt should deduct any epochs spent in the Failing state

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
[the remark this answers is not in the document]
sternhenri: I'm thinking about the withdrawFund call which calls _transferBalance. What should it pass in?

src/systems/filecoin_markets/deal/deal.id
sternhenri: Just added the `@(cached)` call so we know it's just computed from other vlaies

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
sternhenri: take 
`currEpoch == 10`
`MAX_PROVE == 15`
`EndEpoch == 25`

Then in your case it would be triggered whereas in mine it would not, even though EndEpoch is MAX_PROVE away from currEpoch. So we should fix language accordingly: it is either
- at least MAX_PROVE (<), or,
- more than MAX_PROVE (<=)

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
sternhenri: I am saying we should explicitly ensure the lockedFee is the fullFee in the beginning: specifically, the client is not disallowed (I think) from locking more than what the contract entails. This could lead to fees getting stuck here. 
Two options:
1) check and ensure they only lock up full payment (leading to what you say)
2) let them lock up anything, but return full remaining amount (which would be more than fee to end)

sternhenri:
- [ ] Let's add the atomic notes as comments above the method (or make a placeholder noop method for now called `atomic()` called in the beg each time).
- [ ] I think the issue of `remainingFee`vs `totalFee` returned is a real one. There are a couple ways to fix it outlined in comment. 
- [ ] For the `FailingState` edge casee, I do think we should deduct fees from those FailedEpochs (you're paying for storage you didn't get).
- [ ] for the treasury actor I'm just unclear on where `withdrawBalance` is withdrawing from... can you explain? This should be zero-sum so I'm just wondering what the FromB is there.

src/systems/filecoin_markets/storage_market/storage_market_actor.id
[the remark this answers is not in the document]
anorth: FWIW I don't agree the complexity @sternhenri requests is necessary. Failing to prove doesn't mean the data isn't being stored, just that we can't be sure. Miner's can't generally regenerate client deal data - and even if they could, why would the client care? The miner is able to produce their data. IMO only storage power cares about intermittent failure to prove.

Is your rationale that clients are paying for immediate availability of their file at any point during the contract, which they hypothetically would not have had?

zixuanzh:
@sternhenri 

- ack on atomic, cc @jzimmerman if there are any thoughts here
- will fix remaingFee vs totalFee and circle back if there are questions
- I think client is only paying at _onSuccessfulPoSt now so that problem will go away - will go through this agina
- balance is locked in StorageMarketActor, AddBalance send funds to SMA

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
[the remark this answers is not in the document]
zixuanzh: Funds are kept in StorageMarketActor itself, transferBalance is updating the locked vs available fund within StorageMarketActor. WithdrawFund sends fund from StorageMarketActor to the caller.

src/systems/filecoin_markets/storage_market/storage_market_actor.id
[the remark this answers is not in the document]
zixuanzh: I agree with anorth here, as long as the sector is recovered eventually, clients are still paying the miners for the deals contained in the sector.

src/systems/filecoin_markets/storage_market/storage_market_actor.id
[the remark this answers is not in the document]
zixuanzh: Sorry, there was a comment here on this that got removed, perhaps when the method was renamed to ActivateDeals.

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
zixuanzh: I am not sure where the language is coming from, the method is called `_assertDealExpireAfterMaxProveCommit`, in the example you mentioned, the deal is expiring on MaxProveCommit and hence it is not after ProveCommit and hence it will throw.

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
zixuanzh: renaming to _safeGetOnChainDeal and _safeGetBalance

src/systems/filecoin_markets/storage_market/storage_market_actor_state.go
zixuanzh: I just checked, lockedFee is already the full fee in the beginning. Please let me know if this addresses the issue identified.

https://github.com/filecoin-project/specs/blob/master/src/systems/filecoin_markets/storage_market/storage_market_actor_state.go#L195-L196

sternhenri:
Did we close out the above points @zixuanzh ?

zixuanzh:
@sternhenri I believe so maybe except for atomic (they are meant to be atomic but maybe not marked as one)
- remainingFee vs totalFee, dont think it's an issue since lockedFee is already the full fee since the beginning https://github.com/filecoin-project/specs/blob/master/src/systems/filecoin_markets/storage_market/storage_market_actor_state.go#L195-L196
- client only pays at _onSuccessfulPoSt per #696 
- funds are locked in SMA itself

sternhenri:
cool and there's no way for fees to remain locked as i remember we discussed? (ie if remainingFee doesn't add up to totalfee)

PR #630, opened 2019-11-09
Minor VM updates from implementation review

jzimmerman:


src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -366,7 +366,7 @@ func (rtOuter *VMContext) _sendInternal(input InvocInput, ignoreErrors bool) msg
 	rtOuter._refundGasRemaining(toActorMethodGasBound)
 	rtOuter._deductGasRemaining(gasUsed)
 
-	if !ignoreErrors && invocOutput.ExitCode().IsError() {
+	if !catchErrors && invocOutput.ExitCode().IsError() {
```
zixuanzh: To make sure I understand this, as the line suggests, if it is not catching error and there is an error, this function will throw. Correct.

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -275,11 +275,11 @@ func (rt *VMContext) _transferFunds(from addr.Address, to addr.Address, amount a
 
 // TODO: This function should be private (not intended to be exposed to actors).
 // (merging runtime and interpreter packages should solve this)
-func (rt *VMContext) SendToplevelFromInterpreter(input InvocInput, ignoreErrors bool) (
+func (rt *VMContext) SendToplevelFromInterpreter(input InvocInput, catchErrors bool) (
```
jbenet: i think this is a confusing interface, because it will be used with `true` and `false` which will be unclear to read.

```go
rt.SendToplevelFromInterpreter(invIn, true) // what is true?
rt.SendToplevelFromInterpreter(invIn, false) // what is false?
```
for clarity, things like this are clearer:

```go
// way one
rt.SendToplevelFromInterpreter(inv, CatchErrors)
rt.SendToplevelFromInterpreter(inv, ThrowErrors) // or PropagateErrors

// way two
rt.SendToplevelFromInterpreter(inv) // does not catch errors
rt.SendToplevelFromInterpreterCatchErrors(inv)
```

PR #632, opened 2019-11-12
clock: updated code and implementation

jbenet:
This PR cleans up the clock interface:
- use UTCClock to be clear what kind of clock it is (wall clock was ok, but less clear)
- split ChainEpochClock into a separate object.


jbenet:
cc @sternhenri see new interface. 

We should not change this interface further, as (a) this system is now supposed to be done/stable, and (b) many implementors are using it. (def code review and PR fixes if any needed)

PR #637, opened 2019-11-13
VM updates

jzimmerman:
Reading/review notes:
- Switch to using placeholders for method numbers, until the interfaces have stabilized and we can allocate all canonical IDs in a single pass [commit hash 217e0eeb, 56e5560d]
- Various enhancements to runtime functionality:
  - OriginalBlockWinner [85bf288a]
  - OriginalSender [d856886e] as distinct from immediate sender (in the case of nested method invocations)
  - Separate internal vs. toplevel CallSeqNum [6b6029d7]
  - Modify Address to use sum type instead of hardcoded strings [6b6029d7]
  - Serialization/deserialization APIs for message sends, with example usage (submitPoStMessage) [a417af2b, c0c0f94a, 6b75557c]
  - Runtime ABI and gas cost accounting for pure functions (e.g. crypto operations) [1cb8bf74]
- Complete implementations of basic actors: InitActor, CronActor, AccountActor [56e5560d, 66ab7265, 035a2e8e, 9073c663, 8dc07cc3]
- Id language features to support above improvements
  - Tuple return types [c0c0f94a, 0cdec685]
  - Native enums (no longer require unwieldy struct literals) [45cb01c5, 0cdec685]


src/systems/filecoin_vm/runtime/runtime.id
```diff
@@ -17,14 +18,34 @@ type Runtime interface {
...
+    // The actor who mined the block in which the initial on-chain message appears.
+    ToplevelBlockWinner()       addr.Address
+
+    // Top-level call sequence number of the "From" actor in the initial on-chain message.
+    ToplevelSenderCallSeqNum()  actor.CallSeqNum
+
+    // Sequence number representing the total number of calls (to any actor, any method)
+    // during the current top-level message execution.
+    // Note: resets with every top-level message, and therefore not necessarily monotonic.
+    InternalCallSeqNum()        actor.CallSeqNum
```
whyrusleeping: +1

tools/codeGen/util/util.go
```diff
@@ -16,6 +16,8 @@ type Bool bool
 type Int int
 type Any = interface{}
 type String string
+type Bytes = []byte
+type Serialization Bytes
```
zixuanzh: While we are at this, can we also add
```
type Bitfield Bytes
type RLEpBitfield Bitfield
```

mishmosh:
Could you add reading/review notes in the PR description? Will help implementers grok + move quickly.

tools/codeGen/util/util.go
jzimmerman: Done

jzimmerman:
> Could you add reading/review notes in the PR description? Will help implementers grok + move quickly.

Done

PR #640, opened 2019-11-15
Adding Rationality calculation on election post with ticket…

nicola:
Ongoing branch for ubercalc updates.


src/orient/rational-calc.orient
```diff
@@ -26,22 +26,27 @@ ProofOfSpacetime:
...
-      post_cached_lookup_challenges_time = post_cached_regen_hashes * rig_storage_latency
-      // log2(nodes) > post_cached_layers
-
-      post_cached_hashes = expt(2, post_cached_layers) + 1
-      post_cached_remaining_layers = commr_tree_depth - post_cached_layers
-      post_cached = post_cached_hashes * node_size
-      post_cached_gb = post_cached / (1024 * 1024 * 1024)
-      post_cached_fraction = post_cached / sector_size
+      Regeneration:
+        post_mtree_cached_hashnodes = expt(2, post_mtree_layers_cached) + 1
```
porcuquine: I think this should `expt(2, post_mtree_layers_cached) - 1`.

Assumption: we include the root in the number of layers cached. So the root plus one layer would be two layers cached, for a total of 2^2 - 1 = 3 nodes. If I'm misreading your intention, please let me know.

Also, consider adding descriptions. I know this is extra work, but it will really help document, make review easier, and probably help catch bugs along the way.

I consider that to be complete, orient models should have descriptions.

src/orient/rational-calc.orient
```diff
@@ -26,22 +26,27 @@ ProofOfSpacetime:
...
+        post_mtree_layers_deleted = commr_tree_depth - post_mtree_layers_cached
```
porcuquine: Assuming the bottom, data row is not deleted, this looks right.

Example (to make assumptions explicit and check for off-by-one): a tree with 4 rows has depth 3. If the first two (root plus a row of 2 nodes) are cached, then one (the one just above the base layer) is deleted: 3 - 2 = 1.

However, if your assumption is that the bottom, data layer is *not* deleted, then what you have here is correct, since 3 - 2 = 1.

src/orient/rational-calc.orient
```diff
@@ -26,22 +26,27 @@ ProofOfSpacetime:
...
+
+        post_mtree_regen_nodes = expt(2, (log2(nodes) - post_mtree_layers_cached))
```
porcuquine: Isn't `log2(nodes)` just `commr_tree_depth`?

Let's use that instead. But then the whole exponent just becomes `post_mtree_layers_deleted`.

Let's use that instead. But then we can see that this is off by one. I think what we want is `post_mtree_regen_nodes = expt(2, post_mtree_layers_deleted) - 1`.

src/orient/rational-calc.orient
```diff
@@ -26,22 +26,27 @@ ProofOfSpacetime:
...
+        // TODO: check if post_merkle_region_hashes shoulld be *2
```
porcuquine: 'should'

nicola: resolved in latest

src/orient/rational-calc.orient
nicola: resolved in latest

src/orient/rational-calc.orient
nicola: resolved in latest

src/orient/rational-calc.orient
nicola: I will try to add description on the way. There has been a lot of var name changes so I decided to add them once it stabilizes

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,273 @@
...
+
+          // TODO: this is using random access, we should use seq access instead
+          post_mtree_access_challenge_time = post_mtree_regen_hashnodes * rig_storage_latency
+          post_mtree_access_challenges_time = post_mtree_access_challenge_time * post_challenges
+          // TODO: log2(nodes) > post_mtree_layers_cached
+
+    ProofGeneration:
+      describe(post_proof_gen, "Time to generate a PoSt Proof (including SNARKs, generating witnesses and reading from storage the challenged leaves)")
+
+      post_proof_gen = post_snark_time + post_witness_gen
```
porcuquine: `post_snark_time` does not seem to be defined anywhere.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,273 @@
...
+        post_mtree_cached_gb = post_mtree_cached / (1024 * 1024 * 1024)
+        post_mtree_overhead = post_mtree_cached / sector_size
+        post_mtree_cached_hashnodes = expt(2, post_mtree_layers_cached) - 1
+        post_mtree_layers_deleted = commr_tree_depth - post_mtree_layers_cached
+
+      Time:
+        // Note: this assumes reading the cached parts of the merkle tree instantaneous
+        describe(post_witness_gen, "Time to regenerate and to read the merkle tree inclusion proofs for the challenged nodes from memory")
+
+        post_witness_gen = post_mtree_regen_challenges_time + post_mtree_access_challenges_time
```
porcuquine: This accounts for the time required to regenerate a partial merkle tree, but it doesn't account for the time required to regenerate the leaf in the wrapper layer.

src/orient/rational-calc.orient
nicola: yes, I will try to have some form of "time cost per constraint"

src/orient/rational-calc.orient
nicola: It does not, this assumed, you already have the wrapper layer. I will be adding the rest soon.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+
+    describe(window_nodes, "Number of nodes in a window")
+    describe(window_size, "Size of a window (bytes)")
+    describe(window_size_mb, "Size of a window (MiB)")
+    describe(windows, "Number of windows in a sector")
+
+    describe(graph_parents, "Number of parents of a node in the PoRep window graph")
+    describe(drg_parents, "Number of DRG parents of a node in the PoRep window graph")
+    describe(expander_parents, "Number of Chung Expander parents of a node in the PoRep window graph")
+    describe(drg_layers, "Number of layers in the DRG")
```
porcuquine: This naming is a little confusing. The way I see it, there are the key layers, and the wrapper layers (just one). Each of those layers might contain a single DRG. So what you are describing is what I would call a key layer. I'm not saying it's wrong, just surfacing some terminology friction. I do think viewing a DRG as one layer is consistent with names like `drg_parents`.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+    describe(kdf_hash_size_gb, "Size of the input to the KDF hash (GiB)")
+
+    sector_size = sector_size_gb * b_in_gb
+    graph_parents = drg_parents + expander_parents
+    nodes = sector_size / node_size 
+    node_size_gb = node_size / b_in_gb 
+    kdf_hash_size_gb = graph_parents * node_size_gb
+    kdf_hash_size = kdf_hash_size_gb * b_in_gb
+
+    window_size = window_size_mb * 1024 * 1024
```
porcuquine: You might as well create a `b_in_mb` utility constant too, for consistency and to minimize the points where an auditor has to re-verify that the right factors were provided.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+    describe(commr_size, "Size of CommR (bytes)")
+    describe(commd_size, "Size of CommD (bytes)")
+    describe(commrlast_size, "Size of CommRLast (bytes)")
+
+    describe(commr_tree_depth, "Number of layers of the merkle tree with root hash CommR")
+
+    describe(seal_onchain_commitments_size, "Size of the seal commitments", bytes)
+    describe(porep_commit_time, "Time to generate CommC, CommR and CommQ", secs)
+    describe(porep_commit_time_mins, "Time to generate CommC, CommR and CommQ", mins)
+    describe(porep_commit_time_parallel, "Time to generate CommC, CommR and CommQ in parallel", secs)
```
porcuquine: Should these refer to `CommRLast` rather than `CommR`?

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+
+
+    describe(commr_time, "Time to generate CommR (seconds)")
+    describe(commd_time, "Time to generate CommD (seconds)")
+    describe(commc_time, "Time to generate CommC (seconds)")
+
+    commit_size = cid_size
+    // TODO: should be missing *2
+    window_comm_tree_time = merkle_tree_hash_time * window_nodes
```
porcuquine: I think this is what your TODO is approximately communicating, but the number of nodes in a tree will be `window_nodes * 2 - 1`.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+    window_comm_leaves_time = column_leaf_hash_time * windows
+    commr_size = commit_size
+    commd_size = commit_size
+    commrlast_size = commit_size
+
+    seal_onchain_commitments_size = commd_size + commr_size
+    // seal_commitment_time = commr_time
+
+    commd_time = merkle_tree_datahash_time * nodes
```
porcuquine: Here as well, the number of nodes to be hashed in a tree should be `nodes *  2 - 1`.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+
+
+    describe(commr_time, "Time to generate CommR (seconds)")
+    describe(commd_time, "Time to generate CommD (seconds)")
+    describe(commc_time, "Time to generate CommC (seconds)")
+
+    commit_size = cid_size
+    // TODO: should be missing *2
+    window_comm_tree_time = merkle_tree_hash_time * window_nodes
+    window_comm_leaves_time = column_leaf_hash_time * windows
```
porcuquine: Doesn't this also depend on the number of layers in a window? I think column size is `layers * windows`.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+      commr_time = merkle_tree_hash_time * nodes
+      commr_tree_depth = log2(nodes) 
+
+    CommR (ColumnR):
+      commr_time = window_comm_tree_time + window_comm_leaves_time 
+      commr_tree_depth = log2(window_nodes)
+
+    CommC:
+      describe(commc_tree_time, "Time to generate the merkle tree of CommC (after the leaves have been computed)", secs)
+      commc_tree_time = (window_comm_tree_time + window_comm_leaves_time) * drg_layers
```
porcuquine: Okay, I see that you're factoring in `drg_layers` here. It makes more sense to me to put this factor in the definition of `window_comm_leaves_time`, but I think this is equivalent.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+    describe(window_inclusions, "Constraints to check all the required inclusion proofs in a window")
+    describe(column_leaf, "Constraints to check a single leaf generation from a column in a window")
+    describe(column_leaves, "Constraints to check all the leaves generation required in a window")
+    describe(window_column_leaves, "Constraints to check all the leaves generation required for all windows")
+
+    // D
+    // Data layer committed with merkle tree
+    data_inclusion = merkle_tree_datahash_constraints * log2(nodes)
+    data_inclusions = data_inclusion * porep_challenges
+    window_data_inclusions = data_inclusion * porep_challenges * windows
```
porcuquine: Should there be a factor of layers here? (Maybe this will be the same design question as before — just flagging as I read.)

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+
+    porep_snark_partitions = porep_snark_constraints / porep_snark_partition_constraints
+    porep_snark_size = porep_snark_partitions * snark_size
+    porep_snark_size_kb = porep_snark_size / 1024
+
+    Constraints (WindowEncoding):
+      porep_snark_constraints = window_inclusions + window_column_leaves + window_labeling_proof + window_data_inclusions + window_replica_inclusions*2
+
+    Constraints (SectorEncoding):
+      porep_snark_constraints = window_inclusions + window_column_leaves + window_labeling_proof + data_inclusions + replica_inclusions*2
```
porcuquine: Should there be a factor of `windows` in here somewhere?

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+
+    Constraints (Wrapping):
+      porep_snark_constraints = window_inclusions + window_column_leaves + window_labeling_proof + data_inclusions + replica_inclusions*2
```
porcuquine: Same here, should there be a factor of `windows`?

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+  breakeven_nodes = ((cost_gb_per_month *  sector_size_gb) / (cost_node_encoding * challenge_periods_per_month))
+  challenge_periods_per_month = secs_in_month / polling_time
+  regeneration_fraction = breakeven_nodes / nodes
+
+  Costs:
+    Compute:
+      describe(cost_node_encoding, "Cost of encoding a single node")
+      describe(nodes_per_second, "Nodes encoded per second using the fastest hardware")
+      describe(nodes_per_year, "Nodes encoded per year using the fastest hardware")
+      describe(hash_gb_per_second, "GiB that can be hashed per second")
```
porcuquine: Check to make sure the source data you are using is actually in GiB rather than GB — or else provide a conversion. Or maybe you mean GB, as in the variable name. Either way, make the description and name consistent.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+
+      nodes_per_second = hash_gb_per_second / kdf_hash_size_gb
+      nodes_per_year = nodes_per_second * (60 * 60 * 24 * 365)
```
porcuquine: Add a utility constant, please.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+      cost_node_encoding = rig_malicious_cost_per_year / nodes_per_year 
+
+    Storage:
+      describe(cost_gb_per_month, "Cost of storing 1GiB per month")
+      describe(cost_storage_gbs, "Cost of storing 1GiB per second")
```
porcuquine: As above, let's be very precise about the GiB/GB distinction and make sure names and descriptions match. Some things really are denominated in GB (in our source data if nowhere else), and getting this wrong could lead to rude surprises.

src/orient/rational-calc.orient
nicola: added in latest

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+    describe(ticket_constraints, "Constraints of a partial ticket per challenged node (in PoSt)")
+    describe(post_snark_constraints, "Constraints for a PoSt")
+    describe(post_snark_partitions, "Number of partitions of a PoSt proof")
+    describe(post_snark_size, "Size of a PoSt proof (bytes)")
+    describe(post_snark_size_kb, "Size of a Post proof (KiB)")
+    describe(post_snark_partition_constraints, "Number of constraints for a PoSt partition")
+
+    inclusion_proof = merkle_tree_hash_constraints * log2(nodes)
+    post_inclusion_proofs = inclusion_proof * post_challenges
+    ticket_proofs = ticket_constraints * post_challenges
```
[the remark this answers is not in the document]
nicola: which ones are the hashes?

src/orient/rational-calc.orient
nicola: it's a bit unfortunate the way I have defined it, this is as if you were stacking all windows of one layer,

```
commc_tree_time = (window_comm_tree_time + window_comm_leaves_time) * drg_layers 
```

src/orient/rational-calc.orient
nicola: each of those should account for it.

src/orient/rational-calc.orient
nicola: each of those values account for whether or not it's required to do an operation per window or not.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
+// Note: all terms referring to size are in bytes, unless suffixed with _mb, _tb, etc.
+// Note: all terms referring to time are in seconds, unless suffixed with _mins, _days, etc.
+
+Rig:
+  describe(rig_sectors_min, "Minimum sectors require to mine")
+
+  describe(rig_storage_latency, "Random Access latency on the storage medium (honest)")
```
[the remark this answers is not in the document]
nicola: fixed in latest

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+    post_snark_constraints = ticket_proofs + post_inclusion_proofs
+    post_snark_partitions = post_snark_constraints / post_snark_partition_constraints
+    post_snark_size = post_snark_partitions * snark_size
+    post_snark_size_kb = post_snark_size / 1024
+
+ProofOfReplication:
+  Security:
+    describe(porep_challenges, "Number of challenges in a PoRep proof")
+    describe(porep_lambda, "Bits of security of a PoRep proof")
+    describe(spacegap, "Maximum difference in storage between an honest prover and a maliciois prover")
```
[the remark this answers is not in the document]
nicola: fixed in latest

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,354 @@
...
+    porep_challenges = -porep_lambda / (log2(1-spacegap/4))
+
+  Graph:
+    describe(nodes, "Number of nodes in a sector")
+    describe(sector_size, "Size of a sector (bytes)")
+    describe(sector_size_gb, "Size of a sector (GiB)")
+
+    describe(window_nodes, "Number of nodes in a window")
+    describe(window_size, "Size of a window (bytes)")
+    describe(window_size_mb, "Size of a window (MiB)")
```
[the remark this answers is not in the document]
nicola: fixed in latest

src/orient/rational-calc.orient
nicola: fixed in latest

src/orient/rational-calc.orient
nicola: fixed in latest

src/orient/rational-calc.orient
nicola: fixed in latest

src/orient/rational-calc.orient
nicola: fixed in latest

src/orient/rational-calc.orient
nicola: fixed in latest

src/orient/rational-calc.orient
nicola: fixed in latest

src/orient/rational-calc.orient
nicola: same as other comment

src/orient/rational-calc.orient
nicola: same as other comment

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,550 @@
...
+
+ProofOfReplication (StackedChungParameters):
+  describe(chung_a, "Chung A")
+  describe(chung_c, "Chung C")
+  describe(chung_delta, "Chung delta")
+  describe(chung_sigma, "Chung sigma")
+
+  chung_a = spacegap - (chung_c / (chung_c -1)) * chung_delta + ((chung_c-1)/ chung_c) * (chung_sigma - 1)
+
+  stacked_layers = (log2(regeneration_fraction / chung_a) / log2(chung_c)) + 1
```
nikkolasg: This is not conservative at all : we want to set 2 or 3 times this number to be on the safe side, such that even if the cost of encoding goes a bit, it's still safe. We might want to have an extra factor like `marginFactor` that we multiply with regeneration_fraction here.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,550 @@
...
+      describe(cost_node_encoding, "Cost of encoding a single node")
+      describe(nodes_per_second, "Nodes encoded per second using the fastest hardware")
+      describe(nodes_per_year, "Nodes encoded per year using the fastest hardware")
+      describe(hash_gb_per_second, "GB that can be hashed per second")
+      describe(hash_gib_per_second, "GiB that can be hashed per second")
+
+      hash_gib_per_second = hash_gb_per_second * to_i
+      nodes_per_second = hash_gib_per_second / kdf_hash_size_gib
+      nodes_per_year = nodes_per_second * secs_in_year 
+      cost_node_encoding = rig_malicious_cost_per_year / nodes_per_year
```
nikkolasg: This does not include the cost of transferring memory and cost of hashing or does it ? Asking because the latest supranational spreadsheet doesn't use this model of computing the price. Maybe it's hidden in this model but it doesn't appear like it.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,614 @@
...
+  b_in_pib = (1024 * 1024 * 1024 * 1024 * 1024)
+  b_in_tib = (1024 * 1024 * 1024 * 1024)
+  b_in_gib = (1024 * 1024 * 1024)
+  b_in_mib = (1024 * 1024)
+  b_in_kib = 1024
+  secs_in_month = (24 * 60 * 60 * 30)
+  secs_in_year = (24  * 60 * 60 * 365)
+  secs_in_day = (24 * 60 * 60)
+  cid_size = 32 + 1 + 1 + 1
+  to_i = 1024/1000
```
porcuquine: I think this should actually be `(1024 * 1024 * 1024) / (1000 * 1000 * 1000)`.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,614 @@
...
+    describe(post_challenge_read, "Time to read from storage medium data required for a single challenge in PoSt", secs)
+    describe(post_data_access, "Time to retrieve from the storage unit the data for PoSt", secs)
+    describe(post_data_access_parallel, "Time to retrieve from the storage unit the data in parallel for PoSt", secs)
+
+    post_data_access = post_challenge_read * post_challenges
+    post_data_access_parallel = post_data_access / rig_storage_parallelization
+
+    Access (Wrapping):
+      // TODO(enhancement): this is using random access, we should use seq access instead
+      post_challenge_read = (1 + post_mtree_regen_nodes) * rig_storage_latency
```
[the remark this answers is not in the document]
nicola: 1) the tree is cached in SSD, so it should not cost.
2) ok on sequential access

src/orient/rational-calc.orient
nicola: I see. This depends on whether we multiply gbs mbs and kbs

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,704 @@
...
+  describe(b_in_pib, "Number of bytes in a PiB")
+  describe(b_in_tib, "Number of bytes in a TiB")
+  describe(b_in_gib, "Number of bytes in a GiB")
+  describe(b_in_mib, "Number of bytes in a MiB")
+  describe(b_in_kib, "Number of bytes in a KiB")
+  describe(gib_in_tib, "Number of GiB in a TiB")
+  describe(gb_to_gib, "Ratio between GB and GiB")
+  describe(mb_to_mib, "Ratio between MB and MiB")
+
+  b_in_eib = (1024 * 1024 * 1024 * 1024 * 1024 * 1024)
```
schomatis: Can we define each in terms of the previous one?

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,704 @@
...
+
+    Window:
+      describe(window_nodes, "Number of nodes in a window")
+      describe(window_size, "Size of a window", bytes)
+      describe(window_size_mib, "Size of a window", mib)
+      describe(window_size_gib, "Size of a window", mib)
+      describe(windows, "Number of windows in a sector")
+
+      window_size = window_size_mib * b_in_mib 
+      window_size = window_size_gib * b_in_gib
```
schomatis: Why is this defined twice?

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,704 @@
...
+    Window:
+      describe(window_nodes, "Number of nodes in a window")
+      describe(window_size, "Size of a window", bytes)
+      describe(window_size_mib, "Size of a window", mib)
+      describe(window_size_gib, "Size of a window", mib)
+      describe(windows, "Number of windows in a sector")
+
+      window_size = window_size_mib * b_in_mib 
+      window_size = window_size_gib * b_in_gib 
+      windows = sector_size / window_size
```
schomatis: Is this guaranteed to be an integer? Is there a difference between a float and an integer division?

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,704 @@
...
+      describe(node_size_mib, "Size of a node", mib)
+      describe(node_size_gib, "Size of a node", gib)
+      describe(kdf_hash_size, "Size of the input to the KDF hash", bytes)
+      describe(kdf_hash_size_gib, "Size of the input to the KDF hash", gi)
+
+      nodes = sector_size / node_size
+      node_size_mib = node_size / b_in_mib 
+      node_size_gib = node_size / b_in_gib 
+      kdf_hash_size_gib = graph_parents * node_size_gib
+      kdf_hash_size = kdf_hash_size_gib * b_in_gib
```
schomatis: Can this be more direct instead of going through the GiB definition?

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,704 @@
...
+      epost_mtree_read = post_mtree_read
+      epost_data_access = epost_leaves_read + epost_mtree_read
+      epost_data_access_parallel = epost_data_access / rig_storage_parallelization  
+      epost_mtree_read_parallel = epost_mtree_read / rig_storage_parallelization  
+      epost_leaves_read_parallel = epost_leaves_read / rig_storage_parallelization  
+      epost_inclusions_time = post_inclusions_time * epost_challenged_sectors
+      epost_inclusions_time_parallel = epost_inclusions_time / rig_cores
+
+      epost_time = epost_data_access + post_ticket_gen + epost_inclusions_time + post_snark_time
+      epost_time_parallel = epost_data_access_parallel + post_ticket_gen + epost_inclusions_time_parallel + post_snark_time_parallel
```
schomatis: Can `post_ticket_gen` be parallelized?

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,704 @@
...
+
+Rationality:
+  describe(breakeven_nodes, "number of nodes / sector / challenge period after which it's cheaper to store the entire sector, than re-encode those")
+  describe(polling_time, "Time between two PoSt proofs", secs)
+  describe(challenge_periods_per_month, "Number of challenge periods in a month")
+  describe(regeneration_fraction, "Fraction of nodes of a sector for which re-encoding them every `polling_time` costs as much as storing them")
+
+  breakeven_nodes = ((cost_gib_per_month *  sector_size_gib) / (cost_node_encoding * challenge_periods_per_month))
```
schomatis: The multiplication `cost_node_encoding * challenge_periods_per_month` doesn't seem to have matching units.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,704 @@
...
+    describe(post_data_access, "Time to retrieve from the storage unit the data for PoSt", secs)
+    describe(post_data_access_parallel, "Time to retrieve from the storage unit the data in parallel for PoSt", secs)
+    describe(post_leaf_read, "Time to retrieve the challenged leaf in PoSt", secs)
+    describe(post_leaves_read, "Time to retrieve the all the challenged leaves in PoSt", secs)
+    describe(post_leaves_read_parallel, "Time to retrieve the all the challenged leaves in PoSt in parallel", secs)
+
+    post_data_access = post_challenge_read * post_challenges
+    post_leaves_read = post_leaf_read * post_challenges
+    // TODO(enhancement): calculate first mtree and then challenge_read
+    post_mtree_read = (post_challenge_read - post_leaf_read) * post_challenges
```
schomatis: The magnitudes being subtracted were already multiplied by `post_challenges`, I'm not finding the definition of this variable but this seems suspicious.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,704 @@
...
+        post_mtree_overhead = post_mtree_cached / sector_size
+        post_mtree_cached_hashnodes = expt(2, post_mtree_layers_cached) - 1
+        post_mtree_layers_deleted = commr_tree_depth - post_mtree_layers_cached
+
+      Generation:
+        describe(post_mtree_regen_hashnodes, "Number of intermediary nodes to be regenerated for a single replica merkle tree inclusion proof in PoSt")
+        describe(post_mtree_regen_nodes, "Number of leaf nodes to be access for a single replica merkle tree inclusion proof in PoSt")
+        describe(post_inclusion_time, "Time to generate a Merkle tree inclusion proof for PoSt (assuming all data in memory)", secs)
+
+        post_mtree_regen_hashnodes = expt(2, post_mtree_layers_deleted + 1) - 1 // TODO check if -1 should be removed
```
schomatis: Why `+1` in the exponentiation?

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,704 @@
...
+
+        post_mtree_regen_nodes = expt(2, post_mtree_layers_deleted) - 1 // TODO check if -1 should be removed
+        post_inclusions_time = post_inclusion_time * post_challenges
+        post_inclusions_time_parallel = post_inclusions_time / rig_cores
+
+      Generation (Wrapping):
+        post_inclusion_time = (post_mtree_regen_hashnodes * 2) * merkle_tree_hash_time
```
schomatis: Probably related to the previous question, why another `*2`? (compounded to the previous exponentiation increase of `post_mtree_regen_hashnodes`)

src/orient/rational-calc.orient
nicola: this is not a "definition", but is a constraints, and it makes sure that mib and gib are correctly constrained.

src/orient/rational-calc.orient
nicola: no, it can't

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,718 @@
...
+    describe(encoding_time_parallel_asic, "Time to encode a sector in parallel with ASICs", secs)
+    describe(encoding_window_time, "Time to encode a window", secs)
+    describe(encoding_window_time_parallel, "Time to encode a window in parallel", secs)
+    describe(encoding_window_time_asic, "Time to encode a window with an ASIC", secs)
+    describe(encoding_window_time_parallel_asic, "Time to encode a window in parallel with ASICs", secs)
+    describe(kdf_time, "Time to compute a KDF", secs)
+
+    encoding_time = encoding_window_time * windows
+
+    encoding_window_time = window_nodes * (graph_parents * kdf_time) * stacked_layers
```
schomatis: Missing replica id in key generation time (`(graph_parents  + 1) * kdf_time`).

src/orient/rational-calc.orient
schomatis: As per @porcuquine explanation, the +1 in the exponent includes the layer on top of the deleted ones (root node) which will also have to be regenerated.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,718 @@
...
+        post_mtree_layers_deleted = commr_tree_depth - post_mtree_layers_cached
+
+      Generation:
+        describe(post_mtree_regen_hashnodes, "Number of intermediary nodes to be regenerated for a single replica merkle tree inclusion proof in PoSt")
+        describe(post_mtree_regen_nodes, "Number of leaf nodes to be access for a single replica merkle tree inclusion proof in PoSt")
+        describe(post_inclusion_time, "Time to generate a Merkle tree inclusion proof for PoSt (assuming all data in memory)", secs)
+
+        post_mtree_regen_hashnodes = expt(2, post_mtree_layers_deleted + 1) - 1 // TODO check if -1 should be removed
+
+        post_mtree_regen_nodes = expt(2, post_mtree_layers_deleted) - 1 // TODO check if -1 should be removed
```
schomatis: It looks like we could drop the -1 because we don't want the entire tree size, just the last layer. To get to this last leaf nodes layer the exponent (I think) should be increased `+2`: +1 to include the root node of the regenerated sub-tree (as in `post_mtree_regen_hashnodes`) and another +1 to reach the leaf nodes layer (which is not included in `post_mtree_regen_hashnodes` because that layer is not being regenerated).

schomatis: /cc @porcuquine to check construction validity behind this reasoning

src/orient/rational-calc.orient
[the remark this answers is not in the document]
schomatis: Unrelated, why the +1? Is it because of the -1 in the computation of `post_mtree_regen_nodes`? (which might need to be removed)

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,718 @@
...
+    inclusion_proof = merkle_tree_hash_constraints * commr_tree_depth 
+    post_inclusion_proofs = inclusion_proof * post_challenges
+    ticket_proofs = ticket_constraints * commr_column_size * post_challenges
+    // TODO: handle different constructions
+    post_snark_constraints = ticket_proofs + post_inclusion_proofs
+    post_snark_partitions = post_snark_constraints / post_snark_partition_constraints
+    post_snark_size = post_snark_partitions * snark_size
+    post_snark_size_kib = post_snark_size / b_in_kib
+    post_snark_time = snark_constraint_time * post_snark_constraints
+    post_snark_partition_time = snark_constraint_time * post_snark_constraints
```
schomatis: So `post_snark_time` is the same as `post_snark_partition_time`?

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,718 @@
...
+    // Parallelization calculation: it must take at least the time for a single snark
+    // the remaining can be done in parallel
+    // Similar in spirit to:
+    // min(porep_snark_partition_time, porep_snark_time / snarrk_parallelization)
+    post_snark_time_parallel = post_snark_partition_time + (post_snark_time - post_snark_partition_time) / (rig_snark_parallelization)
```
schomatis: The second term in the addition will be zero then.

src/orient/rational-calc.orient
```diff
@@ -0,0 +1,718 @@
...
+Crypto:
+  describe(snark_size, "Size of a SNARK proof", bytes)
+  describe(snark_constraint_time, "Time of proving a SNARK constraint", secs)
+  describe(hashing_amax, "Factor of difference between stock CPU hashing and the fastest hashing")
+  describe(merkle_tree_datahash_constraints, "Constraints for hashing two nodes in CommD")
+  describe(merkle_tree_datahash_time, "Time to hash two nodes in CommD", secs)
+  describe(merkle_tree_hash_constraints, "Constraints for hashing two nodes in CommR")
+  describe(merkle_tree_hash_time, "Time to hash two nodes in CommR", secs)
+  describe(column_leaf_hash_constraints, "Constraints for hashing a single row in a column", secs)
+  describe(column_leaf_hash_time, "Time to compute the hash of a single row in a column", secs)
```
schomatis: Can we clarify this "row" terminology? (which seems to be used only here)

src/orient/viz/rational-calc.notebook.js
```diff
@@ -121,12 +120,12 @@ md`#### Benchmarks`
...
 bench = ({
   "column_leaf_hash_time": 1.7028e-5/2,
-  "kdf_time": 5.4e-7/2,
+  "kdf_time": 5.4e-7,
   "merkle_tree_datahash_time": 1.28e-8/2,
   "merkle_tree_hash_time": 1.7028e-5/2,
   "snark_constraint_time": 3.012e-5/2,
   "ticket_hash": 1.7028e-5/2,
-  "kdf_time_fastcpu": 0.0000000128/2
+  "kdf_time_fastcpu": 0.0000000128 // TODO check if this should be divided by 2
```
porcuquine: I think if this number is for 64 bytes then it (and `kdf_time`) should be divided by 2.

src/orient/rational-calc.orient
```diff
@@ -715,14 +735,23 @@ Filecoin:
...
       epost_cost = epost_inclusions_cost + epost_snark_cost
 
+Timing:
+  describe(malicious_regen_time, "Time for a malicious adversary to regenerate a sector", secs)
+  describe(time_amax, "Factor of how much slower than the blocktime the malicious encoding should take")
+
+Timing (TimingAssumption):
+  RegenAttack (StackedSDRParameters):
+  malicious_regen_time = (encoding_time_asic * drg_d) / (stacked_layers)
+  malicious_regen_time = block_time * time_amax
```
porcuquine: Did you intend for these two lines to be indented so they would require the `StackedSDRParameters` flag?

mishmosh:
@nicola we are trying to land or close old PRs. is this still active?

nicola:
hey @mishmosh, for ubercalc we have been using this branch for all the updates - I am happy to merge this if you prefer not to have this type of long standing PRs

src/orient/rational-calc.orient
nicola: yes, fixed

src/orient/viz/rational-calc.notebook.js
nicola: fixed

mishmosh:
@nicola No worries to continue working as you were. I did add a description to this PR so people know why it's staying open so long.

nicola:
@porcuquine this is ready to merge for now.

PR #641, opened 2019-11-17
minor-fix

steven004:
For Rational PoSt part. 
This part might be replaced by SurprisePoSt and ElectionPost soon, in this case, just close it.


sternhenri:
Thanks @steven004, closing accordingly!

PR #648, opened 2019-11-22
Retrieval Market Spec

hannahhoward:
# Goals

- Describe the retrieval markets in detail to provide sufficient information for implementers to proceed
- Offer an easy path to building the retrieval market module spec compliance while providing an evolutionary path to arbitrary selector retrieval
- Make sure retrieval protocol works in trustless system.
- Stay as close to Lotus implementation as possible while supporting IPLD packed in CAR files

# Implementation

- Flesh out all types
- Define an API for client & provider that works for both an initial spec and an evolved spec without breaking compatibility
- Define specific libp2p protocols used for retrieval
- Flesh out discovery mechanism
- Describe deal flows for initial (V0) and evolved (V1) markets with text and diagrams
- Describe bootstrapping trust mechanism
- Add intermediate voucher & pause/unpause mechanism to data transfer module
- Identify dependencies for Provider & Client 

# For Discussion

1. The biggest challenge here is supporting non-UnixFS dags. We can no longer count on fixed block sizes. This makes paying for chunks ahead of time a bit problematic. My approach is to flip the order -- rather than the client paying a small bit ahead and trusting the provider to send a small bit of data, the provider sends a small bit of data and requests payment before sending more. The provider, if it has information about how much data remains (such as towards the end), can choose to demand payment up front. Data is always sent in a verifiable order (a depth first MerkleDAG traversal).

2. Using the Data Transfer Module and Graphsync would unlock a lot of abilities -- we could use arbitrary selector, and Graphsync already has built in verification mechanisms which would remove the need for us to do verification in Retrieval. But, it's a big change from existing code AND it requires enhancements to both the data transfer module & Graphsync. The idea is to evolve to this kind of a protocol as we approach Mainnet. This is an off chain protocol so it could also update AFTER mainnet. (also, the diagram for V1 may need a little work)


mishmosh:
@ingar can you also read + add any comments or clarification requests from an implementer's perspective?

src/systems/filecoin_markets/retrieval_market/retrieval_provider.md
```diff
@@ -2,4 +2,15 @@
 title: "Retrieval Provider (Miner)"
 ---
 
+### Provider Dependencies
+
+The Retrieval Provider Depends On The Following Dependencies
+**Host**: A libp2p host (set setup the libp2p protocols)
+**Node**: A node implementation to query the chain for pieces and to setup and manage payment channels
+**StorageMiner**: For unsealing sectors
```
zixuanzh: This is fine for now but I think we will define what a StorageMinerNode is (which contains StorageMiningSubsystem that will include the functionality to unseal sectors. Similarly, a Node will become a Filecoin Node.

src/systems/filecoin_markets/retrieval_market/_index.md
```diff
@@ -1,23 +1,75 @@
...
 # Components
 
-Version 0 of the `retrieval market` protocol is what we (tentatively) will launch the filecoin network with. It is version zero because it will only be good enough to fit the bill as a way to pay another node for a file.
+The `retrieval market` refers to the process of negotiating deals for a provider to serve stored data to a client.
 
 The main components are as follows:
 
-- A payment channel actor (See [payment channels](payment-channels.md) for details)
-- 'retrieval-v0' `libp2p` services
+- A payment channel actor (See {{<sref payment_channel_actor "Payment Channel Actor">}} for details)
```
sternhenri: great use of srefs <raised hands>

src/systems/filecoin_markets/retrieval_market/retrieval_flow_v0.mmd
```diff
@@ -0,0 +1,67 @@
...
+  deactivate BSP
+  end
+  loop Retreive In Pieces
+    Provider ->> Client : Request Payment
+    activate Client
+    Client ->> Node : Create Voucher On Payment Channel
+    activate Node
+    Node -->> Client : Message For Voucher Gets On Chain
+    deactivate Node
+    Client -->> Provider : Send Payment
```
sternhenri: wouldn't the client send payment on chain in this case, and the provider redeem the voucher on payment channel (ie get paid) from the chain accordingly?

src/systems/filecoin_markets/retrieval_market/_index.md
```diff
@@ -1,23 +1,75 @@
...
+
+{{< diagram src="retrieval_flow_v0.mmd.svg" title="Retrieval Flow - V0" >}}
+
+The baseline version of proposing and accepting a deal will work as follows:
+
+- The client finds a provider of a given piece with FindProviders.
+- The client queries a provider to see if it meets its retrieval criteria (via Query Protocol)
+- The client creates a payment channel as neccesary and a lane, ensures there are free funds in the channel
+- The client sends a RetrievalDealProposal to the retrieval miner. (via RetrievalProtocol)
+- The providers validates the proposal and rejects it if it is invalid: invalid price per byte, invalid payment channel are reasons for rejection
```
sternhenri: what is an invalid payment channel? missing payment channel?

src/systems/filecoin_markets/retrieval_market/_index.md
```diff
@@ -1,23 +1,75 @@
...
+- If the request is valid, the provider responds to with an accept message
+- The provider may also send a small number of blocks over the protocol before requiring payment
+- The client consumes blocks over the retrieval protocol and manually verifies them
+- When the requires payment to proceed, it sends a request for payment and does not send any more blocks
```
sternhenri: should the rate of payment requests be automated as part of the retrieval deal to prevent a de facto DoS from a hungry provider?

src/systems/filecoin_markets/retrieval_market/_index.md
```diff
@@ -1,23 +1,75 @@
...
+- The client responds with a payment (or cancels the request)
```
sternhenri: Conversely, anything to be done about the client DoSing the provider by opening the channel, having them unseal and then canceling the query? Maybe we should explicitly add a cancellation fee for both

src/systems/filecoin_markets/retrieval_market/_index.md
sternhenri: Likewise, should there be a timeout after which if either party has not acted (paid, sent bytes), the deal is terminated and they are penalized?

src/systems/filecoin_markets/retrieval_market/_index.md
```diff
@@ -1,23 +1,75 @@
...
+
+# Deal Flow (V0)
+
+{{< diagram src="retrieval_flow_v0.mmd.svg" title="Retrieval Flow - V0" >}}
+
+The baseline version of proposing and accepting a deal will work as follows:
+
+- The client finds a provider of a given piece with FindProviders.
+- The client queries a provider to see if it meets its retrieval criteria (via Query Protocol)
+- The client creates a payment channel as neccesary and a lane, ensures there are free funds in the channel
```
sternhenri: what is a lane?

src/systems/filecoin_markets/retrieval_market/_index.md
```diff
@@ -1,23 +1,75 @@
...
+- The client sends a RetrievalDealProposal to the retrieval miner. (via RetrievalProtocol)
+- The providers validates the proposal and rejects it if it is invalid: invalid price per byte, invalid payment channel are reasons for rejection
+- If the request is valid, the provider responds to with an accept message
+- The provider may also send a small number of blocks over the protocol before requiring payment
```
sternhenri: we may want to enforce stricter behavior here. Either it MUST or we drop this line. Why the optionality?

src/systems/filecoin_markets/retrieval_market/_index.md
```diff
@@ -1,23 +1,75 @@
...
+
+The evolved protocol for proposing and accepting a deal will work as follows:
+
+- The client finds a provider of a given piece with FindProviders.
+- The client queries a provider to see if it meets its retrieval criteria (via Query Protocol)
+- The client creates a payment channel as neccesary and a lane, ensures there are free funds in the channel
+- The client schedules a Data Transfer Pull Request passing the RetrievalDealProposal as a voucher.
+- The providers validates the proposal and rejects it if it is invalid: invalid price per byte, invalid payment channel are reasons for rejection
+- If the proposal is valid, the provider responds with an accept message and begins monitoring the data transfer process
+- The provider may begin sending a small amount of bytes before first payment
```
sternhenri: same comment as above. Must or kill

src/systems/filecoin_markets/retrieval_market/retrieval_flow_v1.mmd
```diff
@@ -0,0 +1,70 @@
...
+    Provider ->> DTP : Pause
+    Provider ->> DTP : Send Request Payment Intermediate Voucher
+    DTP ->> DTC : Send Intermediate Voucher
+    DTC ->> Client : Validate Request Payment Voucher
+    activate Client
+    Client ->> Node : Create Voucher On Payment Channel
+    activate Node
+    Node -->> Client : Message For Voucher Gets On Chain
+    deactivate Node
+    Client ->> DTC : Send Payment Intermediate Voucher
```
sternhenri: Again shouldn't this be sent to the chain and accordinglt withdrawn from the client balance?

src/systems/filecoin_markets/retrieval_market/_index.md
```diff
@@ -1,23 +1,75 @@
...
+- When the provider requires payment, it pauses the data transfer and sends a request for payment as an intermediate voucher
+- The client receives the request for payment responds with a payment, also sent as an intermediate voucher (or cancels the request)
+- The provider unpauses the request and data resumes sending and the process continues till the end of the query
+
+### Bootstrapping Trust
```
sternhenri: So this is where I think we can both simplify and add some stuff. Namely:
- I am not sure that enabling a mid-deal change in modalities (specifically price change) is a good idea. This may lead to clients getting held up with half their data downloaded and having to choose between an expensive tear down or racket. Likewise, I wonder if the increment size should not be fixed in terms of the deal (ie front load most of this in the negotiation and then the actual process is just formulaic).
- I would remove choice between pre and post-payment. Let's pick one. Probably favor the client, i.e. I would opt for postpayment given in the first few transfers the data will be unusable as is whereas the payment would not be (for prepayment)
- Finally, I think we are missing some added trust criteria:
    - collateral for either party: the client should potentially lock up negotiated funds on chain and have the vouchers withdraw from there; the provider should lock up a potential cancellation fee in case something goes wrong (all this entails they negotiate ahead of opening a payment channel)
    - timeouts: both for sending data and sending payments. One should not be able to DoS retrieval by simply no longer participating. Both parties should agree to a timeout after which the deal is cancelled and one pays the other a cancellation fee
    - increment size: finally it may be worth having them agree out-of-band on payment frequency and size of data increments (this is my first point again)

src/systems/filecoin_markets/retrieval_market/retrieval_peer_resolver.md
```diff
@@ -0,0 +1,9 @@
+---
+title: "Retrieval Peer Resolver"
+---
+
+The `peer resolver` is a content routing interface to discover retrieval miners that have a given Piece.
```
sternhenri: Is this straightforward to build from this description alone?

src/systems/filecoin_markets/retrieval_market/retrieval_flow_v0.mmd
hannahhoward: sorry bad wording -- this happens through the payment channel -- should read something like
`Client ->> Provider : Send Voucher For Provider to Redeem`

src/systems/filecoin_markets/retrieval_market/_index.md
hannahhoward: yea missing is probably better :) I was sorta spitballing here -- lacking an implementation of this yet.

src/systems/filecoin_markets/retrieval_market/_index.md
hannahhoward: create question -- it's in the Lotus codebase -- perhaps @magik6k has insight here :)

src/systems/filecoin_markets/retrieval_market/_index.md
hannahhoward: I agree on just fixing the requirement on post-payment is best. There is an additional benefit to this: because data is sent in IPLD blocks, the size of which are not known ahead (and which can't be split in at least existing protocols), in a prepayment situation you can't make an agreement for exactly n-bytes.

The one possible downside, and I wonder whether you feel this merits a special case, is the end of the retrieval. Here all the motivation (I think) is for the client, once they've received all the data, to short change the miner. I think this "switch" was sort of what I was trying to get at with all the options.

I'm fine with out of band agreements about payment frequency. The only challenge again is the inexact nature of sending IPLD blocks... you're not just sending a byte stream, so it's tricky to divide things up into exact amounts.

src/systems/filecoin_markets/retrieval_market/retrieval_peer_resolver.md
hannahhoward: it's a bit of an abstract interface -- it's not exactly mean to describe an implementation, because the primary functionality needs to be supplied by the chain node implementer -- there needs to be an API of the sort "FindDealsOnChainByPieceID".

Also this is a fairly undeveloped piece and I just didn't want to flesh it out any more till I had more knowledge of what it will look like.

src/systems/filecoin_markets/retrieval_market/retrieval_flow_v1.mmd
hannahhoward: Yes. Sorry, this is mean to mean "reference (i.e. CID) to payment voucher that provider will redeem"

hannahhoward: needs wording clarification

hannahhoward: and probably a step in the diagram to interact with the chain

src/systems/filecoin_markets/retrieval_market/retrieval_flow_v0.mmd
```diff
@@ -0,0 +1,67 @@
...
+    activate Node
+    Node -->> Client : Providers With PieceID
+    deactivate Node
+  end
+
+  Client ->> Provider : Send Query for Piece
+  activate Provider
+  Provider -->> Client : Send information about Piece (price, size, etc)
+  deactivate Provider
+  Client ->> Node : Create/Activate Payment Channel
```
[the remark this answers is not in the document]
hannahhoward: oh good point :)

src/systems/filecoin_markets/retrieval_market/retrieval_flow_v0.mmd
sternhenri: awesome. That makes sense to me

src/systems/filecoin_markets/retrieval_market/_index.md
sternhenri: Sounds good to me

src/systems/filecoin_markets/retrieval_market/_index.md
sternhenri: yes!

src/systems/filecoin_markets/retrieval_market/_index.md
sternhenri: Good point on the IPLD blocks. Maybe then the frequency is overkill until V1? Or it can be a check for approximate frequency..?

FOr the downside, alas I don't think we can ever really mitigate it: any "fair exchange" solution like this (tit for tat) ends up having someone going on a limb. It's just as well it be the client. This can be derisked by making increments small enough and will have to be taken care of at another layer of the chain (ie through reputation system).

PR #654, opened 2019-11-26
Feat/epost verif

sternhenri:
In this PR, we stub out the full ePoSt verification flow, namely the flow from PoSt reception (either in message for surprise or block header for ePoSt) to proofs library, including:
- [x] rederiving randomness
- [x] (ePoSt) drawing partialTickets again to verify winning tickets posted in header
- [x] (ePoSt) using partialTickets to verify PoStProof
- [x] (surprise) ...

We also renamed some of the proofs related artifacts, namely to `PartialTickets` and `PoStCandidates`.

Note that two choices are worth looking at here:
- Verifying Election PoSt is split into three methods to give the verifier flexibility in processing the block header
  - `VerifyElectionPoSt`
  - `SubmitVerifiedElectionPoSt`
  - `IsValidElection` (forthcoming in future PR)
- proofsLibrary is called directly by the verifier (in case their node does not have a sector subsystem)

Finally, we also fixed up some edge cases around ePost and tracking state:
- namely changing how we update `LastChallengeEndEpoch` to distinguish between successful events (PoSt submissions) and unsuccessful ones (proving periods ending) and eliminate ability for miners to submit surprise after challenge ends (but before cron catches/resets it) or to submit election during their second or third consecutive challenge.

Remains:
- Updates to what is passed in for verifying surprise (pending proofs optimizations)
- Election side verif
- randomness work

Qs for @porcuquine:
- [x] Does the translation of `OnChainPoStVerifyInfo` to `PoStVerifyInfo` get done elsewhere or should we do it in the `StorageMinerActor` ahead of calling the proofs lib?


src/algorithms/post/election_post.md
```diff
@@ -40,32 +40,42 @@ The epoch time is divided into three non-overlapping portions: (1) checking sect
...
     `C_i = HashToBlockNumber(post_randomness || S_j || i)`
     C_i corresponds to a given index of a known-sized data block (the sector)
     - for each C_i, miner reads the specified range of data from disk, which is of size `ChallengeRangeSize : C_i_output`
     - Miner generates a PartialTicket using all the PoSt witnesses
     - `PartialTicket = H(post_randomness || minerID ||S_ j || C_1_Output || … || C_K_Output)`
-1. **(check Challenge Ticket(s) for winners)** Given a returned ChallengeTicket, miner checks it is a winning ticket. Specifically, they do the following:
+
+4. **(check Challenge Ticket(s) for winners)**
+Given a returned PartialTicket, miner checks it is a winning ticket. Specifically, they do the following:
     - `ChallengeTicket = Finalize(PartialTicket).  = H(ChallengeTicket) / 2^len(H)`
```
porcuquine: I think this line may have become garbled.

`ChallengeTicket = Finalize(PartialTicket)` makes sense, but the rest of the line is unclear. To what is the expression `H(ChallengeTicket) / 2^len(H)` being assigned?

Apart from some simple confusion in what was written, I think there may also still be a terminology mismatch. I'm going to make a guess and rewrite this with the added term I think is missing:

```
FinalTicket = Finalize(PartialTicket)
ChallengeTicket = FinalTicket / 2^len(H)
```

In order to be consistent with current implementation (as we decided to do), we need for `Finalize(…)` to correspond to `H(…)`. In order to preserve your meaning for `ChallengeTicket`, we can just call the result of finalization `FinalTicket`.

src/algorithms/post/election_post.md
```diff
@@ -74,26 +84,33 @@ If no one has found a winning ticket in this epoch, increment epoch value as par
...
+3. **Verify `PartialTicket` returned**
     - Prove that the PartialTicket were appropriately derived from the eligible sectors, by submitting all miner sectors along with the wanted number of tickets and verifying that the outputted PartialTicket match.
-    1. **Derive and validate the `ChallengeTicket` from the PartialTickets**
+
+4. **Derive and validate the `ChallengeTicket` from the PartialTickets**
     - Prove the derived ChallengeTicket is below the target
     `{0, 1} ?= (ChallengeTicket < target)`
-    1. Validate the `postProof` using the `PartialTickets`
+
+5. Validate the `postProof` using the `PartialTickets`
```
porcuquine: Can we please 'verify' proofs instead of 'validating' them?

src/systems/filecoin_mining/storage_proving/storage_proving_subsystem.id
```diff
@@ -24,4 +24,9 @@ type StorageProvingSubsystem struct {
         challengeSeed     sector.PoStRandomness
         challengeTickets  [sector.ChallengeTicket]
     ) sector.PoStProof
+
+    ValidatePoStProof(
```
porcuquine: Having `ValidatePoStProof` here is even more confusing when `post_generator.id` contains `VerifyPoStProof`. I'm going to assume this is an oversight and not intentional. If intentional, please explain.

porcuquine:
> Qs for Porcu:

Some answers:

>  I want to try and get rid of all mentions of tickets since the term is overloaded, have thus moved to clearly labeling PartialTickets vs ChallengeTickets plz verif if fits proofs-land

As mentioned inline, I think to fully distinguish all entities, we also need a `FinalTicket` (name not critical, but that one seems apt).

>  Stubbed out verification stubs for storage proving subsystem and PoStGenerator, can you complete/amend?

I can do this when I'm working on PoSt tomorrow.

>  Specifically, does the proofs verif alg need to have proof type passed in?

It should because we may (almost certainly should) differentiate the two cases. At the very least, we can afford larger circuits (and therefore smaller proofs per PoSt) when not under Election PoSt time constraints. This has not yet been implemented but can be transparently if we differentiate the cases. It may make most sense for `filproofs` to just expose distinct methods, though — rather than requiring callers to fuss with marker types.

> For generation the assumption was no since only diff between ePoSt and ticket was number of tickets requesteed of GeneratePoStCandidates (ack)

Yes, but as above, if we optimize proofs for the non-election case we need to know to do so.

>  what does Surprise verification need? This may lead to amendment of what gets put on chain as part of a surprise message... do we need to store all requested partial tickets on statetree?

For now (to be compatible with implementation), it could just take all the tickets — exactly like election PoSt. However, as you point out, this could require a very large amount of on-chain storage which can be eliminated as follows: instead just put a root commitment to all the partial tickets then prove inclusion of each partial ticket in that CommT (say) along with the proof that it (the partial ticket) was derived correctly.

We could accommodate this in a way which would be forward compatible with that optimization by defining CommT as simply `Bytes` with the initial implementation being all partial tickets concatenated. I can try to account for this when I flesh out the verification methods tomorrow.

src/algorithms/post/election_post.md
sternhenri: I thought the partialTicket already corresponded to `H(...)` and Finalization was the act of doing `/2^len(H)` yielding a `ChallengeTicket`

If we can, let's try to only name two values: that yielded by the proofs library and stored on chain (what I understsand as the partial ticket, what you call a finalTicket above) and that which is checked in LE (what I call a ChallengeTicket but we can call it something else)

src/systems/filecoin_mining/storage_proving/storage_proving_subsystem.id
sternhenri: sorry. yes it is. fixed

porcuquine:
@sternhenri 
>  Does the translation of OnChainPoStVerifyInfo to PoStVerifyInfo get done elsewhere or should we do it in the StorageMinerActor ahead of calling the proofs lib?

I think I already did that except for the candidates:
https://github.com/filecoin-project/specs/pull/654/files#diff-7e0cb3d4ab4985461c75059aa7c84087R313-R318 and https://github.com/filecoin-project/specs/pull/654/files#diff-7e0cb3d4ab4985461c75059aa7c84087R356-R360.

Do you mean something different?

This is like how Seal verification is currently done, and it seems like a reasonable model.

For ElectionPoSt, we need to actually get or rebuild the candidates from wherever they were stashed on chain.

Actually, the TODOs in those snippets can be addressed by moving Candidates into onChainPostVerifyInfo. Will push that change now.

Does that address your question, or is there still something else?

src/systems/filecoin_markets/deal/_index.md
```diff
@@ -23,8 +23,7 @@ Filecoin Storage Market Deal Flow
 
 # Receive Challenge
 
-- 5. Miners enter the `Challenged` status whenever `NotifyOfPoStChallenge` is called by the chain. Miners will then have X Epoch as the ProvingPeriod to submit a successful PoSt before `
-CheckPoStSubmissionHappened` is called by the chain. Miners can only get out the challenge with `SubmitPoSt` or `onMissedPoSt`.
+- 5. Miners enter the `Challenged` status whenever `NotifyOfPoStChallenge` is called by the chain. Miners will then have X Epoch as the ProvingPeriod to submit a successful PoSt before `CheckSurprisePoStSubmissionHappened` is called by the chain. Miners can only get out the challenge with `SubmitPoSt` or `onMissedSurprisePoSt`.
```
anorth: `SubmitPoSt` is not an actor method. For clarity this should be "`SubmitSurprisePoSt` or `SubmitElectionPoSt`". Also delete the few other remaining references to `SubmitPost`.

src/systems/filecoin_mining/storage_mining/challenge_status.go
```diff
@@ -1,25 +1,46 @@
...
+	// true if most recent challenge has gone unanswered
+	return cs.LastChallengeEpoch() > cs.LastPoStResponseEpoch()
+}
+
+func (cs *ChallengeStatus_I) CanBeElected(currEpoch block.ChainEpoch) bool {
+	// true if most recent successful post (surprise or election) was recent enough
+	// and not currently getting challenged
+
+	// pull in from consts
+	PROVING_PERIOD := block.ChainEpoch(0)
```
anorth: At least write `TODO` so we have a chance of finding this later.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
```diff
@@ -136,28 +146,31 @@ type StorageMinerActorCode struct {
...
     _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch)
 
-    _verifyPoStSubmission(rt Runtime, postSubmission poster.PoStSubmission) bool
-
     _expirePreCommittedSectors(rt Runtime)
 }
 
 type MinerInfo struct {
+    // The miner actor address assigned by the Storage Power Actor
+    Address                 address.Address
```
anorth: Blocking: this is really weird, different to all the other actors, and a bad smell. The MinerInfo needs to be known before the miner actor can be created. Why is this needed?

We can add an `ActorAddress()` getter to the VM Runtime if you need access to it in actor code.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -78,82 +88,37 @@ func (a *StorageMinerActorCode_I) NotifyOfSurprisePoStChallenge(rt Runtime) Invo
...
-
-// construct PowerReport from SectorTable
-// need lastPoSt to search for new expired deals in _updateSectorUtilization
-// where DealExpirationAMT takes in a range of Epoch and return a list of values that expire in that range
-func (a *StorageMinerActorCode_I) _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch) {
-	h, st := a.State(rt)
-	newExpiredDealIDs := st._updateSectorUtilization(rt, lastPoSt)
-	activePower := st._getActivePower(rt)
-	inactivePower := st._getInactivePower(rt)
+// Called by the cron actor at every tick. Will return immediately if the miner
```
anorth: I was under the impression that we don't need to poll every miner actor at every epoch - that's going to get expensive fast. Surely we only need to poll challenged actors, and it would be worth recording that set somewhere? @zixuanzh

anorth: Also I recommend inserting an `OnCronTick` method, that may just immediately delegate to this one, but provide room for extending the functionality without causing the method name to drift.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -333,61 +244,247 @@ func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime, postSubmission p
...
 	if !a._isChallenged(rt) {
 		// TODO: determine proper error here and error-handling machinery
 		rt.Abort("cannot SubmitSurprisePoSt when not challenged")
 	}
 
-	// Verify correct PoSt Submission
-	isPoStVerified := a._verifyPoStSubmission(rt, postSubmission)
-	if !isPoStVerified {
-		// no state transition, just error out and miner should submitSurprisePoSt again
+	// to pull in from constants
```
anorth: `TODO`

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -333,61 +244,247 @@ func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime, postSubmission p
...
+	PROVING_PERIOD := block.ChainEpoch(0)
+
+	// 2. Check that the challenge has not expired
+	// Check that miner can still submit (i.e. that the challenge window has not passed)
+	// This will prevent miner from submitting a Surprise PoSt past the challenge period
+	_, st := a.State(rt)
+	if rt.CurrEpoch() < st.ChallengeStatus().LastChallengeEpoch()+PROVING_PERIOD {
```
anorth:
```suggestion
	if rt.CurrEpoch() > st.ChallengeStatus().LastChallengeEpoch()+PROVING_PERIOD {
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -333,61 +244,247 @@ func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime, postSubmission p
...
+		rt.Abort("cannot SubmitSurprisePoSt late")
```
anorth: How about just returning `false`? This case doesn't seem like it should be particularly privileged.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -333,61 +244,247 @@ func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime, postSubmission p
...
+	// TODO: validate caller
+	// the caller MUST be the miner who won the block (who won the block should be callable as a a VM runtime call)
+	// we also need to enforce that this call happens only once per block, OR make it not callable by special privileged messages
+	TODO()
+
+	// 1. Check that the miner in question is currently allowed to run election
+	if !a._canBeElected(rt) {
 		// TODO: determine proper error here and error-handling machinery
-		rt.Abort("TODO")
+		rt.Abort("cannot submit an election proof if challenged or having challenge response failure")
```
anorth: Return `false`? Or return some status enumeration?

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -333,61 +244,247 @@ func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime, postSubmission p
...
+	}
+
+	sdr := filproofs.WinSDRParams(&filproofs.SDRCfg_I{SurprisePoStCfg_: &postCfg})
+
+	// 5. Verify the PoSt Proof
+	isPoStVerified := sdr.VerifySurprisePoSt(&pvInfo)
+	return isPoStVerified
+}
+
+func (a *StorageMinerActorCode_I) VerifyElectionPoSt(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo) bool {
```
anorth: @sternhenri do I understand correctly that this method is read-only and will never actually be invoked "on chain", but as part of block validation. I.e. this same code could exist outside the actor itself and everything would be fine.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
sternhenri: mmh yeah, I sort of wung it, so thank you for catching. I should have called attention to it. I don't get why this can't be set when CreateMiner is called on the SPA, but it is used to generate randomness in case different miner actors share owner or worker keys...

What do you recommend?

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
sternhenri: yes, every tick may not be every epoch @anorth (which is why i mucked around with how the state was stored to prevent miners from being able to submit PoSts due to cron delay even if they had passed deadlines.

The issue with the challenged set is that the current method of checking which are challenged uses onChainRandomness from the epoch to answer `ShouldChallenge()`. I think this is worth touching up, but in a different PR.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
sternhenri: good catch!

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
sternhenri: Yeah I think we should do a pass for error handling in general. Don't want to do so now.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
sternhenri: Absolutely right. Given its use of a lot of things from runtime I decided to add it here, but had started implementing it in the `StorageMiningSubsystem` initially. Given your earlier q about querying chainstate, I could move it there. IN general I feel like a lot of read-only stuff is in our actor code which is a mistake. Actor state should be made equally accessible outside the VM.

What do you think?

sternhenri: I just moved it there @anorth. I think it's cleaner to leave readonly methods in the subsystem (where the miner should call them ahead of calling the actor methods anyways)

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
```diff
@@ -125,3 +128,120 @@ func (sms *StorageMiningSubsystem_I) PrepareNewTicket(randomness util.Randomness
 
 	return newTicket
 }
+
+// TODO Import IPLDStore
+var localIPLDStore ipld.GraphStore
+
```
sternhenri: @jzimmerman Check out the below if you can!

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -204,7 +204,7 @@ func (f *Feistel_I) Permute(size UInt, i UInt) UInt {
 	panic("TODO")
 }
 
-func (sdr *WinStackedDRG_I) Seal(sid sector.SectorID, data []byte, randomness sector.SealRandomness) SealSetupArtifacts {
+func (sdr *WinStackedDRG_I) Seal(sid sector.SectorID, data []byte, randomness util.Randomness) SealSetupArtifacts {
```
porcuquine: Why did you remove these types everywhere? This seems like a pure loss of information and makes the filproofs code less clear.

porcuquine: More than less clear: it will make it harder to use the spec to communicate (and ensure) correctness.

src/systems/filecoin_mining/storage_mining/constants.go
```diff
@@ -7,3 +7,4 @@ const MAX_PROVE_COMMIT_SECTOR_PERIOD = block.ChainEpoch(3)    // placeholder
 const MAX_SURPRISE_POST_RESPONSE_PERIOD = block.ChainEpoch(4) // placeholder
 const POST_CHALLENGE_TIME = block.ChainEpoch(1)               // placeholder
 const PROVING_PERIOD = block.ChainEpoch(2)                    // placeholder
+const CHALLENGE_FREQUENCY = 2
```
anorth: Document this: what challenge, in what units?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
anorth: I can't find any reference to this field in this PR, but nevertheless do not add it.

If you need to access it from within actor code, add a `ReceiverAddress() addr.Address` accessor method to `Runtime()`. Otherwise, plumb the address through from wherever you looked up the actor state in the state tree.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -78,82 +88,44 @@ func (a *StorageMinerActorCode_I) NotifyOfSurprisePoStChallenge(rt Runtime) Invo
...
-func (a *StorageMinerActorCode_I) _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch) {
-	h, st := a.State(rt)
-	newExpiredDealIDs := st._updateSectorUtilization(rt, lastPoSt)
-	activePower := st._getActivePower(rt)
-	inactivePower := st._getInactivePower(rt)
+// Will return immediately if the miner
+// has successfully submitted a PoSt (and is no longer challenged)
+// If a Post is missed (either due to faults being not declared on time or
+// because the miner run out of time, every sector is reported as failing
+// for the current proving period.
```
anorth: ❗️ This comment is generally misleading as to the method's behaviour and leads me to think it's not doing what we want it to.

An accurate comment might read:
`// If the miner is currently challenged for surprise PoSt, reports every sector as failing for the current proving period.`

There would seem to be a huge precondition that this is only called when the challenge window has passed, not when it's just opened. But that precondition is not met: this is called on every cron tick.

In any case, details about why PoSt is missed are out of scope here.

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
anorth: If this is the right kind of pattern, then the ipld store should be added as a member of StorageMiningSubsystem and then this would look pretty straightforward.

Can we get some kind of `Get(cid, interface{})` that decodes into the second parameter?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
anorth: Added in https://github.com/filecoin-project/specs/pull/666

sternhenri: Thanks Alex

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
sternhenri: I actually think the IPLD store might be added to the node or some subsystem available to all others so they can all access the single one...

Looking into the Get you mention, it's a good idea. I just worry that .ids don't support generics so passing in the interface type may not work cc @jzimmerman

src/libraries/filcrypto/filproofs/algorithms.go
sternhenri: fixed. sorry

src/systems/filecoin_mining/storage_mining/constants.go
sternhenri: documented in the election_post.md file already but changed to more explicit variable names :)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
sternhenri: Why do you say misleading? I think more details in comments is generally good. (cc @zixuanzh who is the original author I believe)

sternhenri: replaced anyhow

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
jzimmerman: You can definitely pass interface types, even though we don't yet have support for generics -- just use a named Go type such as `util.Any`. However, it's the method implementation itself that concerns me more, since Go doesn't support generics either (and reflection in cases like this seems quite error-prone).

For the moment, I would suggest having `ipldStore.Get(CID)` return a `util.Serialization`, which can then immediately be deserialized via the appropriate, auto-generated type-specific methods (e.g. `Deserialize_StorageMinerActorState` in the code below).

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -78,82 +88,41 @@ func (a *StorageMinerActorCode_I) NotifyOfSurprisePoStChallenge(rt Runtime) Invo
...
-// need lastPoSt to search for new expired deals in _updateSectorUtilization
-// where DealExpirationAMT takes in a range of Epoch and return a list of values that expire in that range
-func (a *StorageMinerActorCode_I) _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch) {
-	h, st := a.State(rt)
-	newExpiredDealIDs := st._updateSectorUtilization(rt, lastPoSt)
-	activePower := st._getActivePower(rt)
-	inactivePower := st._getInactivePower(rt)
+// If the miner is currently challenged for surprise PoSt,
+// reports every sector as failing for the current proving period.
+func (a *StorageMinerActorCode_I) CheckSurprisePoStSubmissionHappened(rt Runtime) InvocOutput {
```
anorth: So the comment is right now, but it still seems very wrong to me that this is called on every cron tick. The miner will be instantly faulted with no window to compute PoSt.

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -1,14 +1,23 @@
...
+	msg "github.com/filecoin-project/specs/systems/filecoin_vm/message"
+
+	addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
+
+	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
+
+	exitcode "github.com/filecoin-project/specs/systems/filecoin_vm/runtime/exitcode"
+
+	util "github.com/filecoin-project/specs/util"
+)
```
anorth: Don't add all these blank lines.

anorth:
I have no further objections after the use of `CheckSurprisePoStSubmissionHappened` is resolved.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
sternhenri: Now I've seen the largeer issue. Sorry I missed your point earlier. I believe the `ExpirePreCommitted` is as intendeed per @zixuanzh. The rest is indeed faulty. Fixed

anorth: The comment could be tweaked for the new better semantics.

zixuanzh: @anorth i think it is okay that miner is immediately faulted when they have missed their SurprisePoSt submission window. It is a separate issue if we think calling this at every Cron tick is too expensive. @jbenet previously thinks this is okay.

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -864,7 +864,10 @@ func (sdr *WinStackedDRG_I) _getChallengedSectors(randomness sector.PoStRandomne
 	panic("TODO")
 }
 
-func (sdr *WinStackedDRG_I) GeneratePoStCandidates(challengeSeed sector.PoStRandomness, eligibleSectors []sector.SectorNumber, candidateCount int, sectorStore sectorIndex.SectorStore) []sector.ChallengeTicket {
+////////////////////////////////////////////////////////////////////////////////
+// Election PoSt
+
+func (sdr *WinStackedDRG_I) GenerateElectionPoStCandidates(challengeSeed sector.PoStRandomness, eligibleSectors []sector.SectorNumber, candidateCount int, sectorStore sectorIndex.SectorStore) []sector.PoStCandidate {
```
zixuanzh: is `eligibleSectors` the same as `ProvingSet`? Let's stick to the same naming?

src/libraries/filcrypto/filproofs/algorithms.id
```diff
@@ -112,14 +112,23 @@ type WinStackedDRG struct {
...
         candidateCount   int
         sectorStore      sectorIndex.SectorStore
-    ) [sector.ChallengeTicket]
-    GeneratePoStProof(privateProofs [sector.PrivatePoStProof]) sector.PoStProof
-    VerifyPoSt(sv sector.PoStVerifyInfo) bool
+    ) [sector.PoStCandidate]
+    GenerateElectionPoStProof(privateProofs [sector.PrivatePoStProof]) sector.PoStProof
+    VerifyElectionPoSt(sv sector.PoStVerifyInfo) bool
+
+    GenerateSurprisePoStCandidates(
```
zixuanzh: why do we have SurprisePoStCandidates and SurprisePoStProofs? I thought it is one single method but I can be totally wrong on this.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
```diff
@@ -73,17 +73,17 @@ func (spc *StoragePowerConsensusSubsystem_I) StoragePowerConsensusError(errMsg s
...
 	activePowerInSector := uint64(block.StoragePower(0))
 
 	// TODO: Send message to SPA
 	totalPower := uint64(block.StoragePower(0))
 
-	electionProof := block.SHA256(challengeTicket.Ticket())
+	challengeTicket := block.SHA256(partialTicket)
 
-	return spc.ec().IsWinningElectionProof(electionProof, activePowerInSector, totalPower)
+	return spc.ec().IsWinningChallengeTicket(challengeTicket, activePowerInSector, totalPower)
```
zixuanzh: Given the distinction between Active and Inactive power and inactive power (committed, recovering, and failing) requires pledge collateral, we might want to do a write up so that we are fully aligned and understood the protocol impact here. I can be the one writing this up too, cc @jzimmerman .

src/systems/filecoin_markets/deal/_index.md
```diff
@@ -23,8 +23,7 @@ Filecoin Storage Market Deal Flow
 
 # Receive Challenge
 
-- 5. Miners enter the `Challenged` status whenever `NotifyOfPoStChallenge` is called by the chain. Miners will then have X Epoch as the ProvingPeriod to submit a successful PoSt before `
-CheckPoStSubmissionHappened` is called by the chain. Miners can only get out the challenge with `SubmitPoSt` or `onMissedPoSt`.
+- 5. Miners enter the `Challenged` status whenever `NotifyOfPoStChallenge` is called by the chain. Miners will then have X Epoch as the ProvingPeriod to submit a successful PoSt before `CheckSurprisePoStSubmissionHappened` is called by the chain. Miners can only get out the challenge with `SubmitSurprisePoSt` or `onMissedSurprisePoSt`.
```
zixuanzh: Miner also enters and exits the Challenged status in the same epoch when generating ElectionPoSt. Is this correct?

src/systems/filecoin_mining/storage_mining/challenge_status.go
```diff
@@ -1,25 +1,53 @@
...
+	// TODO: pull in from consts
+	PROVING_PERIOD := block.ChainEpoch(0)
+	return cs.IsChallenged() && currEpoch > cs.LastChallengeEpoch()+PROVING_PERIOD
+}
+
+func (cs *ChallengeStatus_I) CanBeElected(currEpoch block.ChainEpoch) bool {
+	// true if most recent successful post (surprise or election) was recent enough
+	// and not currently getting challenged
+
+	// TODO: pull in from consts
```
zixuanzh: there is a `constants.go` in storage_mining that this can live in

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
```diff
@@ -125,3 +129,116 @@ func (sms *StorageMiningSubsystem_I) PrepareNewTicket(randomness util.Randomness
...
+	var serializedSubstate Serialization
+	st, err := Deserialize_StorageMinerActorState(serializedSubstate)
+
+	if err == nil {
+		panic("Deserialization error")
+	}
+	return st
+}
+
+func (sms *StorageMiningSubsystem_I) VerifyElectionPoSt(header block.BlockHeader, onChainInfo sector.OnChainPoStVerifyInfo) bool {
```
zixuanzh: 👍seeing these makes me so happy, great work!!

src/systems/filecoin_mining/storage_mining/challenge_status.go
sternhenri: yeah I know. I just don't know how we should go about pulling things in from there (dependency wise). It may be trivial.

src/systems/filecoin_markets/deal/_index.md
sternhenri: yes exactly

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
sternhenri: Happy to review if you do @zixuanzh. I have less of a sense than you here.

src/libraries/filcrypto/filproofs/algorithms.id
sternhenri: no it's a two time operation since you will only generate a proof on a portion of your candidates (the winning ones) in ePost

src/libraries/filcrypto/filproofs/algorithms.go
sternhenri: mmh that is a good point. CC @porcuquine lord of filproofs

PR #656, opened 2019-11-26
Check Pledge Collateral Requirement at PoSt Submission and Pledge Slashing

zixuanzh:
In this PR:

- [x] `EnsurePledgeCollateral` is added to `SubmitSurprise`, a SurprisePoSt is only valid when the miner has sufficient balance in StoragePowerActor. In the event that a SurprisePoSt fails because of insufficient pledge collateral, miners need to add Balance to StoragePowerActor and submit SurprisePoSt again. Note that election PoSt will always go through but some block reward will be sent to LockedBalance in StoragePowerActor in the event of undercollateralization.
- [x] Add slashPledgeForStorageFault, the exact amount to slash and rules in determining how much to slash under different storage faults will be added in a future PR
- [x] Replace `_submitFaultReport` with `_slashCollateralForStorageFaults` that calls `_slashDealsForStorageFault` and `_slashPledgeForStorageFault` depending on the `StorageFaultType`
- [x] Add `_ensurePledgeCollateral` to `WithdrawBalance` 
- [x] Wire collateral slashing between actors with rt.SendCatchingErrors
- [x] Add PledgeCollateralSlashed to SMA (similar to DealCollateralSlashed in StorageMarketActor), we can migrate this to an actor later
- [x] Move `_slashCollateralForStorageFaults` to be invoked after `_submitPowerReport` so that utilizationInfo is up-to-date when checking for affected power

Future PRs:

- [ ] Align on the shape and mechanism for pledge collateral function
- [ ] Decide on how much pledge collateral to slash given different storage fault types
- [ ] Align and decide if we should add a condition to leader election such that Pledge Collateral must be satisfied before being eligible to win a block or add Pledge Collateral check to Block Validation to work well with ElectionPoSt
- [ ] Add block reward garnishing in the event of under collateralization

### DIFF
--- a/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
+++ b/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
@@ -15,9 +15,9 @@ import (
 )
 
 const (
-	MethodProcessPowerReport          = actor.MethodPlaceholder
-	MethodSlashPledgeForStorageFaults = actor.MethodPlaceholder
-	EnsurePledgeCollateralSatisfied   = actor.MethodPlaceholder
+	MethodProcessPowerReport         = actor.MethodPlaceholder
+	MethodSlashPledgeForStorageFault = actor.MethodPlaceholder
+	EnsurePledgeCollateralSatisfied  = actor.MethodPlaceholder
 )
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
+++ b/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
@@ -92,13 +92,14 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, amount actor.Token
 
 	h, st := a.State(rt)
 
-	currEntry, found := st.PowerTable()[minerID]
-	if !found {
-		rt.Abort("minerID not found.")
+	ret := st._ensurePledgeCollateralSatisfied(rt)
+	if !ret {
+		rt.Abort("insufficient pledge collateral.")
 	}
 
+	currEntry := st._safeGetPowerEntry(rt, minerID)
 	if currEntry.AvailableBalance() < amount {
-		rt.Abort("insufficient balance.")
+		rt.Abort("insufficient available balance.")
 	}
 
 	currEntry.Impl().AvailableBalance_ = currEntry.AvailableBalance() - amount
@@ -175,28 +176,19 @@ func (a *StoragePowerActorCode_I) GetTotalPower(rt Runtime) block.StoragePower {
 
 func (a *StoragePowerActorCode_I) EnsurePledgeCollateralSatisfied(rt Runtime) InvocOutput {
 
-	// TODO: convert msgSender to MinerActorID
-	var minerID addr.Address
-
 	h, st := a.State(rt)
-
-	powerEntry := st._safeGetPowerEntry(rt, minerID)
-	pledgeCollateralRequired := st._getPledgeCollateralReq(rt, powerEntry.ActivePower()+powerEntry.InactivePower())
-
-	if pledgeCollateralRequired < powerEntry.LockedPledgeCollateral() {
-		return rt.SuccessReturn()
-	} else if pledgeCollateralRequired < (powerEntry.LockedPledgeCollateral() + powerEntry.AvailableBalance()) {
-		st._lockPledgeCollateral(rt, minerID, (pledgeCollateralRequired - powerEntry.LockedPledgeCollateral()))
-		return rt.SuccessReturn()
-	}
-
+	ret := st._ensurePledgeCollateralSatisfied(rt)
 	UpdateRelease(rt, h, st)
 
-	return rt.ErrorReturn(exitcode.InsufficientPledgeCollateral)
+	if !ret {
+		return rt.ErrorReturn(exitcode.InsufficientPledgeCollateral)
+	}
+
+	return rt.SuccessReturn()
 }
 
 // slash pledge collateral for Declared, Detected and Terminated faults
-func (a *StoragePowerActorCode_I) SlashPledgeForStorageFaults(rt Runtime, affectedPower block.StoragePower, faultType sector.StorageFaultType) {
+func (a *StoragePowerActorCode_I) SlashPledgeForStorageFault(rt Runtime, affectedPower block.StoragePower, faultType sector.StorageFaultType) {
 
 	var msgSender addr.Address // TODO replace this
 
--- a/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
+++ b/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
@@ -10,7 +10,6 @@ import (
 	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
 	addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
 	vmr "github.com/filecoin-project/specs/systems/filecoin_vm/runtime"
-	exitcode "github.com/filecoin-project/specs/systems/filecoin_vm/runtime/exitcode"
 	util "github.com/filecoin-project/specs/util"
 )
 
@@ -18,6 +17,15 @@ const (
 	MethodProcessPowerReport         = actor.MethodPlaceholder
 	MethodSlashPledgeForStorageFault = actor.MethodPlaceholder
 	EnsurePledgeCollateralSatisfied  = actor.MethodPlaceholder
+)
+
+// placeholder values
+// these are the scaling constants for percentage pledge collateral to slash
+// given a miner's affected power and its total power
+const (
+	DeclaredFaultSlashPercent   = 1   // placeholder
+	DetectedFaultSlashPercent   = 10  // placeholder
+	TerminatedFaultSlashPercent = 100 // placeholder
 )
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -57,15 +65,14 @@ func DeserializeState(x Bytes) State {
 
 func (a *StoragePowerActorCode_I) AddBalance(rt Runtime) {
 
-	var msgValue actor.TokenAmount
+	msgValue := rt.ValueReceived()
 
 	// TODO: this should be enforced somewhere else
 	if msgValue < 0 {
 		rt.Abort("negative message value.")
 	}
 
-	// TODO: convert msgSender to MinerActorID
-	var minerID addr.Address
+	minerID := rt.ImmediateCaller()
 
 	h, st := a.State(rt)
 
@@ -87,8 +94,7 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, amount actor.Token
 		rt.Abort("negative amount.")
 	}
 
-	// TODO: convert msgSender to MinerActorID
-	var minerID addr.Address
+	minerID := rt.ImmediateCaller()
 
 	h, st := a.State(rt)
 
@@ -107,7 +113,11 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, amount actor.Token
 
 	UpdateRelease(rt, h, st)
 
-	// TODO: send funds to msgSender
+	// send funds to miner
+	rt.SendPropagatingErrors(&msg.InvocInput_I{
+		To_:    minerID,
+		Value_: amount,
+	})
 }
 
 func (a *StoragePowerActorCode_I) CreateStorageMiner(
@@ -129,9 +139,7 @@ func (a *StoragePowerActorCode_I) CreateStorageMiner(
 	// store ownerAddr and workerAddr there
 	// and return StorageMinerActor address
 
-	// TODO: minerID should be a MinerActorID
-	// which is smaller than MinerAddress
-	var minerID addr.Address
+	minerID := rt.ImmediateCaller()
 
 	h, st := a.State(rt)
 
@@ -145,8 +153,7 @@ func (a *StoragePowerActorCode_I) CreateStorageMiner(
 
 func (a *StoragePowerActorCode_I) RemoveStorageMiner(rt Runtime, address addr.Address) {
 
-	// TODO: make explicit address type
-	var minerID addr.Address
+	minerID := rt.ImmediateCaller()
 
 	h, st := a.State(rt)
 
@@ -174,43 +181,54 @@ func (a *StoragePowerActorCode_I) GetTotalPower(rt Runtime) block.StoragePower {
 	return totalPower
 }
 
-func (a *StoragePowerActorCode_I) EnsurePledgeCollateralSatisfied(rt Runtime) InvocOutput {
+func (a *StoragePowerActorCode_I) EnsurePledgeCollateralSatisfied(rt Runtime) {
 
 	h, st := a.State(rt)
 	ret := st._ensurePledgeCollateralSatisfied(rt)
 	UpdateRelease(rt, h, st)
 
 	if !ret {
-		return rt.ErrorReturn(exitcode.InsufficientPledgeCollateral)
+		rt.Abort("exitcode.InsufficientPledgeCollateral")
 	}
 
-	return rt.SuccessReturn()
 }
 
 // slash pledge collateral for Declared, Detected and Terminated faults
 func (a *StoragePowerActorCode_I) SlashPledgeForStorageFault(rt Runtime, affectedPower block.StoragePower, faultType sector.StorageFaultType) {
 
-	var msgSender addr.Address // TODO replace this
-
-	// placeholder values
-	const unitDeclaredFaultSlash = 1    // placeholder
-	const unitDetectedFaultSlash = 5    // placeholder
-	const unitTerminatedFaultSlash = 10 // placeholder
+	minerID := rt.ImmediateCaller()
 
 	h, st := a.State(rt)
 
+	// TODO: revisit this calculation
+	powerEntry := st._safeGetPowerEntry(rt, minerID)
+	totalPower := powerEntry.ActivePower() + powerEntry.InactivePower()
+	pledgeRequired := st._getPledgeCollateralReq(rt, totalPower)
+	pledgeAffected := actor.TokenAmount(uint64(pledgeRequired) * uint64(affectedPower) / uint64(totalPower))
+
 	switch faultType {
 	case sector.DeclaredFault:
-		amountToSlash := actor.TokenAmount(unitDeclaredFaultSlash * affectedPower)
-		st._slashPledgeCollateral(rt, msgSender, amountToSlash)
+		amountToSlash := actor.TokenAmount(DeclaredFaultSlashPercent * uint64(pledgeAffected) / 100)
+		st._slashPledgeCollateral(rt, minerID, amountToSlash)
 	case sector.DetectedFault:
-		amountToSlash := actor.TokenAmount(unitDetectedFaultSlash * affectedPower)
-		st._slashPledgeCollateral(rt, msgSender, amountToSlash)
+		amountToSlash := actor.TokenAmount(DetectedFaultSlashPercent * uint64(pledgeAffected) / 100)
+		st._slashPledgeCollateral(rt, minerID, amountToSlash)
 	case sector.TerminatedFault:
-		amountToSlash := actor.TokenAmount(unitTerminatedFaultSlash * affectedPower)
-		st._slashPledgeCollateral(rt, msgSender, amountToSlash)
+		amountToSlash := actor.TokenAmount(TerminatedFaultSlashPercent * uint64(pledgeAffected) / 100)
+		st._slashPledgeCollateral(rt, minerID, amountToSlash)
 	}
 
+	// this is done this way because we can't send messages when holding onto the state lock
+	amountToSlash := st.PledgeCollateralSlashed()
+	UpdateRelease(rt, h, st)
+
+	rt.SendPropagatingErrors(&msg.InvocInput_I{
+		To_:    addr.BurntFundsActorAddr,
+		Value_: amountToSlash,
+	})
+
+	h, st = a.State(rt)
+	st.Impl().PledgeCollateralSlashed_ = actor.TokenAmount(0)
 	UpdateRelease(rt, h, st)
 }
 
@@ -218,8 +236,7 @@ func (a *StoragePowerActorCode_I) SlashPledgeForStorageFault(rt Runtime, affecte
 // update miner power based on the power report
 func (a *StoragePowerActorCode_I) ProcessPowerReport(rt Runtime, report PowerReport) {
 
-	// TODO: convert msgSender to MinerActorID
-	var minerID addr.Address
+	minerID := rt.ImmediateCaller()
 
 	h, st := a.State(rt)
 
--- a/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
+++ b/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
@@ -15,9 +15,9 @@ import (
 )
 
 const (
-	MethodProcessPowerReport        = actor.MethodPlaceholder
-	MethodProcessFaultReport        = actor.MethodPlaceholder
-	EnsurePledgeCollateralSatisfied = actor.MethodPlaceholder
+	MethodProcessPowerReport          = actor.MethodPlaceholder
+	MethodSlashPledgeForStorageFaults = actor.MethodPlaceholder
+	EnsurePledgeCollateralSatisfied   = actor.MethodPlaceholder
 )
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -202,17 +202,17 @@ func (a *StoragePowerActorCode_I) EnsurePledgeCollateralSatisfied(rt Runtime) In
 
 // @param fault report
 // slash pledge collateral for Declared, Detected and Terminated faults
-func (a *StoragePowerActorCode_I) ProcessFaultReport(rt Runtime, report sector.FaultReport) {
+func (a *StoragePowerActorCode_I) SlashPledgeForStorageFaults(rt Runtime, faultType sector.StorageFaultType) {
 
-	var msgSender addr.Address // TODO replace this
+	// var msgSender addr.Address // TODO replace this
 
 	h, st := a.State(rt)
 
-	declaredFaultSlash := report.GetDeclaredFaultSlash()
-	detectedFaultSlash := report.GetDetectedFaultSlash()
-	terminatedFaultSlash := report.GetTerminatedFaultSlash()
+	// declaredFaultSlash := report.GetDeclaredFaultSlash()
+	// detectedFaultSlash := report.GetDetectedFaultSlash()
+	// terminatedFaultSlash := report.GetTerminatedFaultSlash()
 
-	st._slashPledgeCollateral(rt, msgSender, (declaredFaultSlash + detectedFaultSlash + terminatedFaultSlash))
+	// st._slashPledgeCollateral(rt, msgSender, (declaredFaultSlash + detectedFaultSlash + terminatedFaultSlash))
 
 	UpdateRelease(rt, h, st)
 }
--- a/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
+++ b/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
@@ -10,12 +10,14 @@ import (
 	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
 	addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
 	vmr "github.com/filecoin-project/specs/systems/filecoin_vm/runtime"
+	exitcode "github.com/filecoin-project/specs/systems/filecoin_vm/runtime/exitcode"
 	util "github.com/filecoin-project/specs/util"
 )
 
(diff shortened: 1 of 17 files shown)


src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -91,13 +93,14 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, amount actor.Token
 
 	h, st := a.State(rt)
 
-	currEntry, found := st.PowerTable()[minerID]
-	if !found {
-		rt.Abort("minerID not found.")
+	ret := st._ensurePledgeCollateralSatisfied(rt)
+	if !ret {
+		rt.Abort("insufficient pledge collateral.")
```
anorth: This is unnecessary. Locked-up pledge collateral should not count as `AvailableBalance` so the check on 102 already covers this.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -91,13 +93,14 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, amount actor.Token
...
 	}
 
+	currEntry := st._safeGetPowerEntry(rt, minerID)
 	if currEntry.AvailableBalance() < amount {
-		rt.Abort("insufficient balance.")
+		rt.Abort("insufficient available balance.")
 	}
 
 	currEntry.Impl().AvailableBalance_ = currEntry.AvailableBalance() - amount
```
anorth: Please satisfy the `// TODO: send funds to msgSender` now

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -172,62 +175,56 @@ func (a *StoragePowerActorCode_I) GetTotalPower(rt Runtime) block.StoragePower {
...
 
-	declaredFaultSlash := report.GetDeclaredFaultSlash()
-	detectedFaultSlash := report.GetDetectedFaultSlash()
-	terminatedFaultSlash := report.GetTerminatedFaultSlash()
+	h, st := a.State(rt)
 
-	st._slashPledgeCollateral(rt, msgSender, (declaredFaultSlash + detectedFaultSlash + terminatedFaultSlash))
+	switch faultType {
+	case sector.DeclaredFault:
+		amountToSlash := actor.TokenAmount(unitDeclaredFaultSlash * affectedPower)
```
anorth: These expressions all need to be rewritten as proportions of the miner's pledge collateral requirement, which is dynamic.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
```diff
@@ -29,14 +26,12 @@ func (st *StoragePowerActorState_I) _slashPledgeCollateral(rt Runtime, address a
 		currEntry.Impl().LockedPledgeCollateral_ = 0
 		// TODO: extra handling of not having enough pledgecollateral to be slashed
 	} else {
-		currEntry.Impl().LockedPledgeCollateral_ = currEntry.LockedPledgeCollateral() - amount
+		currEntry.Impl().LockedPledgeCollateral_ = currEntry.LockedPledgeCollateral() - amountToSlash
 	}
 
-	// TODO: send amountToSlash to TreasuryActor
-	panic(amountToSlash)
+	st.Impl().PledgeCollateralSlashed_ += amountToSlash
```
anorth: So we're not using the burnt funds actor? Will the storage market actor similarly hold on to slashed deal collateral? If so we should delete reference to the burnt funds actor from the spec entirely.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
```diff
@@ -130,3 +118,31 @@ func (st *StoragePowerActorState_I) _sampleMinersToSurprise(rt Runtime, challeng
...
+
+	powerEntry := st._safeGetPowerEntry(rt, minerID)
+	pledgeCollateralRequired := st._getPledgeCollateralReq(rt, powerEntry.ActivePower()+powerEntry.InactivePower())
+
+	if pledgeCollateralRequired < powerEntry.LockedPledgeCollateral() {
+		return true
+	} else if pledgeCollateralRequired < (powerEntry.LockedPledgeCollateral() + powerEntry.AvailableBalance()) {
+		st._lockPledgeCollateral(rt, minerID, (pledgeCollateralRequired - powerEntry.LockedPledgeCollateral()))
+		return true
+	}
```
anorth: Blocking: this whole construction assumes that pledge collateral requirements only increase, which isn't true (miners can reduce stake, which must free up collateral). It would be much clearer to:
- compute requirement
- compute delta between requirement and locked
- if delta < available, update locked by delta (delta may be negative) and return true

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
```diff
@@ -100,45 +100,41 @@ type StorageMinerActorState struct {
 }
 
 type StorageMinerActorCode struct {
+    _isChallenged(rt Runtime) bool
```
anorth: Move private methods to the bottom for consistency with others

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -198,17 +200,17 @@ func (a *StorageMarketActorCode_I) ActivateDeals(rt Runtime, dealIDs []deal.Deal
...
 	// only terminated fault will result in slashing of deal collateral
 
-	switch slashAction {
-	case deal.SlashTerminatedFaults:
-		st._slashTerminatedFaults(rt, dealIDs)
+	switch faultType {
+	case sector.TerminatedFault:
+		st._slashTerminatedFault(rt, dealIDs)
 	default:
 		rt.Abort("sma.ProcessDealSlash: invalid action type")
```
anorth: Don't abort, just do nothing.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -99,31 +99,62 @@ func (a *StorageMinerActorCode_I) _slashDealsFromFaultReport(rt Runtime, sectorN
 
 	Release(rt, h, st)
 
-	// TODO: Send(StorageMarketActor, ProcessDealSlash)
+	processDealSlashParam := make([]actor.MethodParam, 2)
+	rt.SendCatchingErrors(&msg.InvocInput_I{
+		To_:     addr.StorageMarketActorAddr,
+		Method_: market.MethodProcessDealSlash,
+		Params_: processDealSlashParam,
+	})
```
anorth: Blocking: this should not be catching errors. there's no good reason for the call to fail, and if it does (out of gas?) the whole invocation chain should fail too. Defensive programming like this will become really hard to reason about.

If there's some really good justification for why a failure here must be recovered while the rest of this succeeds, it needs a long comment explaining it.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -99,31 +99,62 @@ func (a *StorageMinerActorCode_I) _slashDealsFromFaultReport(rt Runtime, sectorN
...
+	for _, sectorNo := range sectorNumbers {
+
+		utilizationInfo := st._getUtilizationInfo(rt, sectorNo)
+		affectedPower += utilizationInfo.CurrUtilization()
+
+	}
+
+	Release(rt, h, st)
+
+	panic(affectedPower)
```
anorth: Don't panic

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -99,31 +99,62 @@ func (a *StorageMinerActorCode_I) _slashDealsFromFaultReport(rt Runtime, sectorN
...
+	slashPledgeParams := make([]actor.MethodParam, 3)
+
+	rt.SendCatchingErrors(&msg.InvocInput_I{
+		To_:     addr.StoragePowerActorAddr,
+		Method_: power.MethodSlashPledgeForStorageFault,
+		Params_: slashPledgeParams,
+	})
```
anorth: Blocking: more questionable defensive programming

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -138,18 +169,36 @@ func (a *StorageMinerActorCode_I) _submitPowerReport(rt Runtime, lastPoSt block.
...
+		Params_: processPowerReportParam,
+	})
 
 	if len(newExpiredDealIDs) > 0 {
-		// Send(StorageMarketActor, ProcessDealExpiration)
+		rt.SendCatchingErrors(&msg.InvocInput_I{
+			To_:     addr.StorageMarketActorAddr,
+			Method_: market.MethodProcessDealExpiration,
+			Params_: processDealExpirationParam,
+		})
```
anorth: Ditto for both of the above: we have to write code that works, not defensive checks around code that doesn't. These methods should not fail for any reason preventable by the programmer, and unpreventable errors should revert the entire state change.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -684,6 +729,19 @@ func (a *StorageMinerActorCode_I) ProveCommitSector(rt Runtime, info sector.Sect
...
+	emptyParams := make([]actor.MethodParam, 0)
+	ret := rt.SendCatchingErrors(&msg.InvocInput_I{
+		To_:     addr.StoragePowerActorAddr,
+		Method_: power.EnsurePledgeCollateralSatisfied,
+		Params_: emptyParams,
+	})
+
+	if ret.ExitCode() == exitcode.InsufficientPledgeCollateral {
+		rt.Abort("sma._onSuccessfulPoSt: insufficient pledge collateral.")
+	}
```
anorth: What's going on here, it's catching the error only to re-Abort? What other error are you expecting that you don't want to cause an abort?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -91,13 +93,14 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, amount actor.Token
...
-	currEntry, found := st.PowerTable()[minerID]
-	if !found {
-		rt.Abort("minerID not found.")
+	ret := st._ensurePledgeCollateralSatisfied(rt)
+	if !ret {
+		rt.Abort("insufficient pledge collateral.")
 	}
 
+	currEntry := st._safeGetPowerEntry(rt, minerID)
 	if currEntry.AvailableBalance() < amount {
```
jzimmerman: I'm a bit concerned about manual balance accounting -- I think, as much as possible, we should statically enforce that currency is not created or destroyed in any of these operations, but rather is only moved via Send calls (or `_transferFunds` calls, if needed outside the VM).

In this particular case, instead of `AvailableBalance`, I would recommend that the singleton actors each keep a simple running total, per sender-actor, of how much currency they are currently holding on that sender's behalf. Then it's easy to abstract this out, so that we have a single method that ensures this is accurate, rather than custom logic to do the accounting separately for each singleton actor. Separately, the particular actor in question (here StoragePowerActor) can then do collateral arithmetic on top of that running total, to determine whether pledge collateral requirements are satisfied.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -91,13 +93,14 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, amount actor.Token
```
jzimmerman: In general, would be good if we can always have the code error out prominently (e.g., panic("TODO")) whenever something is still incomplete, rather than leaving the variable uninitialized. This also applies to line 61 `var msgValue actor.TokenAmount` and in several other places (I can't seem to add GitHub comments on lines that weren't changed in this diff, though).

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
jzimmerman: +1. These functionalities should now all be doable with the current runtime VM context API (`SendPropagatingErrors`, `ImmediateCaller`)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -172,62 +175,56 @@ func (a *StoragePowerActorCode_I) GetTotalPower(rt Runtime) block.StoragePower {
...
 
-	UpdateRelease(rt, h, st)
-
-	return ret
+	return rt.SuccessReturn()
 }
 
-func (a *StoragePowerActorCode_I) ProcessFaultReport(rt Runtime, report sector.FaultReport) {
+// slash pledge collateral for Declared, Detected and Terminated faults
+func (a *StoragePowerActorCode_I) SlashPledgeForStorageFault(rt Runtime, affectedPower block.StoragePower, faultType sector.StorageFaultType) {
```
jzimmerman: In all of these methods, we should validate the caller (regardless of whether we need to interact with the caller); this is necessary for correctness. (`ValidateImmediateCallerIs`, `ValidateImmediateCallerMatches`, `ValidateImmediateCallerAcceptAny`)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -172,62 +175,56 @@ func (a *StoragePowerActorCode_I) GetTotalPower(rt Runtime) block.StoragePower {
...
 
 	var msgSender addr.Address // TODO replace this
 
-	h, st := a.State(rt)
+	// placeholder values
+	const unitDeclaredFaultSlash = 1    // placeholder
```
jzimmerman: Could we move these constants to the top of the file and given them a brief explanatory comment, with units (e.g., FIL/sector)?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
jzimmerman: I agree this should go to the burnt funds actor (which, I believe, may get renamed to the treasury actor).

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
zixuanzh: Just to be on the same page, the sum of AvailableBalance and LockedBalance is the running total. You are saying we should just check if pledge collateral is satisfied as a function on the total without keeping tracking of Avail vs Locked and we will do this collateral check before withdraw balance as well.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
zixuanzh: will move to what Joe suggested below and only keep one number as the balance if we are in sync.

src/systems/filecoin_mining/sector/sector.go
```diff
@@ -1,20 +1,15 @@
...
 
-func (r *FaultReport_I) GetDetectedFaultSlash() actor.TokenAmount {
-	return actor.TokenAmount(0)
-}
-
-func (r *FaultReport_I) GetTerminatedFaultSlash() actor.TokenAmount {
-	return actor.TokenAmount(0)
-}
+const (
+	DeclaredFault   StorageFaultType = 0
```
jzimmerman: minor nit: prefer to start enums at 1, to avoid accidental uninitialized errors

```
const (
    CaseA EnumType = 1+iota
    CaseB
    CaseC
    // ...
)
```

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
zixuanzh: we will still use that, will add that either in this PR or a future one.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
jzimmerman: +1. Also, with #644 , we now have an extra return value from `SendCatchingErrors` (namely, the error exit code). Calls to `SendCatchingErrors` should never ignore that exit code, except in very special cases such as CronActor.

jzimmerman: (Separately, @anorth currently the plan is for out-of-gas to be a special-cased unconditional failure, which cannot be caught even by `SendCatchingErrors`, except in the case of Cron.)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
jzimmerman: Douglas Adams approved. However, I think this particular panic might be there as a TODO, to remind us to serialize the params into `slashPledgeParams` (@zixuanzh is that correct?). If so, may be helpful to consult `init_actor.go` for relevant examples of current serialization/deserialization APIs.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -99,31 +99,62 @@ func (a *StorageMinerActorCode_I) _slashDealsFromFaultReport(rt Runtime, sectorN
...
 ) {
-	faultReport := &sector.FaultReport_I{
-		NewDeclaredFaults_:   newDeclaredFaults,
-		NewDetectedFaults_:   newDetectedFaults,
-		NewTerminatedFaults_: newTerminatedFaults,
+
+	// only terminatedFault will result in collateral deal slashing
+	if len(newTerminatedFaults) > 0 {
+		a._slashDealsForStorageFault(rt, newTerminatedFaults.SectorsOn(), sector.TerminatedFault)
+		a._slashPledgeForStorageFault(rt, newTerminatedFaults.SectorsOn(), sector.TerminatedFault)
```
jzimmerman: Do these methods need to take `faultType` as an argument? I assume this could just be `_slashDeals(...)`, `_slashPledge(...)` as the behavior is independent of the original cause.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
jzimmerman: +1. Also want to flag that we should carefully analyze what happens upon failure (even out-of-gas failures, which we can't prevent in general), given that these are messages ultimately originated by users. I.e., the consequences of any type of failure should be no worse than the result if the user had just not submitted the message in the first place.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
zixuanzh: that is fine, we will move the logic of computing the amount to slash else where. In `_slashPledge` for example, we expect different amount/proportion depending on the fault type.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
jzimmerman: See also #644 (in particular https://github.com/filecoin-project/specs/pull/644/files#r351364839) where I had to change this call in order to update for the new error handling API -- ideally we should resolve merge conflicts w/that PR as soon as possible.

src/systems/filecoin_vm/runtime/exitcode/vm_exitcodes.go
```diff
@@ -57,7 +57,8 @@ const (
 )
 
 var (
-	InvalidSectorPacking = UserDefinedError(1)
+	InvalidSectorPacking         = UserDefinedError(1)
```
jzimmerman: (also modified in #644)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
zixuanzh: yes, will do.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
zixuanzh: Given the premise that a PoSt is only valid when miner has sufficient balance in StoragePowerActor, LockedBalance is the pledge collateral being locked and it is updated by _ensurePledgeCollateralSatisfied. LockedBalance may get out of sync with what is required at any point in time and hence the invocation of _ensurePledgeCollateralSatisfied here. I think what is missing is to unlock funds when LockedBalance is more than pledge required in _ensurePledgeCollateralSatisfied.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
zixuanzh: ack

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
zixuanzh: @jzimmerman how do we validate that a caller is a StorageMinerAddr since it is a non-singleton actor?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
jzimmerman: Yes -- essentially, using a single tally in StoragePowerActor (or more precisely, a map from actor to tally), rather than using two separate tallies for Available and Locked. I think it's a relatively small change, but conceptually much less error-prone, because if singleton actors in general have a single mechanism for maintaining escrow funds on behalf of individual actors, then we can reuse the same machinery everywhere.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
jzimmerman: `ActorState::CodeID` and the associated methods in `actor.go` should handle it, I think. (There might not be a clean way to get the ActorState of the sender yet, though -- let me know if you hit this issue and I can add some functionality to the runtime API.)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
jzimmerman: Ah, I don't have a strong preference on this; was just wondering, since those flags seemed to be unused in the current PR code. Wherever you think best to implement those logic operations sounds good to me.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
sternhenri: agree. ensure is good safety, it should indeed also unlock as needed. Maybe a more descriptive method name would be nice. 

May also let miner manually unlock if cost is high (this might get called too often).

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
sternhenri: What was the initial purpose of the distinction @zixuanzh ?

sternhenri:
My understanding is we came to an agreement on behavior per your above review comment @anorth on another channel (with @zixuanzh). Documenting as follows:
- Can `ePoSt` with insufficient collateral. Any block reward will go to collateral until it is topped up
- Cannot `sPoSt` with insufficient collateral.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -441,6 +486,8 @@ func (a *StorageMinerActorCode_I) RecoverFaults(rt Runtime, recoveringSet sector
 
 	UpdateRelease(rt, h, st)
 
+	// EnsureDealCollateral
```
anorth: Is this a TODO?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
zixuanzh: @sternhenri this is to accommodate pledge collateral changing dynamically and miners may switch between over and under collateralization quite frequently. Available balance is acting as a buffer or otherwise SPA needs to send and receive collateral very frequently.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -172,62 +182,62 @@ func (a *StoragePowerActorCode_I) GetTotalPower(rt Runtime) block.StoragePower {
...
+		amountToSlash := actor.TokenAmount(DetectedFaultSlashPercent * uint64(affectedPledge) / 100)
+		st._slashPledgeCollateral(rt, minerID, amountToSlash)
+	case sector.TerminatedFault:
+		amountToSlash := actor.TokenAmount(TerminatedFaultSlashPercent * uint64(affectedPledge) / 100)
+		st._slashPledgeCollateral(rt, minerID, amountToSlash)
+	}
 
-	st._slashPledgeCollateral(rt, msgSender, (declaredFaultSlash + detectedFaultSlash + terminatedFaultSlash))
+	// this is done this way because we can't send messages when holding onto the state lock
+	amountToSlash := st.PledgeCollateralSlashed()
```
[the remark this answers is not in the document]
zixuanzh: agree that it is ugly, it was done this way to keep pledge collateral accounting all in one method (_slashPledgeCollateral)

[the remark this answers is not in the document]
zixuanzh: fixing per your suggestion

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
```diff
@@ -9,51 +9,37 @@ import (
...
-	}
+	currEntry := st._safeGetPowerEntry(rt, minerID)
 
 	amountToSlash := amount
 
 	if currEntry.Impl().LockedPledgeCollateral() < amount {
 		amountToSlash = currEntry.Impl().LockedPledgeCollateral_
 		currEntry.Impl().LockedPledgeCollateral_ = 0
-		// TODO: extra handling of not having enough pledgecollateral to be slashed
+		// TODO: extra handling of not having enough pledge collateral to be slashed
```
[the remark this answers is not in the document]
zixuanzh: an inline TODO here is fine so that we are explicit that the spec is not complete. This is more of a TBD, we may not need to do anything extra here.

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -198,19 +205,19 @@ func (a *StorageMarketActorCode_I) ActivateDeals(rt Runtime, dealIDs []deal.Deal
 
 }
 
-func (a *StorageMarketActorCode_I) ProcessDealSlash(rt Runtime, dealIDs []deal.DealID, slashAction deal.StorageDealSlashAction) {
+func (a *StorageMarketActorCode_I) ProcessDealSlash(rt Runtime, dealIDs []deal.DealID, faultType sector.StorageFaultType) {
```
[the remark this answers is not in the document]
zixuanzh: Yes and all we need to do is to verify that it is called by one of the MinerActor addresses. Which miner to slash is contained in deal information.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -86,26 +95,30 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, amount actor.Token
 		rt.Abort("negative amount.")
 	}
 
-	// TODO: convert msgSender to MinerActorID
-	var minerID addr.Address
+	minerID := rt.ImmediateCaller()
```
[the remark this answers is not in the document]
zixuanzh: Yes, they were supposed to be restrictive when we first spec'ed it. I will prefer leaving 1) Add/Withdraw Balance and 2) minerID and addresses to a separate PR.

zixuanzh:
@anorth I have fixed review comments related to pledge collateral requirement and slashing in this PR. Given how fast everything else is moving, I will prefer merging in this PR first since it contains the structure and logic for pledge collateral and we are all aligned on that. Other work this week also depends on the main logic in the PR. We will fix some other review comments in separate and smaller PRs:
- [ ] align on the usage of minerID, worker account, and miner actor addr
- [ ] remove DealCollateralSlashed from Storage Market
- [ ] align on Add/Withdraw Balance API (we wanted them to be more restrictive when we first spec'd them)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -149,19 +151,19 @@ func (a *StorageMinerActorCode_I) _onMissedSurprisePoSt(rt Runtime) {
 
 	Release(rt, h, st)
 
+	a._submitPowerReport(rt, lastPoStResponse)
+
 	// Note: NewDetectedFaults is now the sum of all
 	// previously active, committed, and recovering sectors minus expired ones
 	// and any previously Failing sectors that did not exceed MaxFaultCount
 	// Note: previously declared faults is now treated as part of detected faults
```
anorth: Aside: I would really appreciate a write-up of the fault and slashing rules that could surface things like this more clearly. I just noticed this one and think we should reconsider.

sternhenri: +1, reworking md would be dope

PR #658, opened 2019-11-27
Keep time using an int64

### LINKED ISSUE #653: Clarification around ChainEpochClock Time Type
While implementing the ChainEpochClock in go-filecoin I observed that the type for GenesisTime is a string following ISOFormat "2006-01-02T15:04:05.999999999Z". I want to clarify this is intended as it means [BlockHeaders](https://github.com/filecoin-project/specs/blob/b7f879cc921f8bafab0aad1c6aab8643abacef0d/src/systems/filecoin_blockchain/struct/block/block.id#L30) use a string to represent time instead of what was previously a uint64. Based on the change list in the ChainEpochClock PR this may have been overlooked #632.

After talking with the go-filecoin team we believe this may be unnecessary unless the intent is to track blocktime with nanosecond precision -- gut reaction is this may have some unwanted effects on the size of the chain.

frrist:
### What changed?
- fixes #653: we should not use strings to represent time in block headers.
- simplify UTCClock: nanosecond precision is unnecessary.
- correct parts of spec that referenced time as string (was called `clock.Time`), use `clock.UnixTime` (an int64 under-the-hood) everywhere. 

### Questions:
1. Is returning ChainEpoch as 0 for all times before genesis acceptable? https://github.com/filecoin-project/specs/compare/frrist/fix-chain-epoch-clock?expand=1#diff-7e23591a34c702689c658c6fe73d893dR24-R26
2. BlockHeader Timestamp is an int64, meaning its max value is `9223372036854775807` or `12/04/292277026596 @ ~3:30pm (UTC)`, is this future-poof enough? https://github.com/filecoin-project/specs/compare/frrist/fix-chain-epoch-clock?expand=1#diff-f39f7b025d293796b93d9d8172d323c1R30

### Related Bits:
Here is a reference implementation in go-filecoin that led to this change: https://github.com/filecoin-project/go-filecoin/pull/3669

### DIFF
--- a/src/systems/filecoin_nodes/clock/clock_subsystem.go
+++ b/src/systems/filecoin_nodes/clock/clock_subsystem.go
@@ -9,7 +9,7 @@ var UTCMaxDrift = time.Second
 // source, such as NTP, or a very precise hardware clock.
 var UTCSyncPeriod = time.Hour
 
-// EpochDuration is a constant that represents the time in seconds
+// EpochDuration is a constant that represents the duration in seconds
 // of a blockchain epoch.
 var EpochDuration = UnixTime(15)
 
@@ -17,14 +17,10 @@ func (_ *UTCClock_I) NowUTCUnix() UnixTime {
 	return UnixTime(time.Now().Unix())
 }
 
-// EpochAtTime returns the ChainEpoch corresponding to t.
+// EpochAtTime returns the ChainEpoch corresponding to time `t`.
 // It first subtracts GenesisTime, then divides by EpochDuration
-// and returns the resulting number of epochs. If t is before
-// GenesisTime zero is returned.
+// and returns the resulting number of epochs.
 func (c *ChainEpochClock_I) EpochAtTime(t UnixTime) ChainEpoch {
-	if t <= c.GenesisTime() {
-		return ChainEpoch(0)
-	}
 	difference := t - c.GenesisTime()
 	epochs := difference / EpochDuration
 	return ChainEpoch(epochs)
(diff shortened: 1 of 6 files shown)


frrist:
I _think_ CI is failing because `src/systems/filecoin_blockchain/struct/block/tipset.id` and `src/systems/filecoin_nodes/clock/clock_subsystem.id` are not go formatted properly. I am unsure how to run gofmt over them as they are not go files, any suggestions?

src/systems/filecoin_nodes/clock/clock_subsystem.go
```diff
@@ -9,43 +9,23 @@ var UTCMaxDrift = time.Second
 // source, such as NTP, or a very precise hardware clock.
 var UTCSyncPeriod = time.Hour
 
-// EpochDuration is a constant that represents the UTC time duration
+// EpochDuration is a constant that represents the time in seconds
```
anorth:
```suggestion
// EpochDuration is a constant that represents the duration in seconds
```

src/systems/filecoin_nodes/clock/clock_subsystem.go
```diff
@@ -9,43 +9,23 @@ var UTCMaxDrift = time.Second
...
-	}
-
-	t2, err := time.Parse(ISOFormat, string(t))
-	if err != nil {
-		panic(err)
+// and returns the resulting number of epochs. If t is before
+// GenesisTime zero is returned.
+func (c *ChainEpochClock_I) EpochAtTime(t UnixTime) ChainEpoch {
+	if t <= c.GenesisTime() {
+		return ChainEpoch(0)
```
anorth: No, I think this is better to return negative numbers, so that the first (zero) epoch is distinguishable from negatives. Calling code can deal with negatives if possible in that context.

PR #660, opened 2019-11-27
Using docker orient

nikkolasg:
I messed up the rebase thing in https://github.com/filecoin-project/specs/pull/657 so here is a fresh PR freshly forked of master


nikkolasg:
Note 723ae95 :  we dont need to have an output file depending on the docker since we save it locally from the output of the docker command. Normally it would work the way it was but depending on permissions it might fail, so best to rely on something that always work.

PR #661, opened 2019-11-27
Preprocessing pipeline 

nikkolasg:
This PR brings the ability to pre-process the JSON input files (i.e. `src/orient/filecoin.json`) with pre-processing scripts. Each script takes a JSON file as input and outputs a JSON output. 
This PR brings a first script to compute the expansion that Chung's bipartite expander brings us: it dynamically adds a field to the input JSON computed from the same input JSON.

This PR forces now every input json file in a standard way (with jq)
This PR introduces as well a JSON fmt option. I would like to not add other stuff.

This PR is building on top of #660 .


bin/build-spec-orient.sh
```diff
@@ -11,8 +14,8 @@ else
     diagram_dir=/orientd/$diagram_dir
 fi;
 
-mkdir -p $output_dir
-mkdir -p $diagram_dir
+mkdir -p $output_dir > /dev/null 2>&1
```
nikkolasg: this is to prevent an annoying error message each time you run the script.

bin/preprocessing/preprocessing.sh
```diff
@@ -0,0 +1,28 @@
...
+            continue
+        fi
+        fullpath=$scripts_dir/$script
+        # because the following issue, we need to make a temp file
+        # https://stackoverflow.com/questions/6696842/how-can-i-use-a-file-in-a-command-and-redirect-output-to-the-same-file-without-t
+        ./$fullpath $jinput > $tmp
+        exit_code=$?
+        test $exit_code -eq 0 || (echo "Error with script $(basename $script).  Exit" && exit 1)
+        # Format via jq by default
+        cat $tmp | jq > $jinput
```
nikkolasg: Note that I chose to use jq as a default formatter, but it wil already change some JSON file which are not currently formatted correctly. 
In my opinion, it's time we switch to a standard formatting, and with this pipeline we have no choice anymore: either the scripts do it themselves individually (with the formatting they want) or we use JQ on all outputs. I preferred to go the standard uniform way.

bin/solve-orient.sh
```diff
@@ -13,8 +16,8 @@ else
     output_file=/orientd/$output_file
 fi;
 
-mkdir -p $output_file && rmdir $output_file # Dirty hack to ensure output_file directory exists.
+mkdir -p $output_file > /dev/null 2>&1 && rmdir $output_file > /dev/null 2>&1 # Dirty hack to ensure output_file directory exists.
 
-echo $output_file
+#echo $output_file
```
nikkolasg: Again just nicer & more readable output - just aethestics

bin/preprocessing/preprocessing.sh
nikkolasg: Most annoying effect so far I've seen of that formatting is to change big float numbers into powers of tens `0.000000000154` to `1.25 e-9`

bin/build-spec-orient.sh
porcuquine: Thank you!

bin/preprocessing/preprocessing.sh
porcuquine: Are you saying that `jq` rewrites floats in this way?

nikkolasg: Exactly. by default jq rewrites floats in this form if the decimal form is too long, as well it uses 4indent spaces rather than the ones used in the current json file.

nikkolasg: There may be other differences, but i believe JQ is following the standards for JSON so we should go along with him.

porcuquine:
@nikkolasg Sorry for loss of threading, GH is defeating me: I think you are saying you favor standardizing on jq output for all json files, including the raw json we check into the project. Is that right? If so, a make target equivalent to `fmt-code` would be valuable. If we add, we should also enhance the CI check. If this is what you intend, it makes sense to me. Can you file an issue and/or implement?

nikkolasg:
Closing because @nicola use it already in branch rational-calc-3

PR #662, opened 2019-11-27
Specify the order of tipset message execution, including implicit messages.

anorth:
Prose and code for the execution of a tipset's messages, including message construction
pseudocode for the implicit processing of election post, block reward payment, and
cron tick.

~All messages, including the implicit ones, produce receipts, so we have exactly one receipt per state mutation (this differs from what I'd previously proposed in person).~

Coming later:
- fixes to the rest of vm_interpreter.go
- block/message validation do-over

cc @jzimmerman @acruikshank @icorderi @magik6k

### DIFF
--- a/src/systems/filecoin_vm/interpreter/vm_interpreter.go
+++ b/src/systems/filecoin_vm/interpreter/vm_interpreter.go
@@ -33,12 +33,16 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
 		// Pay block reward.
 		reward := _makeBlockRewardMessage(outTree, blk.Miner())
 		outTree, r = vmi.ApplyMessage(outTree, reward, blk.Miner())
-		receipts = append(receipts, r)
+		if r.ExitCode() != exitcode.OK() {
+			panic("block reward failed")
+		}
 
 		// Process block miner's Election PoSt.
 		epost := _makeElectionPoStMessage(outTree, blk.Miner(), msgs.Epoch(), blk.PoStProof())
 		outTree, r = vmi.ApplyMessage(outTree, epost, blk.Miner())
-		receipts = append(receipts, r)
+		if r.ExitCode() != exitcode.OK() {
+			panic("election post failed")
+		}
 
 		// Process messages from the block.
 		for _, m := range blk.Messages() {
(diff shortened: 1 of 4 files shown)


src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -128,12 +156,90 @@ func (vmi *VMInterpreter_I) ApplyMessage(inTree st.StateTree, message msg.Unsign
...
 	return actor.TokenAmount(util.UVarint(gas) * util.UVarint(price))
 }
+
+// Builds a message for paying block reward from the treasury account to a miner owner.
+func _makeBlockRewardMessage(state st.StateTree, minerActorAddr addr.Address) msg.UnsignedMessage {
+	// networkTreasuryActor := loadActor(NetworkTreasuryActorAddress)
+	// minerActor := loadActorSubstate(minerActorAddr)
+	// minerWorker := loadActor(minerActor.Info.Worker)
+	//return &UnsignedMessage_I{
+	//	From_:       minerActor.Info.Worker,
```
jzimmerman: In general, I'm not sure we want to have the treasury send money in response to messages sent to it by other actors. Could this instead be the execution of a message whose sender is the treasury, and whose accompanying token value is the block reward?

zixuanzh: Treasury actor still needs to know in some way who mines the block tho, how will this be done?

anorth: Top-level message sender must be an account actor, which are the only actors with meaningful CallSeqNum etc, unless we want to break that invariant (I'm not sure what depends on it). Yes, when we write it it will need some protection against being invoked manually. 

But in practice, miners are not going to reach consensus with another that's stealing extra block reward. 

We could invent a new account actor that's solely for the purpose of being the sender of these messages (but it can't have code, which we're now adding to the treasury actor).

anorth: In either case, can I do it in follow-up?

anorth:
> why we apply the block reward first before ePoSt verification

I could go either way. Block reward first provides an opportunity to garnish the reward, possibly preventing the under-collateral state during PoSt. But if we don't prevent block production in that state, it could happen in the other order too.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -1,42 +1,70 @@
...
-}
 
-func _withTransferFundsAssert(tree st.StateTree, from addr.Address, to addr.Address, amount actor.TokenAmount) st.StateTree {
-	// TODO: assert amount nonnegative
-	retTree, err := tree.Impl().WithFundsTransfer(from, to, amount)
-	if err != nil {
-		panic("Interpreter error: insufficient funds (or transfer error) despite checks")
-	} else {
-		return retTree
+	// Invoke cron tick, attributing it to the miner of the first block.
```
sternhenri: a bit late to the party, but just to note that applying `makeCronTickMessage` last means for any epoch-based timeouts, they occur at next epoch number. Ie say my sPost challenge comes in on epoch 400, I'll actually still be able to ePost at 400 and will only get the challenge on 401 if I don't post on 401 given the tick occurs last in the epoch.

It's functionally the same, just something to keep top of mind...

PR #663, opened 2019-11-28
Proofs impl

porcuquine:
This brings proofs implementation into alignment with current (as of now…) construction and implementation. There are still some TODOs in components, and it would be very strange if there weren't some bugs. However, this should be more than plenty for purposes of interfacing with the rest of the spec.

It should also be a good base for a) converging construction (potentially in flux), spec, and implementation in the next period; then b) moving toward executable tests which will confirm having done so.


src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -266,62 +268,59 @@ func (sdr *WinStackedDRG_I) Seal(sid sector.SectorID, data []byte, randomness se
...
-	// FIXME: need to compute commQ.
 	return commC, commQ, commRLast, commR, commCTreePath, commQTreePath, commRLastTreePath
 }
 
-func computeSealSeed(sid sector.SectorID, windowIndex int, randomness sector.SealRandomness, commD sector.UnsealedSectorCID) sector.SealSeed {
-	var proverId []byte // TODO: Derive this from sid.MinerID()
+func getProverID(minerID sector.MinerID) []byte {
+	panic("TODO")
+}
+func computeSealSeed(sid sector.SectorID, randomness sector.SealRandomness, commD sector.UnsealedSectorCID) sector.SealSeed {
```
sternhenri: We should remember to sync on this on spec completion and ensure all randomness generation is accounteed for together in the protocol (e.g. DSTs and the like). may be worth adding a TODO so we can find it later. Fair chance it doesn;t apply for randomness here but wanted to raise it in case.

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -852,66 +862,276 @@ func joinPieceInfos(left PieceInfo, right PieceInfo) PieceInfo {
...
+		sectors = append(sectors, sector)
+	}
+
+	return sectors
 }
 
-////////////////////////////////////////////////////////////////////////////////
-// Election PoSt
+func generateSectorChallenge(randomness sector.PoStRandomness, n int, sectorIDs []sector.SectorID) (sector sector.SectorID) {
+	preimage := append(randomness, littleEndianBytesFromInt(n, 8)...)
```
sternhenri: ibid

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -852,66 +862,276 @@ func joinPieceInfos(left PieceInfo, right PieceInfo) PieceInfo {
...
-	challengedSectors, challenges := sdr._getChallengedSectors(challengeSeed, eligibleSectors, candidateCount)
-	var proofAuxs []sector.PersistentProofAux
+	challengeModulus := new(big.Int)
+	challengeModulus.SetUint64(uint64(len(sectorIDs)))
+
+	sectorIndex := sectorChallenge.Mod(sectorChallenge, challengeModulus)
+	return sectorIDs[int(sectorIndex.Uint64())]
+}
+
+func generateLeafChallenge(randomness sector.PoStRandomness, sectorChallengeIndex UInt, leafChallengeIndex int, nodes int, challengeRangeSize int) UInt {
```
sternhenri: ibid

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -35,6 +35,8 @@ func WinSDRParams(cfg SDRCfg) *WinStackedDRG_I {
 	const NODE_SIZE = 32
```
sternhenri: nit would be awesome to have a proofs const file

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -852,66 +862,276 @@ func joinPieceInfos(left PieceInfo, right PieceInfo) PieceInfo {
...
+	return partialTicket
 }
 
-func (sdr *WinStackedDRG_I) VerifyElectionPoSt(sv sector.PoStVerifyInfo) bool {
+func (sdr *WinStackedDRG_I) _generatePoStCandidates(challengeSeed sector.PoStRandomness, eligibleSectors []sector.SectorID, candidateCount int, sectorStore sector_index.SectorStore) (candidates []sector.PoStCandidate) {
+
+	nodes := int(sdr.Nodes())
+	leafChallengeCount := int(sdr.LeafChallengeCount())
+	challengeRangeSize := int(sdr.ChallengeRangeSize())
+	challengedSectors := sdr._getChallengedSectors(eligibleSectors, challengeSeed, eligibleSectors, candidateCount)
```
sternhenri: passing in too many args. Does this code compile?

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -852,66 +862,276 @@ func joinPieceInfos(left PieceInfo, right PieceInfo) PieceInfo {
...
+
+	return candidates
+}
+
+func (sdr *WinStackedDRG_I) CreatePoStProof(privateCandidateProofs []sector.PrivatePoStCandidateProof, challengeSeed sector.PoStRandomness) sector.PoStProof {
+	privateProof := sdr.CreatePrivatePoStProof(privateCandidateProofs, challengeSeed)
+
+	// var candidates []sector.PoStCandidate // FIXME
+
+	// Hmmmmm: we cannot perform this santiy check without the sectorIDs. Should we require them just for that purpose or perform the check earlier?
```
sternhenri: @dignifiedquire

sternhenri:
For future PR: there are also a number of methods impl'd here (Sha256Hash, uintToLittleEndian, etc) that are useful elsewhere in the spec. Would be great to split them out into a separate general crypto lib for spec-wide use (filproofs alas creates import cycles (e.g. with sector or block).

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -266,62 +268,59 @@ func (sdr *WinStackedDRG_I) Seal(sid sector.SectorID, data []byte, randomness se
...
+func getProverID(minerID sector.MinerID) []byte {
+	panic("TODO")
+}
+func computeSealSeed(sid sector.SectorID, randomness sector.SealRandomness, commD sector.UnsealedSectorCID) sector.SealSeed {
+	proverId := getProverID(sid.MinerID())
 	sectorNumber := sid.Number()
 
 	var preimage []byte
 	preimage = append(preimage, proverId...)
 	preimage = append(preimage, littleEndianBytesFromUInt(UInt(sectorNumber), 8)...)
```
dignifiedquire: Diff: the current implementation uses `BigEndian`

src/libraries/filcrypto/filproofs/algorithms.go
porcuquine: Randomness isn't being created here. Can you be more specific? Or even better, once we know exactly what we're trying to account for later, we can make a pass to identify relevant remaining points in here. The proofs code only ever consumes a few specific random values then uses them as our constructions dictate.

src/libraries/filcrypto/filproofs/algorithms.go
porcuquine: Can move things to a separate file as part of some ubercalc reconciliation pass in the future. FWIW, the reason constants are being defined locally here is that there could be more than one top-level type using them (in fact it's changed several time times since the code was first written), and prematurely using variable names for namespacing seemed time-consuming and error-prone. We can lock things down when they are fully stabilized.

src/libraries/filcrypto/filproofs/algorithms.go
porcuquine: It does compile, so I'm not sure why you say 'too many args'.

src/libraries/filcrypto/filproofs/algorithms.go
porcuquine: Thanks, dig.

porcuquine: Can you confirm that this applies everywhere we are constructing hash preimages? (We discussed, so I think so.)

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -681,30 +700,34 @@ func (proof *SDRColumnProof) Verify(root []byte, challenge UInt) bool {
...
+
+	for i := 0; i < challengeCount; i++ {
+		var preimage []byte
+		preimage = append(preimage, sealSeed...)
+		preimage = append(preimage, randomness...)
+		preimage = append(preimage, littleEndianBytesFromInt(i, 4)...)
 
-		hash := HashBytes_SHA256Hash(bytes)
+		hash := HashBytes_SHA256Hash(preimage)
 		bigChallenge := bigIntFromLittleEndianBytes(hash)
```
dignifiedquire: this one can stay

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -681,30 +700,34 @@ func (proof *SDRColumnProof) Verify(root []byte, challenge UInt) bool {
...
-		}
-		bytes = append(bytes, randomness...)
-		bytes = append(bytes, littleEndianBytesFromInt(i, nodeSize)...)
+	// Maybe factor this into a separate function, since the logic is the same...
+
+	for i := 0; i < challengeCount; i++ {
+		var preimage []byte
+		preimage = append(preimage, sealSeed...)
+		preimage = append(preimage, randomness...)
+		preimage = append(preimage, littleEndianBytesFromInt(i, 4)...)
```
dignifiedquire: this one can stay

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -681,30 +700,34 @@ func (proof *SDRColumnProof) Verify(root []byte, challenge UInt) bool {
...
 
-		hash := HashBytes_SHA256Hash(bytes)
+		hash := HashBytes_SHA256Hash(preimage)
```
dignifiedquire: DIFF: the current implementation does not truncate to Fr size

src/libraries/filcrypto/filproofs/algorithms.go
sternhenri: I would argue the SealSeed is randomness, as in the whole preimage computation and hashing is creating the appropriate randomness value for you. @jzimmerman would push hard to use a CBOR-serialization of an object made up of the sector number window index, commD and proverID per #677 in trying to ensure we are consistent in generating randomness in a same way everywhere in the protocol. Is there a strong reason not to follow that pattern here? (I'm pretty on the fence about the net benefit of enforcing it throughout the protocol fwiw, but if we don't i'd want to weaken it elsewhere as well)...

src/libraries/filcrypto/filproofs/algorithms.go
sternhenri: nvm

src/libraries/filcrypto/filproofs/algorithms.go
porcuquine: I am not sure I agree. I do not think we expect `SealSeed` to be random oracle output. As far as I know, it just needs to be unique. In the past, for example, we used pedersen hashing — which is not pseudorandom, just collision-resistant. This value is itself included in hashes to make their output unique. This operation, which we historically call 'KDF' does need to generate pseudorandom output, and the replicaID is effectively a nonce input.

I think it's important to distinguish between the 'random' values which are inputs to the proofs, and those which proofs per se generates as part of its internal workings. The former can be completely externalized, and the latter cannot be at all. If 'the protocol' (which I distinguish from proofs here, let's not quibble about whether that is meaningful) wants to take ownership of `SealSeed`, it would also need `CommD` for that purpose. I do *not* think `filproofs` should be calling into any general protocol code, nor should it have a dependency (conceptual or software-based on CBOR) — so this means that 'the protocol' would also need to take responsibility for calculating `CommD` — which in  turn will require it to handle preprocessing, etc. Given that `filproofs` depends on `CommD` having been constructed correctly, I think it should continue to own its construction. So this would require an extra library call such that 'whoever' is responsible for packing the sector would need to acquire `CommD`, then use it to generate `SealSeed` and pass it to proofs. This seems like a great deal of extra work, complexity, and places for things to go wrong — especially given that it's not necessary.

Moreover, if we're not prepared to accept that `filproofs` generates some kinds of 'randomness-like things' of its own accord, then we won't be able to perform the proofs at all. Once we accept that there is a qualitative distinction between the proofs inputs and values generated incidentally in the process of proving, then the problem (which I claim never existed) goes away. Much of proofs is necessarily a black box to the rest of the protocol.

sternhenri: I think you're taking this way farther than I intended, and painting this as a very stark choice when it may be much simpler. 

All I'm saying is we should consider having proofs draw/generate randomness under the same structure that the reset of the Filecoin protocol does. Modulo how much work it is (it shouldn't be much imo save for the CBOR), it will help future implementers who may choose to architect the protocol differently. 

I don't think it's a big deal either way, but I hardly think it's tantamount to breaking the abstraction barriers chosen in order to share code modules or something. I'm just saying we could sample randomness the same way for consistency across all parts of FIL

sternhenri: cc @jzimmerman . That said nbd if not someething proofs can do. It could well be that it does not rise to the level of urgency/necessity required for time to be spent on it under the current regimen. I just don't think the options are quite as stark as you present.

porcuquine: As discussed offline, `SealSeed` (was `ReplicaID`) is not 'sampled randomness'. It is an identifier which need only be globally unique in combination with `SealCfg`.

src/libraries/filcrypto/filproofs/algorithms.go
porcuquine: I changed this, resolving the diff in favor of implementation. I'm now only truncating when generating labels. If the implementation does it elsewhere, please let me know, @dignifiedquire.

PR #666, opened 2019-11-28
Add CurrReceiver() to runtime, remove unneeded CurrMethodNum

anorth:
Add `CurrReceiver` to the VM runtime, answering a need from https://github.com/filecoin-project/specs/pull/654#discussion_r351212995.

Remove `CurrMethodNum` from the runtime, which was used only for a runtime check that the caller was the constructor, the need for which was easily removed with a small restructure.


src/systems/filecoin_vm/sysactors/init_actor.go
```diff
@@ -21,14 +21,11 @@ type Runtime = vmr.Runtime
...
-func (a *InitActorCode_I) State(rt Runtime) (vmr.ActorStateHandle, InitActorState) {
+func _loadState(rt Runtime) (vmr.ActorStateHandle, InitActorState) {
 	h := rt.AcquireState()
 	stateCID := ipld.CID(h.Take())
 	if ipld.CID_Equals(stateCID, ipld.EmptyCID()) {
-		if rt.CurrMethodNum() != actor.MethodConstructor {
-			rt.Abort("Actor state not initialized")
-		}
-		return h, nil
+		rt.Abort("Actor state not initialized")
```
jzimmerman: Why is CurrMethodNum (and this logic) no longer needed? This was originally added because, prior to their Constructor method being called, actors have an empty state CID by default (which does not deserialize properly into a value of the desired substate struct type).

anorth: This method is no longer called by the constructor, which ignored the return value anyway. Hence the predicate that was on line 28 would always be satisfied.

PR #667, opened 2019-11-28
Add NewActorAddress() to runtime in place of three different less abstract accessors.

anorth:
`ToplevelSender()`, `InternalCallSeqNum` and `ToplevelSenderCallSeqNum` existed on the runtime only to be used to compute new actor addresses. The runtime should present as abstract an interface as possible, so moving the new address computation inside reduces the surface area.

It may be necessary to add `ToplevelSender` back at some point, but until a concrete case emerges we should keep the interface narrow.

cc @icorderi


anorth:
I think we should also take this further and move address creation inside `CreateActor`.

PR #670, opened 2019-11-28
Proofs impl hs

sternhenri:
fixed randomness issues created by me.

### DIFF
--- a/src/algorithms/post/election_post.md
+++ b/src/algorithms/post/election_post.md
@@ -140,7 +140,7 @@ Miners earn no reward from submitting PoStSurprise messages. This mechanism does
 Surprise PoSt works as follows:
 ```text
 // A number of challenged miners is chosen at every round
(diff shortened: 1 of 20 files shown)


sternhenri:
Fishy change to me may be the change in `orient` which i may have reverted (I've not touched orient)

porcuquine:
The thing to be aware of is that `orient` is a submodule, so if it gets updated somewhere and you don't `git submodule update` to get the latest, then you're likely to accidentally recommit the old version. I don't know what happened here and don't have time to dig in now, but that's just a general PSA. If you see a change to `orient` in your own commits, and you didn't make it intentionally (generally because you've intentionally pulled in the latest `orient/master`), you've likely done this.

sternhenri:
super useful! I've been carrying around an old orient that I've avoided committing forever. Thanks

sternhenri:
Assuming this is now stale @porcuquine

PR #672, opened 2019-11-28
Remove VerifySignature method in AccountActor; replace state with public key

jzimmerman:
- VerifySignature is now handled by the runtime/VM context

### DIFF
--- a/src/systems/filecoin_vm/sysactors/account_actor.id
+++ b/src/systems/filecoin_vm/sysactors/account_actor.id
@@ -1,12 +1,7 @@
-import addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
 import filcrypto "github.com/filecoin-project/specs/algorithms/crypto"
-import vmr "github.com/filecoin-project/specs/systems/filecoin_vm/runtime"
 
-type AccountActorCode struct {
-    VerifySignature(rt vmr.Runtime, sig filcrypto.Signature) InvocOutput
-}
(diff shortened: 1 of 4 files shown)


anorth:
@jzimmerman you can go further and delete all the method invocation boilerplate from this actor, making it very clear that account actors have no code.

src/systems/filecoin_vm/sysactors/account_actor.id
```diff
@@ -1,12 +1,7 @@
...
 
-type AccountActorCode struct {
-    VerifySignature(rt vmr.Runtime, sig filcrypto.Signature) InvocOutput
-}
+type AccountActorCode struct {}
 
 type AccountActorState struct {
-    // normal keypair backed accounts
-    Address addr.Address
+    PublicKey filcrypto.PublicKey
```
whyrusleeping: we just put an address here. In cases where you can do public key recovery, this is just a hash of the public key, in cases where you can't, it contains the whole public key. Fewer types and overhead is better. (plus the interfaces to verifying signatures all take in an address)

jzimmerman: Will go ahead and make this change, but just want to flag that I'm not sure this is correct conceptually -- public keys need not be derivable from addresses in general (and aren't, e.g., in the case of ID-based addresses).

mishmosh:
This appears to be merged in #681. @jzimmerman can you confirm, and close this PR if so?

PR #675, opened 2019-11-29
VM/runtime error-handling improvements

jzimmerman:
- No failure return values (abort should be the only failure mechanism)

### DIFF
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
@@ -38,7 +38,7 @@ func (a *StorageMinerActorCode_I) State(rt Runtime) (vmr.ActorStateHandle, State
 	stateCID := h.Take()
 	stateBytes := rt.IpldGet(ipld.CID(stateCID))
 	if stateBytes.Which() != vmr.Runtime_IpldGet_FunRet_Case_Bytes {
-		rt.Abort("IPLD lookup error")
+		rt.AbortAPI("IPLD lookup error")
 	}
 	state := DeserializeState(stateBytes.As_Bytes())
 	return h, state
@@ -116,7 +116,7 @@ func (a *StorageMinerActorCode_I) _submitFaultReport(
 		NewTerminatedFaults_: newTerminatedFaults,
 	}
 
-	rt.Abort("TODO") // TODO: Send(SPA, ProcessFaultReport(faultReport))
+	panic("TODO") // TODO: Send(SPA, ProcessFaultReport(faultReport))
 	panic(faultReport)
 
 	// only terminatedFault will be slashed
@@ -145,7 +145,7 @@ func (a *StorageMinerActorCode_I) _submitPowerReport(rt Runtime, lastPoSt block.
 
 	Release(rt, h, st)
 
-	rt.Abort("TODO") // TODO: Send(SPA, ProcessPowerReport(powerReport))
+	panic("TODO") // TODO: Send(SPA, ProcessPowerReport(powerReport))
 	panic(powerReport)
 
 	if len(newExpiredDealIDs) > 0 {
@@ -229,7 +229,7 @@ func (a *StorageMinerActorCode_I) _verifyPoStSubmission(rt Runtime, postSubmissi
 	// 2.3 Verify the PoSt Proof
 	// verifyPoSt(challenges, TODO)
 
-	rt.Abort("TODO") // TODO: finish
+	panic("TODO") // TODO: finish
 	return false
 }
 
@@ -280,8 +280,7 @@ func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime, postSubmission p
 	for _, sectorNo := range st.Impl().ProvingSet_.SectorsOn() {
 		sectorState, found := st.Sectors()[sectorNo]
 		if !found {
-			// TODO: determine proper error here and error-handling machinery
-			rt.Abort("Sector state not found in map")
+			rt.AbortStateMsg("Sector state not found in map")
 		}
 		switch sectorState.State().StateNumber {
 		case SectorCommittedSN, SectorRecoveringSN:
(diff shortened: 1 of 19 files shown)


src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -280,8 +280,7 @@ func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime, postSubmission p
 	for _, sectorNo := range st.Impl().ProvingSet_.SectorsOn() {
 		sectorState, found := st.Sectors()[sectorNo]
 		if !found {
-			// TODO: determine proper error here and error-handling machinery
-			rt.Abort("Sector state not found in map")
+			rt.AbortStateMsg("Sector state not found in map")
```
sternhenri: Could you not argue this is an ArgMsg?

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -605,7 +593,15 @@ func (a *StorageMinerActorCode_I) ProveCommitSector(rt Runtime, info sector.Sect
 		// expired
 		delete(st.PreCommittedSectors(), preCommitSector.Info().SectorNumber())
 		UpdateRelease(rt, h, st)
-		return rt.ErrorReturn(exitcode.UserDefinedError(0)) // TODO: user dfined error code?
+
+		TODO()
+		// TODO: if we abort, the above state changes (delete...) will not be committed.
+		// Is this desired? (alternative: support method failure while still committing
+		// state changes if UpdateRelease has already been called)
```
sternhenri: no I believe they should be committed (cc @zixuanzh). Does `UpdateRelease` get aborted?

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -619,16 +615,17 @@ func (a *StorageMinerActorCode_I) ProveCommitSector(rt Runtime, info sector.Sect
...
+			"Seal verification failed")
 	}
 
 	// TODO: check EnsurePledgeCollateralSatisfied
 	// pledgeCollateralSatisfied
 
 	_, utilizationFound := st.SectorUtilization()[onChainInfo.SectorNumber()]
 	if utilizationFound {
-		rt.Abort("sm.ProveCommitSector: sector number found in SectorUtilization")
+		rt.AbortStateMsg("sm.ProveCommitSector: sector number found in SectorUtilization")
```
sternhenri: argMsg?

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
anorth: This is that one case I identified earlier where an actor is attempting to do some clean up even while the method fails. We should rework the code so that this doesn't depend on clean-up happening here. (It's not clear to me that anything does depend on this - it might be ok just to remove the code).

anorth:
My version of @sternhenri's comment: I was expecting InvocOutput to disappear so that actor methods just return Bytes (or nothing).

With regards to abort error codes (which is orthogonal to the return value and could be postponed to another PR), I think we should change `Abort()` to take an error code parameter (which goes on chain) and a message (which doesn't, just for debugging). All codes passed to Abort should be per-actor code, i.e. distinct from the system codes, and starting at value 16. The runtime should clamp any code up to at least 16. These error code definitions should move into the actor files and need not be distinct from each other. Ideally we should have a different code for every call to Abort, but there might be a few where it makes sense to share the code.

The system error codes should only be referenced from inside the VM/interpreter itself, not from within actor code. The dispatch table that's in actor code itself makes that difficult right now, but this is deep in implementation details that IMO the spec really shouldn't be troubling itself with.

jzimmerman:
> * Why split the exit codes out of `invocOutput`?

This is in order to prevent actor code from manually returning failure instead of signalling failure via aborting. With the new structure, the runtime is responsible for determining whether the exit code indicates failure, rather than the actor itself.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
jzimmerman: I agree the distinction is somewhat arbitrary. In general, my heuristic has been to use ArgMsg only when the relevant value is an argument to the public method invocation, and use StateMsg for other internal precondition violations (even if they are "arguments" to an internal method). But I think this is easy enough to change in a followup PR. (cc @zixuanzh who may have more clarity on the intended semantics of the actor code here)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
jzimmerman: Alternatively, we could introduce a variant of Abort that does this (e.g., AbortPreservingUpdates vs. AbortDiscardingUpdates)?

jzimmerman:
> My version of @sternhenri's comment: I was expecting InvocOutput to disappear so that actor methods just return Bytes (or nothing).

That's essentially what we have, I think, except that InvocOutput is now a singleton struct containing Bytes. (We could make InvocOutput a type alias rather than a singleton struct, but this might cause issues with receiver methods on it in Go.)

> With regards to abort error codes (which is orthogonal to the return value and could be postponed to another PR), I think we should change `Abort()` to take an error code parameter (which goes on chain) and a message (which doesn't, just for debugging). All codes passed to Abort should be per-actor code, i.e. distinct from the system codes, and starting at value 16. The runtime should clamp any code up to at least 16. These error code definitions should move into the actor files and need not be distinct between actors. Ideally we should have a different code for every call to Abort, but there might be a few where it makes sense to share the code.

I generally agree, with the additional note that we might want three distinct intervals of error codes (e.g., 0-16 system, 16-32 standard actor, 33+ custom actor). But I also agree that in the interest of velocity we should postpone this to another PR.

> The system error codes should only be referenced from inside the VM/interpreter itself, not from within actor code. The dispatch table that's in actor code itself makes that difficult right now, but this is deep in implementation details that IMO the spec really shouldn't be troubling itself with.

Agreed; would be nice to find a way to avoid the per-type actor boilerplate in Go that really should be part of the runtime.

mishmosh:
This appears to be merged in #684. @jzimmerman can you confirm, and close this PR if so?

PR #676, opened 2019-11-29
WIP: Gas accounting

### LINKED ISSUE #631: Accounting for gas consumption
We need a model for accounting for gas consumption. Which operations are accounted for explicitly? How do we account for direct actor code execution that doesn't cross account-ible boundaries like Runtime calls? Explicit or implicit? How good is good enough for initial network launch?

Precise numbers aren't needed yet, but the potential variance in implementation approaches is huge, and the efficiency gains of knowing where we should head significant.

There's the beginning of a discussion [in slack here](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1572377837107700).

jzimmerman:
(continued from #644, separated from other contributions)

Closes #631

### DIFF
--- a/src/systems/filecoin_vm/interpreter/vm_interpreter.go
+++ b/src/systems/filecoin_vm/interpreter/vm_interpreter.go
@@ -33,14 +33,14 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
 	for _, blk := range msgs.Blocks() {
 		// Pay block reward.
 		reward := _makeBlockRewardMessage(outTree, blk.Miner())
-		outTree, r = vmi.ApplyMessage(outTree, reward, blk.Miner())
+		outTree, r = vmi.ApplyMessage(outTree, reward, 0, blk.Miner())
 		if r.ExitCode() != exitcode.OK() {
 			panic("block reward failed")
 		}
 
 		// Process block miner's Election PoSt.
 		epost := _makeElectionPoStMessage(outTree, blk.Miner(), msgs.Epoch(), blk.PoStProof())
-		outTree, r = vmi.ApplyMessage(outTree, epost, blk.Miner())
+		outTree, r = vmi.ApplyMessage(outTree, epost, 0, blk.Miner())
 		if r.ExitCode() != exitcode.OK() {
 			panic("election post failed")
 		}
@@ -51,16 +51,25 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
 			if found {
 				continue
 			}
-			outTree, r = vmi.ApplyMessage(outTree, m, blk.Miner())
+
+			TODO() // TODO: include signature length?
+			onChainMessageLen := len(msg.Serialize_UnsignedMessage(m))
+
+			outTree, r = vmi.ApplyMessage(outTree, m, onChainMessageLen, blk.Miner())
 			receipts = append(receipts, r)
 			seenMsgs[_msgCID(m)] = struct{}{}
 		}
 	}
 
 	// Invoke cron tick, attributing it to the miner of the first block.
+
+	TODO()
+	// TODO: miners shouldn't be able to trigger cron by sending messages.
+	// Maybe we need a separate ControlActor for this?
+
 	cronSender := msgs.Blocks()[0].Miner()
 	cron := _makeCronTickMessage(outTree, cronSender)
-	outTree, r = vmi.ApplyMessage(outTree, cron, cronSender)
+	outTree, r = vmi.ApplyMessage(outTree, cron, 0, cronSender)
 	if r.ExitCode() != exitcode.OK() {
 		panic("cron tick failed")
 	}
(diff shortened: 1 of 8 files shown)


src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -51,71 +51,129 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
...
 	if toActor == nil {
-		return inTree, _applyError(exitcode.ActorNotFound)
+		return _applyError(inTree, exitcode.ActorNotFound, message.From(), addr.BurntFundsActorAddr)
 	}
 
-	// deduct maximum expenditure gas funds first
-	compTree = _withTransferFundsAssert(compTree, message.From(), addr.BurntFundsActorAddr, maxGasCost)
+	// deduct gas limit funds from sender first
+	compTreePreSend = _withTransferFundsAssert(
+		compTreePreSend, message.From(), addr.BurntFundsActorAddr, gasLimitCost)
```
anorth: nit: it would be more clear why this assertion is ok if you moved it up directly below where the balance is checked.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -51,71 +51,129 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
...
+
+	// Increment sender call sequence number.
+	var err error
+	compTreePreSend, err = compTreePreSend.Impl().WithIncrementedCallSeqNum(message.From())
+	if err != nil {
+		// Note: if actor deletion is possible at some point, may need to allow this case
+		panic("Internal interpreter error: failed to increment call sequence number")
+	}
```
anorth: It's not now. I recommend removing the error return from WithIncrementedCallSeqNum(). That call can panic instead.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -124,51 +182,33 @@ func (vmi *VMInterpreter_I) ApplyMessage(inTree st.StateTree, message msg.Unsign
...
-		TODO() // TODO: assert refundGas is nonnegative
-		outTree = _withTransferFundsAssert(
-			outTree,
-			addr.BurntFundsActorAddr,
-			message.From(),
-			_gasToFIL(refundGas, message.GasPrice()),
-		)
+	ok = _vmiUseGas(sendRet.GasUsed())
+	if !ok {
+		panic("Interpreter error: runtime execution used more gas than provided")
```
anorth: This is not a panic, this is just out of gas (exceeded the pre-set limit). Before execution, it's impossible to know how much will be actually consumed. An efficient implementation will of course halt execution before this point, but IMO the spec doesn't need to. In any case, the semantics are well defined: pay all the gas and revert state changes.

src/systems/filecoin_vm/interpreter/vm_interpreter.id
```diff
@@ -17,6 +17,11 @@ type TipSetMessages struct {
     Epoch   UInt64  // The chain epoch of the blocks
 }
 
+type MessageRef struct {
+    Message          msg.UnsignedMessage
+    MessageSizeOrig  int
+}
+
```
anorth: See my comment in vm_interpreter.go.

You could replace the array elements in BlockMessages.Messages with this, but it's probably better to represent the BLS and SECP messages explicitly, so this spec can also describe which order they are evaluated in.

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -138,6 +139,16 @@ func (rt *VMContext) AbortAPI(msg string) Runtime_AbortAPI_FunRet {
 	return &Runtime_AbortAPI_FunRet_I{}
 }
 
+func (rt *VMContext) CreateActor_DeductGas() Runtime_CreateActor_DeductGas_FunRet {
+	if !rt._actorAddress.Equals(addr.InitActorAddr) {
+		rt.AbortAPI("Only InitActor may call rt.CreateActor_DeductGas")
+	}
```
anorth: This does not seem like the right place for access control. Nothing bad happens anyway.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -51,71 +51,129 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
 			if found {
 				continue
 			}
-			outTree, r = vmi.ApplyMessage(outTree, m, blk.Miner())
+
+			TODO() // TODO: include signature length?
+			onChainMessageLen := len(msg.Serialize_UnsignedMessage(m))
```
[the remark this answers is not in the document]
jzimmerman: (deferring pending resolution of #669)

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -51,71 +51,129 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
...
 			receipts = append(receipts, r)
 			seenMsgs[_msgCID(m)] = struct{}{}
 		}
 	}
 
 	// Invoke cron tick, attributing it to the miner of the first block.
+
+	TODO()
+	// TODO: miners shouldn't be able to trigger cron by sending messages.
+	// Maybe we need a separate ControlActor for this?
```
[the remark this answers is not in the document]
jzimmerman: The TODO isn't intended to remain in the code; it's a placeholder to signal that review is needed (and also serves to ensure that we don't accidentally leave the code incomplete). I can replace the question with a link to the issue, though, if helpful.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
jzimmerman: Unfortunately I think this would cause more confusion due to the out-of-order state tree threading. But I'll add a comment explaining why the assertion is OK.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
jzimmerman: OK -- will modify the API (WithIncrementedCallSeqNum / WithIncrementedCallSeqNum_Assert).

src/systems/filecoin_vm/interpreter/vm_interpreter.go
jzimmerman: It should be impossible for the runtime execution to use _more_ gas than provided -- as specced, it uses `SubtractWhileNonnegative` internally, and bails as soon as the gas usage would exceed the limit (without actually exceeding it).

src/systems/filecoin_vm/interpreter/vm_interpreter.id
jzimmerman: (deferring pending #669)

src/systems/filecoin_vm/runtime/runtime.go
jzimmerman: The desired semantics are for only InitActor to be able to call this. Yes, it might be fine to allow this attack surface, but it might not -- the ability to trigger gas consumption interacts with lots of other potentially complicated logic. I don't think we should allow operations along these lines unless there's a good reason to.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -124,51 +182,33 @@ func (vmi *VMInterpreter_I) ApplyMessage(inTree st.StateTree, message msg.Unsign
...
+		panic("Interpreter error: runtime execution used more gas than provided")
 	}
 
-	outTree, err = outTree.Impl().WithIncrementedCallSeqNum(message.To())
-	if err != nil {
-		// TODO: if actor deletion is possible at some point, may need to allow this case
-		panic("Internal interpreter error: failed to increment call sequence number")
+	ok = _vmiUseGas(gascost.OnChainReturnValue(sendRet.ReturnValue()))
+	if !ok {
+		return _applyError(compTreePreSend, exitcode.OutOfGas, addr.BurntFundsActorAddr, minerAddr)
```
anorth: This is just out-of-gas - the caller should pay gas limit and the miner should not be penalized.

jzimmerman:
Merged in #685.

PR #626, opened 2019-11-07
Domain Separation Tags

nikkolasg:
HS Edit:

follow up to https://github.com/filecoin-project/specs/pull/642

====
This PR adds a Domain Separation Tag for each of the following computation and the verification counter part:
- Ticket generation
- Election Proof 
- Block Signing

*Note 1*: I tried to put it all relevant places, sometimes it added it twice for generating a ticket for example; I am not sure if the redundancy is necessary or not but I put it everywhere.
*Note 2*: Sometimes it is in commented form, but I still added the tag (see BLOCK for example)

PR #677, opened 2019-11-29
Randomness rework

sternhenri:
Continuation of #642 and #626 (cleanly rebased), specifically doing the following:

- Reworking randomness drawing:
  - Ticket production is VRF(last_tix | epoch_wanted)
  - Randomness_drawing is H(tix_from_epoch | epoch)
- specify domain separation tags for VRF
  - specifically little endian uint64 byte of 1 for Ticket, 2 for PoSt, 3 for Block (unused rn)
  - byte delimeter (0)
- Quick touch up of the key_store values to account for real usage (will need further work later)
- fixing some prior erros deleting example code

To Check:
- [ ] How to give the VM Runtime access to `RandomnessAtEpoch()` from `block.go`
- [ ] Whether to add pre set lookbacks to the runtime (as is done in the `Storage_Power_Consensus_Subsystem`s randomness methods) rather than have offset manually added
- [ ] Whether to do Domain Separation Tag for Hashing Randomness (I would think not)
- [x] Whether to change to some `RandomnessRole` usage as recommended


src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
```diff
@@ -86,23 +84,23 @@ func (spc *StoragePowerConsensusSubsystem_I) IsWinningPartialTicket(partialTicke
...
 }
 
-func (spc *StoragePowerConsensusSubsystem_I) GetPoStChallenge(chain block.Chain, minerAddr addr.Address, epoch block.ChainEpoch) util.Randomness {
-	return chain.RandomnessAtEpoch(minerAddr, epoch-SPC_LOOKBACK_POST)
+func (spc *StoragePowerConsensusSubsystem_I) GetPoStChallenge(chain block.Chain, epoch block.ChainEpoch) util.Randomness {
+	return chain.RandomnessAtEpoch(epoch - SPC_LOOKBACK_POST)
 }
 
-func (spc *StoragePowerConsensusSubsystem_I) ValidateElectionProof(height block.ChainEpoch, electionProof block.ElectionProof, workerAddr addr.Address) bool {
+func (spc *StoragePowerConsensusSubsystem_I) ValidateElectionProof(height block.ChainEpoch, electionProof block.ElectionProof, minerAddr addr.Address) bool {
```
anorth: The commented-out body of this function expects to obtain a public key for this address. Miner actor addresses have no such keys. If this is intended to be the address of the actor that signed something, it must be the worker address. Although passing the pubkey would be more useful.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.id
```diff
@@ -23,10 +25,14 @@ type StoragePowerConsensusSubsystem struct {//(@mutable)
...
         electionProof  block.ElectionProof
-        workerAddr     addr.Address
+        minerAddr      addr.Address
     ) bool
 
-    validateTicket(tix block.Ticket, pk filcrypto.VRFPublicKey) bool
+    validateTicket(
+        tix        block.Ticket
+        pk         filcrypto.VRFPublicKey
+        minerAddr  addr.Address
```
anorth: Ditto my comment about this being a worker address if you expect to validate a signature.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
sternhenri: Yes! Sorry, this is all going away in a forthcoming PR to fix up EC for ePoSt so pay it no mind :)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.id
sternhenri: So in this case it's not about a sig but rather about the fact that new tickets use a minerAddress as part of randomness generation. The idea is that if different miner actors use the same worker keys, we would want them to generate different randomness, so we hash along with the minerAddr. No sig validation here.

anorth:
... by which I mean I'm not qualified to approve, but after @jzimmerman does, ship it!

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
```diff
@@ -57,12 +52,15 @@ func (spc *StoragePowerConsensusSubsystem_I) ValidateBlock(block block.Block_I)
...
-	// input.append(Randomness1)
-	// return ticket.Verify(input, pk)
+func (spc *StoragePowerConsensusSubsystem_I) validateTicket(ticket block.Ticket, pk filcrypto.VRFPublicKey, minerAddr addr.Address) bool {
+	randomness1 := spc.GetTicketProductionSeed(spc.blockchain().BestChain(), spc.blockchain().LatestEpoch())
+
+	var input []byte
+	input = append(input, byte(filcrypto.DomainSeparationTag_Case_Ticket))
+	input = append(input, randomness1...)
+	input = append(input, byte(filcrypto.InputDelimeter_Case_Bytes))
+	input = append(input, minerAddr.ToBytes()...)
```
jzimmerman: Hmm... I don't think it's feasible in general to duplicate serialization code for each type involved in randomness generation -- this will tend to cause lots of subtle errors and be very difficult to maintain over time. I've drafted a candidate API that I think solves this in a generic way, and also leverages the existing serialization machinery to avoid potential redundancy there:

https://github.com/filecoin-project/specs/pull/679

Let me know what you think -- if it looks like it will work for these use cases, maybe easiest to rebase (parts of?) this PR on that branch and swap in the new API.

sternhenri: Happy to use your API which is nice though what is the index for in it?

Also I am not sure what you mean wrt my duplicating serialization code. What I've implemented is most of it, as in 
- how to generate new tickets
- how to draw tickets
- how to generate post randomness.

I might be able to generalize the methods a bit (this is where your API is nice) by passing in the tag and serialization... But then some of the types do need to be different (e.g. generation vs sampling).

If you can give me a bit more to go on I'll rebase and try again

jzimmerman: > Happy to use your API which is nice though what is the index for in it?

The index is intended to be used for generation of strings longer than 256 bits (e.g., a 768-bit pseudorandom string may be derived as: `tag.DeriveRandWithIndex(s, 1) || tag.DeriveRandWithIndex(s, 2) || tag.DeriveRandWithIndex(s, 3)` where || indicates concatenation.) I'll update #679 with a comment to this effect.

jzimmerman: > Also I am not sure what you mean wrt my duplicating serialization code. 

I mean writing multiple implementations of transforming input objects to pseudorandom output (which might vary in its details based on the type of the input object, but still duplicates the same basic logic in each case).

> I might be able to generalize the methods a bit (this is where your API is nice) by passing in the tag and serialization... But then some of the types do need to be different (e.g. generation vs sampling).

I think generation should be covered by the API already present in #679. What functionality are you looking for in terms of sampling? Most likely we'll be able to implement this also in a generic way in `randomness.go`.

sternhenri:
Ready for another pass @jzimmerman. Keep in mind that this is not us writing software libraries, simply trying to define an easy-to-follow spec for implementers to follow. To that end, I've eschewed some of your suggestions in the name of keeping things simple and easy to follow, e.g. drawing randomness should not necessitate CBOR-serialization. That seems like overkill, for more complex concats, I agree.

A few questions to keep in mind in this pass:

**For this PR**:
1. DST not strictly nec for ticket drawing. Do we need to use it? I would think it is needlessly complex
1. Do we need DST for VRFs as well? Or just for Hash input to VRF…? I assume the latter
1. How are you planning to implement all the serialization and deserialization methods? I worry that basing randomness on this as we do, given the old way is well defined will delay spec readiness. Done is better than perfect, and few places need randomness (that I can tell) beyond what is here.

**For a forthcoming VM PR** (or happy to do now if you give me a sense of how):
1. How to give the VM Runtime access to RandomnessAtEpoch() from block.go
1. Should we add pre set lookbacks to the runtime (as is done in the Storage_Power_Consensus_Subsystems randomness methods) rather than have offset manually added

src/algorithms/crypto/randomness.go
```diff
@@ -18,7 +18,7 @@ type Serialization = util.Serialization
 // object, define a struct at the .id level that contains those objects as member fields.
 // This will then cause a Serialize_*() method to be generated for the struct type.
 func (tag DomainSeparationTag) DeriveRand(s Serialization) Randomness {
-	return _deriveRandInternal(tag, s, 0)
+	return _deriveRandInternal(tag, s, -1)
```
jzimmerman: What's the rationale for changing the sentinel index to -1 instead of 0? I don't have a strong preference, but if it's just aesthetic, would prefer to leave to a subsequent PR for simplicity.

src/algorithms/crypto/randomness.go
```diff
@@ -33,8 +33,10 @@ func (tag DomainSeparationTag) DeriveRandWithIndex(s Serialization, index int) R
 func _deriveRandInternal(tag DomainSeparationTag, s Serialization, index int) Randomness {
 	buffer := []byte{}
 	buffer = append(buffer, util.IntToBytesLittleEndian(int(tag))...)
-	buffer = append(buffer, util.IntToBytesLittleEndian(int(index))...)
 	buffer = append(buffer, Bytes(s)...)
+	if index != -1 {
+		buffer = append(buffer, util.IntToBytesLittleEndian(int(index))...)
+	}
```
jzimmerman: This change would break the unambiguity proof (https://github.com/filecoin-project/specs/pull/679#discussion_r352290621) -- the serialization may be variable-length.

src/algorithms/crypto/vrf.md
```diff
@@ -18,22 +18,14 @@ BLS signature can be used as the basis to construct a VRF. Filecoin transforms
 the BLS signature scheme it uses (see {{<sref signatures Signatures>}} into a
 VRF, Filecoin uses the random oracle model and deterministically hash the
 signature to produce the final digest. Filecoin uses
-[Blake2b-256](https://blake2.net/) (Blake2b with an output size of 256 bits) as
-the hash function. The algorithm is the following:
+SHA256 as the hash function. The algorithm is the following:
 ```
-VRFOutput = Blake2b-256(DST || BLSSignature(message))
+VRFOutput = SHA256(DST || BLSSignature(message))
```
jzimmerman: Ideally this should be defined in terms of `DeriveRand` also, but I think this can wait for a subsequent PR.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -55,38 +55,47 @@ Whenever comparing tickets is evoked in Filecoin, for instance when discussing s
 While each Filecoin block header contains a ticket field (see {{<sref tickets>}}), it is useful to think of a ticket chain abstraction.
 Due to the nature of Filecoin's Tipsets and the possibility of using tickets from epochs that did not yield leaders to produce randomness at a given epoch, tracking the canonical ticket of a subchain at a given height can be arduous to reason about in terms of blocks. To that end, it is helpful to create a ticket chain abstraction made up of only those tickets to be used for randomness generation at a given height.
 
+To read more about specifically how tickets are processed for randomness, see {{<sref randomness>}}. Note that the ticket output (bytes) can be used directly, unlike with more complex objects where one would use their serialization.
```
jzimmerman: (see below)

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -55,38 +55,47 @@ Whenever comparing tickets is evoked in Filecoin, for instance when discussing s
...
 {{<label ticket_generation>}}
-### Ticket generation
+### Randomness Ticket generation
 
 This section discusses how tickets are generated by EC for the `Ticket` field in every block header.
 
 At round `N`, a new ticket is generated using tickets drawn from the Tipset at round `N-1` (as shown below).
 
-The miner runs the prior ticket through a Verifiable Random Function (VRF) to get a new unique ticket which can later be derived for randomness (as shown above).
+The miner runs the prior ticket through a Verifiable Random Function (VRF) to get a new unique ticket which can later be derived for randomness (as shown above). The prior ticket is prepended with the ticket domain separation tag `1` and concatenated with the miner actor address (to ensure miners using the same worker keys get different randomness).
```
jzimmerman: Would prefer not to hard-code the DST value here.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -55,38 +55,47 @@ Whenever comparing tickets is evoked in Filecoin, for instance when discussing s
...
+
+To generate a ticket for a given epoch n:
+```text
+LastTicket = MinTicketValueAtEpoch(n-1)
+newRandomness = VRF_miner(H(TicketDST || Serialization(pastTicket, minerActorAddress)))
```
jzimmerman: (as above, ideally would be specified in terms of `DeriveRand`)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
```diff
@@ -86,24 +79,24 @@ func (spc *StoragePowerConsensusSubsystem_I) IsWinningPartialTicket(partialTicke
...
-func (spc *StoragePowerConsensusSubsystem_I) GetSealSeed(chain block.Chain, minerAddr addr.Address, epoch block.ChainEpoch) util.Randomness {
-	return chain.RandomnessAtEpoch(minerAddr, epoch-SPC_LOOKBACK_SEAL)
+func (spc *StoragePowerConsensusSubsystem_I) GetSealRand(chain block.Chain, epoch block.ChainEpoch) util.Randomness {
+	return chain.RandomnessAtEpoch(epoch - SPC_LOOKBACK_SEAL)
 }
 
-func (spc *StoragePowerConsensusSubsystem_I) GetPoStChallenge(chain block.Chain, minerAddr addr.Address, epoch block.ChainEpoch) util.Randomness {
-	return chain.RandomnessAtEpoch(minerAddr, epoch-SPC_LOOKBACK_POST)
+func (spc *StoragePowerConsensusSubsystem_I) GetPoStChallengeRand(chain block.Chain, epoch block.ChainEpoch) util.Randomness {
+	return chain.RandomnessAtEpoch(epoch - SPC_LOOKBACK_POST)
```
jzimmerman: These should all use distinct DSTs.

src/systems/filecoin_blockchain/struct/block/election.go
```diff
@@ -10,21 +10,37 @@ func (tix *Ticket_I) ValidateSyntax() bool {
...
 
-func (tix *Ticket_I) DrawRandomness(minerAddr addr.Address, epoch ChainEpoch) util.Bytes {
-	input := tix.Output()
-	input = append(input, addrToLittleEndianBytes(minerAddr)...)
-	input = append(input, epochToLittleEndianBytes(epoch)...)
-	return input
+func (tix *Ticket_I) DrawRandomness(epoch ChainEpoch) util.Bytes {
+	var input []byte
+	input = append(input, tix.Output()...)
+	input = append(input, byte(epoch))
```
jzimmerman: (see below)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
sternhenri: Why?

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: It is in the code. 

Again I don't think we should force people to your abstractions. The prose should also standalone.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: Different conception of Prose. I think the Prose should suffice when needed. Should be changed when necessary. It's more work but worth it to impl imo @anorth ?

src/algorithms/crypto/vrf.md
sternhenri: It is in all the code.

src/algorithms/crypto/randomness.go
sternhenri: ok. I'll flip order again. But couldn't the index? We are talking about chain epochs here which could grow v large.

src/algorithms/crypto/randomness.go
sternhenri: 0 is genesis block. Ambiguous. -1 is a clear flag

jzimmerman:
> I've eschewed some of your suggestions in the name of keeping things simple and easy to follow, e.g. drawing randomness should not necessitate CBOR-serialization.

I think using CBOR serialization is actually the simplest solution by far. Rationale is as follows -- let me know if this seems reasonable:
- We need to be able to serialize/deserialize a wide variety of algebraic datatypes (ADT)
- Any reasonable serialization schema would do fine here, but we've already settled on CBOR in other contexts
- Writing the serialization machinery manually for each ADT is tedious and error-prone, and tends to cause discrepancies between impls; therefore, we should generate it automatically
- Given that we're going to generate CBOR serializers for each ADT automatically, it's effectively free, in the sense that there's no additional engineering overhead to doing CBOR serialization for a new ADT
- Therefore, using CBOR serialization, even in relatively trivial cases, is still easier (and also less error-prone) than writing code for each case manually.

> 1. DST not strictly nec for ticket drawing. Do we need to use it? I would think it is needlessly complex

Yes -- in fact, I would say we should be even _more_ pedantic with DSTs if possible, using a distinct value for every call to `DeriveRand` (and not just distinct values per subsystem or per general application). Potential collisions in randomness derivation can be catastrophic, and I think we should much prefer to eliminate the conceptual overhead of reasoning about them, even at the cost of having a few superfluous DSTs.

> 2. Do we need DST for VRFs as well? Or just for Hash input to VRF…? I assume the latter

If I understand correctly, this is the difference between
`VRFEval(sk, DeriveRand(tag, x))` and
`DeriveRand(VRFEval(sk, ...))`?
If so, I agree only the former is necessary (assuming we use a secure VRF "by the book", and not just e.g. a raw signature scheme).

> 3. How are you planning to implement all the serialization and deserialization methods?

Via codegen from the ADT definitions in id -- will need this fairly soon in any case, so may as well also leverage it for randomness, I think.

> **For a forthcoming VM PR** (or happy to do now if you give me a sense of how):
> 
> 1. How to give the VM Runtime access to RandomnessAtEpoch() from block.go
> 2. Should we add pre set lookbacks to the runtime (as is done in the Storage_Power_Consensus_Subsystems randomness methods) rather than have offset manually added

Let's defer these to a subsequent PR, but re: (2.) I agree with your approach above -- the runtime shouldn't know about the offset logic.

sternhenri:
Thanks @jzimmerman. Will make a final push given all your answers! The one part I disagree with is being super pedantic with the DSTs. While I agree in principle, it makes impl much harder and pushes them to adopt this abstraction. Otherwise I think having a general (`drawRandomness` bucket for basic randomness and specific DSTs as needed for more complex inputs (as with PoSts or generating new tickets) should strike a happy balance. If you think it's insecure I'll defer to you.

A last question for you
4. do you see the key store being modified here (from a wrong v i had) as being conflicting with this one? https://filecoin-project.github.io/specs/#systems__filecoin_nodes__repository__key. How should we adapt/merge?

src/algorithms/crypto/randomness.go
jzimmerman: It's not at all ambiguous in the precise sense of https://github.com/filecoin-project/specs/pull/679#discussion_r352290621 -- but if it's conceptually confusing, I'm fine either way.

src/algorithms/crypto/randomness.go
jzimmerman: No, the index is 64 bits and chain epochs should never exceed that. (Also it's not just a matter of exchanging the order -- we have to always include the index unconditionally in the serialization, whether or not it's -1.)

src/algorithms/crypto/vrf.md
jzimmerman: Understood, just flagging that `SHA256(DST || x)` is technically incorrect (should be `SHA256(DST || index || x)`, whether index ends up being 0 or -1 as per discussion above)

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
jzimmerman: Agreed, but we shouldn't hard-code the number 1 in the prose since this might change -- should either pull it in from the code dynamically using hugo, or (much simpler) avoid mentioning the concrete value in the prose altogether, and just say it uses a DST (link to reference on DSTs).

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
jzimmerman: Up to you, I think it's clearer to the reader if we have a single `DeriveRand` primitive in the prose, but I don't have a strong preference as long as it's functionally correct (see https://github.com/filecoin-project/specs/pull/677#discussion_r352750763).

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
jzimmerman: (discussion continued below)

jzimmerman:
> 4. do you see the key store being modified here (from a wrong v i had) as being conflicting with this one? https://filecoin-project.github.io/specs/#systems__filecoin_nodes__repository__key. How should we adapt/merge?

I'm not sure I have enough context on this, but in general I like the new KeyStore much better, since it makes the roles of the keys more explicit rather than relying on a string-based map. Definitely worth another pass in a subsequent PR, though, and in particular, we should make sure we're disambiguating keys correctly (and not, e.g., reusing a BLS key both for signatures and for the VRF).

src/algorithms/crypto/randomness.go
sternhenri: nit: it is ambiguous in that it has meaning beyond being a sentinel

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
sternhenri: This is the only one where I strong disagreee,. These methods are just convenience to automatically apply proper offsets for drawing the right ticket. The DSTs should be applied when constructing the seed properly (e.g. PoStRandomness or SealRandomness) as is done (eg) in storage_mining_subsystem for PoSt.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -55,38 +55,47 @@ Whenever comparing tickets is evoked in Filecoin, for instance when discussing s
...
+Set referenceTipsetOffset = 0
+While true:
+    Set referenceTipsetHeight = n - referenceTipsetOffset
+    If blocks were mined at referenceTipsetHeight:
+        ReferenceTipset = TipsetAtHeight(referenceTipsetHeight)
+        Select the block in ReferenceTipset with the smallest final ticket, return its value (pastTicket).
+    If no blocks were mined at referenceTipsetHeight:
+        Increment referenceTipsetOffset
+        (Repeat)
+newRandomness = H(TicketDrawDST || epoch || Serialization(pastTicketOutput))
```
jzimmerman: Would prefer to use `H(DST || index || Serialization(epoch || pastTicketOutput))` where index = -1, as above, to be conceptually consistent -- not critical though.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
jzimmerman: What if some of the offsets (`SPC_LOOKBACK_TICKET`, `SPC_LOOKBACK_RANDOMESS`, ...) happen to be equal -- wouldn't this cause us to collide in the randomness output?

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: So when would you ever use the index?

jzimmerman: If you want to derive a longer pseudorandom string from a single input -- e.g. if you need 768 bits generated from `struct { epoch, ticketOutput }`, you would concatenate the results with indices (0, 1, 2).

sternhenri:
@anorth could I ask for your ack on the final solution here plz? I am a bit worried we overengineered this vs what is needed rn and will slow down impl work by needlessly future proofing here.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
sternhenri: yes but we apply extra transformations to the randomness (eg for tickets we use epochID, for Posts we use miner addr) so the actual used randomness is processed further. Beside if it wasn't I don't see why that would be so bad, as long as the randomness is deterministically drawn...?

jzimmerman: I agree that there's no obvious attack -- but I think it would be much less work to prove security if we didn't have to do the case-by-case analysis each time, and instead could just rely on the general composability of the randomness primitive (which only holds if we invoke it with distinct tags for each use).

sternhenri: I'm sensitive to that. Let's get feedback from @anorth and @whyrusleeping on #677. If 👍 I'm fine to add. Your point is well taken, I just worry about preemptive complexity and excessive future proofing though that might be an engineering mindset.

anorth:
> @anorth could I ask for your ack on the final solution here plz

I don't think I understand the requirements or the intent. I certainly can't derive them from the spec-code. So I can't judge whether it's overengineered. It's definitely opaque though.

PR #678, opened 2019-11-29
[To be closed; ignore] Simplify how Surprise PoSts occur within a PP

sternhenri:
In this PR I simplify the surprise sampling mechanism. This is how it worked:
- At every cron tick, `SURPRISE_CHALLENGE_FREQUENCY * len(PT) / ProvingPeriod` miners are randomly chosen for sampling (meaning each should be chosen SURPRISE_POST_FREQUENCY times per proving period on expectation).
- If a chosenMiner should be challenged (`ShouldChallenge`), then it is issued the surprise.
  - `ShouldChallenge` checks that the miner is not already being challenged, and that more than `SURPRISE_NO_CHALLENGE_PERIOD` has passed since their last successful PoSt (ePoSt or sPoSt) submission. 
  - We set `SURPRISE_NO_CHALLENGE_PERIOD = PROVING_PERIOD / SURPRISE_CHALLENGE_FREQUENCY` meaning that if SURPRISE_CHALLENGE_FREQ = 2, then a miner will not get surprised in the first half of their PP.
- Finally, a miner cannot do an ePoSt after they have been challenged.

This is how this PR modifies it:
1. A miner can now ePoSt even after being challenged (and can do that instead of replying to the challenge) the first time (i.e. if they fail the challenge they can no longer do so).
2. There is no more `SURPRISE_NO_CHALLENGE_PERIOD`: a miner can get surprised at any time during a proving period, and given a certain `SURPRISE_CHALLENGE_FREQUENCY` may get challenged more than once per PP. (In fact in order to ensure that each miner gets at least 1 challenge per PP, SCF should be set high enough).

It is suboptimal but removes unnecessary complexity from the process.

### DIFF
--- a/src/algorithms/post/election_post.md
+++ b/src/algorithms/post/election_post.md
@@ -148,12 +148,12 @@ For i=0; i < challNumber; i++:
     ranHash = H(ticket, i)
     ranIndex = HashToInt(ranHash) mod len(PowerTable)
     chosenMiner = PowerTable[ranIndex].address
-    // a miner should only be challenged if they have not submitted a post in ProvingPeriod/SURPRISE_CHALLENGE_FREQUENCY epochs and are not currently challenged
-    if chosenMiner.shouldChallenge(SURPRISE_NO_CHALLENGE_PERIOD):
+    // a miner can be challenged at any time in the proving period
+    if chosenMiner.shouldChallenge():
         sampledMiners.append(chosenMiner)
 ```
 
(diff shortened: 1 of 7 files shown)


src/algorithms/post/election_post.md
```diff
@@ -148,12 +148,12 @@ For i=0; i < challNumber; i++:
     ranHash = H(ticket, i)
     ranIndex = HashToInt(ranHash) mod len(PowerTable)
```
nikkolasg: Imo we must fix this biased way of generating a random index.

sternhenri: do you have a suggestion?

sternhenri:
After discussion, PR will be closed pending resolution of outstanding comments. We will keep the NO_POST_PERIOD in order to protect on-chain footprint.

sternhenri:
closed in favor of https://github.com/filecoin-project/specs/pull/718

PR #679, opened 2019-11-29
Randomness API with domain separation

jzimmerman:
Establish a single consistent API for generating randomness with domain separation. (Note: the same domain separation enum should also apply for signatures.)

### DIFF
--- a/src/algorithms/crypto/domain.go
+++ b/src/algorithms/crypto/domain.go
@@ -1,0 +1,9 @@
+package crypto
+
+type DomainSeparationTag int
+
+const (
+	DomainSeparationTag_TicketProduction DomainSeparationTag = 1 + iota
+	DomainSeparationTag_ElectionPoSt
+	// ...
+)
--- a/src/algorithms/crypto/hash.go
+++ b/src/algorithms/crypto/hash.go
@@ -1,0 +1,11 @@
+package crypto
+
+import (
+	util "github.com/filecoin-project/specs/util"
+)
+
+type Bytes = util.Bytes
+
+func SHA256(Bytes) Bytes {
+	panic("TODO")
+}
--- a/src/algorithms/crypto/randomness.go
+++ b/src/algorithms/crypto/randomness.go
@@ -1,0 +1,40 @@
+package crypto
+
+import (
+	util "github.com/filecoin-project/specs/util"
+)
+
+type Randomness = util.Randomness
+type Serialization = util.Serialization
+
+// Derive a random byte string from a domain separation tag and an arbitrary
+// serializable object.
+//
+// Note: to produce values of type Serialization, use the auto-generated method prototypes
+//   Serialize_T(T) Serialization
+// for each type T defined in .id.
+//
+// In order to derive randomness from a collection of objects, rather than just a single
+// object, define a struct at the .id level that contains those objects as member fields.
+// This will then cause a Serialize_*() method to be generated for the struct type.
+func (tag DomainSeparationTag) DeriveRand(s Serialization) Randomness {
+	return _deriveRandInternal(tag, s, 0)
+}
+
+// As in DeriveRand(), but additionally accepts an index into the implicit pseudorandom stream.
+// Index must be strictly positive.
+func (tag DomainSeparationTag) DeriveRandWithIndex(s Serialization, index int) Randomness {
+	if index <= 0 {
+		panic("DeriveRandWithIndex only accepts indices > 0")
+	}
+	return _deriveRandInternal(tag, s, index)
+}
+
+func _deriveRandInternal(tag DomainSeparationTag, s Serialization, index int) Randomness {
+	buffer := []byte{}
+	buffer = append(buffer, util.IntToBytesLittleEndian(int(tag))...)
+	buffer = append(buffer, util.IntToBytesLittleEndian(int(index))...)
+	buffer = append(buffer, Bytes(s)...)
+	ret := SHA256(buffer)
+	return Randomness(ret)
+}
--- a/tools/codeGen/util/util.go
+++ b/tools/codeGen/util/util.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"bytes"
+	"encoding/binary"
 	"io"
 	"log"
 	big "math/big"
@@ -90,6 +91,13 @@ func TextAbbrev(s string, maxLen int) string {
 		return s
 	}
 	return s[:maxLen-3] + "..."
+}
+
+func IntToBytesLittleEndian(x int) Bytes {
+	buf := bytes.NewBuffer(make([]byte, 0, 8))
+	err := binary.Write(buf, binary.LittleEndian, x)
+	Assert(err == nil)
+	return buf.Bytes()
 }
 
 func IntMin(x, y int) int {


src/algorithms/crypto/domain.go
```diff
@@ -0,0 +1,9 @@
+package crypto
+
+type DomainSeparationTag int
+
+const (
+	DomainSeparationTag_TicketProduction DomainSeparationTag = 1 + iota
+	DomainSeparationTag_ElectionProof
```
sternhenri: Remove election proof (stale) and replace with post, will also need one for block signing

src/algorithms/crypto/hash.go
```diff
@@ -0,0 +1,11 @@
+package crypto
+
+import (
+	util "github.com/filecoin-project/specs/util"
+)
+
+type Bytes = util.Bytes
+
+func SHA256(Bytes) Bytes {
```
sternhenri: There's an impl in the filproofs lib (algorithms.go) but it causes some import cycles so this is fine. May be worth moving theirs here and having them import it (with Porcu's consent)

src/algorithms/crypto/randomness.go
```diff
@@ -0,0 +1,40 @@
...
+// In order to derive randomness from a collection of objects, rather than just a single
+// object, define a struct at the .id level that contains those objects as member fields.
+// This will then cause a Serialize_*() method to be generated for the struct type.
+func (tag DomainSeparationTag) DeriveRand(s Serialization) Randomness {
+	return _deriveRandInternal(tag, s, 0)
+}
+
+// As in DeriveRand(), but additionally accepts an index into the implicit pseudorandom stream.
+// Index must be strictly positive.
+func (tag DomainSeparationTag) DeriveRandExtended(s Serialization, index int) Randomness {
```
sternhenri: The method name is not super descriptive. But I really like these interfaces.

Can you add an optional field as well in case we need it (e.g. including minerAddress in ticket production to distinguish tickets generated by different miners with the same key

src/algorithms/crypto/randomness.go
```diff
@@ -0,0 +1,40 @@
...
+	if index <= 0 {
+		panic("DeriveRandExtended only accepts indices > 0")
+	}
+	return _deriveRandInternal(tag, s, index)
+}
+
+func _deriveRandInternal(tag DomainSeparationTag, s Serialization, index int) Randomness {
+	buffer := []byte{}
```
sternhenri: can you add a field delimeter between each field ('0' byte)?

src/algorithms/crypto/domain.go
jzimmerman: > Remove election proof (stale) and replace with post

i.e., replace `ElectionProof` with `ElectionPoSt`?

jzimmerman: > will also need one for block signing

Will leave this for you to fill in in subsequent PR -- presumably many more tags will be needed (this is just one example), and the intent is for clients of the API to just add them here.

src/algorithms/crypto/hash.go
jzimmerman: OK -- am inclined to leave as is for merge velocity, but we should definitely coalesce the various impls soon (I think there are other stub impls throughout the codebase as well).

src/algorithms/crypto/randomness.go
jzimmerman: > The method name is not super descriptive.

Good point; how about `DeriveRandWithIndex`?

jzimmerman: > Can you add an optional field as well in case we need it

I've experimented with variants of this, but I think it would incur too much complexity. If we want to derive randomness from the conjunction of type `A` and optionally type `B`, we can just create a struct `type C = struct { A; B? }` and use the serialization of `C` as the argument here.

src/algorithms/crypto/randomness.go
jzimmerman: I'm not sure what purpose this would serve. The entries are already unambiguously delimited, by virtue of the fact that the first two are both constant-length (8 bytes). (And interposing a zero byte wouldn't help with ambiguity if they weren't already unambiguous -- e.g., some of the entries might just happen to have zero bytes in them anyway.)

src/algorithms/crypto/domain.go
sternhenri: Yes, it would be `PoSt` more generally since could be `electionPoSt` or `SurprisePoSt`.

Your second point sgtm

src/algorithms/crypto/randomness.go
sternhenri: I disagree only insofar as what the abstraction provides here is inner working of how we form the randomness, e.g. using delimeters, type of encoding, etc. 

If people have to much with that outside the class it defeats part of the purpose imo and remains as error prone as it was before

src/algorithms/crypto/randomness.go
sternhenri: cc @whyrusleeping who likely won't review just yet. But you could try and find meaningful collisions by picking keys right/entries right. 

You are right they are constant for now but as a general rule... Costs nothing

src/algorithms/crypto/randomness.go
jzimmerman: I don't quite understand this point. The definition of `struct C` doesn't need to mention or reason about delimiters or encoding at all; it's just defined the same way as any ordinary struct.

src/algorithms/crypto/randomness.go
jzimmerman: This doesn't quite make sense to me either. Backtracking a moment, just in case I misunderstood:

> can you add a field delimeter between each field ('0' byte)?

Does this mean adding two field delimiters to separate the three entries listed (i.e., tag, index, and serialization `s`), or does it mean adding field delimiters _within_ the serialization `s` itself?

src/algorithms/crypto/randomness.go
sternhenri: fair enough point. yes.

src/algorithms/crypto/randomness.go
sternhenri: depends on how you build the serialization but i would say in both cases.

jzimmerman: I think it's important to separate the two cases out. In general, the property we'd like to prove is _unambiguity_. More precisely: for two inputs `x_0 = (tag_0, index_0, obj_0)` and `x_1 = (tag_1, index_1, obj_1)`, suppose `DeriveRand(x_i) = H(f(x_i))`. Then we want to prove:

> [Property A]:  If `x_0 != x_1`, then `f(x_0) != f(x_1)`.

from which the desired pseudorandomness will follow assuming we model the hash function H as a random oracle.

As defined above, we have

`f(tag, index, obj) = (tag || index || S(obj))`

where `S` denotes serialization of an arbitrary object. Since tag and index are both fixed-length, I claim property A holds for `f` trivially, _assuming_ property A holds for `S`. Moreover, if a serialization method `S` is functionally correct (meaning that it has a corresponding deserialization method that is its left-inverse), then this condition holds for `S`. Thus, we may take any serialization method `S` that is functionally correct under this definition.

In order to minimize redundancy, I suggest using the same serialization method `S` that we use elsewhere, i.e., currently CBOR-based serialization defined on algebraic datatypes. But I've left the specification parametric over `S`, in the sense that the `Serialize_T(T)` method prototypes auto-generated from .id files are still stubs at the moment. I think we should plan to update this in a future PR.

sternhenri: Makes sound sense to me. Thanks for walking through that.

PR #680, opened 2019-11-30
ElectionPoSt Leader Election

sternhenri:
Added code for doing ePoSt LE verification, tying into appropriate interfaces.

Needs a lot of review, including
- [ ] how SPA and SMA are brought in to the subsystems
- [ ] how sector utilization is done unsafely (ie without runtime or error checking)
- [ ] some type conversion hacks

TODO in future PRs:
- [ ] add code for consensus fault checking
- [ ] add code for computing chain weight


src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
```diff
@@ -238,6 +238,12 @@ func (st *StorageMinerActorState_I) _getUtilizationInfo(rt Runtime, sectorNo sec
 	return utilizationInfo
 }
 
+func (st *StorageMinerActorState_I) _getCurrUtilization(sectorNo sector.SectorNumber) block.StoragePower {
```
zixuanzh: there is a `_getUtilizationInfo` that does something similar in storage_miner_actor_state

src/systems/filecoin_blockchain/storage_power_consensus/expected_consensus.go
```diff
@@ -32,22 +34,29 @@ func (self *ExpectedConsensus_I) IsValidConsensusFault(faults ConsensusFaultType
 	// && abs(block1.Epoch - block2.Epoch) == 1
 }
 
-func (self *ExpectedConsensus_I) IsWinningChallengeTicket(electionProof []byte, targetNum util.UVarint, targetDen util.UVarint) bool {
+func (self *ExpectedConsensus_I) IsWinningChallengeTicket(challengeTicket util.Bytes, maxTicket util.Bytes, sectorPower block.StoragePower, totalPower block.StoragePower, sampleNum util.UVarint, sampleDenom util.UVarint) bool {
 	// Conceptually we are mapping the pseudorandom, deterministic hash output of the challenge ticket onto [0,1]
 	// by dividing by 2^HashLen and comparing that to the sector's target.
 	// if the challenge ticket hash is smaller than active power in the sector size / network total power * sectorSampled * ec.ExpectedLeaders
```
sternhenri: 1/sectorsSampled

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
sternhenri: I know but it requires the runtime, which I am not sure how to pass in from the subsystem. That's what I mean above by "unsafe" sector utilization

src/algorithms/post/election_post.md
```diff
@@ -63,34 +66,35 @@ The miner calls `GenerateCandidates` from proofs with their non-faulted (declare
 
 4. **(check Challenge Ticket(s) for winners)**
 Given a returned PartialTicket, miner checks it is a winning ticket. Specifically, they do the following:
-    - `ChallengeTicket = Finalize(PartialTicket).  = H(ChallengeTicket) / 2^len(H)`
+
+    - `ChallengeTicket = Finalize(PartialTicket) = H(ChallengeTicket) / 2^len(H)`
     - Check that `ChallengeTicket < Target`
+        - `Target = activePowerInSector/networkPower * 1/sectorsSampled * EC.ExpectedLeaders`.
+        - Put another way check `challengeTicket_num * networkPower * sectorsSampled_denom < activePowerInSector * sectorsSampled_num * EC.ExpectedLeaders * challengeTicket_denom`
```
jzimmerman: nit: should sectorsSampled_num here be sectorsSampled_denom and vice versa?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
```diff
@@ -66,17 +69,45 @@ func (spc *StoragePowerConsensusSubsystem_I) StoragePowerConsensusError(errMsg s
...
+}
 
-	return spc.ec().IsWinningChallengeTicket(challengeTicket, activePowerInSector, totalPower)
+// TODO: fix linking here
+var node node_base.FilecoinNode
+
+func (spc *StoragePowerConsensusSubsystem_I) _getStoragePowerActorState(stateTree stateTree.StateTree) StoragePowerActorState {
+	// powerAddr := stateTree.GetActorAddress(stateTree.ActorName_StoragePowerActor)
+	// TODO Fix
+	var powerAddr addr.Address
```
jzimmerman: I think this can just be `addr.StoragePowerActorAddr`.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
jzimmerman: I think this is correct -- it should be a pure function and shouldn't need the runtime. (Why is this "unsafe"?)

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
```diff
@@ -79,11 +79,16 @@ func (sms *StorageMiningSubsystem_I) _tryLeaderElection() {
...
 
+	// TODO Fix
+	var currState stateTree.StateTree
 	winningCandidates := make([]sector.PoStCandidate, 0)
+	st := sms._getStorageMinerActorState(currState, sms._keyStore().MinerAddress())
 
 	for _, candidate := range candidates {
-		// TODO align on worker address
-		if sms._consensus().IsWinningPartialTicket(candidate.PartialTicket()) {
+		// TODO: fix
```
jzimmerman: Ideally this type of TODO should be accompanied by a panic (or util.TODO call), since otherwise we may risk missing it and assuming the functionality is ready.

src/systems/filecoin_vm/actor/address/address.go
```diff
@@ -55,6 +55,11 @@ func (a *Address_I) IsKeyType() bool {
 	panic("TODO")
 }
 
+func (a *Address_I) ToBytes() util.Bytes {
```
jzimmerman: `addr.Serialize_Address_Compact` should be used here, no need for a new method I think.

src/systems/filecoin_vm/actor/address/address.go
```diff
@@ -55,6 +55,11 @@ func (a *Address_I) IsKeyType() bool {
...
+	ret := make([]byte, 0)
+	return ret
```
jzimmerman: (see above -- would be much better if there were a conspicuous panic or util.TODO here)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
sternhenri: how should the actor behave if there is no corresponding sectorNo? `_getUtilizationInfo` uses runtime to handle this

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
sternhenri: sure will fix. THough we should check all TODOs

sternhenri: @porcuquine have a sense of where I might get the SectorNumber given the PostVerifyInfo?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
sternhenri: I don't understand. Can you expand? Have a sense of how I can get it? Likewise for the sister metehod in storage_miner_subsystem which isn't quite right

src/algorithms/post/election_post.md
sternhenri: I don't believe so. E.g. we might get 1/25 as sectors sampled or num = 1, denom = 25. We want to increase likelihood by the inverse, as in multiply left side by denom. 

So It's not nit but it is correct i think

jzimmerman: I don't think that's right -- multiplying the left hand side by 25 would make it 25x _less_ likely for any individual sample to hit the target.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
jzimmerman: `addr.StoragePowerActorAddr` is just a constant in the addr package, which is what powerAddr is intended to be here (if I understand correctly?)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
jzimmerman: Ah, if failure is possible, then I think `_getCurrUtilization` should return `(block.StoragePower, error)`, and the caller must check the error. If the caller is actor code, it can then propagate the error via rt.Abort; if it's system code, then panic.

src/algorithms/post/election_post.md
sternhenri: yes indeed. thanks

src/systems/filecoin_blockchain/storage_power_consensus/expected_consensus.go
```diff
@@ -32,22 +35,29 @@ func (self *ExpectedConsensus_I) IsValidConsensusFault(faults ConsensusFaultType
...
+	rhs := new(big.Int).Lsh(util.BigFromUint64(uint64(sectorPower)), uint(SHA256Len))
+	rhs = rhs.Mul(rhs, util.BigFromUint64(uint64(sampleDenom)))
+	rhs = rhs.Mul(rhs, big.NewInt(int64(self.expectedBlocksPerEpoch())))
+
+	// lhs < rhs?
+	return lhs.Cmp(rhs) == -1
 }
 
-func (self *ExpectedConsensus_I) GetBlockRewards(electionProof block.ElectionProof, minerPower block.StoragePower, totalPower block.StoragePower) util.UVarint {
+func (self *ExpectedConsensus_I) GetBlockRewards(ePoStInfo sector.OnChainPoStVerifyInfo) util.UVarint {
```
zixuanzh: this method might need to be moved else where.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
zixuanzh: @sternhenri ok, then maybe we should change `_getUtilizaitonInfo` to `_safeGetUtilizationInfo` and make it call `_getCurrUtilization` but `_getCurrUtilization` will probably need to change to `_getUtilizationInfo`?

PR #684, opened 2019-11-30
[workaround] Temporary PR to merge #675 with correct branch

jzimmerman:


src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -616,7 +608,15 @@ func (a *StorageMinerActorCode_I) ProveCommitSector(rt Runtime, info sector.Sect
 		// expired
 		delete(st.PreCommittedSectors(), preCommitSector.Info().SectorNumber())
 		UpdateRelease(rt, h, st)
-		return rt.ErrorReturn(exitcode.UserDefinedError(0)) // TODO: user dfined error code?
+
+		TODO()
+		// TODO: if we abort, the above state changes (delete...) will not be committed.
+		// Is this desired? (alternative: support method failure while still committing
+		// state changes if UpdateRelease has already been called)
```
anorth: @zixuanzh can you comment on the necessity or otherwise of this cleanup code? Can/will it be cleaned up somewhere else if not here? I think it should be the same whether a miner submits a too-late proof or withholds it indefinitely.

src/systems/filecoin_vm/runtime/exitcode/vm_exitcodes.go
```diff
@@ -26,38 +31,45 @@ const (
...
-var (
-	InvalidSectorPacking = UserDefinedError(1)
+const (
+	InsufficientFunds_User = UserDefinedErrorCode_Placeholder + iota
+	InvalidArguments_User
+	InconsistentState_User
+
+	InvalidSectorPacking
+	SealVerificationFailed
+	DeadlineExceeded
```
anorth: As mentioned elsewhere, I think that all actor-specified error codes should move into the actors, and the system errors be ideally never referenced from actor code (only from within the VM internals). That can wait for a follow-up though and I'm aware of the APIError problem.

src/systems/filecoin_vm/sysactors/init_actor.go
```diff
@@ -133,45 +133,38 @@ func _codeIDSupportsExec(codeID actor.CodeID) bool {
...
-		if err != nil {
-			return argError
-		}
+		_checkArgs(err == nil)
 		params = params[1:]
 
-		if len(params) != 0 {
-			return argError
-		}
+		_checkArgs(len(params) == 0)
```
anorth: I don't get this construction - it's unnecessarily complicated. GetActorIDForAddress takes a single parameter. Just assert `len(params) == 1` and be done!

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
zixuanzh: ProveCommit should fail when it arrives later than MAX_PROVE_COMMIT. I agree that garbage collection (together with some slashing for missing MAX_PROVE_COMMIT) should happen whether ProveCommit arrives too late or it never happens but I am not sure where and how we should do this. We also need to clean up deals in these PreCommitted sectors.

jzimmerman: Presumably some amount of cleanup will need to happen via cron anyway, if miners withhold proofs indefinitely, so maybe we can add a TODO in cron to implement this (and any other) cleanup there, and remove the cleanup here?

src/systems/filecoin_vm/sysactors/init_actor.go
jzimmerman: The goal is that by following a standard pattern, even if it's slightly more verbose, the code can be less susceptible to errors when argument specs change over time. I've just updated the PR refactoring this logic out, though, so hopefully will be much cleaner.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
jzimmerman: (Just updated PR with this change.)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -523,12 +527,13 @@ func (a *StorageMinerActorCode_I) _verifySeal(rt Runtime, onChainInfo sector.OnC
 	info := st.Info()
 	sectorSize := info.SectorSize()
 	dealIDs := onChainInfo.DealIDs()
-	params := make([]actor.MethodParam, 1+len(dealIDs))
+	params := make([]util.Serialization, 1+len(dealIDs))
```
zixuanzh: not sure if this is right, since dealIDs should be one argument.

PR #685, opened 2019-11-30
[workaround] Merge #676 (gas accounting) with correct branch

### LINKED ISSUE #631: Accounting for gas consumption

jzimmerman:
Adds model for gas accounting. Closes #631.

### DIFF
--- a/src/systems/filecoin_vm/interpreter/vm_interpreter.go
+++ b/src/systems/filecoin_vm/interpreter/vm_interpreter.go
@@ -2,6 +2,7 @@ package interpreter
 
 import (
 	ipld "github.com/filecoin-project/specs/libraries/ipld"
+	storage_mining "github.com/filecoin-project/specs/systems/filecoin_mining/storage_mining"
 	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
 	addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
 	msg "github.com/filecoin-project/specs/systems/filecoin_vm/message"
@@ -14,14 +15,15 @@ import (
 
 type Bytes = util.Bytes
 
+var Assert = util.Assert
 var TODO = util.TODO
 var IMPL_FINISH = util.IMPL_FINISH
 
+type SenderResolveSpec int
+
 const (
-	// TODO: reduce these when expected gas costs are known
-	sumbitElectionPostGasLimit = 1e18
-	payBlockRewardGasLimit     = 1e1
-	cronTickGasLimit           = 1e18
+	SenderResolveSpec_OK SenderResolveSpec = 1 + iota
+	SenderResolveSpec_Invalid
 )
 
 // Applies all the message in a tipset, along with implicit block- and tipset-specific state
@@ -29,21 +31,17 @@ const (
 func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSetMessages) (outTree st.StateTree, receipts []vmr.MessageReceipt) {
 	outTree = inTree
 	seenMsgs := make(map[ipld.CID]struct{}) // CIDs of messages already seen once.
-	var r vmr.MessageReceipt
+	var receipt vmr.MessageReceipt
+
 	for _, blk := range msgs.Blocks() {
-		// Pay block reward.
-		reward := _makeBlockRewardMessage(outTree, blk.Miner())
-		outTree, r = vmi.ApplyMessage(outTree, reward, blk.Miner())
-		if r.ExitCode() != exitcode.OK() {
-			panic("block reward failed")
-		}
+		minerOwner := storage_mining.GetMinerOwnerAddress_Assert(inTree, blk.Miner())
 
 		// Process block miner's Election PoSt.
-		epost := _makeElectionPoStMessage(outTree, blk.Miner(), msgs.Epoch(), blk.PoStProof())
-		outTree, r = vmi.ApplyMessage(outTree, epost, blk.Miner())
-		if r.ExitCode() != exitcode.OK() {
-			panic("election post failed")
-		}
+		epostMessage := _makeElectionPoStMessage(outTree, blk.Miner(), msgs.Epoch(), blk.PoStProof())
+		outTree = _applyMessageBuiltinAssert(outTree, epostMessage, blk.Miner())
+
+		minerPenaltyTotal := actor.TokenAmount(0)
+		var minerPenaltyCurr actor.TokenAmount
 
 		// Process BLS messages from the block.
 		for _, m := range blk.BLSMessages() {
@@ -51,10 +49,13 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
 			if found {
 				continue
 			}
-			outTree, r = vmi.ApplyMessage(outTree, m, blk.Miner())
-			receipts = append(receipts, r)
+			onChainMessageLen := len(msg.Serialize_UnsignedMessage(m))
+			outTree, receipt, minerPenaltyCurr = vmi.ApplyMessage(outTree, m, onChainMessageLen, blk.Miner())
+			minerPenaltyTotal += minerPenaltyCurr
+			receipts = append(receipts, receipt)
 			seenMsgs[_msgCID(m)] = struct{}{}
 		}
+
 		// Process SECP messages from the block.
 		for _, sm := range blk.SECPMessages() {
 			m := sm.Message()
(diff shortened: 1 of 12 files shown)


anorth:
This PR has two commits from #684, I'll review only the new commit (gas accounting)

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -50,71 +51,116 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
 			if found {
 				continue
 			}
-			outTree, r = vmi.ApplyMessage(outTree, m, blk.Miner())
+
+			TODO() // TODO: include signature length?
+			onChainMessageLen := len(msg.Serialize_UnsignedMessage(m))
```
anorth: Now that #669 has merged, please you can discharch this TODO and measure the BLS and SECP message separately.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -123,51 +169,48 @@ func (vmi *VMInterpreter_I) ApplyMessage(inTree st.StateTree, message msg.Unsign
...
-		minerAddr, // TODO: may not exist
-		_gasToFIL(sendRet.GasUsed(), message.GasPrice()),
+		message.From(),
+		_gasToFIL(refundGas, message.GasPrice()),
 	)
 
-	return outTree, sendRet
+	return _applyReturn(
+		compTreeRet, vmr.InvocOutput_Make(sendRet.ReturnValue()), sendRet.ExitCode(),
+		addr.BurntFundsActorAddr, minerAddr)
```
anorth: Blocking: the miner address is the wrong place to send funds. The fees go to the miner actor's owner address (along with the block reward).

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -123,51 +169,48 @@ func (vmi *VMInterpreter_I) ApplyMessage(inTree st.StateTree, message msg.Unsign
...
-	// If nonzero, there is not guaranteed to be a nonzero gas balance from which to deduct it.
-	gasUsed := gascost.ApplyMessageFail
-	TODO()
-	return vmr.MessageReceipt_MakeSystemError(errCode, gasUsed)
+func _stateTreeWithGasUsedFundsTransfer(
+	tree st.StateTree, gasUsed msg.GasAmount, gasPrice msg.GasPrice,
+	fundsFrom addr.Address, fundsTo addr.Address) st.StateTree {
+
+	TODO() // TODO: what if the miner has insufficient funds?
+	// Should we instead aggregate these deltas, and SubtractWhileNonnegative at the end of the tipset?
```
anorth: Blocking: please resolve this in this PR.

Note that the miner (owner) will just have been paid the block reward before processing any other messages, so can be penalised at least up to that amount. But this code should still define a state transition that will succeed.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -50,71 +51,116 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
...
 
 	// WithActorForAddress may create new account actors
-	compTree, toActor = compTree.Impl().WithActorForAddress(message.To())
+	compTreePreSend := inTree
+	compTreePreSend, toActor := compTreePreSend.Impl().WithActorForAddress(message.To())
 	if toActor == nil {
-		return inTree, _applyError(exitcode.ActorNotFound)
+		// Execution error; receiver actor does not exist (and could not be implicitly created)
+		// at time of message execution.
+		return _applyError(inTree, exitcode.ActorNotFound, message.From(), addr.BurntFundsActorAddr)
```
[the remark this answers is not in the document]
jzimmerman: Yes, the intent is for the sender to pay here, since the sender should know in advance whether the receiver they're trying to send to exists.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -24,162 +25,252 @@ const (
...
+
+	_vmiUseGas := func(amount msg.GasAmount) (vmiUseGasOK bool) {
+		vmiGasRemaining, vmiUseGasOK = vmiGasRemaining.SubtractWhileNonnegative(amount)
+		return
+	}
+
+	ok := _vmiUseGas(gascost.OnChainMessage(onChainMessageSize))
+	if !ok {
+		// Invalid message; insufficient gas limit to pay for the on-chain message size.
+		_applyError(inTree, exitcode.OutOfGas, nil, nil, MinerPenaltySpec_ChargeUsedGas)
```
anorth: This will result in zero penalty because used gas is zero (b/c SubtractWhileNonNegative failed). The miner needs to pay more than the gas limit here: the actual message size cost.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -24,162 +25,252 @@ const (
...
+
+	vmiGasRemaining := message.GasLimit()
 
-	compTree := inTree
-	var outTree st.StateTree
-	var toActor actor.ActorState
-	var err error
+	_applyReturn := func(
+		tree st.StateTree, invocOutput vmr.InvocOutput, exitCode exitcode.ExitCode,
+		gasUsedFrom *addr.Address, gasUsedTo *addr.Address, minerPenaltySpec MinerPenaltySpec) {
```
anorth: Nit: This method is either called with final parameters `(..., nil, nil, MinerPenaltySpec_ChargeUsedGas)` or non-nils and `MinerPenaltySpec_ChargeUsedGas`, so it would be less noisy and easier to get right if there were separate methods and no penalty spec type.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -24,162 +25,252 @@ const (
...
+		panic("Interpreter error: runtime execution used more gas than provided")
+	}
+
+	ok = _vmiUseGas(gascost.OnChainReturnValue(sendRet.ReturnValue()))
+	if !ok {
+		// In compTreePreSend payment for the gas limit has been transferred to burnt funds actor.
+		// When the return value serialization overflows the gas limit, pay the full balance to the miner,
+		// refunding none to the sender (even though the offending return-value won't go on-chain), and then fail.
+		_applyError(compTreePreSend, exitcode.OutOfGas,
+			addr.BurntFundsActorAddr.Ref(), minerAddr.Ref(), MinerPenaltySpec_None)
```
anorth: This still isn't quite right. `compTreePreSend` has the sender paying gas, but the CallSeqNum is updated only in `compTreePostSend`. These two things should come together: the actor pays gas if and only if its CallSeqNum is updated.

I think this is made difficult by the CallSeqNum bump happening inside `_applyMessageInternal`: it would be easier to get right if it was pulled up here to line 192 so that both state changes happened together, or not at all.

Alternatively, the contract for _applyMessageInternal should be to return the input tree if the exit code is nonzero, and then there'll be no need to hold on to the pre-send one here, or the exit code check below.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -24,162 +25,252 @@ const (
...
+
+	_applyReturn(
+		compTreeRet, vmr.InvocOutput_Make(sendRet.ReturnValue()), sendRet.ExitCode(),
+		addr.BurntFundsActorAddr.Ref(), minerAddr.Ref(), MinerPenaltySpec_None)
+	return
+}
+
+func _applyMessageBuiltinAssert(tree st.StateTree, message msg.UnsignedMessage, topLevelBlockWinner addr.Address) st.StateTree {
+	gasRemainingInit := msg.GasAmount_SentinelUnlimited()
+	Assert(gasRemainingInit.Equals(message.GasLimit()))
```
anorth: Just use message.GasLimit? I don't understand why a sentinel or "unlimited" is involved, this seems to be duplicating a concept that is already expressed at the builtin message construction.

src/systems/filecoin_vm/message/message.go
```diff
@@ -63,18 +63,57 @@ func Verify(message SignedMessage, publicKey filcrypto.PublicKey) (UnsignedMessa
...
+}
+
+func GasAmount_FromInt(x int) GasAmount {
+	IMPL_FINISH()
+	panic("")
+}
+
+func GasAmount_SentinelUnlimited() GasAmount {
+	// Amount of gas larger than any feasible execution; meant to indicated unlimited gas
+	// (e.g., for builtin system method invocations).
```
anorth: I don't think we need this. We should set a real limit based on expected max cost, placeholders for which are already in the interpreter.

src/systems/filecoin_vm/state_tree/state_tree.go
```diff
@@ -30,6 +33,18 @@ func (st *StateTree_I) GetActorState(a addr.Address) actor.ActorState {
 	panic("TODO")
 }
 
+// Get the owner account address associated to a given miner actor.
+func (st *StateTree_I) GetMinerOwnerAddress(a addr.Address) (addr.Address, error) {
```
anorth: Convenient as this looks, it's a huge dependency inversion to have the state tree depending on the miner actor substate. It's not obvious because the absent implementation means you didn't have to do an `import`. This method sends the wrong signal about the dependency relationships.

Please don't commit this: add a helper function (unimplemented) in the interpreter instead (which still has a circular dep problem, but at least at a less deep level). See the existing commented-out calls to `loadActorSubstate`.

src/systems/filecoin_vm/state_tree/state_tree.id
```diff
@@ -25,6 +25,10 @@ type StateTree struct {
 
     GetActorState(a addr.Address) actor.ActorState
 
+    // Get the owner account address associated to a given miner actor.
+    GetMinerOwnerAddress(a addr.Address) (addr.Address, error)
+    GetMinerOwnerAddress_Assert(a addr.Address) addr.Address
```
anorth: See note above: these don't belong here.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
jzimmerman: This is due to the fact that it's a system invocation and we should never allow it to run out of gas (moreover, it's not clear who we would charge the gas to, if it did).

src/systems/filecoin_vm/message/message.go
jzimmerman: (see above)

src/systems/filecoin_vm/state_tree/state_tree.go
jzimmerman: The Go dependency graph is a bit difficult to resolve here, but I've tentatively moved it to StorageMinerActor for now.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -14,172 +15,255 @@ import (
...
+
+		switch senderResolveSpec {
+		case SenderResolveSpec_OK:
+			// In this case, the sender is valid and has already transferred funds to the burnt funds actor
+			// sufficient for the gas limit. Thus, we may refund the unused gas funds to the sender here.
+			Assert(!message.GasLimit().LessThan(vmiGasUsed))
+			Assert(message.GasLimit().Equals(vmiGasUsed.Add(vmiGasRemaining)))
+			vmiGasRemainingFIL := _gasToFIL(vmiGasRemaining, message.GasPrice())
+			tree = _withTransferFundsAssert(tree, addr.BurntFundsActorAddr, message.From(), vmiGasRemainingFIL)
+			retMinerPenalty = actor.TokenAmount(0)
```
anorth: The miner hasn't been paid!

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -14,172 +15,255 @@ import (
...
+		return
+	}
+
+	// Deduct an amount of gas corresponding to costs already incurred, and for which the
+	// gas cost must be paid even if it would cause the gas used to exceed the limit.
+	_vmiBurnGas := func(amount msg.GasAmount) (vmiBurnGasOK bool) {
+		vmiGasUsedPre := vmiGasUsed
+		vmiBurnGasOK = _vmiAllocGas(amount)
+		if !vmiBurnGasOK {
+			Assert(vmiGasRemaining.Equals(msg.GasAmount_Zero()))
```
anorth: That's not how `SubtractIfNonnegative` works. `vmiGasRemaining` will be its prior value, not zero.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
anorth: Hard disagree. We have to put an actual number into the real computer! There should be a real limit and our blockchain should halt if it's exceeded (it's gunna halt anyway), and we can fix it at the social consensus layer of humans.

Please pick a number after which "unlimited" runs out, preferably by restoring the constants I had at the top of this file. It can be large enough to be practically "never", but you can't IMPL_TODO explicitly dodging this.

src/systems/filecoin_vm/sysactors/init_actor.go
```diff
@@ -72,6 +72,8 @@ func (a *InitActorCode_I) Exec(rt Runtime, codeID actor.CodeID, constructorParam
 		rt.AbortArgMsg("cannot exec an actor of this type")
 	}
 
+	rt.CreateActor_DeductGas()
```
anorth: It already calls `rt.CreateActor`: move the charge there and simplify the Runtime interface.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
jzimmerman: Yes, that's a bug -- just fixed. (I had missed this earlier since it's still very counterintuitive to me that a miner causes the network to incur costs here, and then gets paid for doing so!)

mishmosh:
@jzimmerman saw that you pushed some changes. Does that address all of @anorth's comments so far; is it ready for re-review? (Please explicitly add a comment pinging your reviewer(s) in the future.)

anorth:
To follow up:
- update the prose for tipset message execution to match the new ordering of block reward last and garnishing of penalty gas fees
- restore actual numbers for implicit message gas limits
- maybe try to restructure for legibility.

PR #688, opened 2019-12-02
Message description and validation text.

anorth:
Cleaned up the UnsignedMessage to more closely align with the existing implementations.
Added prose for message syntactic validity, and interpreter notes about condition that
will cause execution to fail.

Reviewers: please take a moment to think about what's _not_ written but should be.

@jzimmerman please check the notes on the interpreter. I didn't touch the code so as not to clash with your #676, but we should align them.

### DIFF
--- a/src/systems/filecoin_vm/interpreter/_index.md
+++ b/src/systems/filecoin_vm/interpreter/_index.md
@@ -50,6 +50,40 @@ The sequence of executions for a tipset is thus summarised:
 - [... subsequent blocks ...]
 - cron tick 
 
+# Message validity and failure
+Every message in a valid block can be processed and produce a receipt (note that block validity
+implies all messages are syntactically valid and correctly signed). However, execution may
+or may not succeed, depending on the state to which the message is applied. If the execution
+of a message fails, the corresponding receipt will carry a non-zero exit code. 
+
+If a message fails due to a reason that can reasonably be attributed to the miner including a
+message that could never have succeeded in the parent state, or because the sender lacks funds
+to cover the maximum message cost, then the miner pays a penalty by burning the gas fee 
+(rather than the sender paying fees to the block miner).
+
+The only state changes resulting from a message failure are either:
+
+- incrementing of the sending actor's `CallSeqNum`, and payment of gas fees from the sender to the owner of the miner of the block including the message; or
+- a penalty equivalent to the gas fee for the failed message, burnt by the miner (sender's `CallSeqNum` unchanged).
+ 
+A message execution will fail if, in the immediately preceding state:
+
+- the `From` actor does not exist in the state (miner penalized),
+- the `From` actor is not an account actor (miner penalized),
+- the `CallSeqNum` of the message does not match the `CallSeqNum` of the `From` actor (miner penalized),
+- the `To` actor does not exist in state and the `To` address is not a pubkey-style address (miner penalized),
+- the `To` actor does not exist in state and the message has a non-zero `MethodNum` (miner penalized),
+- the `To` actor exists but does not have a method corresponding to the non-zero `MethodNum`,
+- deserialized `Params` is not an array of length matching the arity of the `To` actor's `MethodNum` method,
+- deserialized `Params` are not valid for the types specified by the `To` actor's `MethodNum` method,
+- the `From` actor does not have sufficient balance to cover the sum of the message `Value` plus the
+maximum gas cost, `GasLimit * GasPrice` (miner penalized),
+- the invoked method consumes more gas than the `GasLimit` allows, or
+- the invoked method exits with a non-zero code (via `Runtime.Abort()`).
+
+Note that if the `To` actor does not exist in state and the address is a valid `H(pubkey)` address, 
+it will be created as an account actor (only if the message has a MethodNum of zero).
+
 (You can see the _old_ VM interpreter [here](docs/systems/filecoin_vm/vm_interpreter_old) )
 
 # `vm/interpreter` interface
--- a/src/systems/filecoin_vm/message/_index.md
+++ b/src/systems/filecoin_vm/message/_index.md
@@ -4,6 +4,64 @@ statusIcon: 🔁
 title: VM Message - Actor Method Invocation
 ---
 
+A message is the unit of communication between two actors, and thus the primitive cause of changes
+in state. A message combines:
+
+- a token amount to be transferred from the sender to the receiver, and
+- a method with parameters to be invoked on the receiver (optional).
+
+Actor code may send additional messages to other actors while processing a received message. 
+Messages are processed synchronously: an actor waits for a sent message to complete before resuming control.
+
+The processing of a message consumes units of computation and storage denominated in gas. 
+A message's gas limit provides an upper bound on its computation. The sender of a message pays 
+for the gas units consumed by a message's execution (including all nested messages) at a
+gas price they determine. A block producer chooses which messages to include in a block and is
+rewarded according to each message's gas price and consumption, forming a market.
+
+# Message syntax validation
+
+A syntactically invalid message must not be transmitted, retained in a message pool, or
+included in a block.
+
+A syntactically valid `UnsignedMessage`:
+
+- has a well-formed, non-empty `To` address,
+- has a well-formed, non-empty `From` address, 
+- has a non-negative `CallSeqNum`,
+- has `Value` no less than zero and no greater than the total token supply (`2e9 * 1e18`), and
+- has a non-negative `MethodNum`,
+- has non-empty `Params` only if `MethodNum` is zero,
+- has non-negative `GasPrice`,
+- has `GasLimit` that is at least equal to the gas consumption associated with the message's serialized bytes,
+- has `GasLimit` that is no greater than the block gas limit network parameter
+- has a total serialized size no greater than 32 KiB.
+
+# Message semantic validation
+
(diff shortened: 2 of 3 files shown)


src/systems/filecoin_vm/interpreter/_index.md
```diff
@@ -50,6 +50,36 @@ The sequence of executions for a tipset is thus summarised:
...
+ 
+The sender pays the gas fee to the miner if and only if sending actor's `CallSeqNum` is incremented.
+Otherwise, the miner pays (burns) the gas fee as penalty for including the failing message in a block.
+ 
+A message execution will fail if:
+
+- the `From` actor does not exist in the state (miner penalized),
+- the `From` actor is not an account actor (miner penalized),
+- the `CallSeqNum` of the message does not match the `CallSeqNum` of the `From` actor (miner penalized),
+- the `To` actor does not exist in state and the `To` address is not an pubkey-style address (miner penalized),
```
sternhenri: nit: typo `a pubkey`

src/systems/filecoin_vm/interpreter/_index.md
```diff
@@ -50,6 +50,36 @@ The sequence of executions for a tipset is thus summarised:
...
+- the `To` actor does not exist in state and the message has a non-zero `Method` (miner penalized),
```
sternhenri: worth mentioning what the zero method does?

src/systems/filecoin_vm/interpreter/_index.md
```diff
@@ -50,6 +50,36 @@ The sequence of executions for a tipset is thus summarised:
...
+- deserialized `Params` is not an array of length matching the arity of the `To` actor's `MethodNum` method,
+- deserialized `Params` are not valid for the types specified by the `To` actor's `MethodNum` method,
+- the `From` actor does not have sufficient balance to cover the sum of the message `Value` plus the
+maximum gas cost, `GasLimit * GasPrice` (miner penalized),
+- the invoked method consumes more gas than the `GasLimit` allows, or
+- the invoked method exits with a non-zero code (via `Runtime.Abort()`).
+
+Note that if the `To` actor does not exist in state and the address is a valid `H(pubkey)` address, 
+it will be created as an account actor (only if the message has a MethodNum of zero).
+
```
sternhenri: also worth giving a heuristic at the top of when the miner is penalized vs not. Looks a bit random without that explanation up front.

src/systems/filecoin_vm/interpreter/_index.md
```diff
@@ -50,6 +50,36 @@ The sequence of executions for a tipset is thus summarised:
...
 (You can see the _old_ VM interpreter [here](docs/systems/filecoin_vm/vm_interpreter_old) )
```
sternhenri: is it useful?

src/systems/filecoin_vm/message/_index.md
```diff
@@ -4,6 +4,47 @@ statusIcon: 🔁
 title: VM Message - Actor Method Invocation
 ---
 
+A message is the unit of communication between two actors, and thus the primitive cause of changes
+in state. A message combines:
+
+- a token amount to be transferred to be transferred from the sender to the receiver, and
```
sternhenri: dup

src/systems/filecoin_vm/message/_index.md
```diff
@@ -4,6 +4,47 @@ statusIcon: 🔁
...
+- a method with parameters to be invoked on the receiver (optional).
+
+Actor code may send additional messages to other actors while processing a received message. 
+Messages are processed synchronously: an actor waits for a sent message to complete before resuming control.
+
+The processing of a message consumes units of computation and storage denominated in gas. 
+A message's gas limit provides an upper bound on its computation. The sender of a message pays 
+for the gas units consumed by a message's execution (including all nested messages) at a
+gas price they determine.
+
```
sternhenri: maybe add a note on how miners choosing which messages to include in their blocks creates an effective market for gas price (out of scope here but the `they determine` begs the question)

src/systems/filecoin_vm/interpreter/_index.md
```diff
@@ -50,6 +50,36 @@ The sequence of executions for a tipset is thus summarised:
 - [... subsequent blocks ...]
 - cron tick 
 
+# Message validity and failure
+Every syntactically valid message can be processed and produce a receipt. However, execution may
```
sternhenri: worth linking to message/index.id after `syntactically valid` so readers can quickly move between syntactic and semantic validity

src/systems/filecoin_vm/message/_index.md
```diff
@@ -4,6 +4,47 @@ statusIcon: 🔁
...
+# Message validation
+
+A syntactically invalid message should not be transmitted, retained in the message pool, or
+included in a block.
+
+A syntactically valid message:
+
+- has a well-formed, non-empty `To` address ,
+- has a well-formed, non-empty `From` address, 
+- has distinct `To` and `From ` addresses,
```
sternhenri: may be a dumb question but is that truly necessary? Could an actor never call a method on itself to update its own state?

src/systems/filecoin_vm/message/_index.md
```diff
@@ -4,6 +4,47 @@ statusIcon: 🔁
...
+- has a `CallSeqNum` no less than zero,
+- has `Value` no less than zero and no greater than the total token supply (`2e9 * 1e18`), and
+- has a `MethodNum` no less than zero,
+- has non-empty `Params` only if `MethodNum` is zero,
+- has `GasPrice` no less than zero,
```
sternhenri: can it be 0?

src/systems/filecoin_vm/message/_index.md
```diff
@@ -4,6 +4,47 @@ statusIcon: 🔁
...
+- has `GasLimit` that is at least equal to the gas consumption associated with the message's serialized bytes,
+- has `GasLimit` that is no greater than the block gas limit network parameter
+- has a total serialized size no greater than 32 KiB.
+
```
sternhenri: per the below:
- has a syntactically valid signature

src/systems/filecoin_vm/interpreter/_index.md
```diff
@@ -50,6 +50,36 @@ The sequence of executions for a tipset is thus summarised:
...
+- the `To` actor does not exist in state and the `To` address is not an pubkey-style address (miner penalized),
+- the `To` actor does not exist in state and the message has a non-zero `Method` (miner penalized),
+- the `To` actor does not have a method corresponding to `MethodNum`,
+- deserialized `Params` is not an array of length matching the arity of the `To` actor's `MethodNum` method,
+- deserialized `Params` are not valid for the types specified by the `To` actor's `MethodNum` method,
+- the `From` actor does not have sufficient balance to cover the sum of the message `Value` plus the
+maximum gas cost, `GasLimit * GasPrice` (miner penalized),
+- the invoked method consumes more gas than the `GasLimit` allows, or
+- the invoked method exits with a non-zero code (via `Runtime.Abort()`).
+
```
sternhenri: The message signature is not validly signed by the `From` actor's associated pk

src/systems/filecoin_vm/message/_index.md
```diff
@@ -4,6 +4,47 @@ statusIcon: 🔁
...
+A syntactically valid signed message must carry a signature that verifies the payload as having
+been signed by the public key associated with the message's `From` address.
+
+There is no semantic validation (i.e. requiring access to information outside the message itself) 
+of a message that can cause a block including the message to be invalid. Every syntactically valid
+message can be included in a block and will produce a receipt from execution. However, a message
+may fail to execute to completion, in which case it will not effect the desired state change.
```
sternhenri: the above is confusing for me: given the number of failures which cause miner penalization, i considered them semantic validation issues (e.g state issues). How do you distinguish the two? And is it fair to say "every syntactically valid message can be included in a block" when such inclusion could lead to punishment?

anorth: Your last statement: by definition!

I see what you're saying though. There might be room to simplify. The goal is to avoid any situation that requires executing the messages of an individual block outside the context of executing them as a tipset. The state to which a block's messages apply is _not known when the block is produced_. So my thinking was, if there exists some possible state to which the message would apply correctly, we can't invalidate a block for it.

Given the monotonic nature of actor creation, you might think the checks of the `To` and `From` actor validity could be moved here, and performed by a miner _against the parent state declared by the block_ (not the state the message will actually execute against) without actually executing the whole message. But that would preclude a perfectly reasonable sequence of first creating an actor and then sending a message to it in the same block. The semantic validity of the second message depends on the first, so you end up having to execute each block's messages in isolation as well as for the whole tipset, only to achieve an _approximation_ of semantic validity.

Our decision at present is to avoid the double-execution of messages. Even if we included it, we'd still have to handle non-monotonic cases like mismatched `CallSeqNum` and insufficient funds during tipset execution.

anorth:
> how message signing works for secp vs bls

That'll be in block validation. The `Message` type doesn't know about BLS. Maybe we should even move `SignedMessage` out to the block package.

src/systems/filecoin_vm/message/_index.md
sternhenri: understood. reinforces the need to define when miner is punished vs not (or that intent as explained here) imo, but makes sense to me

sternhenri:
I wouldn;'t move it out because it's where people would expect to find it...

PR #691, opened 2019-12-03
Integrate minimal usage of dynamic external orient transformations or Chung.

porcuquine:
This PR integrates Orient changes to support dynamic external transformations (https://github.com/filecoin-project/orient/pull/44).

I have minimally adapted @nikkolasg's preprocessing PR in order to demonstrate usage and ensure this works for our intended purpose. This satisfies the minimum requirement @nicola requested to unblock modeling with Chung.

I also changed script paths to use `/usr/bin/env` because I was getting errors on MacOS when testing.


porcuquine:
This was inadvertently created against `master` rather than `preprocessing`. I am closing in favor of #691, which corrects that error.

PR #692, opened 2019-12-03
Integrate minimal usage of dynamic external orient transformations for Chung

porcuquine:
This PR integrates Orient changes to support dynamic external transformations (https://github.com/filecoin-project/orient/pull/44).

I have minimally adapted @nikkolasg's preprocessing PR in order to demonstrate usage and ensure this works for our intended purpose. This satisfies the minimum requirement @nicola requested to unblock modeling with Chung.

I also changed script paths to use `/usr/bin/env` because I was getting errors on MacOS when testing.

This transcript shows the effect of these changes:

```
(base) ➜  specs git:(dynamic-chung) ✗ make orient
bin/solve-orient.sh src/orient/chung.orient src/orient/chung.json build/orient/chung.orient.json
(base) ➜  specs git:(dynamic-chung) ✗ cat build/orient/chung.orient.json
[
[{"alpha":65,"beta":0,"expander_degree":8}]
]
```

Fair warning: I will probably change this syntax as the interface for external engines improves and evolves over time. What you see here is a minimal consistent solution which handles what we need for Chung now.

### DIFF
--- a/bin/preprocessing/chung_spec.py
+++ b/bin/preprocessing/chung_spec.py
@@ -90,6 +90,7 @@ def main():
     jinput, alphaT,betaT, degreeT = parse()
     alpha = extract_value(jinput,alphaT)
     degree = extract_value(jinput,degreeT)
+
     if alpha is None or degree is None:
         # default behavior: return same thing if nothing to be done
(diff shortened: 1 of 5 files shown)


mishmosh:
@porcuquine is this still active?

PR #693, opened 2019-12-03
Clean up interplay between InitActor.Exec and Runtime.CreateActor.

anorth:
This change started out with an implementation-driven delta from @icorderi, which I then adjusted. The new arrangement clearly distinguishes:
1. Allocating an actor ID
1. Creating empty actor state in the state tree, keyed by that ID address
1. Calling the constructor

They have to happen in that order. @icorderi the snippet you sent me attempted to send to the constructor method by the actor address before the init actor had persisted the mapping to ID address.

### DIFF
--- a/src/systems/filecoin_vm/runtime/runtime.go
+++ b/src/systems/filecoin_vm/runtime/runtime.go
@@ -148,26 +148,24 @@ func (rt *VMContext) _rtAllocGasCreateActor() {
 	rt._rtAllocGas(gascost.ExecNewActor)
 }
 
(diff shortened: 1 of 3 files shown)


anorth:
Edit: I plan to pull NewActorAddress out of the runtime, it doesn't need to be privileged.

anorth:
Update: no I don't. `NewActorAddress` needs to stay on the runtime, otherwise we have to expose the top-level and immediate caller seq nums etc: this gives a more abstract runtime interface.

PR #694, opened 2019-12-03
clean block header files

sternhenri:
**Done:**
- Remove conflict between files
- consolidate within a single one

**TODO:**
- clean up the rest of the data structures listing nightmare of staleness


src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -7,25 +7,24 @@ import addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address
...
-type MessageRoot ipld.CID
-type ReceiptRoot ipld.CID
 type ChainWeight UVarint
 type ChainEpoch UVarint
+type MessageReceipt util.Bytes
 
 // On-chain representation of a block header.
 type BlockHeader struct {
     // Chain linking
     Parents          Tipset
```
anorth: This should be `[&BlockHeader]`. There's a rabbit hole here because some existing code uses this wrongly.

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -7,25 +7,24 @@ import addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address
...
     Epoch            ChainEpoch
 
     // Miner Info
     MinerAddress     addr.Address
 
     // State
-    StateTree        st.StateTree
-    Messages         MessageRoot
-    MessageReceipts  ReceiptRoot
+    StateTree        st.StateTree
```
anorth: Move the StateTree and MessageRecipts up next to the parents/weight, and document it as the state tree after processing of the parent tipset.

The Messages are proposed messages for this epoch.

anorth: Move BLSAggregateSig next to Messages

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -7,25 +7,24 @@ import addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address
...
     // State
-    StateTree        st.StateTree
-    Messages         MessageRoot
-    MessageReceipts  ReceiptRoot
+    StateTree        st.StateTree    
+    Messages         &TxMeta
+    MessageReceipts  [&MessageReceipt]
 
     // Consensus things
     Timestamp        clock.UnixTime
```
anorth: I suggest moving Epoch down near these (or these up to the top).

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -40,11 +39,15 @@ type BlockHeader struct {
     //	ComputeUnsignedFingerprint() []
 }
 
+type TxMeta struct {
+    BLSMessages [msg.UnsignedMessage]
+    SECPMessages [msg.SignedMessage]
```
anorth: These are both AMTs, the values of which are the CIDs of the messages. I'm not sure how best to represent that here, maybe `&[&msg.UnsignedMessage] // array-mapped trie`

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -40,11 +39,15 @@ type BlockHeader struct {
...
+}
+
 // Internal representation of a full block, with all messages.
 type Block struct {
     Header        BlockHeader
-    BLSMessages   [msg.UnsignedMessage]
-    SECPMessages  [msg.SignedMessage]
+    Messages      TxMeta
```
anorth: I suggest reverting this one back: it's a more convenient internal representation.

src/listings/data_structures.md
```diff
@@ -2,197 +2,8 @@
...
-A type to represent an amount of Filecoin tokens.
-{{<goFile TokenAmount>}}
-
-
-## SectorID
-Uniquely identifies a miner's sector.
-{{<goFile SectorID>}}
-
-
 # RLE+ Bitset Encoding
```
sternhenri: @anorth @jzimmerman I left the following below because I do not know whether still relevant to the protocol and/or captured elsewhere in the spec (when needed). If you know for a fact that some of these are stale, please remove before merge.

src/systems/filecoin_blockchain/struct/block/block.id
jzimmerman: `&[&msg.UnsignedMessage]` seems right to me, but as @anorth points out above, this may be a rabbit hole we don't want to deal with yet -- there are many places we currently use type T where we really mean &T, and it might be easier to address all of those in a single subsequent PR.

sternhenri:
Can I delete the whole of data-structures.md?

src/systems/filecoin_blockchain/struct/block/block.id
sternhenri: As you say, we use a lot of the `Tipset` convenience methods from `Tipset.id` elsewhere. Should i go down the rabbit hole and fix throughout or leave as is for now?

src/listings/data_structures.md
zixuanzh: We use RLE+ bitfields quite extensively and it looks right but not sure about the rest. The listing section is supposed to link concepts and terms back to different parts of the spec where they are defined. We can remove most of them here and populate this section later when the spec stabilizes.

anorth:
> Can I delete the whole of data-structures.md?

Maybe un-render, link to and rename to data-structures_old.md

src/systems/filecoin_blockchain/struct/block/block.id
anorth: Yes, please fix. I don't mind of those fixes involve deleting or disabling the downstream code that breaks. Implementers need this header structure to be exact and precise, but not so for other code.

src/systems/filecoin_blockchain/struct/block/block.id
anorth: This is a place where the distinction between an embedded and referenced object makes a big difference. Implementations disagree about whether some of these things are CIDs or embedded - the spec needs to be clear.

I used `&` as a shorthand for CID that I'd seen elsewhere, but just as happy with `[cid.CID] // AMT<&UnsignedMessage]> array-mapped trie`

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -7,34 +7,35 @@ import addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address
...
     Ticket
     ElectionProof
 
+    MinerAddress           addr.Address
+    // ElectionPoSt           OnChainPoStVerifyInfo
+
+    // Proposed update
+    BLSAggregateSig        filcrypto.Signature
+    BLSMessages            &[&msg.UnsignedMessage]  // array-mapped trie
+    SECPMessages           &[&msg.SignedMessage]
```
anorth: `// array-mapped trie`

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -43,8 +44,8 @@ type BlockHeader struct {
 // Internal representation of a full block, with all messages.
 type Block struct {
     Header        BlockHeader
-    BLSMessages   [msg.UnsignedMessage]
-    SECPMessages  [msg.SignedMessage]
+    BLSMessages   [&msg.UnsignedMessage]
+    SECPMessages  [&msg.SignedMessage]
```
anorth: These ones can be embedded, not CIDs.

src/systems/filecoin_blockchain/struct/block/block.id
sternhenri: Ok. done. I feel the cached tipset methods from `tipset.id` are trivial to understand with a tipset initialized with [&BlockHeader]. If you don't agree, I can implement them all and show a Tipset constructor to really hammer home the point... Unec. i think

src/systems/filecoin_blockchain/struct/block/block.id
sternhenri: That is the larger question. If & is a standin for CID that should be explicitly stated in the spec. I have added accordingly.

src/listings/data_structures.md
sternhenri: will leave as is for now then unless I hear otherwise.

sternhenri:
Almost there I think @anorth. 

2 Points.
- Renamed `BLSMessages` and `SECPMessages` in the BlockHeader to `BLSMessagesCIDs` and `SECPMessagesCIDs` respectively to disambiguate with those in the block itself.
- Check out the change to `CID.md` related to treatment of `&` references being interchangeable with CIDs. 
- Note my choice not to explicitly define the relationship between `[&BlockHeader]` and `Tipset` thinking most of the impl is self-evident. Lmk if disagree and I'll do a quick job of showing tipset construction and derivation of cached methods.

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -1,40 +1,39 @@
...
 
     // Consensus things
-    Timestamp        clock.UnixTime
+    Epoch                  ChainEpoch
+    Timestamp              clock.UnixTime
     Ticket
     ElectionProof
 
+    MinerAddress           addr.Address
+    // ElectionPoSt           OnChainPoStVerifyInfo
```
sternhenri: any read on how i can bring this in without creating a circular dependency with sector?

hugo/data/sref_locations.yaml
```diff
@@ -36,6 +36,7 @@ chain_sync: [systems, filecoin_blockchain, chainsync]
 block_validation: [systems, filecoin_blockchain, chainsync, block_validation]
 
 ### Message Pool
+message: [systems, filecoin_vm, message]
```
anorth: Move this up to line 10, matching it's location in TOC

src/systems/filecoin_blockchain/message_pool/message_syncer.md
```diff
@@ -11,7 +11,7 @@ TODO:
 
 # Message Propagation
 
-Messages are propagated over the libp2p pubsub channel `/fil/messages`. On this channel, every [serialised `SignedMessage`](data-structures.md#messages) is announced.
+Messages are propagated over the libp2p pubsub channel `/fil/messages`. On this channel, every serialised `SignedMessage` is announced (see {{<sref message>}}).
```
anorth: `message_syncer` is a better reference

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -1,40 +1,39 @@
...
-    // Miner Info
-    MinerAddress     addr.Address
-
+    Parents                [&BlockHeader]
+    ParentWeight           ChainWeight
     // State
-    StateTree        st.StateTree
-    Messages         MessageRoot
-    MessageReceipts  ReceiptRoot
+    ParentStateTree        st.StateTree
```
anorth: This is a CID too, so maybe `&st.StateTree`

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -1,40 +1,39 @@
...
+    ParentMessageReceipts  [&MessageReceipt]
```
anorth: This must be a reference to an array-mapped-trie structure too, not an embedded list.

src/systems/filecoin_blockchain/struct/block/block.id
anorth: I propose duplicating the structure here.

src/systems/filecoin_blockchain/message_pool/message_syncer.md
sternhenri: but this is the message_syncer file already so useless to self-link. leaving as is

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -1,40 +1,39 @@
...
     Ticket
     ElectionProof
 
+    MinerAddress           addr.Address
+    ElectionPoSt           OnChainPoStVerifyInfo
+
+    // Proposed update
+    BLSAggregateSig        filcrypto.Signature
+    BLSMessageCIDs         &[&msg.UnsignedMessage]  // array-mapped trie
+    SECPMessageCIDs        &[&msg.SignedMessage]  // array-mapped trie
```
anorth: Blocking: The `TxMeta` structure from the data_structures file exists in both implementations and was probably put there for a reason, probably header size reduction. Please obtain @whyrusleeping's approval before removing it (or just put it back).

PR #696, opened 2019-12-03
Miner actor cleanup

zixuanzh:
**In this PR:**
- [x] Add Send calls to `StorageMinerActorCode`
- [x] Remove Send comments from `StorageMinerActorState`
- ~~[x] Add `ProcessDealPayment` to `_onSuccessfulPoSt` and add flag to `_getActivePower` (deal payment is only triggered in `_onSuccessfulPoSt`)~~
- [x] Add `_claimDealPayments` and invoke `_claimDealPayments` at `_onSuccessfulPoSt`
- [x] Remove comment on depledge (we dont need this as a separate mechanism), miners' pledge collateral requirement gets updated based on the power they have and they can call `WithdrawBalance` from StoragePowerActor (depledge)
- [x] Add Serialization helpers for Send calls

### DIFF
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
@@ -106,7 +106,7 @@ type StorageMinerActorState struct {
     _updateSectorUtilization(rt Runtime, lastPoStResponse block.ChainEpoch) [deal.DealID]
     _initializeUtilizationInfo(rt Runtime, deals [deal.OnChainDeal]) sector.SectorUtilizationInfo
 
-    _getActivePower(rt Runtime) block.StoragePower
+    _getActivePower(rt Runtime, getActiveDealIDs bool) (activePower block.StoragePower, activeDealIDs [deal.DealID])
     _getInactivePower(rt Runtime) block.StoragePower
     _getUtilizationInfo(rt Runtime, sectorNo sector.SectorNumber) sector.SectorUtilizationInfo
     _getCurrUtilization(sectorNo sector.SectorNumber) (utilization block.StoragePower, found bool)
@@ -145,7 +145,7 @@ type StorageMinerActorCode struct {
         faultType      sector.StorageFaultType
     )
 
-    _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch)
+    _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch, getActiveDealIDs bool)
     _expirePreCommittedSectors(rt Runtime)
     _ensurePledgeCollateralSatisfied(rt Runtime)
 }
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
@@ -106,7 +106,7 @@ type StorageMinerActorState struct {
     _updateSectorUtilization(rt Runtime, lastPoStResponse block.ChainEpoch) [deal.DealID]
     _initializeUtilizationInfo(rt Runtime, deals [deal.OnChainDeal]) sector.SectorUtilizationInfo
 
-    _getActivePower(rt Runtime, getActiveDealIDs bool) (activePower block.StoragePower, activeDealIDs [deal.DealID])
+    _getActivePower(rt Runtime) block.StoragePower
     _getInactivePower(rt Runtime) block.StoragePower
     _getUtilizationInfo(rt Runtime, sectorNo sector.SectorNumber) sector.SectorUtilizationInfo
     _getCurrUtilization(sectorNo sector.SectorNumber) (utilization block.StoragePower, found bool)
@@ -122,8 +122,6 @@ type StorageMinerActorCode struct {
     SubmitVerifiedElectionPoSt(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo)
 
     _checkSurprisePoStSubmissionHappened(rt Runtime)
-
-    // TODO: should depledge be in here or in storage market actor?
 
     DeclareFaults(rt Runtime, failingSet sector.CompactSectorSet)
     RecoverFaults(rt Runtime, recoveringSet sector.CompactSectorSet)
@@ -145,6 +143,7 @@ type StorageMinerActorCode struct {
         faultType      sector.StorageFaultType
     )
 
+    _claimDealPayments(rt Runtime)
     _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch, getActiveDealIDs bool)
     _expirePreCommittedSectors(rt Runtime)
     _ensurePledgeCollateralSatisfied(rt Runtime)
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
@@ -121,7 +121,7 @@ type StorageMinerActorCode struct {
     SubmitVerifiedSurprisePoSt(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo)
     SubmitVerifiedElectionPoSt(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo)
 
-    CheckSurprisePoStSubmissionHappened(rt Runtime)
+    _checkSurprisePoStSubmissionHappened(rt Runtime)
 
     // TODO: should depledge be in here or in storage market actor?
 
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
@@ -165,7 +165,7 @@ type MinerInfo struct {
     PeerId                  libp2p.PeerID
 
     // Amount of space in each sector committed to the network by this miner.
-    SectorSize              util.BytesAmount
+    SectorSize              sector.SectorSize
     WindowCount             UVarint
     SealPartitions          UVarint
     ElectionPoStPartitions  UVarint
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
@@ -144,7 +144,7 @@ type StorageMinerActorCode struct {
     )
 
     _claimDealPayments(rt Runtime)
-    _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch, getActiveDealIDs bool)
+    _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch)
     _expirePreCommittedSectors(rt Runtime)
     _ensurePledgeCollateralSatisfied(rt Runtime)
 }
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
@@ -153,7 +153,7 @@ func (a *StorageMinerActorCode_I) _onMissedSurprisePoSt(rt Runtime) {
 
 	Release(rt, h, st)
 
-	a._submitPowerReport(rt, lastPoStResponse)
+	a._submitPowerReport(rt, lastPoStResponse, false)
 
 	// Note: NewDetectedFaults is now the sum of all
 	// previously active, committed, and recovering sectors minus expired ones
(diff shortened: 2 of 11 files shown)


src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -526,21 +530,22 @@ func (a *StorageMinerActorCode_I) _verifySeal(rt Runtime, onChainInfo sector.OnC
 	h, st := a.State(rt)
 	info := st.Info()
 	sectorSize := info.SectorSize()
-	dealIDs := onChainInfo.DealIDs()
-	params := make([]util.Serialization, 1+len(dealIDs))
+
+	// @param sectorSize util.UVarint
+	// @param dealIDs []deal.DealID onChainInfo.DealIDs()
+	getPieceInfoParams := make([]util.Serialization, 2)
 	panic("TODO: serialize arguments")
```
sternhenri: maybe can take care of above using the auto-generated serialization method

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -557,9 +562,26 @@ func (a *StorageMinerActorCode_I) _verifySeal(rt Runtime, onChainInfo sector.OnC
 		WindowCount_: info.WindowCount(),
 		Partitions_:  info.SealPartitions(),
 	}
+
+	// @param address addr.Address info.Worker()
+	getActorIDParams := make([]util.Serialization, 1)
```
sternhenri: ibid, actually use the serialization stub in gen.go file

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -557,9 +562,26 @@ func (a *StorageMinerActorCode_I) _verifySeal(rt Runtime, onChainInfo sector.OnC
...
+	getActorIDRet := getActorIDReceipt.ReturnValue()
+	minerID, err := addr.Deserialize_Address(getActorIDRet)
+	if err != nil {
+		rt.AbortStateMsg("sm.verifySeal: failed to get actor id")
+	}
+
 	svInfo := sector.SealVerifyInfo_I{
 		SectorID_: &sector.SectorID_I{
-			MinerID_: info.Worker(), // TODO: This is actually miner address. MinerID needs to be derived.
+			MinerID_: minerID,
```
sternhenri: are you sure we want the minerID, not workerinfo?

We should homogeonize throughout the spec whether we call it `minerID` or `minerAddr` both are used and mean same. Based on your choice, please change accordingly in the `keyStore` file

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -598,12 +620,20 @@ func (a *StorageMinerActorCode_I) PreCommitSector(rt Runtime, info sector.Sector
 
 	st._assertSectorDidNotExist(rt, info.SectorNumber())
 
+	// @param dealIDs []deal.DealID info.DealIDs()
```
sternhenri: ibid, see #677 for example of how to serialize args (in that case with randomness args)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -674,10 +704,19 @@ func (a *StorageMinerActorCode_I) ProveCommitSector(rt Runtime, info sector.Sect
...
 	}
 
-	// ActivateStorageDeals
-	// Ok if deal has started
-	// Send(SMA.ActivateDeals(onChainInfo.DealIDs())
-	// abort if activation failed
+	// Activate storage deals, abort if activation failed
+	// @param dealIDs []deal.DealID onChainInfo.DealIDs()
+	activateDealsParams := make([]util.Serialization, 1)
+	panic("TODO: serialize arguments")
```
sternhenri: ibid

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
```diff
@@ -176,9 +179,7 @@ func (st *StorageMinerActorState_I) _updateFailSector(rt Runtime, sectorNo secto
 	}
 
 	if newFaultCount > MAX_CONSECUTIVE_FAULTS {
-		// TODO: heavy penalization: slash pledge collateral and delete sector
-		// TODO: SendMessage(SPA.SlashPledgeCollateral)
-
+		// slashing is done at _slashCollateralForStorageFaults
```
sternhenri: where is it called?

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
```diff
@@ -106,7 +106,7 @@ type StorageMinerActorState struct {
     _updateSectorUtilization(rt Runtime, lastPoStResponse block.ChainEpoch) [deal.DealID]
     _initializeUtilizationInfo(rt Runtime, deals [deal.OnChainDeal]) sector.SectorUtilizationInfo
 
-    _getActivePower(rt Runtime) block.StoragePower
+    _getActivePower(rt Runtime, getActiveDealIDs bool) (activePower block.StoragePower, activeDealIDs [deal.DealID])
```
sternhenri: this sort of breaks a software pattern regarding simplicity of `get`. I'd recommend splitting into two methods, calling one or two based on case (rather than passing in bool). Simpler

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -175,10 +177,10 @@ func (a *StorageMinerActorCode_I) _onMissedSurprisePoSt(rt Runtime) {
 // construct PowerReport from SectorTable
 // need lastPoSt to search for new expired deals in _updateSectorUtilization since the lastPost
 // where DealExpirationAMT takes in a range of Epoch and return a list of values that expire in that range
-func (a *StorageMinerActorCode_I) _submitPowerReport(rt Runtime, lastPoStResponse block.ChainEpoch) {
+func (a *StorageMinerActorCode_I) _submitPowerReport(rt Runtime, lastPoStResponse block.ChainEpoch, processDealPayment bool) {
 	h, st := a.State(rt)
 	newExpiredDealIDs := st._updateSectorUtilization(rt, lastPoStResponse)
-	activePower := st._getActivePower(rt)
+	activePower, _ := st._getActivePower(rt, processDealPayment)
```
sternhenri: I don't understand why you send it across here but don't harvest the return...?

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -191,6 +193,8 @@ func (a *StorageMinerActorCode_I) _submitPowerReport(rt Runtime, lastPoStRespons
 	processPowerReportParam := make([]util.Serialization, 1)
 	// @param dealIDs []deal.DealID
 	processDealExpirationParam := make([]util.Serialization, 1)
+	// @params dealIDs []deal.DealID activeDealIDs
+	processDealPaymentParam := make([]util.Serialization, 1)
```
sternhenri: ibid

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
anorth: +1

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
```diff
@@ -144,7 +144,7 @@ type StorageMinerActorCode struct {
         faultType      sector.StorageFaultType
     )
 
-    _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch)
+    _submitPowerReport(rt Runtime, lastPoSt block.ChainEpoch, getActiveDealIDs bool)
```
anorth: The argument is actually called `processDealPayment`, which is very different!

Split this into two methods, this and `claimDealPayments`, the second of which is only called after successful PoSt.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
zixuanzh: @sternhenri I agree but in this case splitting into two will mean that we have to loop through the active sectors twice. Are we okay with that?

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -244,38 +255,34 @@ func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime, onChainInfo sect
...
-			// TODO: determine proper error here and error-handling machinery
 			rt.AbortStateMsg("Sector state not found in map")
 		}
 		switch sectorState.State().StateNumber {
 		case SectorCommittedSN, SectorRecoveringSN:
 			st._updateActivateSector(rt, sectorNo)
 		case SectorActiveSN:
-			// do nothing
-			// deal payment is made at _onSuccessfulPoSt
+			// do nothing, deal payment is made at the end
```
anorth:
```suggestion
			// do nothing, deal payment is made at the end of this method
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
zixuanzh: we implicitly use this in serialization, this is _ so that compiler doesnt complain.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
zixuanzh: this is now called in `_slashCollateralForStorageFaults` right after `_submitPowerReport`

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
zixuanzh: ack, will do

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -375,6 +382,8 @@ func (a *StorageMinerActorCode_I) RecoverFaults(rt Runtime, recoveringSet sector
 
 			// SendMessage(sma.PublishStorageDeals) or sma.ResumeStorageDeals?
 			// throw if miner cannot cover StorageDealCollateral
+			// TODO: EnsureDealCollateral
+			panic("TODO")
```
anorth: Note: dealing with dynamic collateral sort of works for us here. We could add a call to SPA to `tryLockCollateral(expectedPower)` that locks up collateral for the expected power after recovery, in advance of marking the sectors recovered. If the call fails, refuse to recover the sectors until the miner deposits enough that the tryLock succeeds. For a follow-up

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
sternhenri: Yes. I think that's fine. It's either two calls in the loop or one call in two loops. Now yes, the seek time for the object will be longer but leave to impl to optimize code. Let's make the spec as legible as possible

zixuanzh: sounds good

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
zixuanzh: will leave this to @porcuquine

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -216,7 +218,31 @@ func (a *StorageMinerActorCode_I) _submitPowerReport(rt Runtime, lastPoStRespons
 			Params_: processDealExpirationParam,
 		})
 	}
+}
+
+func (a *StorageMinerActorCode_I) _claimDealPayments(rt Runtime) {
```
sternhenri: This is a larger issue to be zoomed out but methods like these are going to be very expensive to run if run too often. (obviously has been brought up before)

sternhenri: To that larger design question:

This method is called on `_onSuccessfulPoSt`. My q is this. Why do we need to cycle through every deal in the network on every successful post? Why not just deal with the deals for the sector that has been posted only?

Or conversely only run this version of `_claimDealPayments`every X epochs (via cron) rather than onSuccessfulPoSt.

I think we should move toward one of those two solutions.

zixuanzh: I dont think we are cycling through every deal in the network on every successful PoSt. This is every active deal in this StorageMinerActor that just submitted its PoSt successfully.

zixuanzh:
@sternhenri 
- [x] Use SectorSize every where instead of UInt whereas SectorSize is of type UInt cc @porcuquine 
- [x] `recoverFaults` is left as a follow up
- [x] I think `_claimDealPayments` is right per comment above

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
zixuanzh: ah, this is something else, pledge collateral is checked at PoSt submission. This is on deal collateral which is a bit more complicated to account for..

zixuanzh: good news, we actually dont need `EnsureDealCollateral` here because storage deal collateral is not slashed at Declared/Detected Faults. It is only slashed in TerminatedFaults and all will be irrecoverably slashed and hence we dont need a top up. cc @sternhenri @jzimmerman

Issue #458, opened 2019-08-27
Undefined: block timestamp validation logic


whyrusleeping:
Its not clear how to validate a timestamp on a block header. The spec says it must be at least one block time past the previous block, and 'not in the future'.

I think those rules are likely incorrect, as clock skew would likely cause miners to drop a lot of blocks.

anorth:
There's a `ALLOWABLE_CLOCK_DRIFT` in https://github.com/filecoin-project/specs/blob/master/mining.md#block-validation, but IMO the logic there should be expressed only once, in prose, in https://github.com/filecoin-project/specs/blob/master/validation.md (which I think is where you were looking)

sternhenri:
Yes. Consolidation is right. Good catch (I’d never seen validation.md :b)

ZenGround0:
@whyrusleeping should be closing this out in #697, will post a link once merged.

PR #697, opened 2019-12-03
Spec/clock

ZenGround0:
*In this PR*

- [x] cleanup deprecated text and concepts from clock module
- [x] make explicit proposed validation rules by adding to chain sync *BV* series
- [x] explain timing model in "Mining Cycle" using diagram

This seems to address #458 as checking timestamps against the future is replaced with checking timestamp inclusion in the epoch which gets rid of the problem of skew dropping lots of good blocks.


src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -345,22 +347,23 @@ State Machine:
...
-  must exist in the lookback state). This requires some chain state, enough to establish plausibility levels of each of these values. A node should be able to estimate valid ranges for `b.ChainEpoch` based on the `LastTrustedCheckpoint`. `b.ChainWeight` is easy if some of the relatively recent chain is available, otherwise hard.
-  - **BV2 - Signature Validation**: Verify `b.BlockSig` is correct.
-  - **BV3 - Verify ElectionProof**: Verify `b.ElectionProof` is correct. This requires having state for relevant lookback parameters.
-  - **BV4 - Verify Ancestry links to chain**: Verify ancestry links back to trusted blocks. If the ancestry forks off before finality, or does not connect at all, it is a bad block.
-  - **BV4 - Verify MessageSigs**: Verify the signatures on messages
-  - **BV5 - Verify StateTree**: Verify the application of `b.Parents.Messages()` correctly produces `b.StateTree` and `b.MessageReceipts`
+  - **BV1 - Epoch Window Validation**: Validate that block was produced during the current epoch.  Validate that the block was received before `EPOCH_CUTOFF`.  This stage only applies during `CHAIN_FOLLOW`.
```
ZenGround0: Note to future spec writers: as we may continue to split up and add to these the current number system and presentation format could use some rethinking.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -345,22 +347,23 @@ State Machine:
...
+  - **BV2 - Light Consensus State Checks**: Validate `b.ChainWeight`, `b.ChainEpoch`, `b.MinerAddress`, `b.Timestamp`, are plausible (some ranges of bad values can be detected easily, especially if we have the state of the chain at `b.ChainEpoch - consensus.LookbackParameter`. Eg Weight and Epoch have well defined valid ranges, and `b.MinerAddress`
+  must exist in the lookback state). This requires some chain state, enough to establish plausibility levels of each of these values. A node can determine the exact valid range for `b.ChainEpoch` based on the `LastTrustedCheckpoint`, and a synchronized clock (see {{<sref clock>}}). `b.ChainWeight` is easy if some of the relatively recent chain is available, otherwise hard.
```
ZenGround0: HELP ACTION REQUIRED: right now this link does not render and idk how to fix.

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -345,22 +347,23 @@ State Machine:
 - Validation computation is considerable, and a serious DOS attack vector.
 - Secure implementations must carefully schedule validation and minimize the work done by pruning blocks without validating them fully.
 - `ChainSync` SHOULD keep a cache of unvalidated blocks (ideally sorted by likelihood of belonging to the chain), and delete unvalidated blocks when they are passed by `FinalityTipset`, or when `ChainSync` is under significant resource load.
-- It is key to note that any block received after the `ROUND_CUTOFF` time must be automatically discarded by the miner until the start of the next epoch.
+- It is key to note that any block received after the `EPOCH_CUTOFF` time must be automatically discarded by the miner until the start of the next epoch.
```
sternhenri: @ZenGround0 touch this up per notes we shared, related to potential DoSing issue. Specifically it is a weaker rule: no block received after the cutoff should be considered for the HEAD. But it should not be discarded. Same as with 294 (no blocks received `Epoch_cutoff` time after start are considered...)

src/systems/filecoin_blockchain/chainsync/_index.md
sternhenri: I feel like BV1 may be more subtle (assuming I got it):
- discard any block from future or from past validation ago
- cache other blocks (in case returning to CATCH UP mode) but do not consider them for inclusion in head

src/systems/filecoin_blockchain/chainsync/_index.md
sternhenri: see below

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -290,6 +290,8 @@ State Machine:
   - `ChainSync` is well-bootstrapped, and has an initial **trusted** `StateTree` to start from.
   - `ChainSync` fetches and validates blocks (see _Block Fetching and Validation_).
   - `ChainSync` is receiving and validating latest `Blocks` from `BlockPubsub`
+  - `ChainSync` validation enforces that incoming blocks belong to the current epoch
```
sternhenri: see below

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -290,6 +290,8 @@ State Machine:
...
+  - `ChainSync` validation enforces that no blocks are received `EPOCH_CUTOFF` time after start of epoch.
   - `ChainSync` DOES NOT have unvalidated blocks between `ChainSync.FinalityTipset` and `ChainSync.TargetHeads`
```
sternhenri: is the above still true given clock conversation?

src/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -345,22 +347,23 @@ State Machine:
...
+  - **BV2 - Light Consensus State Checks**: Validate `b.ChainWeight`, `b.ChainEpoch`, `b.MinerAddress`, `b.Timestamp`, are plausible (some ranges of bad values can be detected easily, especially if we have the state of the chain at `b.ChainEpoch - consensus.LookbackParameter`. Eg Weight and Epoch have well defined valid ranges, and `b.MinerAddress`
+  must exist in the lookback state). This requires some chain state, enough to establish plausibility levels of each of these values. A node can determine the exact valid range for `b.ChainEpoch` based on the `LastTrustedCheckpoint`, and a synchronized clock (see {{<sref clock>}}). `b.ChainWeight` is easy if some of the relatively recent chain is available, otherwise hard.
+  - **BV3 - Signature Validation**: Verify `b.BlockSig` is correct.
+  - **BV4 - Verify ElectionProof**: Verify `b.ElectionProof` is correct. This requires having state for relevant lookback parameters.
+  - **BV5 - Verify Ancestry links to chain**: Verify ancestry links back to trusted blocks. If the ancestry forks off before finality, or does not connect at all, it is a bad block.
```
sternhenri: Does this fit within the `candidateHeads` and `validationFringe` model? My sense is your worked forward, not bacl.

src/systems/filecoin_blockchain/struct/block_producer/_index.md
```diff
@@ -22,23 +22,34 @@ This process is repeated until either a winning ticket is found (and block publi
...
 
-Anytime a miner receives new blocks, it should evaluate what is the heaviest Tipset it knows about and mine atop it.
+Anytime a miner receives new valid blocks, it should evaluate what is the heaviest Tipset it knows about and mine atop it.
+
+### Timing
+
+{{< diagram src="../../../../diagrams/timing/timing.png" title="Mining Cycle Timing" >}}
+
+
+The mining cycle relies on receiving and producing blocks concurrently.  The sequence of these events in time is given by the timing diagram above.  The upper row represents the conceptual consumption channel marked by a period of block reception `Rx` followed by an inter-epoch buffer `Q`.  The lower row is the conceptual production channel made up of a period of mining followed by a period of transmission.  The lengths of the periods are not to scale.
```
sternhenri: period of mining 'M' followed by a period of transmission 'Tx'

src/systems/filecoin_blockchain/struct/block_producer/_index.md
```diff
@@ -22,23 +22,34 @@ This process is repeated until either a winning ticket is found (and block publi
...
+
+Blocks are received and validated during `Rx` up to `Q`.  The heaviest tipset at the beginning of `Q` is then used as the next head to mine against during `M`.  If mining is successful a block is transmitted during `Tx`.  The epoch boundaries are as shown.  
+
+This sequence of events applies only when the node is in the `CHAIN_FOLLOW` syncing mode.  Nodes in other syncing modes receive and process blocks differently and do not mine blocks.
```
sternhenri: can you further define "differently"?

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -2,26 +2,29 @@
...
 ---
+{{<label clock>}}
 
 {{< readfile file="clock_subsystem.id" code="true" lang="go" >}}
 {{< readfile file="clock_subsystem.go" code="true" lang="go" >}}
 
-Filecoin assumes weak clock synchrony amongst participants in the system. That is, the system relies on participants having access to a globally synchronized clock, tolerating bounded delay in honest clock lower than epoch time (more on this in a forthcoming paper).
 
-Filecoin relies on this system clock in order to secure consensus, specifically ensuring that participants are only running leader elections once per epoch and enabling miners to catch such deviations from the protocol. Given a system start and epoch time by the genesis block, the system clock allows miners to associate epoch and wall clock time, thereby enabling them to reason about block validity and give the protocol liveness.
+Filecoin assumes weak clock synchrony amongst participants in the system. That is, the system relies on participants having access to a globally synchronized clock (tolerating some bounded drift).
```
sternhenri: why remove the upper bound on tolerated drift?

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -2,26 +2,29 @@
...
+
 
 ## Clock uses
-Specifically, the Filecoin system clock is used:
+The Filecoin system clock is used:
 
-- to validate incoming blocks and ensure they were mined in the appropriate round, looking at the wall clock time in conjunction with the block's epoch number (see {{<sref block_validation>}}).
-- to help protocol convergence by giving miners a specific cutoff after which to reject incoming blocks in this round (see {{<sref chain_sync>}}).
-- to maintain protocol liveness by allowing participants to try leader election in the next round if no one has produced a block in this round (see {{<sref storage_power_consensus>}}).
+- by syncing nodes to validate that incoming blocks were mined in the appropriate epoch given their time stamp (see {{<sref block_validation>}}).  This is possible because the system clock maps all times to a unique epoch number totally determined by the start time in the genesis block.
```
sternhenri: nit: timestamp one word

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -2,26 +2,29 @@
...
+- by syncing nodes in the `CHAIN_FOLLOW` state to validate that incoming blocks were mined during the current epoch, not in the past or future.
```
sternhenri: further define past if agreed with above comments

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -2,26 +2,29 @@
...
+- by syncing nodes in the `CHAIN_FOLLOW` state to determine the cutoff period during each epoch after which they must reject incoming blocks for that epoch (see {{<sref chain_sync>}}).
```
sternhenri: ibid, change based on "reject" rule

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -2,26 +2,29 @@
...
+- by mining nodes to maintain protocol liveness by allowing participants to try leader election in the next round if no one has produced a block in the current round (see {{<sref storage_power_consensus>}}).
 
 In order to allow miners to do the above, the system clock must:
 
-1. have low clock drift: at most on the order of 1s (i.e. markedly lower than epoch time) at any given time.
-2. maintain accurate network time over many epochs: resyncing and enforcing accurate network time.
-3. set epoch number on client initialization equal to `epoch ~= (current_time - genesis_time) / epoch_time`
+1. Have low enough clock drift relative to other nodes so that blocks produced at and propagated from one node arrive within the same epoch boundary on the other node.  Maintain this throughout operation
+2. Set epoch number on node initialization equal to `epoch = Floor[(current_time - genesis_time) / epoch_time]`
```
sternhenri: why the removal of 2.?

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -39,14 +42,13 @@ Computer-grade clock crystals can be expected to have drift rates on the order o
   - See how [go-ethereum does this](https://github.com/ethereum/go-ethereum/blob/master/p2p/discv5/ntp.go) for inspiration
 - clients CAN consider using cesium clocks instead for accurate synchrony within larger mining operations
 
-Assuming a majority of rational participants, the above should lead to relatively low skew over time, with seldom more than 10-20% clock skew that should be rectified periodically by the network, as is the case in other networks. This assumption can be tested over time by ensuring that:
-
-- (real-time) epoch time is as dictated by the protocol
-- (historical) the current epoch number is as expected
+Assuming a majority of rational participants, the above should lead to relatively low skew over time, with seldom more than 10-20% clock skew that should be rectified periodically by the network, as is the case in other networks. This assumption will be maintained over time with both the real-time -- blocks with timestamps outside of current epoch time are rejected -- and historical -- blocks with timestamps that don't match protocol specified epochs are rejected -- validation rules.
```
sternhenri: assumption will be maintained makes it seem like less of an assumption. Do we know how to rectify? If not, the "tested" signifies we should be ensuring the assumption holds, else speccing a solution

src/systems/filecoin_nodes/clock/_index.md
```diff
@@ -39,14 +42,13 @@ Computer-grade clock crystals can be expected to have drift rates on the order o
...
 
 ## Future work
 
 If either of the above metrics show significant network skew over time, future versions of Filecoin may include potential timestamp/epoch correction periods at regular intervals.
 
+When recoverying from exceptional chain halting events (for example all implementations panic on a given block) the network will potentially opt for special purpose "dead zone" rules banning the authoring of blocks during the outage epochs to prevent attack vectors related to unmined epochs during chain restart.
```
sternhenri: Would seek ack from all impls and @jbenet on this

src/systems/filecoin_blockchain/chainsync/_index.md
sternhenri: could make the last sentence wrt chainWeight more precise. Eg. "Verifying `b.ChainWeight` will require computation that scales linearly with the length of the chain to verify"

sternhenri: must add `clock` to `hugo/data/sref_titles.yaml` and `hugo/data/sref_locations.yaml` in order for the label to be recognized

src/systems/filecoin_blockchain/chainsync/_index.md
ZenGround0: OK I'm going to go back to work on this one.

Thinking this over more with help from @jzimmerman the right place to encode this information is in the chain selection function (set the chain to head only if its height is our current epoch).  We really want to avoid adding more complexity (what is this cache?  how does it work? how is it used?) directly to the chain sync system.

src/systems/filecoin_blockchain/chainsync/_index.md
ZenGround0: Everything past BV1 is just copy pasted from shifting the number, so idk. The spec's intent might be to sync forwards but currently I read it as leaving it up to implementation.  and go-filecoin for sure syncs backwards.

src/systems/filecoin_nodes/clock/_index.md
ZenGround0: Its just a relevant idea, not a commitment (as go-filecoin repr I ack 👿 ).  Might be best to just nuke "future work" from the spec.

src/systems/filecoin_nodes/clock/_index.md
ZenGround0: I stated what seemed to be the same information in the parentheses at the end of the new paragraph "(tolerating some bounded drift)".  I may have cut out information that I didn't understand from reading the old version.  If so let me know and I'll put it back

src/systems/filecoin_blockchain/chainsync/_index.md
sternhenri: I would agree. The cache I mean is you should keep blocks around to not request them from the network if you get a block much heavier than yours in a given epoch from a subchain you didn't know about. You probably already have part of the chain so you want to request as few as possible when going back to catch up. This is what Juan meant I believe when saying "only discard past finality".

So agree all in all. the effect of your call is just that chain head will only be set to head if validated to current epoch. Else youll fall behind enough to move back to catch up and sync to new best chain

src/systems/filecoin_nodes/clock/_index.md
sternhenri: Weird... I thought it stated:
> can tolerate drift up to an epoch length
but can't find it

Ignore unless you think useful

src/systems/filecoin_nodes/clock/_index.md
sternhenri: Sorry I hadn't seen the header. You're all set. Thanks for pointing out.

src/systems/filecoin_nodes/clock/_index.md
ZenGround0: old bullet 2 was refining and describing implementation of old bullet 1 so I consolidated.

ZenGround0:
Build seems to be broken on master, trying to debug now.

hugo/data/sref_locations.yaml
```diff
@@ -61,5 +61,6 @@ post: [algorithms, proof_of_spacetime]
 election_post: [algorithms, proof_of_spacetime, election_post]
 
 # nodes
+clock: [filecoin_blockchain]
```
sternhenri: does it work? no need for
```suggestion
clock: [systems, filecoin_nodes, clock]
```
?

src/systems/filecoin_blockchain/blockchain.id
```diff
@@ -45,10 +42,10 @@ type BlockchainSubsystem struct @(mutable) {
     // and verify state root
     TryGenerateStateTree(block block.Block) st.StateTree
 
-    // call by clock in BlockReception after Round Cutoff
+    // call by clock in BlockReception upon new epoch
     // TODO define clock here
```
sternhenri: can remove TODO?

src/systems/filecoin_blockchain/chainsync/_index.md
jzimmerman: I believe syncing backwards is correct -- nodes shouldn't waste resources on validating chains that might not lead to heads the network believes to be current.

ZenGround0:
@jzimmerman thanks for brining this up.  Because I haven't done anything in this PR other than shift the BV numbers this is wasn't introduced here and so is out of scope for merging this PR.  This is probably worth running through as part of another work item (this seems relevant for you @anorth).

Quick thoughts on this: I'm skeptical we can't or shouldn't achieve signature validation before the actor is created -- we've been doing this in go-filecoin for a while.  Isn't it necessary to carry a non-ID address with an actual public key in order to instantiate a new account actor?  If so do we still have a problem?

anorth:
@ZenGround0 @jzimmerman I have already specified a behaviour for this case: https://filecoin-project.github.io/specs/#message-semantic-validation

- the sending account must exist in the parent state referenced by the block, so signatures *are* checkable
- actor creation and sending in one block is explicitly disallowed (precisely b/c the send signature cannot be verified)

PR #698, opened 2019-12-03
add label before spec version date

mishmosh:
Make the last-updated date more obvious.


mishmosh:
Tagging @sternhenri as current publishing master. Not urgent, review & merge when you have time.

PR #700, opened 2019-12-04
Prose for block syntactic and semantic validation

anorth:
FYI @ZenGround0

### DIFF
--- a/src/systems/filecoin_blockchain/struct/block/_index.md
+++ b/src/systems/filecoin_blockchain/struct/block/_index.md
@@ -12,19 +12,72 @@ A block header contains information relevant to a particular point in time over 
 **Note:** A block is functionally the same as a block header in the Filecoin protocol. While a block header contains Merkle links to the full system state, messages, and message receipts, a block can be thought of as the full set of this information (not just the Merkle roots, but rather the full data of the state tree, message tree, receipts tree, etc.). Because a full block is quite large, our chain consists of block headers rather than full blocks. We often use the terms `block` and `block header` interchangeably.
 {{% /notice %}}
 
-{{<label block_validation>}}
-## Block Validation Rules
-  - **BV0 - Syntactic Validation**: Validate data structure packing and ensure correct typing.
-  - **BV1 - Future Epoch Validation**: Validate that block does not claim to be from a future epoch. 
-  - **BV2 - Light Consensus State Checks**: Validate `b.ChainWeight`, `b.ChainEpoch`, `b.MinerAddress`, `b.Timestamp`, are plausible (some ranges of bad values can be detected easily, especially if we have the state of the chain at `b.ChainEpoch - consensus.LookbackParameter`. Eg Weight and Epoch have well defined valid ranges, and `b.MinerAddress`
-  must exist in the lookback state). This requires some chain state, enough to establish plausibility levels of each of these values. A node can determine the exact valid range for `b.Timestamp`  based on the `LastTrustedCheckpoint`, `b.ChainEpoch`, and a synchronized clock (see {{<sref clock>}}).
-  - **BV3 - Signature Validation**: Verify `b.BlockSig` is correct.
-  - **BV4 - Verify ElectionPoSt**: Verify `b.ElectionPoStOutput` was correctly generated and yielding winning `PartialTickets`.
-  - **BV5 - Verify Ancestry links to chain**: Verify ancestry links back to trusted blocks. If the ancestry forks off before finality, or does not connect at all, it is a bad block.
-  - **BV6 - Verify MessageSigs**: Verify the signatures on messages
-  - **BV7 - Verify StateTree**: Verify the application of `b.Parents.Messages()` correctly produces `b.StateTree` and `b.MessageReceipts`
-
 {{< readfile file="block.id" code="true" lang="go" >}}
 
-{{< readfile file="block.go" code="true" lang="go" >}}
+{{< readfile file="election.id" code="true" lang="go" >}}
+
+# Block syntax validation
+
+Syntax validation refers to validation that may be performed on a block and its messages 
+without reference to outside information such as the parent state tree.
+
+An invalid block must not be transmitted or referenced as a parent.
+
+A syntactically valid block header must decode into fields matching the type definition below. 
+
+A syntactically valid header must have:
+
+- a non-empty array of `Parents` CIDs if `Epoch` is greater than zero,
+- a non-negative `ParentWeight`,
+- a `Miner` address which is an ID-address,
+- a non-negative `Epoch`,
+- a positive `Timestamp`,
+- a `Ticket` with non-empty `VRFResult`,
+- `ElectionPoStOutput` containing:
+   - a non-empty `Candidates` array
+   - a non-empty `PoStRandomness` field
+   - a non-empty `Proof` field
+   
+A syntactically valid full block must have:
+
+- all referenced messages syntactically valid,
+- all referenced parent receipts syntactically valid,
+
+Note that validation of the block signature requires access to the miner worker address and
+public key from the parent tipset state, so signature validation forms part of semantic validation. 
+Similarly, message signature validation requires lookup of the public key associated with 
+each message's `From` account actor in the block's parent state.
+
+# Block semantic validation
+
+Semantic validation refers to validation that requires reference to information outside the block
+header and messages themselves, in particular the parent tipset and state on which the block is built.
+
+A semantically valid block must have:
+
+- `Parents` listed in lexicographic order of their header's `Ticket`,
+- `Parents` all reference valid blocks and form a valid {{<sref tipset>}},
+- `ParentState` matching the state tree produced by executing the parent tipset's messages (as defined by the VM interpreter) against that tipset's parent state,
+- `ParentMessageReceipts` identifying the receipt list produced by parent tipset execution, with one receipt for each unique message from the parent tipset, 
+- `ParentWeight` matching the weight of the chain up to and including the parent tipset,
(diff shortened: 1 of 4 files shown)


src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -7,5 +7,66 @@ The Block is a unit of the Filecoin blockchain.
...
+
+A syntactically valid header must have:
+
+- a non-empty array of `Parents` CIDs if `Epoch` is greater than zero,
+- a non-negative `ParentWeight`,
+- a `Miner` address which is an ID-address,
+- a non-negative `Epoch`,
+- a positive `Timestamp`,
+- a `Ticket` with non-empty `VRFResult`,
+- an `ElectionProof` with non-empty `VRFResult`.
```
sternhenri: could you please change the `ElectionProof` bit given new ePoSt work merged in #705 
```suggestion
- `ElectionPoStOutput` containing:
  - a non-empty `Candidates` array
  - a non-empty `PoStRandomness` field
  - a non-empty `Proof` field
```

src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -7,5 +7,66 @@ The Block is a unit of the Filecoin blockchain.
...
+
+A syntactically valid full block must have:
+
+- all referenced messages syntactically valid,
+- all referenced parent receipts syntactically valid,
+
+Note that validation of the block signature requires access to the miner worker address from the
+parent tipset state, so signature validation forms part of sematic validation. Similarly, message
```
sternhenri: typo s/sematic/semantic

src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -7,5 +7,66 @@ The Block is a unit of the Filecoin blockchain.
...
+- a positive `Timestamp`,
+- a `Ticket` with non-empty `VRFResult`,
+- an `ElectionProof` with non-empty `VRFResult`.
+
+A syntactically valid full block must have:
+
+- all referenced messages syntactically valid,
+- all referenced parent receipts syntactically valid,
+
+Note that validation of the block signature requires access to the miner worker address from the
```
sternhenri: to the miner worker address and public key

src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -7,5 +7,66 @@ The Block is a unit of the Filecoin blockchain.
...
+
+# Block semantic validation
+
+Semantic validation refers to validation that requires reference to information outside the block
+header and messages themselves, in particular the parent tipset and state on which the block is built.
+
+A semantically valid block must have:
+
+- `Parents` listed in lexicographic order of their header's `Ticket`,
+- `Parents` all reference valid blocks and form a valid tipset,
```
sternhenri:
```suggestion
- `Parents` all reference valid blocks and form a valid {{<sref tipset>}},
```

src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -7,5 +7,66 @@ The Block is a unit of the Filecoin blockchain.
...
+- `ParentStateRoot` identifying the state tree produced by executing the parent tipset's messages,
+- `ParentReceiptRoot` identifying the receipt list produced by parent tipset execution, with one receipt for each non-duplicate message from the parent tipset,
```
sternhenri: for each non-duplicate is wrong. Should be for each unique (i.e. the first duplicate gets one too)

src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -7,5 +7,66 @@ The Block is a unit of the Filecoin blockchain.
...
+Semantic validation refers to validation that requires reference to information outside the block
+header and messages themselves, in particular the parent tipset and state on which the block is built.
+
+A semantically valid block must have:
+
+- `Parents` listed in lexicographic order of their header's `Ticket`,
+- `Parents` all reference valid blocks and form a valid tipset,
+- `ParentStateRoot` identifying the state tree produced by executing the parent tipset's messages,
+- `ParentReceiptRoot` identifying the receipt list produced by parent tipset execution, with one receipt for each non-duplicate message from the parent tipset, 
+- `ParentWeight` matching the weight of the chain up to the parent tipset,
```
sternhenri: up to and including the parent tipset

src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -7,5 +7,66 @@ The Block is a unit of the Filecoin blockchain.
...
+- `Epoch` greater than that of its parents,
+- `Miner` that is active in the storage power table in the parent tipset state,  
+- a `Ticket` derived from the minimum ticket from the parent tipset's block headers, 
+    - `Ticket.VRFResult` validly signed by the `Miner` actor's worker account public key,
+- `ElectionProof` that is a winner given the `Miner` actor's storage power fraction in the parent state, 
+    - `ElectionProof.VRFResult` validly signed by the `Miner` actor's worker account public key,
```
sternhenri: I would suggest
```
- `ElectionPoStOutput` yielding winning partial tickets that were generated validly
  - `ElectionPoSt.Randomness` is well formed and appropriately drawn from a past tipset according to the PoStLookback
  - `ElectionPoSt.Proof` is a valid proof verifying the generation of the `ElectionPoSt.Candidates` from the `Miner`'s eligible sectors.
  - `ElectionPoSt.Candidates` contains well formed `PoStCandidate`s each of which has a `PartialTicket` yielding a winning `ChallengeTicket` in Expected Consensus.

File: src/systems/filecoin_blockchain/struct/block/_index.md [hunk 8]
```diff
@@ -7,5 +7,66 @@ The Block is a unit of the Filecoin blockchain.
...
+- a `Timestamp` in seconds lying within the quantized window implied by the genesis block's timestamp and the
+block's `Epoch`, and not in the future,
```
sternhenri: May want an extra one here per @ZenGround0 's ongoing work.

src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -7,5 +7,66 @@ The Block is a unit of the Filecoin blockchain.
...
+- all SECP messages correctly signed by their sending actor's key,
```
sternhenri:
```suggestion
- all SECP messages correctly signed by their sending actor's worker account key,
```

src/systems/filecoin_blockchain/struct/block/_index.md
sternhenri: nit: maybe move part of this to the epoch, st the verif is that it is larger than its parents and corresponds to the timestamp (then have extra check in timestamp re clock synchronization).

That would mirror what we would get were we to use a VDF (ie lighter timestamp verif, and added epoch verif)

src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -7,5 +7,66 @@ The Block is a unit of the Filecoin blockchain.
...
+# Block semantic validation
+
+Semantic validation refers to validation that requires reference to information outside the block
+header and messages themselves, in particular the parent tipset and state on which the block is built.
+
+A semantically valid block must have:
+
+- `Parents` listed in lexicographic order of their header's `Ticket`,
+- `Parents` all reference valid blocks and form a valid tipset,
+- `ParentStateRoot` identifying the state tree produced by executing the parent tipset's messages,
```
sternhenri: add an extra line specifically mentioning that val here means repeating the message processing of grandparent messages (per vm interpreter) to yield `ParentStateRoot` and then comparing both state trees

src/systems/filecoin_blockchain/struct/block/_index.md
ZenGround0: The timestamp validation fits neatly into two rules
(1) the timestamp fits within the epoch boundary 
(2) the block's epoch is not in the future as determined by the node's local reading of the current epoch using its clock.

src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -13,5 +13,71 @@ A block header contains information relevant to a particular point in time over
...
+Similarly, message signature validation requires lookup of the public key associated with 
+each message's `From` account actor in the block's parent state.
+
+# Block semantic validation
+
+Semantic validation refers to validation that requires reference to information outside the block
+header and messages themselves, in particular the parent tipset and state on which the block is built.
+
+A semantically valid block must have:
+
```
ZenGround0: Since we are planning to remove duplication by putting these all in one place (https://github.com/filecoin-project/specs/pull/709), we need a way to reference these rules succinctly from the chain sync section.  The `BV*` method used in the chain sync doc worked ok.  Consider reinstating this grouping.  If you want to avoid grouping here then each bullet here could be a number / BV string and we could do the grouping in the chain sync doc.

src/systems/filecoin_blockchain/struct/block/_index.md
anorth: Done, moved the "future" up to the epoch line.

src/systems/filecoin_blockchain/struct/block/_index.md
anorth: I didn't see #709 and I don't think the progressive notes belong here. I have moved them back.

I see a clear distinction between the prescriptive validation rules described here, and the recommendation of progressive validation in the chain sync process.

PR #704, opened 2019-12-04
add precommit deposit and garbage collection

zixuanzh:
**In this PR**

- [x] Add PRECOMMIT_DEPOSIT
- [x] Miners need to include at least that amount in PreCommitSector
- [x] Miners get back the PRECOMMIT_DEPOSIT at ProveCommitSector
- [x] PRECOMMIT_DEPOSIT will be burned when PreCommit expires before receiving a ProveCommit
- [x] Burn PRECOMMIT_DEPOSIT at `_expirePreCommittedSectors` and clear inactive deals in PreCommittedSectors as a form of garbage collection

### DIFF
--- a/src/systems/filecoin_mining/storage_mining/constants.go
+++ b/src/systems/filecoin_mining/storage_mining/constants.go
@@ -1,6 +1,9 @@
 package storage_mining
 
-import block "github.com/filecoin-project/specs/systems/filecoin_blockchain/struct/block"
+import (
+	block "github.com/filecoin-project/specs/systems/filecoin_blockchain/struct/block"
+	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
+)
 
 // TODO: placeholder epoch value -- this will be set later
 const MAX_PROVE_COMMIT_SECTOR_PERIOD = block.ChainEpoch(3)    // placeholder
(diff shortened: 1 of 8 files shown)


src/systems/filecoin_mining/storage_mining/constants.go
```diff
@@ -11,3 +14,7 @@ const PROVING_PERIOD = block.ChainEpoch(2)                    // placeholder
 const SURPRISE_CHALLENGE_FREQUENCY = 2 // placeholder
 // how far into their PP does a miner get their first challenge
 const SUPRISE_NO_CHALLENGE_PERIOD = PROVING_PERIOD / SURPRISE_CHALLENGE_FREQUENCY
+
+// FIL deposit per sector precommit in Interactive PoRep
+// refunded after ProveCommit but burned if PreCommit expires
+const PRECOMMIT_DEPOSIT = actor.TokenAmount(0) // placeholder
```
anorth: Should this be a function of sector size, so a deposit per byte or per GiB?

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -580,19 +580,21 @@ func (a *StorageMinerActorCode_I) _verifySeal(rt Runtime, onChainInfo sector.OnC
...
-
 	// can be called regardless of Challenged status
 
-	// TODO: might take collateral in case no ProveCommit follows within sometime
-	// TODO: collateral also penalizes repeated precommit to get randomness that one likes
 	// TODO: might be a good place for Treasury
 
 	h, st := a.State(rt)
 
+	msgValue := rt.ValueReceived()
```
anorth: I think it's worth documenting somewhere up the top that the balance of a storage miner actor should equal exactly the sum of precommit deposits not yet either returned or burnt.

zixuanzh: ack, will do

src/systems/filecoin_mining/storage_mining/constants.go
zixuanzh: Not sure, but think it's a good suggestion, I can add that for now, simple change to add/remove later.

zixuanzh: cc @jzimmerman @nicola

PR #707, opened 2019-12-05
Min Miner Size

sternhenri:
**In This PR**:
- [x] Minimum Miner Power for SPC explained in detail, including
  - [x] relation to block production and EPoSt
  - [x] implications for network bootstrap
  - [x] implications for miner onboarding

**Open Questions**:
- Should the min be expressed in terms of active power or total power? Given incentive structure we want, the former seems right but it does incentivize self-deals.
- Should the largerThanMinMinerSize abstraction be defined in the SPA or SMA? I'd argue it's a SPA component...


src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -112,7 +112,21 @@ Each Ticket should be generated from the prior one in the ticket-chain and verif
...
+## Minimum Miner Size
+
+{{% notice placeholder %}}
+The below values are currently placeholders.
+{{% /notice %}}
+
+In order to secure Storage Power Consensus, the system defines a minimum miner size required to participate in consensus.
+
+Specifically, miners must have either at least 100TB of active power (i.e. storage power currently used in storage deals) or 1/3 of the network's active storage power to participate in SPC.
+Miners smaller than this cannot mine blocks and earn block rewards. Put another way, this means that network liveness requires at least 3 active miners (running storage deals) or 300TB of active power.
```
ZenGround0: Why at least 3 active miners?  One miner with all power has > 1/3 power right?

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -112,7 +112,21 @@ Each Ticket should be generated from the prior one in the ticket-chain and verif
...
+
+Accordingly, to bootstrap the network, the genesis block must include 3 valid storage deals made between the genesis miners(along with appropriate committed storage).
```
ZenGround0: Is the 1/3 edge case just for the genesis block?  If so it seems like we could cheat for early testnet and transition to genesis miners with 100s of TBs

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -112,7 +112,21 @@ Each Ticket should be generated from the prior one in the ticket-chain and verif
...
+{{<label min_miner_size>}}
+## Minimum Miner Size
+
+{{% notice placeholder %}}
+The below values are currently placeholders.
+{{% /notice %}}
+
+In order to secure Storage Power Consensus, the system defines a minimum miner size required to participate in consensus.
+
+Specifically, miners must have either at least 100TB of active power (i.e. storage power currently used in storage deals) or 1/3 of the network's active storage power to participate in SPC.
```
anorth: TiB or TB? Probably best to just spell out a number of bytes.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
anorth: Here again I agree with @ZenGround0 that a sole miner is automatically a valid block producer. Of course we don't expect to start mainnet that way, but it's fine for testnets.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
```diff
@@ -10,6 +10,20 @@ import (
 	util "github.com/filecoin-project/specs/util"
 )
 
+func (st *StoragePowerActorState_I) PowerIsSmallerThanMin(minPower block.StoragePower) bool {
```
anorth: Is `minPower` supposed to be miner power? A very confusing name in the context.

I think this functionality would be clearer inverted: `PowerMeetsConsensusMinimum()`

src/systems/filecoin_mining/storage_mining/constants.go
```diff
@@ -10,6 +12,8 @@ const PROVING_PERIOD = block.ChainEpoch(2)                    // placeholder
 // how many times per PP should a miner get challenged on expectation
 const SURPRISE_CHALLENGE_FREQUENCY = 2 // placeholder
 // how far into their PP does a miner get their first challenge
-const SUPRISE_NO_CHALLENGE_PERIOD = PROVING_PERIOD / SURPRISE_CHALLENGE_FREQUENCY
-const EPOST_SAMPLE_NUM = 1
-const EPOST_SAMPLE_DENOM = 25
+const SUPRISE_NO_CHALLENGE_PERIOD = PROVING_PERIOD / SURPRISE_CHALLENGE_FREQUENCY // placeholder
+const EPOST_SAMPLE_NUM = 1                                                        // placeholder
+const EPOST_SAMPLE_DENOM = 25                                                     // placeholder
+const MIN_MINER_SIZE_STOR = 100                                                   // placeholder
```
anorth: In bytes please

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
```diff
@@ -1,16 +1,54 @@
...
+	// TODO fix conversion to bytes
+	panic(substate)
+	var serializedSubstate util.Serialization
+	spa, err := spc.Deserialize_StoragePowerActorState(serializedSubstate)
+
+	if err == nil {
+		panic("Deserialization error")
+	}
+	return spa
+}
```
anorth: Blocking: I believe this pattern of inspecting another actor's state should be explicitly banned. Write a method/message to query the SPA for the desired information. In this case I think this entire addition should migrate to SPA.

src/algorithms/post/election_post.md
```diff
@@ -196,4 +202,10 @@ Each reported fault carries a penality with it.
 
 Note that surprise PoSt will frequently be used by small miners who do not win blocks, and by miners as they are onboarding power to the network, since they will not be able to win any blocks (they have no power) and hence gain power through ElectionPoSt.
 
-While it resets a miner’s ElectionWindow, miners earn no reward from submitting SurprisePoSt messages. This mechanism does forces a miner to announce their faults and clean up the Power Table accordingly.
+While it resets a miner’s ElectionWindow, miners earn no reward from submitting SurprisePoSt messages. This mechanism does force a miner to announce their faults and clean up the Power Table accordingly.
+
+# Miner Onboarding
+
+Storage Power Consensus participants are subject to a {{<sref min_miner_size>}}, meaning miners smaller than `MIN_MINER_SIZE_STOR` of active (or in-deal) storage or whose active storage represents less than `MIN_MINER_SIZE_PERC`% of the network's cannot produce valid electionPoSts.
```
anorth: This parenthetical is the only definition of active power in the spec. Please elevate it somewhere appropriate, along with a description of what the difference is from total power, inactive power etc.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: duh. went through a whole cycle of weird useless calculations clearly. Lesson not to spec late

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: The point is rather more to ensure liveness in case the network shrinks too much. Ie no need to reboot FIL if there remains a single active participant. We intend to have large genesis miners. Will revise

src/algorithms/post/election_post.md
steven004: Should be `and`? 
However, it does not make sense to me to have the 2nd condition. If we asked one miner to have more than 1/3 power to start mining, the first one might be able to control the network. We may need to have a start point for all miners to prevent it.

zixuanzh: In-deal or not does not determine whether power is active or inactive. Power has to come from in-deal data. Active Power comes from data in Active sectors (and hence the deals are also active). Inactive Power comes from data in Committed, Recovering and Failing sectors. Miners still need to maintain pledge collateral for Inactive Power.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
```diff
@@ -65,18 +103,18 @@ func (st *StorageMinerActorState_I) _updateSectorUtilization(rt Runtime, lastPoS
 
 }
 
-func (st *StorageMinerActorState_I) _getActivePower(rt Runtime) block.StoragePower {
+func (st *StorageMinerActorState_I) _getActivePower() (block.StoragePower, error) {
 	var activePower = block.StoragePower(0)
 
 	for _, sectorNo := range st.SectorTable().Impl().ActiveSectors_.SectorsOn() {
 		utilizationInfo, found := st.SectorUtilization()[sectorNo]
```
zixuanzh: please use `_getUtilizationInfo` here which does this safely and is used everywhere else

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
zixuanzh: We should also be more explicit here with regard to what Power we are talking about? TotalPower, ActivePower, or InactivePower? I think we should talk through this.

src/algorithms/post/election_post.md
sternhenri: Thanks for clarifying that @zixuanzh. i was imprecise (read plain wrong) on my distinction. I will fix accordingly

sternhenri: To your point that this needs a clear single def @anorth. Will add it.

@steven004 no I stand by `or`. This is more about liveness than control @steven004. If your powerful miner is the only miner, then yes they control the network. But if any other miners have `>MIN_MINER_SIZE_PERC` or `>MIN_MINER_SIZE_STOR` they can also publish blocks, so control is dependent on power, as it should be.

The second condition is mostly here to ensure that if the network drops to a size `< MIN_MINER_SIZE_STOR/MIN_MINER_SIZE_PERC`, the chain can still grow (otherwise you lose liveness entirely).

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
sternhenri: But this is safe, instead of dealing with error internally though I return it so it can be handled by the caller appropriately (rt error for actors, something else for subsystems). Can you explain why `_getUtitlizationInfo` is more appropriate?

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -30,6 +30,17 @@ We must distinguish between both types of "miners" (storage and block miners). {
...
 
+## On Power
+
+Per the above, we also clearly distinguish putting storage on-chain from gaining power in consensus (sometimes called "Storage Power") as follows:
+
+**Consensus power in Filecoin is determined by in-deal storage**. For instance, if a miner had a 32GB sector, 20 of which were used as part of storage deals, only those 20 would contribute to said miner's *Storage Power*.
+
+Thereafter, power can be either *active* or *inactive* as defined in {{<sref storage_mining_subsystem>}}.
+
+In-deal or not does not determine whether power is active or inactive. Power has to come from active in-deal data. Active Power comes from deal data in Active sectors (and hence the deals are also active). Inactive Power comes from data in Committed, Recovering and Failing sectors. Miners still need to maintain pledge collateral for Inactive Power.
```
zixuanzh: 👍

src/systems/filecoin_mining/storage_mining/_index.md
```diff
@@ -9,21 +9,47 @@ entries:
...
+- Continously proving storage ({see {<sref election_post>}})
+- Declaring storage faults and recovering from them.
+
+## Storage States
+
+When managing their storage sectors as part of Filecoin mining, storage providers will account for where in the mining cycle their sectors are. For instance, has a sector been committed? Does it need a new PoSt?
+
+Through this cycle, sectors will fall into either a miner's:
+- *PreCommitted Sectors*: New sectors to be committed using PoRep
+- `Proving Set`: Committed sectors to be proven over time using PoSts
```
zixuanzh: Blocking: `ProvingSet` includes ActiveSector, CommittedSector, and RecoveringSector. This is the set of sectors that the miners are supposed to generate proofs against.

src/systems/filecoin_mining/storage_mining/_index.md
```diff
@@ -9,21 +9,47 @@ entries:
...
+  - Sectors can be faulted by the miner itself (`Declared Faults`).
```
zixuanzh: Blocking: add that Sectors can also be detected by the chain (`Detected Faults`), a sector that is in the failing state for three consecutive Epoch will result in termination (`Terminated Faults`)

src/systems/filecoin_mining/storage_mining/_index.md
```diff
@@ -9,21 +9,47 @@ entries:
...
+- `Fault Set`: Committed sectors that have failed to be proven.
```
zixuanzh: Blocking: remove this, there is no more notion of a Fault Set as a state, a Fault Set is only used in `DeclareFaults` as a struct. We are using `Failing` sector state in place of the old Fault Set.

src/systems/filecoin_mining/storage_mining/_index.md
```diff
@@ -9,21 +9,47 @@ entries:
...
+  - Fault recovery allows sectors to return to the `Proving Set`.
```
zixuanzh: Blocking: to be precise, `RecoverFaults`, which can be called any time by the miner, will move sectors from `Failing` to `Recovering` which is part of the `ProvingSet`.

src/systems/filecoin_mining/storage_mining/_index.md
```diff
@@ -9,21 +9,47 @@ entries:
...
+
+Through the above workflows, Filecoin {{<sref sector "Sectors">}} (or Storage more generally) can be describede as beeing in the following states:
+
+- *Not-on-Fil*: This includes *Precommitted sectors* that have yet to be proven (and do not account for storage in Filecoin).
+- **Active**: This includes sectors in the `Proving Set`.
+- **Inactive**: This includes sectors in the `Fault Set` (i.e. through `Declared Faults`) and faults detected by the system (`Detected Faults`) which render all of a miner's storage inactive.
```
zixuanzh: Blocking: please update this according to above comments. Ack that `DetectedFaults` will move all miner sectors into the Failing state and all power inactive.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
```diff
@@ -65,18 +67,18 @@ func (st *StorageMinerActorState_I) _updateSectorUtilization(rt Runtime, lastPoS
...
 
-func (st *StorageMinerActorState_I) _getActivePower(rt Runtime) block.StoragePower {
+func (st *StorageMinerActorState_I) _getActivePower() (block.StoragePower, error) {
 	var activePower = block.StoragePower(0)
 
 	for _, sectorNo := range st.SectorTable().Impl().ActiveSectors_.SectorsOn() {
-		utilizationInfo, found := st.SectorUtilization()[sectorNo]
-		if !found {
-			rt.AbortStateMsg("sm._getActivePower: sectorNo not found in SectorUtilization")
+		utilInfo, err := st._getUtilizationInfo(sectorNo)
```
zixuanzh: Blocking: Any reason why we are not using safeGet here? we are returning an error to be thrown later, in SubmitPowerReport

zixuanzh: I see, saw why we are returning an error below. In this case, can you also change `_getInactivePower(rt Runtime)` to `_getInactivePower()` in this same PR?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
```diff
@@ -10,6 +10,20 @@ import (
 	util "github.com/filecoin-project/specs/util"
 )
 
+func (st *StoragePowerActorState_I) ActivePowerMeetsConsensusMinimum(minerPower block.StoragePower) bool {
+	totPower := st._getActivePower()
```
zixuanzh: nit: call this `totalActivePower` instead of `totPower`?

src/systems/filecoin_mining/storage_mining/_index.md
sternhenri: why do we actually need the `Recovering` state? Just an optimization? (We could always test every failing sector each time)...

src/systems/filecoin_mining/storage_mining/_index.md
sternhenri: ack. Let's separate power and sector state here though. I think that is part of the confusion.

I hope this helps clarify that power is an SPC thing, sector state and SMS thing.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
sternhenri: Done.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
sternhenri: feel it's obvs from function names all around :b. Will do if I touch up again!

src/systems/filecoin_mining/storage_mining/_index.md
zixuanzh: Recovering state will be transitioned into Active State after a PoSt, it's a required state.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -38,8 +39,9 @@ Per the above, we also clearly distinguish putting storage on-chain from gaining
 
 Thereafter, power can be either *active* or *inactive* as defined in {{<sref storage_mining_subsystem>}}.
 
-In-deal or not does not determine whether power is active or inactive. Power has to come from active in-deal data. Active Power comes from deal data in Active sectors (and hence the deals are also active). Inactive Power comes from data in Committed, Recovering and Failing sectors. Miners still need to maintain pledge collateral for Inactive Power.
+It is important to remember that storage contributing to power and storage state are orthogonal notions. Specifically, power comes from storage being used in live-deals in active sectors (as opposed to non-deal active storage or in-deal inactive storage).
```
zixuanzh: Blocking: Sorry I think this is more confusing than before. Power comes from active in-deal storage (the sector can be inactive, e.g. a Committed sector is inactive, deals can be already active in those sectors, and it counts for inactive power).

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -38,8 +39,9 @@ Per the above, we also clearly distinguish putting storage on-chain from gaining
...
 
+Miners need to maintain pledge collateral for both active and inactive in-deal storage.
```
zixuanzh: Blocking: lets not create a new concept here called in-deal storage? The previous paragraph here seems fine for now, we might need to update this again in the next few days.

src/systems/filecoin_mining/storage_mining/_index.md
```diff
@@ -24,22 +24,20 @@ The above involves a number of steps to putting on and maintaining online storag
...
-- **Inactive**: This includes sectors in the `Fault Set` (i.e. through `Declared Faults`) and faults detected by the system (`Detected Faults`) which render all of a miner's storage inactive.
+A sector that is in the `Failing` state for three consecutive `Proving Period`s will be terminated (`Terminated Faults`) meaning its data will be deemed unrecoverable by PoSt and the sector will have to be SEALed once more (as part of PoRep).
 
-All of these states end up affecting both a miner's ability to serve as a storage provider in the Filecoin Storage Market, and their ability to participate in Filecoin consensus through Storage Power Consensus (see {{<sref storage_power>}}). We illustrate this below.
+Through the above workflows, Filecoin (or Storage more generally) can more generally be described as being **Active**, or **Inactive** (taking together `Committed` and `Recovering` sectors). Specifically, active storage is used in SPC (and in-deal active storage counts toward {{<sref storage_power>}}) whereas inactive storage is not. Both active and inactive storage is tracked by the system and affect a miner's ability to serve as a storage provider in the Filecoin Storage Market.
```
zixuanzh: Blocking: and also `Failing` for Inactive. The distinction here is that:
ProvingSet: Active, Committed, Recovering
FailingSet (not such explicit concept): Failing
ActivePower: Active
InactivePower: Committed, Recovering, Failing

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: Deleted paragraph, will take care of in futre PR

src/systems/filecoin_mining/storage_mining/_index.md
sternhenri: why not move `Failing` into `Active` automatically in that instance is what I mean?

PR #708, opened 2019-12-05
Remove GetSignaturePublicKey from Actor

anorth:
Remove GetSignaturePublicKey (unimplemented) from Actor. The missing implementation obscured the fact that this can't be implemented here. GetSignaturePublicKey does not belong on the base actor. Only account actors have keys, and not even all of them do. The account actor's address/key is stored in the account actor's substate.

If we need on-chain access to keys, the actor account might need a method to either fetch the key or verify a signature. If only off-chain, the VM can inspect the account actor's state.

Also remove legacy cron reserved method number.

### DIFF
--- a/src/systems/filecoin_vm/actor/actor.id
+++ b/src/systems/filecoin_vm/actor/actor.id
@@ -1,7 +1,6 @@
 // This contains actor things that are _outside_ of VM exection.
 // The VM uses this to execute actors.
 
-import filcrypto "github.com/filecoin-project/specs/algorithms/crypto"
 import ipld "github.com/filecoin-project/specs/libraries/ipld"
 
 // TokenAmount is an amount of Filecoin tokens. This type is used within
@@ -74,24 +73,18 @@ type BuiltinActorID enum {
 // to Actors in the Actor Model (programming), or Objects in Object-
 // Oriented Programming, or Ethereum Contracts in the EVM.
 //
-// Questions for id language:
-// - we should not do inheritance, we should do composition.
-//   but we should make including actor state nicer.
-//
-// ActorState represents the on-chain storage actors keep.
+// ActorState represents the on-chain storage all actors keep.
 type ActorState struct {
-    // common fields for all actors
-
+    // Identifies the code this actor executes.
     CodeID
-    // use a CID here, load it in interpreter.
-    // Alternative is to use ActorState here but tricky w/ the type system
-    State                    ActorSubstateCID
-
-    Balance                  TokenAmount
-    CallSeqNum  // FKA Nonce
-
-    // Note: may be nil if actor has no public key
-    GetSignaturePublicKey()  filcrypto.PublicKey
+    // CID of the root of optional actor-specific sub-state.
+    State       ActorSubstateCID
+    // Balance of tokens held by this actor.
+    Balance     TokenAmount
+    // Expected sequence number of the next message sent by this actor.
+    // Initially zero, incremented when an account actor originates a top-level message.
+    // Always zero for other actors.
+    CallSeqNum
 }
 
 type ActorSystemStateCID ipld.CID
--- a/src/systems/filecoin_vm/runtime/runtime.go
+++ b/src/systems/filecoin_vm/runtime/runtime.go
@@ -1,7 +1,6 @@
 package runtime
 
 import (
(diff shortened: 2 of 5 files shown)


src/systems/filecoin_vm/actor/actor.id
```diff
@@ -74,24 +73,18 @@ type BuiltinActorID enum {
...
-
-    // Note: may be nil if actor has no public key
-    GetSignaturePublicKey()  filcrypto.PublicKey
+    // CID of the root of optional actor-specific sub-state.
+    State       ActorSubstateCID
+    // Balance of tokens held by this actor.
+    Balance     TokenAmount
+    // Expected sequence number of the next message sent by this actor.
+    // Initially zero, incremented when an account actor originates a top-level message.
+    // Always zero for other actors.
```
jzimmerman: Note: also incremented for ControlActor (or whichever actor we want to use for this role); see e.g. `_applyMessageBuiltinAssert` in vm_interpreter.

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -522,19 +521,6 @@ func (rt *VMContext) AcquireState() ActorStateHandle {
...
-	if st == nil {
-		rt.AbortAPI("VerifySignature: signer actor not found")
-	}
-	pk := st.GetSignaturePublicKey()
-	if pk == nil {
-		rt.AbortAPI("VerifySignature: signer actor has no public key")
-	}
-	ret := rt.Compute(ComputeFunctionID_VerifySignature, []Any{pk, sig, m})
-	return ret.(bool)
-}
```
jzimmerman: [Blocking] I think we need this method. Actors clearly need to verify each other's signatures, and we don't want to do this by sending a message to the originating actor asking it to verify its own signature. Maybe it could just abort if the address provided isn't an AccountActor?

jzimmerman: (please consider the PR approved conditional on this -- GitHub doesn't seem to let me update the review state to reflect that)

src/systems/filecoin_vm/actor/actor.id
anorth: The system actor will be an account actor, so this remains true.

src/systems/filecoin_vm/runtime/runtime.go
anorth: > Actors clearly need to verify each other's signatures

When? I removed this runtime call without any actor code breaking. I don't think we should specify runtime methods that are not used. What's the use case for in-actor signature verification?
- message and block signatures are checked prior to evaluation, assumed correct within the VM (which can't see the signatures)
- tickets and election post are verified outside actor code

There could well be something, but we should know what it is before implementing this (correctly).

jzimmerman: I'm ok with removing this for now, since we're refactoring the checking of actor principals anyway. Ideally with the new design we won't require any signature verifications, since the origin of a message (as supplied by the VM context) is sufficient authentication. There might be some corner cases we're missing, though.

src/systems/filecoin_vm/actor/actor.id
jzimmerman: Ah, didn't realize the control/system actor was intended to be an account actor. Is there a reason this has to be the case? It seems confusing (and potentially hazardous), given that it places the system actor in the same category as the current family of AccountActors that may be user-created.

anorth: Well, it's precisely an actor with no code and with `CallSeqNum` that increments. The system right now has an invariant that only account actors can inject messages into the system, and that's exactly what we need to do.

IMO it's simpler than having a new type that behaves exactly the same way as an existing thing, and updating all our invariants about that to be "xxx _and the control actor_".

jzimmerman: I'm not sure about that -- for one thing, the control actor shouldn't have a public key, whereas all account actors do. Also, there are a number of places where we check (or intend to check) that an actor is an AccountActor, and in such a case, the control actor would unintentionally pass that check.

sternhenri: yikes. Sorry I jumped the gun. Let's finish discussion here so we can PR against master when you have a resolution.

PR #711, opened 2019-12-05
Spec/network name

ZenGround0:
In this PR
- [x] Land the change proposed here: https://github.com/filecoin-project/specs/pull/456, putting the network name inside the init actor's state.
- [x] Spec the network protocol id requirement differentiating network protocols (1) from non-filecoin networks (2) filecoin networks with different network names

### DIFF
--- a/src/systems/filecoin_nodes/network/_index.md
+++ b/src/systems/filecoin_nodes/network/_index.md
@@ -10,13 +10,27 @@ Filecoin nodes use the libp2p protocol for peer discovery, peer routing, and mes
 
 Here is the list of libp2p protocols used by Filecoin.
 
-- Graphsync: [Draft spec](https://github.com/ipld/specs/blob/master/block-layer/graphsync/graphsync.md)
-- Bitswap:  TODO
-- Gossipsub: block headers and messages are broadcasted through a Gossip PubSub protocol where nodes can subscribe to topics such as `NewBlock`, `BlockHeader`, `BlockMessage`, etc and receive messages in those topics. When receiving messages related to a topic, nodes processes the message and forwards it to its peers who also subscribed to the same topic.  Spec is [here](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub)
-- KademliaDHT: Kademlia DHT is a distributed hash table with a logarithmic bound on the maximum number of lookups for a particular node. Kad DHT is used primarily for peer routing as well as peer discovery in the Filecoin protocol.  Spec TODO [reference implementation](https://github.com/libp2p/go-libp2p-kad-dht)
-- Bootstrap List: Bootstrap is a list of nodes that a new node attempts to connect upon joining the network. The list of bootstrap nodes and their addresses are defined by the users.
-- Peer Exchange: spec [TODO](https://github.com/libp2p/specs/issues/222)
+- Graphsync: 
+	- Graphsync is used to transfer blockchain and user data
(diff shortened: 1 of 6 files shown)


src/systems/filecoin_nodes/network/_index.md
```diff
@@ -17,4 +17,6 @@ Here is the list of libp2p protocols used by Filecoin.
 - Bootstrap List: Bootstrap is a list of nodes that a new node attempts to connect upon joining the network. The list of bootstrap nodes and their addresses are defined by the users.
 - Peer Exchange: spec [TODO](https://github.com/libp2p/specs/issues/222)
 - DNSDiscovery: Design and spec needed before implementing
-- HTTPDiscovery: Design and spec needed before implementing 
\ No newline at end of file
+- HTTPDiscovery: Design and spec needed before implementing 
+
+Filecoin network protocols should be differentiated from non-filecoin network protocols.  Additionally different filecoin networks should differentiate protocols from each other.  To achieve this filecoin network protocol IDs MUST adhere to the following format: `/fil/<libp2p-protocol>/<network-name>`.  For example: `fil/kad-dht/mainnet`.
```
whyrusleeping: We should definitely not do this for all protocols, things like bitswap, graphsync, and the hello protocol should work between all peers.

Maybe this makes sense for other protocols like deal making, or payment channel negotiation, or retrieval. But I think a better solution would be to include the network identifier in the message itself instead (though i feel less strongly about those cases)

ZenGround0: Any opinions on gossipsub, and dht?

ZenGround0: Also it sounds like this objection has to do with interop between filecoin and non-filecoin nodes.  It would be great to hear a quick sketch of intended use cases for the network that motivate this in the particular protocols you have flagged.  I'm guessing this is IPFS interop related?

anorth: I concur that we should not namespace protocols that are not specific to a filecoin network, including bitswap and gossipsub.

I think we *should* namespace:
- the DHT
- Hello
- The message and block pubsub channels and blocksync
- Deal exchange, retrieval and data transfer (probably)

@ZenGround0 could you please expand this to an explicit list of protocols (either here or where the protocols are or should be defined)

@whyrusleeping we seem to disagree about hello protocol. What's the use case where we'd want nodes from different networks (e.g. a testnet and mainnet) be able to exchange hello messages?

ZenGround0: Heard back indirectly regarding hello.  @whyrusleeping wants to keep network name in hello message, potentially for error handling.

Benefit still unclear clear but this is a minor difference and trying to optimize by shutting down hello connections at protocol id reading time looks premature.  Also we need to spec the hello protocol.

ZenGround0:
Ping @anorth ready for review after syncing up with @mishmosh.  @whyrusleeping has some ideas for hello protocol that I don't care much about one way or the other.  Best reasons against seem like optimizations we will be able to leave for the future.

ZenGround0:
> Regarding deal exchange and retrieval I really do think we want those protocols scoped by network name.

Thanks, I'll add this in.

ZenGround0:
@anorth there is no mention of any libp2p protocols in the storage deal section of the spec and adding that section seems out of scope so I'm leaving that out of this PR.

PR #712, opened 2019-12-05
fix(deps): use current released mermaid

hannahhoward:
# Goals

Use the current mermaid.js that is reflected in online documentation

# Implementation

So the origin of this is I was trying to use this lovely feature in a sequence diagram: https://mermaid-js.github.io/mermaid/#/sequenceDiagram?id=background-highlighting -- and found my diagrams would not compile. A lot of digging helped me identify that mermaid.cli is hard coded to use mermaid 8.0.0 -- a super old version (~2 years). Unfortunately, mermaid.cli appears to have gone stale. So:

- I forked mermaid.cli, and used some ideas of another fork to move away from having a fixed version.
- This allows us to move the mermaid version as needed
- I can submit back but the project is as I said, dead

# For discussion

- Ack using a forked version of a package! Not great. Alternatively, we can move it over to filecoin-shipyard and publish it under the name space but that probably needs @alanshaw ? who I think owns that repo.

- Ack I hate the javascript ecosystem!


jbenet:
yay thanks for updating, that probably fixed other problems too that we were running into w/ the older mermaid version. and yeah im fine w/ the forked package

alanshaw:
Sorry I can't help you I have no admin on `filecoin-shipyard`. @mishmosh and @pooja are owners.

alanshaw:
> Ack I hate the javascript ecosystem!

Hey!

PR #717, opened 2019-12-08
Miners smaller than min miner size don't count toward total power

sternhenri:
Per your comment Magik.

**In this PR**
- miner storage for miners whose active power < min miner size does not count toward system total power

Let me know if this is what you intend. Further looking a bit more into it @magik6k, we have Storage Miner Actor's `_getActivePower` method which adds up its sector's power and then uses `submitPowerReport` to update the miner's `ActivePower` in their PT `PowerTableEntry`. FYI


src/algorithms/post/election_post.md
```diff
@@ -208,6 +208,6 @@ While it resets a miner’s ElectionWindow, miners earn no reward from submittin
 
 Storage Power Consensus participants are subject to a {{<sref min_miner_size>}}, meaning miners smaller than `MIN_MINER_SIZE_STOR` of active (or in-deal) storage or whose active storage represents less than `MIN_MINER_SIZE_PERC`% of the network's cannot produce valid electionPoSts. 
 
-However, these miners still have active power and can still run and transmit SurprisePosts as messages to be added on-chain. These miners can also be faulted as usual for lacking to prove their power in a proving period. 
+These miners' power does not count as part of the total network power, nor are they able to sumit electionPoSts but they can still run and transmit SurprisePosts as messages to be added on-chain. These miners can also be faulted as usual for lacking to prove their power after `MAX_FAULT_COUNT` challenges.
```
zixuanzh: Blocking: These miners can also be faulted as usual *and be terminated* for failing in `MAX_FAULT_COUNT` consecutive epochs.

sternhenri: good catch. If you're ok I'll actually simplify by removing max fault count here since it's mentioned in a ton of places.

PR #719, opened 2019-12-09
Add block size and gas limit to validation

anorth:


src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -42,6 +42,8 @@ A syntactically valid full block must have:
 
 - all referenced messages syntactically valid,
 - all referenced parent receipts syntactically valid,
+- the sum of the serialized sizes of all messages is no greater than `block.BlockMaxSize`,
```
sternhenri: Does it not include the header ie
```suggestion
- the sum of the serialized sizes of all messages and the header is no greater than `block.BlockMaxSize`,
```

src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -42,6 +42,8 @@ A syntactically valid full block must have:
...
+- the sum of the gas limit of all messages is no greater than `block.BlockGasLimit`.
```
sternhenri: may want to remind reader to include cron-activated messages (i.e. implicit messages) w gas cost

sternhenri:
Also fixing of compilation obviously :b

src/systems/filecoin_blockchain/struct/block/_index.md
anorth: I intend to explicitly exclude the implicit messages, which will have large, fixed gas limits.

PR #721, opened 2019-12-09
Defined the SystemActor as sender of implicit messages.

anorth:
Interpreter code now constructs the implicit messages in full (except EPost params).
System message receivers verify the sender account is the system actor.
Defined concrete addresses for all singleton actors.

### DIFF
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	Method_StorageMinerActor_SubmitSurprisePoSt = actor.MethodPlaceholder + iota
+	Method_StorageMinerActor_SubmitVerifiedSurprisePoSt = actor.MethodPlaceholder + iota
 	Method_StorageMinerActor_SubmitVerifiedElectionPoSt
 	Method_StorageMinerActor_NotifyOfSurprisePoStChallenge
 )
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
@@ -18,6 +18,7 @@ import (
 
 const (
 	Method_StorageMinerActor_SubmitSurprisePoSt = actor.MethodPlaceholder + iota
+	Method_StorageMinerActor_SubmitVerifiedElectionPoSt
 	Method_StorageMinerActor_NotifyOfSurprisePoStChallenge
 )
 
@@ -357,11 +358,9 @@ func (a *StorageMinerActorCode_I) SubmitVerifiedSurprisePoSt(rt Runtime, onChain
 // Assume ElectionPoSt has already been successfully verified when the function gets called.
 // Likewise assume that the rewards have already been granted to the storage miner actor. This only handles sector management.
 func (a *StorageMinerActorCode_I) SubmitVerifiedElectionPoSt(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo) InvocOutput {
-
-	// TODO: validate caller
-	// the caller MUST be the miner who won the block (who won the block should be callable as a a VM runtime call)
-	// we also need to enforce that this call happens only once per block, OR make it not callable by special privileged messages
-	TODO()
+	rt.ValidateImmediateCallerIs(addr.SystemActorAddr)
+	// The receiver must be the miner who produced the block for which this message is created.
+	Assert(rt.ToplevelBlockWinner() == rt.CurrReceiver())
 
 	// we do not need to verify post submission here, as this should have already been done
 	// outside of the VM, in StoragePowerConsensus Subsystem. Doing so again would waste
--- a/src/systems/filecoin_vm/interpreter/_index.md
+++ b/src/systems/filecoin_vm/interpreter/_index.md
@@ -26,7 +26,8 @@ For each tipset:
 
 - the `CronActor`'s tick method is invoked implicitly after all the blocks' messages.
 
-These implicit messages specify a gas price of zero, but must be included in the computation.
+These implicit messages have a `From` address being the distinguished system account actor.
+They specify a gas price of zero, but must be included in the computation.
 They must succeed (have an exit code of zero) in order for the new state to be
 computed. Receipts for them are excluded from the receipt list; only explicit messages have an 
 explicit receipt. 
--- a/src/systems/filecoin_vm/interpreter/_index.md
+++ b/src/systems/filecoin_vm/interpreter/_index.md
@@ -11,36 +11,49 @@ producing a new state and a sequence of message receipts. The CIDs of this new s
 collection are included in blocks from the subsequent epoch, which must agree about those CIDs 
 in order to form a new tipset.
 
+Every state change is driven by the execution of a message.
 The messages from all the blocks in a tipset must be executed in order to produce a next state.
 All messages from the first block are executed before those of second and subsequent blocks in the
 tipset. For each block, BLS-aggregated messages are executed first, then SECP signed messages.
 
-In addition, for each block:
+# Implicit messages
 
-- the block reward is paid to the miner owner account, and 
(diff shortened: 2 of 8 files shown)


src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -18,6 +18,7 @@ import (
 
 const (
 	Method_StorageMinerActor_SubmitSurprisePoSt = actor.MethodPlaceholder + iota
```
sternhenri:
```suggestion
	Method_StorageMinerActor_SubmitVerifiedSurprisePoSt = actor.MethodPlaceholder + iota
```

Note that we may want to change it, but current assumption is that the verifier will use the subsystem verification methods ahead of submitting the messages. We could do all in one go.

src/systems/filecoin_vm/actor/actor.go
```diff
@@ -11,7 +11,6 @@ type Serialization = util.Serialization
 const (
 	MethodSend        = MethodNum(0)
 	MethodConstructor = MethodNum(1)
-	MethodCron        = MethodNum(2)
```
sternhenri: This is so we can explicitly include them as coming from the `cronactor`?

src/systems/filecoin_vm/interpreter/_index.md
```diff
@@ -26,7 +26,8 @@ For each tipset:
 
 - the `CronActor`'s tick method is invoked implicitly after all the blocks' messages.
 
-These implicit messages specify a gas price of zero, but must be included in the computation.
+These implicit messages have a `From` address being the distinguished system account actor.
+They specify a gas price of zero, but must be included in the computation.
```
sternhenri: The decision was made to make them free? How do we deal with overflows or long-running computation from cron (for future PR I reckon)

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -71,7 +70,7 @@ func (vmi *VMInterpreter_I) ApplyTipSetMessages(inTree st.StateTree, msgs TipSet
 		}
 
 		// Pay block reward.
-		rewardMessage := _makeBlockRewardMessage(outTree, minerOwner, minerPenaltyTotal)
+		rewardMessage := _makeBlockRewardMessage(outTree, blk.Miner(), minerPenaltyTotal)
```
sternhenri: must ensure that blk.Miner() in block formation is set as referring to the owner (and not the worker). I believe there are some conflicts with code elsewhere where blk.Miner() is used to get workerKeys() (e.g. in the PoSt verification methods I wrote). Let's agree on a way to resolve that before we merge this in...

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -285,72 +280,58 @@ func _gasToFIL(gas msg.GasAmount, price msg.GasPrice) actor.TokenAmount {
...
-	//	Method_:     StorageMinerActor.SubmitElectionPoSt,
-	//	Params_:     serialize([PoStSubmission{postProof, epoch}]),
-	//	CallSeqNum_: minerWorker.CallSeqNum,
-	//	Value_:      0,
-	//	GasPrice_:   0,
-	//	GasLimit_:   sumbitElectionPostGasLimit,
-	//}
-	panic("TODO: implement when necessary dependencies are importable")
+func _makeElectionPoStMessage(state st.StateTree, minerActorAddr addr.Address, epoch UInt64) msg.UnsignedMessage {
+	// TODO: determine parameters necessary for this message.
```
sternhenri: What input is needed (proofs convo end?)?

src/systems/filecoin_vm/sysactors/init_actor.go
```diff
@@ -58,7 +58,7 @@ func (a *InitActorCode_I) Constructor(rt Runtime) InvocOutput {
 	h := rt.AcquireState()
 	st := &InitActorState_I{
 		AddressMap_: map[addr.Address]addr.ActorID{}, // TODO: HAMT
-		NextID_:     addr.ActorID(0),
+		NextID_:     addr.ActorID(addr.FirstNonSingletonActorId),
```
sternhenri: Shouldn't this get incremented afterward?

src/systems/filecoin_vm/sysactors/reward_actor.id
```diff
@@ -22,8 +22,10 @@ type RewardActorState struct {
 type RewardActorCode struct {
     Constructor(rt vmr.Runtime)
 
-    // put Reward in RewardMap
-    MintReward(rt vmr.Runtime)
+    // Allocates a block reward to the owner of a miner, less
+    // - penalty, which is burnt
```
sternhenri:
```suggestion
    // - penalty used to pay for some messages executed in blocks which is burnt
```

Or something correct. I.e. where does penalty come from?

src/systems/filecoin_vm/sysactors/reward_actor.go
```diff
@@ -70,7 +74,8 @@ func (a *RewardActorCode_I) Constructor(rt vmr.Runtime) InvocOutput {
 	panic("TODO")
 }
 
-func (a *RewardActorCode_I) MintReward(rt vmr.Runtime) {
+func (a *RewardActorCode_I) AwardBlockReward(rt vmr.Runtime, miner addr.Address, penalty actor.TokenAmount) {
```
sternhenri: should add a placeholder to handle penalty larger than blockReward (as deleted from `vm_interpreter.go` L296)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
anorth: Done (this is unused anyway).

src/systems/filecoin_vm/actor/actor.go
anorth: This is debt. The CronActor has already been updated to be able to call any method registered with it. We don't need a special reserved method id.

src/systems/filecoin_vm/interpreter/_index.md
anorth: Yes. Overflows are not handled yet (the gas limit is really, really big but we can probably do better - add it to the list of parameters TBD).

src/systems/filecoin_vm/interpreter/vm_interpreter.go
anorth: Right here we're just sending a message to the RewardActor. Responsibility for looking up the owner lies at the receiver. It will need the actor address in order to check for under-collateralisation etc. 

IMO the actor address should be passed ~everywhere, and resolved to an owner or worker last minute.

anorth: Wait, I might have misunderstood. `blk.Miner` needs to be the miner *actor* address, which is the only unique identifier.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
anorth: You tell me!

src/systems/filecoin_vm/sysactors/init_actor.go
anorth: This is just initialization. `Exec` increments it when a new actor is created.

src/systems/filecoin_vm/sysactors/reward_actor.id
anorth: It comes from what would have been paid to the miner as block reward. This actor doesn't care why.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
sternhenri: yes, ok I agree with that. I expect some cleanup will be required elsewhere...

src/systems/filecoin_vm/interpreter/vm_interpreter.go
sternhenri: will handle in #724 once alignment is found

anorth:
FYI @icorderi

PR #723, opened 2019-12-09
fix target calculation in readme and code

sternhenri:
EC Target calculation fixed per lotus bug (see https://github.com/filecoin-project/lotus/pull/808)

Specifically, rather than use the `sampleRate` to offset sampling (i.e. advantaging miners by giving them extra power).
e.g. for `sampleRate = 1/25`, miner with 1 sector gets same power as miner with 2 sectors gets the same as miner with 24 sectors), we use the numMinerSectors/numSectorsSampled to account for total sectors appropriately.


src/algorithms/expected_consensus/_index.md
```diff
@@ -77,12 +77,22 @@ As discussed in {{<sref election_post>}}, a miner will use the challenge ticket
 
 The miner gets to draw one such challenge ticket per sector they have committed and must then compare the value derived from the challenge ticket against a target to determine whether they are eligible to mine. This is called finding a winning ticket.
 
-The target is set as follows, for each ticket and its associated sector:
-`target = activePowerInSector/networkPower * sectorsSampled * EC.ExpectedLeaders`.
+The target is set as follows, for each sector sampled for a ticket from the miner's `ProvingSet` and the number of sectors in the set:
```
zixuanzh: I think this should be `ActiveSectors` see comments below, either case there is an inconsistency.

PR #724, opened 2019-12-09
[WIP] remove unnecessary params from post verifs

sternhenri:
Remove `OnChainPoStVerifyInfo` from `_onSuccessfulPoSt` and upstream methods

### DIFF
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
@@ -123,8 +123,8 @@ type StorageMinerActorCode struct {
     PreCommitSector(rt Runtime, info sector.SectorPreCommitInfo)  // TODO: check with Magik on sizes
     ProveCommitSector(rt Runtime, info sector.SectorProveCommitInfo)
 
-    SubmitVerifiedSurprisePoSt(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo)
-    SubmitVerifiedElectionPoSt(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo)
+    SubmitVerifiedSurprisePoSt(rt Runtime)
+    SubmitVerifiedElectionPoSt(rt Runtime)
 
     _checkSurprisePoStSubmissionHappened(rt Runtime)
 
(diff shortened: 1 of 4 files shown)


src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -258,7 +258,7 @@ func (a *StorageMinerActorCode_I) _claimDealPayments(rt Runtime) {
 //       - Failing / Recovering / Active / Committed -> Cleared
 //     - Remove SectorNumber from Sectors, ProvingSet
 // - Update ChallengeEndEpoch
-func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo) InvocOutput {
+func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime) InvocOutput {
 	h, st := a.State(rt)
 
 	// TODO add info on chain
```
anorth: If this TODO will require at least some parameter, I suggest we add that now rather than over-simplify.

For context, I need to thread whatever is required here from the block headers through the interpreter to construct the implicit EPost processing method.

sternhenri: Agree. Halted this PR after I read through more carefully (but not before i put it up. Derp)

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
```diff
@@ -125,8 +125,8 @@ type StorageMinerActorCode struct {
     PreCommitSector(rt Runtime, info sector.SectorPreCommitInfo)  // TODO: check with Magik on sizes
     ProveCommitSector(rt Runtime, info sector.SectorProveCommitInfo)
 
-    SubmitVerifiedSurprisePoSt(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo)
-    SubmitVerifiedElectionPoSt(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo)
+    SubmitVerifiedSurprisePoSt(rt Runtime)
+    SubmitVerifiedElectionPoSt(rt Runtime)
```
anorth: Maybe `Submit` is not the best name here - nothing is being put up for any kind of inspection. `Handle` or `Process`?

PR #725, opened 2019-12-10
Clean up state tree to be just a HAMT of actor states.

anorth:
Replace incorrect implicit receiver actor creation with a placeholder and comment.
Fix accidental persistence of CallSeqNum and gas burn when implicit receiver creation fails.

FYI @icorderi

### DIFF
--- a/src/systems/filecoin_vm/interpreter/vm_interpreter.go
+++ b/src/systems/filecoin_vm/interpreter/vm_interpreter.go
@@ -310,7 +310,8 @@ func _makeBlockRewardMessage(state st.StateTree, minerAddr addr.Address, penalty
 	params[0] = addr.Serialize_Address(minerAddr)
 	params[1] = actor.Serialize_TokenAmount(penalty)
 
-	sysActor := state.GetActorState(addr.SystemActorAddr)
+	sysActor, ok := state.GetActor(addr.SystemActorAddr)
+	Assert(ok)
 	return &msg.UnsignedMessage_I{
 		From_:       addr.SystemActorAddr,
 		To_:         addr.RewardActorAddr,
@@ -328,7 +329,8 @@ func _makeElectionPoStMessage(state st.StateTree, minerActorAddr addr.Address) m
 	// TODO: determine parameters necessary for this message.
 	params := make([]util.Serialization, 0)
 
-	sysActor := state.GetActorState(addr.SystemActorAddr)
+	sysActor, ok := state.GetActor(addr.SystemActorAddr)
+	Assert(ok)
 	return &msg.UnsignedMessage_I{
 		From_:       addr.SystemActorAddr,
 		To_:         minerActorAddr,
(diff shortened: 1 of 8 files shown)


src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -174,32 +180,27 @@ func (vmi *VMInterpreter_I) ApplyMessage(
 		return
 	}
 
-	// make sure this is the right message order for fromActor
-	if message.CallSeqNum() != fromActor.CallSeqNum() {
-		_applyError(inTree, exitcode.InvalidCallSeqNum, SenderResolveSpec_Invalid)
+	// Check receiving actor and method exists, possibly creating an account actor.
+	// If this succeeds, compTreePreSend will become a state snapshot which includes
+	// implicit receiver creation, the sender (rather than miner) paying gas, and the sender's
+	// CallSeqNum being incremented; at least that much state change will be persisted even if the
```
sternhenri: What's the rationale for persisting that state?

anorth: Someone has to pay gas. This marks the transition from the miner being on the hook for a bad message, to the sender for a possibly-ok one.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -225,6 +226,29 @@ func (vmi *VMInterpreter_I) ApplyMessage(
...
+	}
+	if !message.To().IsKeyType() {
+		// Don't implicitly create an account actor for an address without an associated key.
+		return tree, false
+	}
+	if message.Method() != actor.MethodSend {
+		// Don't implicitly create account actor if message was expecting to invoke a method.
+		return tree, false
+	}
+	// TODO Create a new account actor via the InitActor, which will receive a new ID address
```
jzimmerman: nit: would be better to use `util.TODO()` or `panic` here, to make absolutely clear that this TODO is mandatory

src/systems/filecoin_vm/runtime/runtime.go
```diff
@@ -429,7 +433,7 @@ func (rtOuter *VMContext) _sendInternal(input InvocInput, errSpec ErrorHandlingS
 		rtOuter._toplevelSender,
 		rtOuter._toplevelBlockWinner,
 		rtOuter._toplevelSenderCallSeqNum,
-		rtOuter._internalCallSeqNum+1,
+		rtOuter._internalCallSeqNum,
```
jzimmerman: [Blocking] Why was the +1 removed? The internal CallSeqNum needs to be incremented somewhere, independently from the toplevel CallSeqNum.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
anorth: The implementation is coming in the direct follow-up, so I beg off this time.

src/systems/filecoin_vm/runtime/runtime.go
anorth: Sorry, I misread. Reverted.

PR #727, opened 2019-12-10
Enforce that SendCatchingErrors is callable only by Cron actor.

anorth:


anorth:
@sternhenri @jzimmerman I would like to rename `SendPropagatingErrors` to just `Send`, since it's unambiguous for all callers except CronActor.

sternhenri:
I like the duality of propagating vs catching errors but up to you

PR #733, opened 2019-12-11
Patch/min miner size

sternhenri:
- added a note about future solutions to network reboot in case of total loss of liveness
- changed min miner rule to be about largest X miners rather than a percentage

Note that: 
- I don't show impl how to set `_largestMinerSize`
- conversely I do show how to check if miner is in top `MIN_MINER_SIZE_TARG` of the PT

Q: @anorth should I replace the entire body with `IMPL()` to signal that it's impl-choice, or conversely should I show how you set _lMS?

### DIFF
--- a/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
+++ b/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
@@ -15,13 +15,18 @@ func (st *StoragePowerActorState_I) ActivePowerMeetsConsensusMinimum(minerPower 
 	MIN_MINER_SIZE_STOR := block.StoragePower(0)
 	MIN_MINER_SIZE_TARG := 0
 
-	// if a miner meets min power requirement check whether this one does
+	// if miner is larger than min power requirement, we're set
+	if minerPower >= MIN_MINER_SIZE_STOR {
+		return true
+	}
+
+	// otherwise, if another miner meets min power requirement check whether this one does
 	if st._minersLargerThanMin() > util.UVarint(0) {
 		return minerPower >= MIN_MINER_SIZE_STOR
 	}
 
 	// else check whether in MIN_MINER_SIZE_TARG miners
-	if len(st.PowerTable()) < MIN_MINER_SIZE_TARG {
+	if len(st.PowerTable()) <= MIN_MINER_SIZE_TARG {
 		// miner should pass
 		return true
 	}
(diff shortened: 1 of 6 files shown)


src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
```diff
@@ -9,17 +11,28 @@ import (
...
-	MIN_MINER_SIZE_PERC := 0
+	MIN_MINER_SIZE_TARG := 0
+
+	// if a miner meets min power requirement check whether this one does
+	if st._minersLargerThanMin() > util.UVarint(0) {
+		return minerPower >= MIN_MINER_SIZE_STOR
+	}
+
+	// else check whether in MIN_MINER_SIZE_TARG miners
+	if len(st.PowerTable()) < MIN_MINER_SIZE_TARG {
```
anorth:
```suggestion
	if len(st.PowerTable()) <= MIN_MINER_SIZE_TARG {
```

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
```diff
@@ -9,17 +11,28 @@ import (
...
-	// if miner smaller than both min size in bytes and min percentage
-	if (int(minerPower)*MIN_MINER_SIZE_PERC < int(totPower)*100) && minerPower < MIN_MINER_SIZE_STOR {
-		return false
+	// get size of MIN_MINER_SIZE_TARGth largest miner
+	minerSizes := make([]block.StoragePower, 0, len(st.PowerTable()))
+	for _, v := range st.PowerTable() {
+		minerSizes = append(minerSizes, v.ActivePower())
 	}
-	return true
+	sort.Slice(minerSizes, func(i, j int) bool { return int(i) < int(j) })
```
anorth:
```suggestion
	sort.Slice(minerSizes, func(i, j int) bool { return int(i) > int(j) })
```
... for descending order

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
```diff
@@ -9,17 +11,33 @@ import (
...
+	MIN_MINER_SIZE_TARG := 0
+
+	// if miner is larger than min power requirement, we're set
+	if minerPower >= MIN_MINER_SIZE_STOR {
+		return true
+	}
+
+	// otherwise, if another miner meets min power requirement check whether this one does
+	if st._minersLargerThanMin() > util.UVarint(0) {
+		return minerPower >= MIN_MINER_SIZE_STOR
```
anorth: This must be `false`

sternhenri: thanks for that. derp.

PR #729, opened 2019-12-10
Add pieces and sectors diagrams

mishmosh:
- Add pieces diagram and intro text to 2.2.2 Piece.
- Add sectors diagram and intro text to 2.6.2 Sector.

PR #734, opened 2019-12-11
per @vmx review of #729. Thanks

sternhenri:
- typo
- adding encoding format


sternhenri:
the small `(using CBOR)`

PR #735, opened 2019-12-12
remove sector utilization

zixuanzh:
**In this PR**
- [x] Every sector now contains an Activation epoch (when ProveCommit hits the chain) and an Expiration epoch (when the last deal in the sector expires)
- [x] Every sector is assigned a power that remains constant throughout its life time, this power is a function of duration of the sector, size of the sector, and deals' start and end times within the sector
- [x] Remove SectorUtilization from StorageMiner and DealExpirationAMT from Deal

**EDIT: In future PR:**
- [ ] Deal payments and expiration are evaluated lazily but remaining payment and expiration will be done at sector expiration  (still need some work)

### DIFF
--- a/src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
+++ b/src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
@@ -113,7 +113,7 @@ func (a *StorageMarketActorCode_I) AddBalance(rt Runtime) {
 
 func (a *StorageMarketActorCode_I) PublishStorageDeals(rt Runtime, newStorageDeals []deal.StorageDeal) []PublishStorageDealResponse {
 
-	TODO() // verify StorageMinerActor
+	TODO() // verify StorageMinerActor (note client can put the signed message onchain too)
 
 	h, st := a.State(rt)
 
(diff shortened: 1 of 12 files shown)


src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -187,12 +188,24 @@ func (a *StorageMarketActorCode_I) ActivateDeals(rt Runtime, dealIDs []deal.Deal
...
+
+		if dealEnd > lastDealExpiration {
+			lastDealExpiration = dealEnd
+		}
+
+		dealPs = append(dealPs, dealP)
+		st._activateDeal(rt, publishedDeal)
 	}
 
+	st._assertEpochEqual(rt, lastDealExpiration, sectorExpiration)
```
jzimmerman: Should be less-than-or-equal?

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -225,11 +238,15 @@ func (a *StorageMarketActorCode_I) ProcessDealSlash(rt Runtime, dealIDs []deal.D
...
+	TODO() // verify caller must be StorageMinerActor
+
 	h, st := a.State(rt)
 
 	for _, dealID := range dealIDs {
-		deal := st._safeGetOnChainDeal(rt, dealID)
-		st._assertActiveDealState(rt, dealID)
+		deal, ok := st._getOnChainDeal(dealID)
+		if !ok {
+			continue
```
jzimmerman: Is this failure allowed (e.g. because the deal has already expired, but we are computing expiration lazily)? If so, would be good to add a comment

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
```diff
@@ -148,8 +143,13 @@ type StorageMinerActorCode struct {
         faultType      sector.StorageFaultType
     )
 
-    _claimDealPayments(rt Runtime)
+    ClaimDealPaymentsForSector(
```
jzimmerman: I think this doesn't need to be a public method - it's subsumed by `Withdraw`?

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
zixuanzh: currently, we expect sector Activation to be when ProveCommit happens and sector Expiration to be when the last deal expires, hence the equality check.

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
zixuanzh: yes, it depends on if we delete the deal when it expires which i think we should. This is left for a follow-up.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
zixuanzh: Potentially, we still need a way for StorageMinerActor to trigger payment updates in the BalanceTable. We should touch this up in a follow-up.

PR #736, opened 2019-12-12
Add CPU cost.

porcuquine:
Add CPU cost based on the benchmark machine we use: https://www.packet.com/cloud/servers/g2-large/


nicola:
added in cbe57d67

PR #737, opened 2019-12-13
add soft finality check to block validation

sternhenri:
I looked through it @anorth and thought it wouldn't be that additive to add this in code, given the rest of block val is all in md (and clear as is). Will code up if you disagree.

### DIFF
--- a/src/systems/filecoin_blockchain/struct/block/_index.md
+++ b/src/systems/filecoin_blockchain/struct/block/_index.md
@@ -66,6 +66,7 @@ A semantically valid block must have:
 - `Epoch` greater than that of its parents, and 
     - not in the future according to the node's local clock reading of the current epoch,
     - not farther in the past than the soft finality as defined by SPC ${{<sref finality>}},
+        - this rule only applied when receiving new gossip blocks (i.e. from the current chain head), not when syncing to the chain for the first time (e.g.)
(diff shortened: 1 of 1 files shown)


src/systems/filecoin_blockchain/struct/block/_index.md
```diff
@@ -63,7 +63,9 @@ A semantically valid block must have:
 - `ParentState` matching the state tree produced by executing the parent tipset's messages (as defined by the VM interpreter) against that tipset's parent state,
 - `ParentMessageReceipts` identifying the receipt list produced by parent tipset execution, with one receipt for each unique message from the parent tipset, 
 - `ParentWeight` matching the weight of the chain up to and including the parent tipset,
-- `Epoch` greater than that of its parents, and not in the future according to the node's local clock reading of the current epoch,
+- `Epoch` greater than that of its parents, and 
+    - not in the future according to the node's local clock reading of the current epoch,
+    - not farther in the past than the soft finality as defined by SPC ${{<sref finality>}},
```
anorth: This only applies when receiving a new block gossiped over the network. It does not apply, for example, when validating a chain from the beginning. So I think either:
- add a `(when receiving new gossip blocks)`
- move this into the chain sync section (where it may already be stated)
- Add a sentence after this bullet list explaining this additional contextual rule.

PR #738, opened 2019-12-13
Patch/surprise target

sternhenri:
**In this PR**:
- placeholder for using a target with SurprisePoSt (TBD in research)
- moved surprise PoSt verif to the actor > subsystem

**Needed in future PR**:
- Randomness from RT
- Pulling miner actor address easily
- Getting worker public keys from miner actor address as a facility in VM

### DIFF
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
@@ -1,6 +1,7 @@
 package storage_mining
 
 import (
+	filcrypto "github.com/filecoin-project/specs/algorithms/crypto"
 	filproofs "github.com/filecoin-project/specs/libraries/filcrypto/filproofs"
 	ipld "github.com/filecoin-project/specs/libraries/ipld"
 	spc "github.com/filecoin-project/specs/systems/filecoin_blockchain/storage_power_consensus"
@@ -348,19 +349,6 @@ func (a *StorageMinerActorCode_I) _onSuccessfulPoSt(rt Runtime) InvocOutput {
 	return rt.SuccessReturn()
 }
 
-// called by verifier to update miner state on successful surprise post
-// after it has been verified in the storage_mining_subsystem
-func (a *StorageMinerActorCode_I) ProcessVerifiedSurprisePoSt(rt Runtime) InvocOutput {
-	TODO() // TODO: validate caller
-
-	// Ensure pledge collateral satisfied
-	// otherwise, abort ProcessVerifiedSurprisePoSt
-	a._ensurePledgeCollateralSatisfied(rt)
-
-	return a._onSuccessfulPoSt(rt)
-
-}
-
 // Called by StoragePowerConsensus subsystem after verifying the Election proof
 // and verifying the PoSt proof in the block header.
 // Assume ElectionPoSt has already been successfully verified (both proof and partial ticket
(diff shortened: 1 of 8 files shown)


src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -381,6 +369,103 @@ func (a *StorageMinerActorCode_I) ProcessVerifiedElectionPoSt(rt Runtime) InvocO
...
+	// otherwise, abort ProcessVerifiedSurprisePoSt
+	a._ensurePledgeCollateralSatisfied(rt)
+
+	return a._onSuccessfulPoSt(rt)
+
+}
+
+func (a *StorageMinerActorCode_I) GetWorkerKeyByMinerAddress(minerAddr addr.Address) filcrypto.VRFPublicKey {
+	panic("TODO")
+}
```
anorth: This doesn't make sense here, and it looks like you removed calls to it.

What would make sense is
```
func (a *SMAC_I) GetWorkerKey() {
    h, st := a.State(rt)
    w := st.Info().Worker
    Release(...)
    return w
}
```

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -381,6 +369,103 @@ func (a *StorageMinerActorCode_I) ProcessVerifiedElectionPoSt(rt Runtime) InvocO
...
+	randomness := rt.Randomness(randomnessEpoch, SPC_LOOKBACK_POST)
+	panic(randomness)                       // ignore circular dependency
+	var postRandomnessInput util.Randomness // sms.PreparePoStChallengeSeed(randomness, actorAddr)
+
+	postRand := &filcrypto.VRFResult_I{
+		Output_: onChainInfo.Randomness(),
+	}
+
+	// TODO: get worker key from minerAddr
+	var workerKey filcrypto.VRFPublicKey
```
anorth: See above, this is easy.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -381,6 +369,103 @@ func (a *StorageMinerActorCode_I) ProcessVerifiedElectionPoSt(rt Runtime) InvocO
...
+
+	// verify the partialTickets themselves
+	// 4. Verify appropriate randomness
+
+	// pull from consts
+	SPC_LOOKBACK_POST := uint64(0)
+	randomnessEpoch := st.ChallengeStatus().LastChallengeEpoch()
+	randomness := rt.Randomness(randomnessEpoch, SPC_LOOKBACK_POST)
+	panic(randomness)                       // ignore circular dependency
+	var postRandomnessInput util.Randomness // sms.PreparePoStChallengeSeed(randomness, actorAddr)
```
anorth: I don't quite understand what's going on here, what's in/complete and whether it should be straightforward in implementation.

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
```diff
@@ -220,21 +216,29 @@ func (sms *StorageMiningSubsystem_I) VerifyElectionPoSt(header block.BlockHeader
...
-	postRandomnessInput := sector.PoStRandomness(sms._preparePoStChallengeSeed(randomness, header.Miner()))
+	postRandomnessInput := sector.PoStRandomness(sms.PreparePoStChallengeSeed(randomness, header.Miner()))
 
 	postRand := &filcrypto.VRFResult_I{
 		Output_: onChainInfo.Randomness(),
 	}
 
-	if !postRand.Verify(postRandomnessInput, sms.GetWorkerKeyByMinerAddress(header.Miner())) {
+	// TODO: get worker key from minerAddr
+	var workerKey filcrypto.VRFPublicKey
```
anorth: This is straightforward when you have `sma`, though you might need to add an accessor method.

anorth:
> Getting worker public keys from miner actor address as a facility in VM

I don't think we should do this. Send a message to the actor instead.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
sternhenri: it should be but I'll use this to clean up randomness in the vm

PR #739, opened 2019-12-13
Separate Runtime interface depended on by actors from its implementation.

anorth:
This separation is necessary so that the vm/runtime implementation can depend on the
InitActor for address lookup, without a circular dependency from the init actor
back on the Runtime interface which all actors consume.

The implementations already do something like that (and consumer-defined interfaces (e.g. if the Runtime interface was in the actor package) is canonical Go).

FYI @jzimmerman

### DIFF
--- a/src/systems/filecoin_vm/runtime/impl/runtime.go
+++ b/src/systems/filecoin_vm/runtime/impl/runtime.go
@@ -1,11 +1,15 @@
-package runtime
+package impl
 
 import (
+	"bytes"
+	"encoding/binary"
+
 	ipld "github.com/filecoin-project/specs/libraries/ipld"
 	block "github.com/filecoin-project/specs/systems/filecoin_blockchain/struct/block"
 	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
 	addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
 	msg "github.com/filecoin-project/specs/systems/filecoin_vm/message"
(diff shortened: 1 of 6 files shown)


src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -508,13 +491,15 @@ func (rt *VMContext) Randomness(e block.ChainEpoch, offset uint64) util.Randomne
...
-	}
-	hash := addr.ActorExecHash(Serialize_ActorExecAddressSeed(seed))
-
+	buf := new(bytes.Buffer)
+	_, err := buf.Write(addr.Serialize_Address(rt._immediateCaller))
+	util.Assert(err != nil)
+	err = binary.Write(buf, binary.BigEndian, rt._toplevelSenderCallSeqNum)
+	util.Assert(err != nil)
+	err = binary.Write(buf, binary.BigEndian, rt._internalCallSeqNum)
+	util.Assert(err != nil)
```
sternhenri: why the shift from the hash of the serialized seed object to the buffer? I don't mind it. Just wondering. 

@jzimmerman please confirm there's no possible collision here. I don't see any given fixed input sizes

anorth: Thanks, I should have called this semantic change out explicitly.

The ActorExecAddressSeed type is unnecessary, we can very simply define the output here without it. I copied the implementation from Lotus.

sternhenri: SGTM

PR #742, opened 2019-12-16
added more details to epoch validation

sternhenri:
- added appropriate delay in all parts of spec
- added details on non-validation of future blocks.


sternhenri:
please approve and merge @zixuanzh when able

PR #744, opened 2019-12-16
Remove ValidateImmediateCallerAcceptAnyOfType and all caller-type checks.

anorth:
The checks for account-type actors are flawed because they make multi-sig actors
unable to act as accounts, e.g. miner owners. The others are unnecessary and similarly restrictive of future design space.

In general, objects and functions that make assertions about who uses them an how are very brittle, since the developer cannot predict all the useful ways something might be used. IMO objects should define clear semantics and invariants about their behaviour, and be unopinionated about how they might be composed into a larger system.

There is follow-up work to change `_code.go` methods to not panic when a key is not found: they should return an error and the actor code can abort.

FYI @sternhenri

### DIFF
--- a/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.id
+++ b/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor.id
@@ -42,8 +42,8 @@ type StoragePowerActorState struct {
 }
 
 type StoragePowerActorCode struct {
-    AddBalance(rt Runtime)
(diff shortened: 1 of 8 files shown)


sternhenri:
@jzimmerman to merge and close

jzimmerman:
> @jzimmerman to opine and merge/close

I think this PR would need to be modified significantly given changes in the interim, and may no longer be required. (@anorth, please feel free to update or close if you agree.)

anorth:
This was clobbered by subsequent changes. I have filed #780 outlining the most important change.

PR #745, opened 2019-12-16
Runtime resolves receiver addresses through InitActor state.

anorth:
Implicitly created account actors are now also correctly mapped to ID by InitActor.

Note that this changed the behaviour when:
- the `To` actor does not exist and is not a pubkey address, or
- the `To` actor is implicitly created but the message has a non-zero MethodID
 
the caller now pays gas for a failed message, rather than the miner be penalized.

The rationale for this change is to unify the behaviour between top-level and nested sends (if we get to a nested send, the sender is definitely paying gas). It's pretty reasonable for senders to be confident the receiver exists and is of the right type before sending a message.

Follow-on work is to map sender addresses through InitActor too.

FYI @jzimmerman

### DIFF
--- a/src/systems/filecoin_vm/interpreter/vm_interpreter.go
+++ b/src/systems/filecoin_vm/interpreter/vm_interpreter.go
@@ -1,13 +1,13 @@
 package interpreter
 
 import (
-	"github.com/filecoin-project/specs/libraries/ipld"
-	"github.com/filecoin-project/specs/systems/filecoin_mining/storage_mining"
-	"github.com/filecoin-project/specs/systems/filecoin_vm/actor"
+	ipld "github.com/filecoin-project/specs/libraries/ipld"
+	storage_mining "github.com/filecoin-project/specs/systems/filecoin_mining/storage_mining"
+	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
 	addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
 	msg "github.com/filecoin-project/specs/systems/filecoin_vm/message"
 	vmr "github.com/filecoin-project/specs/systems/filecoin_vm/runtime"
-	"github.com/filecoin-project/specs/systems/filecoin_vm/runtime/exitcode"
+	exitcode "github.com/filecoin-project/specs/systems/filecoin_vm/runtime/exitcode"
 	gascost "github.com/filecoin-project/specs/systems/filecoin_vm/runtime/gascost"
 	vmri "github.com/filecoin-project/specs/systems/filecoin_vm/runtime/impl"
 	st "github.com/filecoin-project/specs/systems/filecoin_vm/state_tree"
--- a/src/systems/filecoin_vm/interpreter/vm_interpreter.go
+++ b/src/systems/filecoin_vm/interpreter/vm_interpreter.go
@@ -1,13 +1,13 @@
 package interpreter
 
 import (
-	ipld "github.com/filecoin-project/specs/libraries/ipld"
-	storage_mining "github.com/filecoin-project/specs/systems/filecoin_mining/storage_mining"
-	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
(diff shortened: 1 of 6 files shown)


src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -180,25 +176,11 @@ func (vmi *VMInterpreter_I) ApplyMessage(
...
-		return
-	}
-
-	// Deduct gas limit funds from sender.
-	// (This should always succeed, due to the sender balance check above.)
-	compTreePreSend = _withTransferFundsAssert(
-		compTreePreSend, message.From(), addr.BurntFundsActorAddr, gasLimitCost)
-
-	// Increment sender CallSeqNum.
+	compTreePreSend := _withTransferFundsAssert(inTree, message.From(), addr.BurntFundsActorAddr, gasLimitCost)
```
sternhenri: why the ability to remove `_ensureReceiver` here?

src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -470,8 +471,61 @@ func (rtOuter *VMContext) _sendInternal(input InvocInput, errSpec ErrorHandlingS
...
+		return act, targetIdAddr
+	}
+
+	if !targetRaw.IsKeyType() {
+		// Don't implicitly create an account actor for an address without an associated key.
+		rt._throwError(exitcode.SystemError(exitcode.ActorNotFound))
+	}
+
+	// Allocate an ID address from the init actor and map the pubkey To address to it.
+	newIdAddr := initSubState.MapAddressToNewID(targetRaw)
```
sternhenri: why not do this as part of the `_createActor` call?

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -226,31 +208,12 @@ func (vmi *VMInterpreter_I) ApplyMessage(
...
-	// The init actor will maintain a map from pubkey address to ID address.
-	return tree, true
-}
-
 func _applyMessageBuiltinAssert(tree st.StateTree, message msg.UnsignedMessage, topLevelBlockWinner addr.Address) st.StateTree {
 	Assert(message.From().Equals(addr.SystemActorAddr))
+	// Note: this message CallSeqNum is never checked (b/c it's created in this file), but probably should be.
+	// Since it changes state, we should be sure about the state transition.
+	// Alternatively we could special-case the system actor and declare that its CallSeqNumber
+	// never changes (saving us the state-change overhead).
```
sternhenri: We should probably do one or the other. I'd opt against special case but either seems good

src/systems/filecoin_vm/sysactors/init_actor.id
```diff
@@ -4,11 +4,14 @@ import vmr "github.com/filecoin-project/specs/systems/filecoin_vm/runtime"
...
-    NetworkName      string
+    AddressMap   {addr.Address: addr.ActorID}
+    NextID       addr.ActorID
+    NetworkName  string
 
-    _assignNextID()  addr.ActorID
+    // Allocates a new ID and maps another address to it. Returns the ID-address.
+    MapAddressToNewID(address addr.Address) addr.Address
+    // Resolves an address to an ID-address, if possible. Returns the argument if un-mapped.
+    ResolveAddress(address addr.Address) addr.Address
```
sternhenri: Could/should we make this accessible from actors directly. Specifically, there are multiple places in which a getter for StorageMinerActor ID-addresses is necessary.

anorth: Other actors shalt not access this state, but the InitActor can expose a method. However I think there are better solutions. I have seen one such place and plan to fix it.

src/systems/filecoin_vm/runtime/impl/runtime.go
sternhenri: :D glad I could help and thanks for the thoughtful answer

PR #747, opened 2019-12-17
Charge gas for state updates in implicit actor creation, plus method invocation.

anorth:
Also more commentary for gas cost primitives.

For follow-up:
There's something not quite right in the costs for send, transfer, and substate update. I think they're operating at slightly too high a level of abstraction. E.g., when an actor receives value and mutates its state, the new actor state is written just once, but we charge twice. Similarly the sender's value and CallSeqNum are written together.

I think we might want to rearrange the gascost units to have one just for actor state update, and charge it just once when we actually commit that update, be it for CallSeqNum, balance, or substate CID. The actor-substate is being charged per byte, which is correct but undocumented, so I'm not sure if it was intentional.

### DIFF
--- a/src/systems/filecoin_vm/runtime/impl/runtime.go
+++ b/src/systems/filecoin_vm/runtime/impl/runtime.go
@@ -73,6 +73,9 @@ type VMContext struct {
 	_running            bool
 	_actorAddress       addr.Address
 	_actorStateAcquired bool
+	// Tracks whether actor substate has changed in order to charge gas just once
+	// regardless of how many times it's written.
+	_actorSubstateUpdated bool
 
 	_immediateCaller addr.Address
 	// Note: This is the actor in the From field of the initial on-chain message.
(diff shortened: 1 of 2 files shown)


src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -389,9 +387,6 @@ func _invokeMethodInternal(
 	rt._running = true
 	ret, exitCode = _catchRuntimeErrors(func() InvocOutput {
 		methodOutput := actorCode.InvokeMethod(rt, method, params)
-		if rt._actorSubstateUpdated {
-			rt._rtAllocGas(gascost.UpdateActorSubstate)
-		}
```
jzimmerman: This was intended to partially mitigate the problem you mentioned in the top-level comment, viz., an actor should only be charged once even if the substate is updated multiple times. I agree it doesn't completely solve the problem, but I think it is a better approximation to the true cost than the implied policy if we remove it (as in this PR)?

anorth: Ok, agreed.

PR #748, opened 2019-12-17
Remove unnecessary invocation of InitActor to resolve an address to an ID.

anorth:


anorth:
FYI @zixuanzh

src/systems/filecoin_mining/sector/sector.id
```diff
@@ -30,8 +29,8 @@ type SectorSize UInt
 
 // TODO make sure this is globally unique
 type SectorID struct {
-    MinerID
-    Number SectorNumber
+    Miner   addr.ActorID
```
sternhenri: this is great. much cleaner/legible

src/systems/filecoin_vm/sysactors/init_actor.go
```diff
@@ -98,12 +98,13 @@ func (a *InitActorCode_I) Exec(rt Runtime, codeID actor.CodeID, constructorParam
 		Bytes(addr.Serialize_Address_Compact(idAddr)))
 }
 
-func (a *InitActorCode_I) GetActorIDForAddress(rt Runtime, address addr.Address) InvocOutput {
-	h, st := _loadState(rt)
-	actorID := st.AddressMap()[address]
-	Release(rt, h, st)
-	return rt.ValueReturn(Bytes(addr.Serialize_ActorID(actorID)))
-}
+// This method is disabled until proven necessary.
```
sternhenri: I thought you had written this method @anorth? If not in use anywhere, why not remove?

sternhenri: I guess I don't understand how it is not redundant with your new code: your code derives the id from the address, this one takes the address and runtime and returns the id... (I'm making an assumption about how values in the addressmap are generated) but it seems to me both of these should be the same.. no?

anorth: They do different things. This method looks up a *non-ID-address* in the InitActor's address table, and returns the associated ID address if found.

PR #749, opened 2019-12-17
Add skeleton for global statistics/indices

jzimmerman:


anorth:
Registering my objection here for posterity. Having the runtime tightly coupled with the internals of the InitActor is bad enough - at least that's not application-specific. Coupling it also to the internals of the Reward and StoragePower actors is way worse.

Since this hasn't come with any documentation of the justification, design for how to use these values, or assessment of the alternatives, it's hard for me to mount a technical argument here, but I urge taking a different path if possible.

PR #750, opened 2019-12-17
Clean up StorageMarketActor structure (state vs. runtime functions)

jzimmerman:


src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -250,170 +208,166 @@ func (a *StorageMarketActorCode_I) TerminateDealsOnSlashProviderSector(rt Runtim
...
 
-	if deal.SectorStartEpoch() == block.ChainEpoch_None {
-		// Not yet appeared in proven sector; check for timeout.
-		if dealP.StartEpoch() >= epoch {
-			st._rtProcessDealInitTimedOut(rt, dealID)
-		}
-		return
-	}
+	var relevantDealIDs []deal.DealID
+	// Populate with the set of all elements in st.Deals() in which addr is one of the parties
```
zixuanzh: We might not have an efficient way to do this, deals are not indexed by the parties. We can consider allowing this function to take in an input array of DealIDs.

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -250,170 +208,166 @@ func (a *StorageMarketActorCode_I) TerminateDealsOnSlashProviderSector(rt Runtim
...
+	deal, dealP, found = st._getOnChainDeal(dealID)
+	if !found {
+		rt.AbortStateMsg("dealID not found in Deals.")
+	}
+	return
 }
 
-func (a *StorageMarketActorCode_I) OnEpochTickEnd(rt Runtime) {
-	h, st := a.State(rt)
+func (st *StorageMarketActorState_I) _rtLockBalanceUntrusted(rt Runtime, addr addr.Address, amount actor.TokenAmount) {
```
zixuanzh: should we rename this to `_rtLockBalanceOrAbort` for consistency?

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
jzimmerman: They're not indexed by the parties on-chain, but implementations can maintain a local index for this -- I'll add a note to this effect. (Though, we might want to consider partitioning the on-chain data structure by party also.)

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -250,170 +208,166 @@ func (a *StorageMarketActorCode_I) TerminateDealsOnSlashProviderSector(rt Runtim
...
-		// Not yet appeared in proven sector; check for timeout.
-		if dealP.StartEpoch() >= epoch {
-			st._rtProcessDealInitTimedOut(rt, dealID)
-		}
-		return
-	}
+	var relevantDealIDs []deal.DealID
+	// Populate with the set of all elements in st.Deals() in which addr is one of the parties
+	// (either client or provider).
+	IMPL_FINISH()
```
anorth: It is unacceptable for actor code to have IMPL_FINISH(). The spec must must completely specify every operation.

Also, if some actor state needs to be indexed for on-chain query, it must be indexed in the on-chain state. Or we need to conjure up yet another Runtime facility. There's no existing operation where the actor can consult some magic efficient but unspecified index.

PR #751, opened 2019-12-17
StorageMarketActor: handle non-miner (client) callers; finish caller validation

jzimmerman:
**Note:** please do not merge this PR directly (it's based on a non-master branch for clarity in review).

### DIFF
--- a/src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
+++ b/src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
@@ -7,7 +7,6 @@ import (
 	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
 	addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
 	ai "github.com/filecoin-project/specs/systems/filecoin_vm/actor_interfaces"
-	vmr "github.com/filecoin-project/specs/systems/filecoin_vm/runtime"
 	util "github.com/filecoin-project/specs/util"
 )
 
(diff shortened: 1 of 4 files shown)


src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -7,82 +7,60 @@ import (
...
 	// Before any operations that check the balance tables for funds, execute all deferred
 	// deal state updates.
 	//
 	// Note: as an optimization, implementations may cache efficient data structures indicating
 	// which of the following set of updates are redundant and can be skipped.
-	amountSlashedTotal += st._rtUpdatePendingDealStatesForParty(rt, minerAddr)
+	amountSlashedTotal += st._rtUpdatePendingDealStatesForParty(rt, entryAddr)
 
-	minBalance := st._getLockedReqBalanceInternal(minerAddr)
+	minBalance := st._getLockedReqBalanceInternal(entryAddr)
```
zixuanzh: Lets confirm the math here is correct. It needs to first deduct amountSlashedTotal from the BalanceTable and then extract amount no?

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -7,82 +7,60 @@ import (
...
 	newTable, amountExtracted, ok := actor.BalanceTable_WithExtractPartial(
-		st.EscrowTable(), minerAddr, amountRequested, minBalance)
+		st.EscrowTable(), entryAddr, amountRequested, minBalance)
 	Assert(ok)
```
zixuanzh: should it be an Abort instead of an Assert depending on how we handle the above? If a miner requests more than it has after subtracting the slashAmount, it shouldnt crash the VM?

jzimmerman:
Rebased in https://github.com/filecoin-project/specs/pull/752.

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
```diff
@@ -7,82 +7,60 @@ import (
...
-func (a *StorageMarketActorCode_I) AddBalance(rt Runtime, minerAddr addr.Address) {
-	TODO() // TODO: must also handle case of non-miner (client) sender
-
-	ownerAddr, workerAddr := _rtGetMinerAccountsAssert(rt, minerAddr)
-	rt.ValidateImmediateCallerInSet([]addr.Address{ownerAddr, workerAddr})
-	TODO() // TODO: should workerAddr be permitted here? Depends on actor principal security assumptions.
-
-	msgValue := rt.ValueReceived()
+func (a *StorageMarketActorCode_I) AddBalance(rt Runtime, entryAddr addr.Address) {
+	_rtValidateImmediateCallerDetermineRecipient(rt, entryAddr)
```
zixuanzh: I think we want to allow any one to AddBalance to a miner actor or a client account actor. Not sure if we need this check but having this check is a stricter requirement even though not the one that I thought we agreed on.

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
jzimmerman: No, I don't think so -- `_rtUpdatePendingDealStatesForParty` mutates the balance table in-place, so it should already reflect the slashed amounts.

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
jzimmerman: Ah, but this shouldn't be possible in StorageMarketActor -- a party can never be undercollateralized, because the collateral requirement for existing items doesn't change over time.

src/systems/filecoin_markets/storage_market/storage_market_actor_code.go
jzimmerman: Interesting. I interpreted it to mean either of the two relevant parties (owner or worker), but not necessarily any party in the entire network. Let's make a note to revisit this once the code stabilizes?

PR #752, opened 2019-12-17
StorageMarketActor: handle non-miner (client) callers; finish caller validation; minor edits

jzimmerman:


jzimmerman:
Subsumed by #753.

PR #753, opened 2019-12-17
Finish StorageMarketActor open items

jzimmerman:
- Generate storage deal IDs
- Placeholders for partial slash and other parameters
- Handle non-miner (client) callers
- Finish caller validation
- Minor edits and cleanup

PR #754, opened 2019-12-18
spa cleanup

zixuanzh:
In this PR
- [x] remove rt as input to pure functions in SPA
- [x] pull out ValidateImmediateCallerDetermineRecipient etc from SMA to runtime
- [x] add caller verification to SPA
- [x] resolve a couple TODOs in SPA

For upcoming PR
- [ ] fix pledge collateral requirement

### DIFF
--- a/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
+++ b/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
@@ -17,11 +17,8 @@ import (
 ////////////////////////////////////////////////////////////////////////////////
 
 func (a *StoragePowerActorCode_I) AddBalance(rt Runtime, minerAddr addr.Address) {
-	if !rt.ValidateIsStorageMiner(minerAddr) {
-		rt.AbortArgMsg("spa.AddBalance: invalid minerAdder.")
-	}
-	// either worker or owner can add balance to a miner
-	rt.ValidateImmediateCallerDetermineRecipient(minerAddr)
+	// caller verification
+	TODO()
 
 	msgValue := rt.ValueReceived()
 
@@ -40,11 +37,8 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, minerAddr addr.Add
 		rt.AbortArgMsg("spa.WithdrawBalance: negative amount.")
 	}
 
-	if !rt.ValidateIsStorageMiner(minerAddr) {
-		rt.AbortArgMsg("spa.WithdrawBalance: invalid minerAddr.")
-	}
-	// either worker or owner can withdraw balance from a miner
-	rt.ValidateImmediateCallerDetermineRecipient(minerAddr)
+	// caller verification
+	TODO()
 	msgSender := rt.ImmediateCaller()
 
 	h, st := a.State(rt)
@@ -99,9 +93,11 @@ func (a *StoragePowerActorCode_I) CreateStorageMiner(
 }
 
 func (a *StoragePowerActorCode_I) RemoveStorageMiner(rt Runtime) {
-	rt.ValidateImmediateCallerAcceptAnyOfType(actor.BuiltinActorID_StorageMiner)
 
 	minerAddr := rt.ImmediateCaller()
+
+	// caller verification
+	TODO()
 
 	h, st := a.State(rt)
 
@@ -136,10 +132,9 @@ func (a *StoragePowerActorCode_I) RemoveStorageMiner(rt Runtime) {
 func (a *StoragePowerActorCode_I) EnsurePledgeCollateralSatisfied(rt Runtime) {
 
 	minerAddr := rt.ImmediateCaller()
-	ok := rt.ValidateIsStorageMiner(minerAddr)
-	if !ok {
-		rt.AbortStateMsg("spa.EnsurePledgeCollateralSatisfied: caller must be StorageMiner.")
-	}
+
+	// caller verification
+	TODO()
 
 	h, st := a.State(rt)
 	pledgeReq := st._getPledgeCollateralReqForMiner(minerAddr)
@@ -157,10 +152,9 @@ func (a *StoragePowerActorCode_I) EnsurePledgeCollateralSatisfied(rt Runtime) {
 func (a *StoragePowerActorCode_I) SlashPledgeForStorageFault(rt Runtime, affectedPower block.StoragePower, faultType sector.StorageFaultType) {
 
 	minerAddr := rt.ImmediateCaller()
-	ok := rt.ValidateIsStorageMiner(minerAddr)
-	if !ok {
-		rt.AbortStateMsg("spa.SlashPledgeForStorageFault: caller must be StorageMiner.")
-	}
+
+	// caller verification
+	TODO()
 
 	h, st := a.State(rt)
 
@@ -182,10 +176,9 @@ func (a *StoragePowerActorCode_I) SlashPledgeForStorageFault(rt Runtime, affecte
 func (a *StoragePowerActorCode_I) ProcessPowerReport(rt Runtime, report PowerReport) {
 
 	minerAddr := rt.ImmediateCaller()
-	ok := rt.ValidateIsStorageMiner(minerAddr)
-	if !ok {
-		rt.AbortStateMsg("spa.ProcessPowerReport: caller must be StorageMiner.")
-	}
+
+	// caller verification
+	TODO()
 
 	h, st := a.State(rt)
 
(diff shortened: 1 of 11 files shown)


src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -3,108 +3,49 @@ package storage_power_consensus
...
-	minerAddr := rt.ImmediateCaller()
-
-	var ownerAddr addr.Address
-	TODO() // Determine owner address from miner
+	if !rt.ValidateIsStorageMiner(minerAddr) {
+		rt.AbortArgMsg("spa.WithdrawBalance: invalid minerAddr.")
+	}
+	// either worker or owner can withdraw balance from a miner
+	rt.ValidateImmediateCallerDetermineRecipient(minerAddr)
+	msgSender := rt.ImmediateCaller()
```
anorth: The name `ValidateImmediateCallerDetermineRecipient` suggests that the return value should be used as the recipient. If we want different logic here then that points to it being a bad function to have on the Runtime since it's so special-case. Don't attempt to share the wrong function, write the correct one here.

Also I'm pretty sure we had decided in conversation that pledge collateral could only be withdrawn to the owner.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -192,10 +134,12 @@ func (a *StoragePowerActorCode_I) RemoveStorageMiner(rt Runtime) {
...
 
 func (a *StoragePowerActorCode_I) EnsurePledgeCollateralSatisfied(rt Runtime) {
-	rt.ValidateImmediateCallerAcceptAnyOfType(actor.BuiltinActorID_StorageMiner)
 
-	TODO() // TODO: principal access control
 	minerAddr := rt.ImmediateCaller()
+	ok := rt.ValidateIsStorageMiner(minerAddr)
+	if !ok {
+		rt.AbortStateMsg("spa.EnsurePledgeCollateralSatisfied: caller must be StorageMiner.")
+	}
```
anorth: Unnecessary! If the caller has an entry in the collateral table, they they're fine.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -212,35 +156,40 @@ func (a *StoragePowerActorCode_I) EnsurePledgeCollateralSatisfied(rt Runtime) {
 // slash pledge collateral for Declared, Detected and Terminated faults
 func (a *StoragePowerActorCode_I) SlashPledgeForStorageFault(rt Runtime, affectedPower block.StoragePower, faultType sector.StorageFaultType) {
 
-	minerID := rt.ImmediateCaller()
+	minerAddr := rt.ImmediateCaller()
+	ok := rt.ValidateIsStorageMiner(minerAddr)
+	if !ok {
+		rt.AbortStateMsg("spa.SlashPledgeForStorageFault: caller must be StorageMiner.")
+	}
```
anorth: Ditto

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -212,35 +156,40 @@ func (a *StoragePowerActorCode_I) EnsurePledgeCollateralSatisfied(rt Runtime) {
...
 // @param PowerReport with ActivePower and InactivePower
 // update miner power based on the power report
 func (a *StoragePowerActorCode_I) ProcessPowerReport(rt Runtime, report PowerReport) {
 
-	minerID := rt.ImmediateCaller()
+	minerAddr := rt.ImmediateCaller()
+	ok := rt.ValidateIsStorageMiner(minerAddr)
+	if !ok {
+		rt.AbortStateMsg("spa.ProcessPowerReport: caller must be StorageMiner.")
+	}
```
anorth: Ditto

src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -308,6 +309,38 @@ func (rt *VMContext) _checkNumValidateCalls(x int) {
...
 }
 
+func (rt *VMContext) ValidateIsStorageMiner(minerAddr addr.Address) bool {
+	codeID, ok := rt.GetActorCodeID(minerAddr)
+	Assert(ok)
+	if !codeID.IsBuiltin() {
+		return false
+	}
+	return (codeID.As_Builtin() == actor.BuiltinActorID_StorageMiner)
+}
```
anorth: Strong objection! Actors should not inspect the code type of other actors. This is terribly brittle and a crutch for things which can be better represented by actors themselves.
This is also redundant given that #753 gives access to the actor code ID.

src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -308,6 +309,38 @@ func (rt *VMContext) _checkNumValidateCalls(x int) {
...
+		// Storage miner actor; implied funds recipient is the associated owner address.
+		ownerAddr, workerAddr := rt.GetMinerAccountsAssert(entryAddr)
+		rt.ValidateImmediateCallerInSet([]addr.Address{ownerAddr, workerAddr})
+		return ownerAddr
+	} else {
+		// Account actor (client); funds recipient is just the entry address itself.
+		rt.ValidateImmediateCallerAcceptAnyOfType(actor.BuiltinActorID_Account)
+		return entryAddr
+	}
+}
```
anorth: No! The Runtime should not have such application-specific coupling to the actors. This is easily implemented outside of the runtime and so should not be here.

src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -308,6 +309,38 @@ func (rt *VMContext) _checkNumValidateCalls(x int) {
...
+
+func (rt *VMContext) GetMinerAccountsAssert(minerAddr addr.Address) (ownerAddr addr.Address, workerAddr addr.Address) {
+	ownerAddr = addr.Deserialize_Address_Compact_Assert(
+		rt.SendQuery(minerAddr, ai.Method_StorageMinerActor_GetOwnerAddr, []util.Serialization{}))
+
+	workerAddr = addr.Deserialize_Address_Compact_Assert(
+		rt.SendQuery(minerAddr, ai.Method_StorageMinerActor_GetWorkerAddr, []util.Serialization{}))
+
+	return
+}
```
anorth: This should not be on the runtime! Do not couple the runtime to application specifics, and this can perfectly well be implemented outside the runtime.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -3,108 +3,43 @@ package storage_power_consensus
...
-func (a *StoragePowerActorCode_I) State(rt Runtime) (vmr.ActorStateHandle, State) {
-	h := rt.AcquireState()
-	stateCID := h.Take()
-	stateBytes := rt.IpldGet(ipld.CID(stateCID))
-	if stateBytes.Which() != vmr.Runtime_IpldGet_FunRet_Case_Bytes {
-		rt.AbortAPI("IPLD lookup error")
-	}
-	state := DeserializeState(stateBytes.As_Bytes())
-	return h, state
-}
```
anorth: Note to self: conflict with https://github.com/filecoin-project/specs/pull/757

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -123,11 +58,8 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, amountRequested ac
 
 	UpdateRelease(rt, h, st)
 
-	// send funds to miner
-	rt.SendPropagatingErrors(&vmr.InvocInput_I{
-		To_:    ownerAddr,
-		Value_: amountExtracted,
-	})
+	// send funds to sender (pledge collateral)
+	rt.SendFunds(msgSender, amountExtracted)
```
anorth: It's still preferable to leave this sending to the (uninitialized) owner address than sender.

zixuanzh: I think it makes sense for reward to always go to the owner. Pledge collateral is basically working capital, any reason why we would not want that to go the worker?

jzimmerman: Not sure we need to decide this now, but remitting to owner would be more consistent with SMA.

anorth:
Noting for history that in subsequent conversation we determined that inspecting the code type of caller actors was acceptable - redundant but defence in depth against attempted malicious calls.

PR #757, opened 2019-12-18
Add IPLDStore placehold interface in runtime impl and make Runtime.IpldGet perform deserialization

anorth:
IpldGet is now symmetric with IpldPut in performing de/serialization, and more useful.
Cleaned up a bunch of IPLD access, which revealed a few un-checked incompatibilities
between actor state structure definitions and their code. The new methods with `panic("TODO")` are added just to satisfy the compiler, but obviously they need implementing later.

The motivation for this is to make the IPLD store accessible in the VM runtime so that I can look up the init actor state for address resolution.

FYI @zixuanzh

### DIFF
--- a/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
+++ b/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
@@ -117,7 +117,7 @@ func (st *StoragePowerActorState_I) GetCurrPledgeForMiner(minerAddr addr.Address
 	return actor_util.BalanceTable_GetEntry(st.EscrowTable(), minerAddr)
 }
 
-func _getStorageFaultSlashPledgePercent(faultType sector.StorageFaultType) int {
+func (st *StoragePowerActorState_I) _getStorageFaultSlashPledgePercent(faultType sector.StorageFaultType) int {
 	PARAM_FINISH() // TODO: instantiate these placeholders
 	panic("")
 
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
@@ -217,6 +217,11 @@ func (st *StorageMinerActorState_I) _getSectorWeight(sectorNo sector.SectorNumbe
 	return sectorInfo.SectorWeight(), true
 }
 
+func (st *StorageMinerActorState_I) _getSectorPower(sectorNo sector.SectorNumber) (power block.StoragePower, ok bool) {
+	TODO()
+	panic("")
+}
+
 func (st *StorageMinerActorState_I) _getSectorDealIDs(sectorNo sector.SectorNumber) (dealIDs deal.DealIDs, ok bool) {
 	sectorInfo, found := st._getSectorOnChainInfo(sectorNo)
 	if !found {
--- a/src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
+++ b/src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
@@ -5,24 +5,10 @@ import (
 	deal "github.com/filecoin-project/specs/systems/filecoin_markets/deal"
 	sector "github.com/filecoin-project/specs/systems/filecoin_mining/sector"
 	actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
-	addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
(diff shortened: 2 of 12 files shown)


src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
```diff
@@ -127,6 +135,10 @@ func (st *StoragePowerActorState_I) _selectMinersToSurprise(challengeCount int,
 	return selectedMiners
 }
 
+func (st *StoragePowerActorState_I) _shouldChallenge(rt Runtime, networkPower block.StoragePower) bool {
+	panic("TODO")
+}
+
```
sternhenri: Why all the additions in this file? Some of these are defined in `ChallengeStatus.go` (no reason you should have known that and we should probably split out interfaces which have corresponding go files). I assume it was left in there by @zixuanzh since it is technically "actor code."

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
```diff
@@ -40,6 +40,10 @@ func (st *StorageMinerActorState_I) _shouldChallenge(currEpoch block.ChainEpoch)
 	return st.ChallengeStatus().ShouldChallenge(currEpoch)
 }
 
+func (st *StorageMinerActorState_I) _verifySeal(rt Runtime, onChainInfo sector.OnChainSealVerifyInfo) bool {
```
sternhenri: It's a monster of a function but I think this may be redundant with `ProveCommitSector` (could be factored out of there so PCS is a shorter method) though I can't tell for sure. @zixuanzh ?

src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -534,12 +548,10 @@ func (rt *VMContext) _resolveReceiver(targetRaw addr.Address) (actor.ActorState,
...
-	initStateBytes := rt.IpldGet(ipld.CID(initState.State()))
-	initSubState, err := sysactors.Deserialize_InitActorState(util.Serialization(initStateBytes.As_Bytes()))
-	if err != nil {
-		rt.AbortAPI("State deserialization error")
-	}
-	return initSubState
+	var initSubState sysactors.InitActorState_I
+	ok = rt.IpldGet(ipld.CID(initState.State()), &initSubState)
+	util.Assert(ok)
+	return &initSubState
```
sternhenri: much nicer ^

src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -618,28 +630,20 @@ func (rt *VMContext) NewActorAddress() addr.Address {
...
-	panic("") // retrieve serializedSize
-
-	rt._rtAllocGas(gascost.IpldGet(serializedSize))
-
-	IMPL_FINISH()
-	panic("") // return item
+func (rt *VMContext) IpldGet(c ipld.CID, o ipld.Object) bool {
+	serialized, ok := rt._store.Get(c)
+	if ok {
+		rt._rtAllocGas(gascost.IpldGet(len(serialized)))
```
sternhenri: Where is `gascost` getting pulled in from? This is a random question but wouldn't you want to do the `rtAllocGas` in get even if the object isn't found? This is still a DoS vector...

src/systems/filecoin_vm/runtime/runtime.id
```diff
@@ -105,7 +105,9 @@ type Runtime interface {
         address  addr.Address
     )
 
-    IpldGet(c ipld.CID) union {Bytes, error}
+    // Retrieves and deserializes an object from the store into o. Returns whether successful.
+    IpldGet(c ipld.CID, o ipld.Object) bool
```
sternhenri: cool

src/systems/filecoin_vm/runtime/impl/runtime.go
anorth: maybe there should be a gas cost for a failed lookup, yes. But it can't depend on the object's size.

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_state.go
anorth: Because it didn't compile without them

src/systems/filecoin_vm/runtime/impl/runtime.go
sternhenri: fair enough.

PR #759, opened 2019-12-18
Resolve message sender addresses through InitActor at top of interpreter.

anorth:
Also document and check a few address-type invariants.

Note: this branch is based on #757, rebase before merging.

FYI @icorderi

### DIFF
--- a/src/systems/filecoin_vm/runtime/runtime.id
+++ b/src/systems/filecoin_vm/runtime/runtime.id
@@ -14,8 +14,9 @@ type Runtime interface {
     // Randomness returns a (pseudo)random stream (indexed by offset) for the current epoch.
     Randomness(e block.ChainEpoch, offset UInt) util.Randomness
 
-    // Note: This is the _immediate_ caller.
+    // The address of the immediate calling actor.
     // Not necessarily the actor in the From field of the initial on-chain Message.
+    // Always an ID-address.
     ImmediateCaller() addr.Address
     ValidateImmediateCallerIs(caller addr.Address)
     ValidateImmediateCallerInSet(callers [addr.Address])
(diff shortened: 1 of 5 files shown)


src/systems/filecoin_vm/runtime/runtime.id
```diff
@@ -95,13 +97,14 @@ type Runtime interface {
...
     // different actor after messages are re-ordered).
+    // Always an ActorExec address.
     NewActorAddress() addr.Address
 
     // Creates an actor in the state tree, with empty state. May only be called by InitActor.
     CreateActor(
         // The new actor's code identifier.
         codeId   actor.CodeID
-        // Address under which the new actor's state will be stored.
+        // Address under which the new actor's state will be stored. Must be an ID-address.
```
sternhenri: You should also be able to pass in a key address if the actor has one

src/systems/filecoin_vm/runtime/runtime.id
```diff
@@ -24,10 +25,11 @@ type Runtime interface {
     ValidateImmediateCallerAcceptAny()
     ValidateImmediateCallerMatches(CallerPattern)
 
-    // The address of the actor receiving the message.
+    // The address of the actor receiving the message. Always an ID-address.
     CurrReceiver()         addr.Address
 
     // The actor who mined the block in which the initial on-chain message appears.
+    // Always an ID-address.
     ToplevelBlockWinner()  addr.Address
```
sternhenri: how would you verify message sigs if passing in the id addr here? Lookup of miner key based on id addr each time..?

sternhenri: seems convoluted. A good rule might be sender --> use key addr, receiver use id addr...

jzimmerman: I think actors should be able to treat addresses as black-boxes for the purposes of signature verification, i.e., shouldn't rely on being able to inspect them manually to obtain the public key. Rather they should be able to invoke a runtime method to verify a signature given an actor address, or (if we're willing to designate AccountActor code as trusted) by sending a message to the actor to verify its own signature. The former seems more intuitive to me, but I don't have a strong preference.

src/systems/filecoin_vm/runtime/runtime.id
anorth: No, this cannot be used for creating account actors, which is why I made it explicit.

src/systems/filecoin_vm/runtime/runtime.id
anorth: Signature verification is always off-chain. The node needs to look up the pubkey-address from account actor state (which is why the account actor has state at all).

PR #763, opened 2019-12-18
same file for docker and local

nikkolasg:


mishmosh:
@nikkolasg we are trying to either land or close old PRs. where would you like to go next with this one?

nikkolasg:
@nicola Did you already integrated that in your branch or not ? If not I think that's safe to merge ?

PR #765, opened 2019-12-18
Add singleton SystemActor for VM-interpreter-builtin messages

jzimmerman:
- Also refine documentation for TODO vs. IMPL_FINISH

### DIFF
--- a/src/systems/filecoin_vm/interpreter/vm_registry.go
+++ b/src/systems/filecoin_vm/interpreter/vm_registry.go
@@ -15,6 +15,7 @@ var (
 
 // CodeIDs for system actors
 var (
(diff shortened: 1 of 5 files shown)


src/systems/filecoin_vm/interpreter/vm_registry.go
```diff
@@ -15,6 +15,7 @@ var (
 
 // CodeIDs for system actors
 var (
+	SystemActorCodeID         = actor.CodeID_Make_Builtin(actor.BuiltinActorID_System)
```
anorth: In line with codeID checks elsewhere, you probably want to check for this code in the places where the `addr.SystemActorAddr` is currently checked. Otherwise this change has zero semantic impact.

PR #768, opened 2019-12-19
Pledge collateral & reward actor

zixuanzh:
- [x] connect pledge collateral with spa
- [x] add reward garnishing to reward actor
- [x] add general reward vesting function to reward actor

### DIFF
--- a/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
+++ b/src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
@@ -41,15 +41,9 @@ func (a *StoragePowerActorCode_I) WithdrawBalance(rt Runtime, minerAddr addr.Add
 	// caller verification
 	TODO()
 	msgSender := rt.ImmediateCaller()
+	minBalance := a._rtGetPledgeCollateralReqForMinerOrAbort(rt, minerAddr)
 
 	h, st := a.State(rt)
-
-	minerPowerTotal, ok := st._getPowerTotalForMiner(minerAddr)
-	if !ok {
-		rt.AbortArgMsg("spa.WithdrawBalance: Miner not found.")
-	}
-
-	minBalance := st._getPledgeCollateralReq(minerPowerTotal)
 	newTable, amountExtracted, ok := actor_util.BalanceTable_WithExtractPartial(
 		st.EscrowTable(), minerAddr, amountRequested, minBalance)
 	if !ok {
@@ -102,11 +96,11 @@ func (a *StoragePowerActorCode_I) RemoveStorageMiner(rt Runtime) {
 
 	h, st := a.State(rt)
 
-	minerPowerTotal, ok := st._getPowerTotalForMiner(minerAddr)
+	activePower, inactivePower, ok := st._getPowerTotalForMiner(minerAddr)
 	if !ok {
 		rt.AbortArgMsg("spa.RemoveStorageMiner: miner entry not found in PowerTable.")
 	}
-	if minerPowerTotal > 0 {
+	if activePower > 0 || inactivePower > 0 {
 		rt.AbortStateMsg("spa.RemoveStorageMiner: power still remains.")
 	}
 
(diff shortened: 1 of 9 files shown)


src/systems/filecoin_vm/sysactors/reward_actor.go
```diff
@@ -36,6 +39,18 @@ func DeserializeState(x Bytes) RewardActorState {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+func (r *Reward_I) AmountVested(elapsedEpoch block.ChainEpoch) actor.TokenAmount {
+	switch r.VestingFunction() {
+	case VestingFunction_None:
+		return r.Value()
+	case VestingFunction_Linear:
+		vestedProportion := math.Max(1.0, float64(elapsedEpoch)/float64(r.StartEpoch()-r.EndEpoch()))
```
jzimmerman: Let's add a TODO here to make sure we don't use floating-point arithmetic.

src/systems/filecoin_vm/sysactors/reward_actor.go
```diff
@@ -74,12 +89,56 @@ func (a *RewardActorCode_I) Constructor(rt vmr.Runtime) InvocOutput {
...
 	rt.ValidateImmediateCallerIs(addr.SystemActorAddr)
-	// block reward function should live here
-	// handle penalty greater than reward
-	// put Reward into RewardMap
-	panic("TODO")
+
+	inds := rt.CurrIndices()
+	pledgeReq := inds.BlockReward_PledgeCollateralReq(minerActivePower, minerInactivePower, currPledge)
+	currReward := inds.BlockReward_GetCurrRewardForMiner(minerActivePower, currPledge)
+	underPledge := math.Max(float64(actor.TokenAmount(0)), float64(pledgeReq-currPledge)) // 0 if over collateralized
```
jzimmerman: (same as above re: floating-point)

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_actor_code.go
```diff
@@ -240,16 +242,38 @@ func (a *StoragePowerActorCode_I) Surprise(rt Runtime) {
...
+	if !ok {
+		rt.AbortArgMsg("spa._rtGetPledgeCollateralReqForMinerOrAbort: miner not in PowerTable.")
+	}
+
+	currPledge, ok := st._getCurrPledgeForMiner(minerAddr)
+	if !ok {
+		rt.AbortArgMsg("spa._rtGetPledgeCollateralReqForMinerOrAbort: miner not in EscrowTable.")
+	}
+
+	minBalance := inds.BlockReward_PledgeCollateralReq(activePower, inactivePower, currPledge)
```
anorth: Why is the current pledge a parameter to the pure function that computes the pledge requirement?

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -275,12 +275,14 @@ func _makeBlockRewardMessage(state st.StateTree, minerAddr addr.Address, penalty
 	params[0] = addr.Serialize_Address(minerAddr)
 	params[1] = actor.Serialize_TokenAmount(penalty)
 
+	TODO() // serialize other inputs to BlockRewardMessage or get this from query in RewardActor
+
```
anorth: The reward actor should query for the miner's power and pledge, not trust the caller. So IMO there should be no more parameters.

src/systems/filecoin_vm/sysactors/reward_actor.id
```diff
@@ -3,11 +3,23 @@ import actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
...
+    // TODO: potential options
+    // PieceWise
+    // Quadratic
+    // Exponential
+}
+
 type Reward struct {
+    VestingFunction
     StartEpoch       block.ChainEpoch
+    EndEpoch         block.ChainEpoch
```
anorth: This change makes it hard to avoid floating point arithmetic. Why change it from the release rate?

src/systems/filecoin_vm/sysactors/reward_actor.go
```diff
@@ -74,12 +89,56 @@ func (a *RewardActorCode_I) Constructor(rt vmr.Runtime) InvocOutput {
...
+
+	h, st := a.State(rt)
+	if actualReward > 0 {
+		// put Reward into RewardMap
+		newReward := &Reward_I{
+			StartEpoch_:      rt.CurrEpoch(),
+			EndEpoch_:        rt.CurrEpoch(),
+			Value_:           actualReward,
+			AmountWithdrawn_: actor.TokenAmount(0),
+			VestingFunction_: VestingFunction_None,
```
anorth: Noting that if we actually decide to have no vesting, we should skip all this complexity and just pay the reward to the owner immediately. This is very heavy on chain.

zixuanzh: Definitely

src/systems/filecoin_vm/sysactors/reward_actor.id
zixuanzh: We wanted to accommodate vesting function other than linear and hence the change. Thanks for flagging 👍

PR #772, opened 2019-12-20
Add algorithm to sector configuration

porcuquine:
A sector configuration now includes the `SealAlgorithm` which should be used when sealing. The chosen algorithm needs to be remembered in the sector store's metadata so PoSt can discriminate on it when generating private proofs. Private proofs are grouped by algorithm type when generating circuit proofs, which are aggregated into a single 'circuit proof' containing candidates and proof bytes.

TODO (but not necessarily in this PR): some kind of framing to associate candidates with corresponding circuit proofs. With current PoSt, there is a 1-1 relationship, but if/when that changes, then grouping may be perturbed by the proof-combining operation. Some such mechanism will be required anyway, if/when PoSt becomes more sophisticated, so this can be considered an extension which will also be required then.

UPDATE: I added circuit type fields to both PoRep and PoSt. To support this, PoSt creation now always returns an array of `PoStProof`s. PoSt verification was also missing the map from sector IDs to CommRs. I corrected that.


src/systems/filecoin_mining/storage_proving/poster/post_generator.id
```diff
@@ -24,11 +24,13 @@ type PoStGenerator struct {
...
 
     CreateElectionPoStProof(
-        witness sector.PoStWitness
+        randomness  sector.PoStRandomness
+        witness     sector.PoStWitness
     ) sector.PoStProof
 
     CreateSurprisePoStProof(
-        witness sector.PoStWitness
+        randomness  sector.PoStRandomness
```
sternhenri: nice catch

src/systems/filecoin_mining/storage_proving/storage_proving_subsystem.go
```diff
@@ -64,17 +67,14 @@ func (sps *StorageProvingSubsystem_I) GenerateSurprisePoStCandidates(challengeSe
...
-func (sps *StorageProvingSubsystem_I) CreateSurprisePoStProof(challengeSeed sector.PoStRandomness, challengeTickets []sector.PoStCandidate) sector.PoStProof {
+func (sps *StorageProvingSubsystem_I) CreateSurprisePoStProof(challengeSeed sector.PoStRandomness, candidates []sector.PoStCandidate) sector.PoStProof {
 	witness := &sector.PoStWitness_I{
-		Candidates_: challengeTickets,
+		Candidates_: candidates,
 	}
 
 	var poster = sps.PoStGenerator()
-	return poster.CreateSurprisePoStProof(witness)
+	return poster.CreateSurprisePoStProof(challengeSeed, witness)
```
sternhenri: if you're able to review terminology in the `election_post.md` doc, "challengeSeed" is not used, "challengeTicket" is -- would be good to align. Not urgent.

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -971,42 +1024,37 @@ func computePartialTicket(randomness sector.PoStRandomness, sectorID sector.Sect
...
 
-	return candidates
-}
-
-func (sdr *WinStackedDRG_I) CreatePoStProof(privateCandidateProofs []sector.PrivatePoStCandidateProof, challengeSeed sector.PoStRandomness) sector.PoStProof {
-	privateProof := sdr.CreatePrivatePoStProof(privateCandidateProofs, challengeSeed)
+	var circuitProofs []sector.PoStProof
+	for algorithm, proofs := range proofsMap {
+		privateProof := createPrivatePoStProof(algorithm, proofs, challengeSeed)
+		// Hmmmmm: we cannot perform this santiy check without the sectorIDs. Should we require them just for that purpose or perform the check earlier?
```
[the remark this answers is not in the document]
porcuquine: The proposed verification is of the private proof which is an intermediate product in creation of the `PoStProof` this function returns. It *does* happen outside the scope of the function which creates it (`createPrivatePoStProof`), and the context of the verification (the `PrivatePoStProof`) never escapes `CreatePoStProof` anyway. The right answer is probably just to delete the comment and the proposed verification, since it's not required for specification and introduces this awkwardness.

src/libraries/filcrypto/_index.md
```diff
@@ -0,0 +1,8 @@
+---
+menuTitle: Filcrypto
+title: Cryptographic libraries used in Filecoin
+entries:
+- filproofs
+---
+
+{{< incTocMap "/docs/libraries/filcrypto" >}}
```
sternhenri: have you checked that this works @porcuquine ?

src/systems/filecoin_mining/sector/sealing.id
```diff
@@ -35,12 +40,14 @@ type SealVerifyInfo struct {
...
-    DealIDs           deal.DealIDs
+    SealedCID             SealedSectorCID  // CommR
+    SealEpoch             block.ChainEpoch
+    InteractiveEpoch      block.ChainEpoch
+    Proof                 SealProof
+    DealIDs               deal.DealIDs
     SectorNumber
+
+    IsValidAtSealEpoch()  bool
 }
```
sternhenri: @porcuquine Do you not need the randomness on chain for seal?

src/libraries/filcrypto/_index.md
porcuquine: It did for me.

src/systems/filecoin_mining/sector/sealing.id
porcuquine: Did I remove it? I think both randomness values are derived from the corresponding epochs when needed.

porcuquine:
@henri
> Note that I don't see `CreateSealProofInputs` or `SealInputs` getting passed to the library from the non-proofs parts of the spec... Ie the storage_mining_subsystem call that leads to the beginning of sealing and then begins interactivePoRep. That said those can be added later once this is stabilized.

In general, I agree that we can fix incidental plumbing later if needed.

In specific, I think the `CreateSealProofInputs` are being taken apart and fed individually to `filproofs.CreateSealProof` in `sealer.CreateSealProof`. Does that address your concern, or did you mean something else?

src/systems/filecoin_mining/sector/sealing.id
sternhenri: Yes. We chose to explicitly include randomness in the ePoSt case given the randomness is a VRF output. Note that there is a potential space saving step we hadn't considered if we choose to do the surprisePoSt without the VRF prior, in which case we can just use epoch number for surprisePoSt onchain output. The downside being it means the two posts are now meaningfully different.

sternhenri:
Yeah I meant the plumbing isn't done ;). We should do it once the dust settles on interfaces.

src/systems/filecoin_mining/sector/sealing.id
porcuquine: Got it. Let's fix later, since it's unrelated to this change set we need for other reasons?

sternhenri:
Agree

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -26,16 +26,67 @@ type PieceInfo = sector.PieceInfo
...
+type PrivatePostCandidateProof = sector.PrivatePoStCandidateProof
+
 const WRAPPER_LAYER_WINDOW_INDEX = -1
 
-func WinSDRParams(cfg SDRCfg) *WinStackedDRG_I {
+const NODE_SIZE = 32
+const ELECTION_POST_PARTITIONS = 1
+const SURPRISE_POST_PARTITIONS = 1
+const POST_LEAF_CHALLENGE_COUNT = 66
+const POST_CHALLENGE_RANGE_SIZE = 1
```
jbenet: some/all of these constants seem to be part of a single configuration, they should probably be bundled as such

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -726,7 +770,24 @@ func (proof *SDRColumnProof) Verify(root []byte, challenge UInt) bool {
...
 	// publicInputs := GeneratePublicInputs()
+
 	panic("TODO")
+	var bytes []byte
+
+	var registeredProof sector.RegisteredProof
+
+	switch sdr.Cfg().SealCfg().SectorSize() {
+	case GIB_32:
+		registeredProof = sector.RegisteredProof_WinStackedDRG32GiBSeal
```
jbenet: is something like this needed?
```
default:
  panic("no valid configuration")
```

src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -929,7 +988,24 @@ func generateLeafChallenge(randomness sector.PoStRandomness, sectorChallengeInde
...
 
-func generateCandidate(randomness sector.PoStRandomness, aux sector.PersistentProofAux, sectorID sector.SectorID, sectorChallengeIndex UInt, leafChallengeCount int, nodes int, challengeRangeSize int) sector.PoStCandidate {
+func generateCandidate(algorithm sector.ProofAlgorithm, cfg sector.PoStCfg, randomness sector.PoStRandomness, aux sector.PersistentProofAux, sectorID sector.SectorID, sectorChallengeIndex UInt) sector.PoStCandidate {
+	var candidate sector.PoStCandidate
+	switch algorithm {
+	case sector.ProofAlgorithm_StackedDRGSeal:
+		panic("TODO")
+	case sector.ProofAlgorithm_WinStackedDRGSeal:
+		sdr := WinStackedDRG_I{}
+		candidate = sdr._generateCandidate(cfg, randomness, aux, sectorID, sectorChallengeIndex)
```
jbenet:
```go
default:
    panic("invalid proofs algorithm")
```

jbenet:
I think this definitely improves things :+1:

Im not sure, but i think this is still very hard to change to add a new type of proof protocol, especially something that differs substantially. 

I would have expected parametrization to yield different types to exist for each of the relevant algorithms, and parametrizing function calls to the functions common to all algorithms. the current approach explodes out details of the algorithm all over, which puts type switches in a bunch of critical sections to use one detail or another. it is not clear if a substantially different algorithm can be added easily.

---

This approach:

```go
// in a lot of places:
thing() {
  if algo == x {
    do_this()
  } else if algo == y {
    do_that()
  }
}

thing2() {
  switch algo {
  case x:
    // do something thing2 and x related
  case y:
    // do something thing2 and y related
  }
}
```

this gets nasty, hard to follow, and error prone quickly

---

another approach:
```go
// in common file
type PoRepAlgo interface {
  thing()
  thing2()
}

thing() {
  cfg.algorithm.thing()
}

thing2() {
  cfg.algorithm.thing2()
}
```
```go
// in one algorithm file
type X {
  thing()
  thing2()
}

func (x *X) thing() {
  do_this()
}

func (x *X) thing2() {
  // do something thing2 and x related
}
```
```go
// in another algorithm file
type Y {
  thing()
  thing2()
}

func (y *Y) thing() {
  do_that()
}

func (y *Y) thing2() {
  // do something thing2 and y related
}

```

---

hopefully that communicates the point -- i can write a more concrete example, with the relevant algorithm functions if useful

src/libraries/filcrypto/filproofs/algorithms.go
porcuquine: > I think this definitely improves things 👍
> 
> Im not sure, but i think this is still very hard to change to add a new type of proof protocol, especially something that differs substantially.
> 
> I would have expected parametrization to yield different types to exist for each of the relevant algorithms, and parametrizing function calls to the functions common to all algorithms. the current approach explodes out details of the algorithm all over, which puts type switches in a bunch of critical sections to use one detail or another. it is not clear if a substantially different algorithm can be added easily.
 
I avoided Go interfaces in favor of `.id` unions because I thought significant data structure were required to be defined in `.id`. If idiomatic Go is a viable solution, I agree what you suggest would be better, and we can make that change later if we decide it's worthwhile.

As discussed elsewhere, I'm going to merge this now in order to facilitate other refactoring and spec work. We can clean up loose ends then.

PR #774, opened 2019-12-21
StoragePowerActor, StorageMinerActor et al.

jzimmerman:
This PR resolves most of the outstanding TODOs in the storage/mining-related actors, and introduces several updates, including:
- Cron dispatch via StoragePowerActor
- Sector/miner fault state machine isolation
- Improved event handling
- TemporaryFault logic
- Caller verification and associated machinery
- Exec for StorageMinerActors
- Funds checks for TemporaryFault declarations
- Power accounting improvements and fixes

PR #775, opened 2019-12-27
Complete VM Randomness methods

sternhenri:
This PR simplifies the Randomness sections of the VM to get the simplest possible interfaces currently.

It also undoes parts of #774 where deemed unnecessary (specifically with regards to DSTs).


src/systems/filecoin_vm/runtime/runtime.id
```diff
@@ -2,28 +2,24 @@ import actor "github.com/filecoin-project/specs/systems/filecoin_vm/actor"
...
-    // Randomness returns a (pseudo)random string for the given epoch and tag.
-    Randomness(tag filcrypto.DomainSeparationTag, epoch block.ChainEpoch) util.Randomness
-    RandomnessWithAuxSeed(
-        tag      filcrypto.DomainSeparationTag
-        epoch    block.ChainEpoch
-        auxSeed  util.Serialization
-    ) util.Randomness
+    // Chain returns the associated chain which can be used to
+    // derive randomness in this runtime
+    Chain()            chain.Chain
```
anorth: No no absolutely not! The runtime must abstract away the chain and provide a high level interface to what the caller actually needs. Actors do not need access to the full chain.

sternhenri: I don't see how we can get access to the appropriate randomness at various points in the chain without passing in the chain (an abstraction atop a tipset basically)... This is how lotus does it if I'm not mistaken.

sternhenri:
I understand your point @jzimmerman. I happen to think it's overly complicated but I'll add back all the tags...

src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -636,16 +640,8 @@ func (rt *VMContext) ValueReceived() actors.TokenAmount {
...
-	IMPL_TODO()
-	panic("")
-}
-
-func (rt *VMContext) RandomnessWithAuxSeed(
-	tag filcrypto.DomainSeparationTag, epoch block.ChainEpoch, auxSeed util.Serialization) util.Randomness {
-
-	IMPL_TODO()
-	panic("")
+func (rt *VMContext) Chain() chain.Chain {
```
anorth: It looks like this is never called, but GetRandomness is unimplemented, but should be done so in terms of chain.

sternhenri:
Argh. I had to squash this when I rebased. Will redo. Thank you

src/systems/filecoin_blockchain/struct/block/election.id
```diff
@@ -15,12 +15,6 @@ type Ticket struct {
...
-    MinerAddr   addr.Address
-}
-
-type TicketDrawingSeedInput struct {
-    PastTicket  util.Randomness
-    Epoch       abi.ChainEpoch
-}
+type BytesAmount UVarint
+type StoragePower BytesAmount
+type SectorWeight BytesAmount
```
anorth: There are in `abi` now.

sternhenri: bad rebase. Thanks for catching!

src/actors/builtin/storage_miner/storage_miner_actor_code.go
```diff
@@ -59,12 +59,8 @@ func (a *StorageMinerActorCode_I) OnSurprisePoStChallenge(rt Runtime) {
 		numConsecutiveFailures = st.PoStState().As_DetectedFault().NumConsecutiveFailures()
 	}
 
-	IMPL_TODO() // Determine auxiliary seed input to ensure uniqueness
-	challengedSectorsRandomness := rt.RandomnessWithAuxSeed(
-		filcrypto.DomainSeparationTag_SurprisePoStSampleSectors,
-		rt.CurrEpoch()-node_base.SPC_LOOKBACK_POST,
-		addr.Serialize_Address_Compact(rt.CurrReceiver()),
-	)
+	randomnessK := rt.GetRandomness(rt.CurrEpoch() - node_base.SPC_LOOKBACK_POST)
```
jzimmerman: Removing the other seed parameters seems error-prone here, as it encourages users of the API to make calls with potentially duplicated inputs.

src/actors/builtin/storage_miner/storage_miner_actor_code.go
```diff
@@ -626,20 +622,10 @@ func (a *StorageMinerActorCode_I) _rtVerifySurprisePoStOrAbort(rt Runtime, onCha
...
-	postRandomnessInput := rt.RandomnessWithAuxSeed(
-		filcrypto.DomainSeparationTag_SurprisePoStVRFRandomnessInput,
-		challengeEpoch-node_base.SPC_LOOKBACK_POST,
-		addr.Serialize_Address_Compact(rt.CurrReceiver()),
-	)
-
-	if !postRand.Verify(postRandomnessInput, info.WorkerVRFKey()) {
-		rt.AbortStateMsg("Invalid Surprise PoSt. Invalid randomness.")
-	}
+	randomnessK := rt.GetRandomness(challengeEpoch - node_base.SPC_LOOKBACK_POST)
```
jzimmerman: (same as above)

src/algorithms/crypto/randomness.go
```diff
@@ -1,40 +1,41 @@
...
 }
 
-// As in DeriveRand(), but additionally accepts an index into the implicit pseudorandom stream.
-// Index must be strictly positive.
-func DeriveRandWithIndex(tag DomainSeparationTag, s Serialization, index int) Randomness {
-	if index <= 0 {
-		panic("DeriveRandWithIndex only accepts indices > 0")
-	}
-	return _deriveRandInternal(tag, s, index)
+func DeriveRandWithEpoch(tag DomainSeparationTag, tix abi.Randomness, epoch int) Randomness {
```
jzimmerman: Why are we special-casing `Epoch` and `MinerAddr` as input seeds to derive randomness? The previous code allowed randomness derivation from any serializable object, with much less duplication of logic.

src/algorithms/crypto/randomness.id
```diff
@@ -1,8 +0,0 @@
-type DomainSeparationTag enum {
```
jzimmerman: Why did this disappear?

src/actors/builtin/storage_miner/storage_miner_actor_code.go
sternhenri: It's a good point. I thought about it from the standpoint of separating where you get your random seed from and how you compose it to get the wanted random input. will look at refactoring (it's been a bit so wanna convince myself i should)

src/algorithms/crypto/randomness.go
sternhenri: Legibility. These are the only two inputs that we use to derive randomness at this point. I thought the duplication worth abstracting away the internals and making the spec easier to read in terms of what we're basing the randomness on.

src/algorithms/crypto/randomness.id
sternhenri: Cleanly defined in `randomness.go` so again a question of legibility (minimizing imports to the md)

src/algorithms/crypto/randomness.go
anorth: This makes the call sites easier to read. Even if they just wrapped a `DeriveRandWithCBOR`.

src/actors/builtin/storage_miner/storage_miner_actor_code.go
anorth: I like the decomposition - the code behind the runtime API should be as minimal as possible. Perhaps some type definition could provide the prompt to then combine that with other information. Call the thing from the chain a randomness seed.

src/algorithms/crypto/randomness.go
sternhenri: yeah they more or less do just that, wrapping a `_deriveRandInternal` call

sternhenri: And the day we want to add another source of entropy we can just add a new wrapper

src/algorithms/crypto/randomness.go
```diff
@@ -1,40 +1,43 @@
...
-	return _deriveRandInternal(tag, s, index)
+func DeriveRandWithEpoch(tag DomainSeparationTag, tix abi.RandomnessSeed, epoch int) Randomness {
+	buffer := _deriveRandInternal(tag, tix, -1)
+	buffer = append(buffer, BigEndianBytesFromInt(epoch)...)
+	ret := SHA256(buffer)
+	return Randomness(ret)
 }
 
-func _deriveRandInternal(tag DomainSeparationTag, s Serialization, index int) Randomness {
+func _deriveRandInternal(tag DomainSeparationTag, tix abi.RandomnessSeed, index int) util.Bytes {
```
jzimmerman: This parameter doesn't always have to be a ticket, so should remain as `s Serialization`.

src/algorithms/crypto/randomness.go
```diff
@@ -1,40 +1,43 @@
...
 	buffer := []byte{}
 	buffer = append(buffer, BigEndianBytesFromInt(int(tag))...)
 	buffer = append(buffer, BigEndianBytesFromInt(int(index))...)
-	buffer = append(buffer, util.Bytes(s)...)
-	ret := SHA256(buffer)
-	return Randomness(ret)
+	buffer = append(buffer, util.Bytes(tix)...)
```
jzimmerman: [Blocking] We lost the SHA invocation here...

sternhenri: Yeah, given the addition of appropriate entropy in the public methods, the SHA invocations have to be done there. It's a drawback of this interface granted.

jzimmerman: I see, the usage has changed then; in that case, this method definitely shouldn't be called `_deriveRandInternal` as it doesn't derive randomness at all.

(FWIW, the solution I had in mind was to keep `_deriveRandInternal` as is, and have the wrapper functions pass in the serialization of a struct that contains all the relevant parameters. Then the randomness derivation itself is isolated to `_deriveRandInternal`, with the wrappers just packaging up structs.)

sternhenri: Changed accordingly! Fair enough

PR #776, opened 2019-12-28
Worker Key change, id to addr APIs [WIP]

sternhenri:
- remove old keyStore stale data
- Specc’d distinction of ActorID, ActorAddr in md
- Clear API for moving between ID/Addr and getting keys [draft, needs discussion, work]
- Changing worker keys

### DIFF
--- a/src/actors/builtin/storage_miner/storage_miner_actor_code.go
+++ b/src/actors/builtin/storage_miner/storage_miner_actor_code.go
@@ -1,6 +1,7 @@
 package storage_miner
 
 import (
+	"errors"
 	"math/big"
 
 	abi "github.com/filecoin-project/specs/actors/abi"
@@ -25,7 +26,7 @@ const epochUndefined = abi.ChainEpoch(-1)
 
 func (a *StorageMinerActorCode_I) StageWorkerKeyChange(rt Runtime, key filcrypto.VRFPublicKey) (err error) {
 	h, st := a.State(rt)
-	rt.ValidateImmediateCallerIs(st.Info().GetWorker(rt.CurrEpoch()))
+	rt.ValidateImmediateCallerIs(st.Info().Owner())
 	util.IMPL_TODO() // verify public key syntax
 
 	keyAddr, err := addr.Address_Make_Key(node_base.NETWORK, addr.KeyHash(key))
@@ -41,6 +42,23 @@ func (a *StorageMinerActorCode_I) StageWorkerKeyChange(rt Runtime, key filcrypto
 
 	// note that this may replace another pending key change
 	st.Info()._updateMinerPendingKeyChange(keyChange)
+	Release(rt, h, st)
+	return nil
+}
+
+func (a *StorageMinerActorCode_I) CommitWorkerKeyChange(rt Runtime) (err error) {
+	h, st := a.State(rt)
+
+	if st.Info().PendingKeyChange() == nil {
+		return errors.New("No pending key change.")
+	}
+
+	// lazy updates to worker keys if at effectiveness
+	if st.Info().PendingKeyChange().EffectiveAt() <= rt.CurrEpoch() {
+		st.Info()._updateMinerKey()
+		util.TODO()
+		// remains to update the key in every deal, or associate deals to minerActorID instead of workerKey
+	}
 	Release(rt, h, st)
 	return nil
 }
@@ -98,7 +116,7 @@ func (a *StorageMinerActorCode_I) OnSurprisePoStChallenge(rt Runtime) {
 // Invoked by miner's worker address to submit a response to a pending SurprisePoSt challenge.
 func (a *StorageMinerActorCode_I) SubmitSurprisePoStResponse(rt Runtime, onChainInfo sector.OnChainPoStVerifyInfo) {
 	h, st := a.State(rt)
-	rt.ValidateImmediateCallerIs(st.Info().GetWorker(rt.CurrEpoch()))
+	rt.ValidateImmediateCallerIs(st.Info().Worker())
 
 	if !st.PoStState().Is_Challenged() {
 		rt.AbortStateMsg("Not currently challenged")
@@ -149,7 +167,7 @@ func (a *StorageMinerActorCode_I) OnVerifiedElectionPoSt(rt Runtime) {
 // Optimization: PreCommitSector could contain a list of deals that are not published yet.
 func (a *StorageMinerActorCode_I) PreCommitSector(rt Runtime, info sector.SectorPreCommitInfo) {
 	h, st := a.State(rt)
-	rt.ValidateImmediateCallerIs(st.Info().GetWorker(rt.CurrEpoch()))
+	rt.ValidateImmediateCallerIs(st.Info().Worker())
 
 	if _, found := st.Sectors()[info.SectorNumber()]; found {
 		rt.AbortStateMsg("Sector number already exists in table")
@@ -199,8 +217,7 @@ func (a *StorageMinerActorCode_I) PreCommitSector(rt Runtime, info sector.Sector
 
 func (a *StorageMinerActorCode_I) ProveCommitSector(rt Runtime, info sector.SectorProveCommitInfo) {
 	h, st := a.State(rt)
-	workerAddr := st.Info().GetWorker(rt.CurrEpoch())
-	rt.ValidateImmediateCallerIs(workerAddr)
+	rt.ValidateImmediateCallerIs(st.Info().Worker())
 
 	preCommitSector, found := st.Sectors()[info.SectorNumber()]
 	if !found || preCommitSector.State() != SectorState_PreCommit {
@@ -278,7 +295,7 @@ func (a *StorageMinerActorCode_I) ProveCommitSector(rt Runtime, info sector.Sect
 	)
 
 	// Return PreCommit deposit to worker upon successful ProveCommit.
-	rt.SendFunds(workerAddr, preCommitSector.PreCommitDeposit())
+	rt.SendFunds(st.Info().Worker(), preCommitSector.PreCommitDeposit())
 }
 
 /////////////////////////
@@ -289,7 +306,7 @@ func (a *StorageMinerActorCode_I) ExtendSectorExpiration(rt Runtime, sectorNumbe
 	storageWeightDescPrev := a._rtGetStorageWeightDescForSector(rt, sectorNumber)
 
 	h, st := a.State(rt)
-	rt.ValidateImmediateCallerIs(st.Info().GetWorker(rt.CurrEpoch()))
+	rt.ValidateImmediateCallerIs(st.Info().Worker())
 
 	sectorInfo, found := st.Sectors()[sectorNumber]
 	if !found {
@@ -321,7 +338,7 @@ func (a *StorageMinerActorCode_I) ExtendSectorExpiration(rt Runtime, sectorNumbe
 
 func (a *StorageMinerActorCode_I) TerminateSector(rt Runtime, sectorNumber sector.SectorNumber) {
 	h, st := a.State(rt)
-	rt.ValidateImmediateCallerIs(st.Info().GetWorker(rt.CurrEpoch()))
+	rt.ValidateImmediateCallerIs(st.Info().Worker())
 	Release(rt, h, st)
 
 	a._rtTerminateSector(rt, sectorNumber, autil.SectorTerminationType_UserTermination)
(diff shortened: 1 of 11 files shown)


src/appendix/address.md
```diff
@@ -4,6 +4,8 @@ title: "Address"
 
 A Filecoin address is an identifier that refers to an actor in the Filecoin state. All actors (miner actors, the storage market actor, account actors) have an address. This address encodes information about the network to which an actor belongs, the specific type of address encoding, the address payload itself, and a checksum. The goal of this format is to provide a robust address format that is both easy to use and resistant to errors.
 
+Note that each `ActorAddress` in the protocol contains a unique `ActorID` given to it by the `InitActor`. Throughout the protocol, actors are referenced by their IDs rather than full addresses for brevity. However, addresses and IDs are accessible in constant-time using the `AddressMap` (to go from ID to addr) and `IDMap` (in oppositee direction). Using the `IDMap` in conjunction with address resolvers should enable easy lookup of any actor's public keys (should they exist).
```
jzimmerman: typo: `opposite direction`

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -626,7 +641,12 @@ func (a *StorageMinerActorCode_I) _rtVerifySurprisePoStOrAbort(rt Runtime, onCha
 		addr.Serialize_Address_Compact(rt.CurrReceiver()),
 	)
 
-	if !postRand.Verify(postRandomnessInput, info.WorkerVRFKey()) {
+	minerKey, err := info.GetWorker(rt.CurrEpoch()).GetKey()
+	if err != nil {
+		rt.AbortStateMsg("Couldn't get worker key to verify surprise PoSt.")
+	}
```
jzimmerman: Can this ever fail? If not, let's make it an assert. We'll need to decide later whether assertion failures should crash the VM or not, but in any case, we're currently distinguishing between errors that might legitimately arise from user input (as `rt.Abort`) vs. errors that indicate a software implementation bug (as `util.Assert`).

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
```diff
@@ -115,17 +115,37 @@ func MinerInfo_New(
...
+
+	// lazy updates to worker keys if at effectiveness
+	if mi.PendingKeyChange().EffectiveAt() <= currEpoch {
+		mi._worker_ = mi.PendingKeyChange().NewKey()
+		mi.PendingKeyChange_ = nil
+		util.TODO()
+		// remains to update the key in every deal, or associate deals to minerID
+	}
+	return mi._worker()
+}
```
jzimmerman: We can't do state updates here -- it's used as a pure function elsewhere (e.g., in `TerminateSector`), where we call `Release` instead of `UpdateRelease`. Let's split this stanza out into a separate function, if we need to do this lazy update. (It's not clear to me that we do, though -- isn't the logic identical if we just return the would-be updated worker here, without actually committing any state updates?)

src/systems/filecoin_nodes/key_store/_index.md
```diff
@@ -17,7 +17,9 @@ Filecoin storage miners rely on three main components:
 
 While miner addresses are unique, multiple storage miner actors can share an owner public key or likewise a worker public key.
 
+The process for chaing the worker keypairs on-chain (i.e. the workerKey associated to a storage miner) is specified in {{<sref storage_miner_actor>}}. Note that this is a two-step process. First a miner stages a change by sending a message to the chain. When received, the key change is staged to occur in twice the randomness lookback parameter number of epochs, to prevent adaptive key selection attacks.
```
jzimmerman: typo: `chaing the worker keypairs`

src/systems/filecoin_vm/sysactors/init_actor.id
```diff
@@ -5,6 +5,7 @@ import vmr "github.com/filecoin-project/specs/systems/filecoin_vm/runtime"
 type InitActorState struct {
     // responsible for create new actors
     AddressMap   {addr.Address: addr.ActorID}
+    IDMap        {addr.ActorID: [addr.Address]}
```
anorth: We went to great effort to remove this, it's not necessary. Given an actor id, it's address is `makeIdAddress(id)`.

The AddressMap above maps _non-ID addresses_ to ids, but there is no reason for the reverse map. An actors ID address is its canonical address.

src/appendix/address.md
anorth: This sounds confused and some bits contradict my understanding and the implementations.
- Almost universally, actors are referenced by ID-address, not ID. We could change this but the observation isn't true. E.g. state tree keys are ID-addresses and the runtime always resolves to ID addresses before exposing them to actor code.
- ID-addresses are computable from IDs, and vice versa.
- Most actors have an alternative address which is *not* their key in the state tree, but is resolved to that address during message processing. Accounts have a key-based address, and non-singleton actors have a temporary reorg-proof address.
- An account actor's crypto-address (for signature verification) is found by looking up its actor state, keyed by the canonical ID-address. If you already had the alternative pubkey address, there's nothing to look up. There is no map from ID-address to pubkey address because we don't need one.

In order to be informative I think this would need to be a much more detailed description of the various address types, why some actors have multiple, the state tree keys etc.

src/systems/filecoin_mining/storage_mining/storage_miner_actor.id
```diff
@@ -82,18 +81,25 @@ type StorageMinerActorCode struct {
...
     // Worker account for this miner.
     // This will be the key that is used to sign blocks created by this miner, and
     // sign messages sent on behalf of this miner to commit sectors, submit PoSts, and
     // other day to day miner activities.
-    Worker                  address.Address
-    WorkerVRFKey            filcrypto.VRFPublicKey
+    _worker  addr.Address
+
+    _updateMinerPendingKeyChange(mkc MinerKeyChange)
+    GetWorker(epoch block.ChainEpoch) addr.Address
```
anorth: The parameter to this method is unnecessary. 100% of the invocations pass the current epoch, the epoch of the state being inspected. There understandably seems to be no reason to want to know a miner's key some epoch in the future. 

I see further down that it has side-effects which I think we should strictly separate from access, and are ineffective in read-only contexts anyway. Given this I think the method can be removed and the member exported and consulted directly.

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
anorth: Agreed we can't do state updates here, and I think the silent nature is confusing. We should make this change quite explicit.

My preferred approach would be to add a CommitKeyChange explicit method which will persist the change if called after suitable delay

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
```diff
@@ -20,11 +20,26 @@ import (
 // Actor methods
 ////////////////////////////////////////////////////////////////////////////////
 
-func (a *StorageMinerActorCode_I) GetWorkerVRFKey(rt Runtime) filcrypto.VRFPublicKey {
+func (a *StorageMinerActorCode_I) StageWorkerKeyChange(rt Runtime, key filcrypto.VRFPublicKey) (err error) {
 	h, st := a.State(rt)
-	ret := st.Info().WorkerVRFKey()
+	rt.ValidateImmediateCallerIs(st.Info().GetWorker(rt.CurrEpoch()))
```
anorth: This must be the *owner*, not the worker, so the owner can recover from compromised or lost keys.

src/appendix/address.md
sternhenri: Fair enough. I fixed to include your comments @anorth. Clearly I'm not quite paged in here. Where can I pull in the full details from? Lotus or go-fil (eg)?

src/systems/filecoin_mining/storage_mining/storage_miner_actor_code.go
sternhenri: Thanks for catching that!

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
sternhenri: Changed accordingly.

A question though, if the `CommitKeyChange()` method is called explicitly, then there may be some number of epochs in which the key could be changed but is not changed yet, thereby allowing the storageMiner to "straddle" two worker keys momentarily and only make the switch when relevant. 

I claim that unless the systemActor basically queues up key changes to push them at the appropriate epoch with the cron tick, then the rational strategy is to always run two keys to gain an advantage. Do you agree @anorth @jzimmerman ?
If so, it's worth reworking this so the cron actor prompts the key change on time.

src/actors/builtin/storage_miner/storage_miner_actor_code.go
```diff
@@ -23,11 +24,43 @@ import (
...
+
+	// note that this may replace another pending key change
+	st.Info()._updateMinerPendingKeyChange(keyChange)
 	Release(rt, h, st)
-	return ret
+	return nil
+}
+
+func (a *StorageMinerActorCode_I) CommitWorkerKeyChange(rt Runtime) (err error) {
+	h, st := a.State(rt)
```
sternhenri: Should we verify the caller here @anorth ? I think not

anorth: Probably yes (as the owner)

src/systems/filecoin_mining/storage_mining/storage_miner_actor_state.go
anorth: Fair concern. I was hoping to avoid the cron because it'll involve more complexity. But I think in the giant actor PR of late December a job queue type thing was added to SPA, so this could register for a callback from that.

src/actors/builtin/storage_miner/storage_miner_actor_code.go
```diff
@@ -23,11 +24,43 @@ import (
 // Actor methods
 ////////////////////////////////////////////////////////////////////////////////
 
-func (a *StorageMinerActorCode_I) GetWorkerVRFKey(rt Runtime) filcrypto.VRFPublicKey {
+func (a *StorageMinerActorCode_I) StageWorkerKeyChange(rt Runtime, key filcrypto.VRFPublicKey) (err error) {
```
anorth: Actors never return errors, they only `rt.Abort()`

src/actors/builtin/storage_miner/storage_miner_actor_code.go
```diff
@@ -23,11 +24,43 @@ import (
...
+
+	if st.Info().PendingKeyChange() == nil {
+		return errors.New("No pending key change.")
+	}
+
+	// lazy updates to worker keys if at effectiveness
+	if st.Info().PendingKeyChange().EffectiveAt() <= rt.CurrEpoch() {
+		st.Info()._updateMinerKey()
+		util.TODO()
+		// remains to update the key in every deal, or associate deals to minerActorID instead of workerKey
```
anorth: IMO deals should be associated with actors and the key looked up any time it is needed. cc @hannahhoward

src/appendix/address.md
```diff
@@ -4,6 +4,14 @@ title: "Address"
 
 A Filecoin address is an identifier that refers to an actor in the Filecoin state. All actors (miner actors, the storage market actor, account actors) have an address. This address encodes information about the network to which an actor belongs, the specific type of address encoding, the address payload itself, and a checksum. The goal of this format is to provide a robust address format that is both easy to use and resistant to errors.
 
+Note that each `ActorAddress` in the protocol contains a unique `ActorID` given to it by the `InitActor`. Throughout the protocol, actors are referenced by their ID-addresses. ID-addresses are computable from IDs (using `makeIdAddress(id)`), and vice versa (using the `AddressMap` to go from addr to ID).
```
anorth:
```suggestion
Note that each `ActorAddress` in the protocol contains a unique `ActorID` given to it by the `InitActor`. Throughout the protocol, actors are referenced by their ID-addresses. ID-addresses are computable from IDs (using `makeIdAddress(id)`), and vice versa.
```

src/appendix/address.md
```diff
@@ -4,6 +4,14 @@ title: "Address"
...
+
+Most actors have an alternative address which is not their key in the state tree, but is resolved to their ID-address during message processing.
```
anorth:
```suggestion
Most actors have an alternative address which is not their key in the state tree, but is resolved to their ID-address during message processing via the InitActor's address map.
```

src/actors/builtin/storage_miner/storage_miner_actor_code.go
sternhenri: duh. Thank you!

src/actors/builtin/storage_miner/storage_miner_actor_code.go
sternhenri: checking it's `addr.CronActorAddr` given https://github.com/filecoin-project/specs/pull/776#discussion_r364480899

src/actors/builtin/storage_miner/storage_miner_actor_code.go
hannahhoward: i will admit being a bit out of context here, but if the question is do we need to be able to, at any time, query the chain to determine "what deals are currently on chain for miner" the answer is yes.

hannahhoward: or query the actor... sorry again, out of context and my area of expertise

PR #777, opened 2019-12-29
Payment Channels [WIP]

sternhenri:
Lays out how payment channels will work in FIL, based on the lotus impl
(https://github.com/filecoin-project/lotus/blob/master/chain/types/voucher.go, https://github.com/filecoin-project/lotus/blob/master/chain/actors/actor_paych.go). Interface is included, actual actor impl left out for now.

Some details remain to be worked out with lotus:
- their use of preimage and data in each voucher
- their ordering of vouchers and lanes
- the reason for the timelock
- the reason for the extra field

Once done, both spec and lotus can converge, and spec can be fleshed out farther


mishmosh:
What's next for this PR @sternhenri?

anorth:
The actor code will shortly disappear to the `specs-actors` repository, so that change will need to be applied there.

sternhenri:
Superceded by #854

Issue #621, opened 2019-11-05
Encoding for on-chain structures


anorth:
The spec is very light on details about the serialization/encoding of on-chain structures. At present, CBOR is not mentioned but a few structs have a `// representation tuple` comment.

I believe the intention at present is for all structures to be CBOR-tuple encoded (i.e. a CBOR array with items corresponding to struct fields in their order of declaration). This is efficient but has some potential problems. I'm filing this issue so that we have them written down somewhere.

@jbenet's most recent [declaration](
https://filecoinproject.slack.com/archives/CHMNDCK9P/p1571738348373600?thread_ts=1571629767.290400) is:
> - i'm OK with tuple encoding for testnet
> - we MAY ship mainnet w/ tuple encoding
> - we MAY have to change from tuple to int-keyed map for structs for mainnet
> - we will prioritize this along other changes that come out of security review during testnet
> - IF [int-keyed maps are already implemented] we can motivate realignment to that now.
> - IF NOT (gfc has string maps, but not int-keyed maps and those would be a lot of work), proceed w/ tuple. but keep it easy to change this

anorth:
Problem: future-proofing.

[Quoth](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1571633041295500?thread_ts=1571629767.290400&cid=CHMNDCK9P) @jbenet 

> in light of evolving protocols, security oriented protocols that serialize into non-self-describing formats take great care to ensure fields are appropriately tagged to ensure the right serialized field value is serialized/deserialized into the right in-memory field. protobuf, capnp, and more enforce this, and have for decades, for precisely protocol evolution and security. deserializing field A into field B is a class of bug trivially defeated and not worth exposing ourselves to.
>
> - this compounds as formats change and programs (which do not all update in lockstep_ continue to read old and new versions of structures).
> - this is made specially worse in hash-linked data structures which cannot be upgraded by migrating data, but instead tend to force all programs in the future to read old structure versions. field tagging is key for secure schema evolution

anorth:
Another annoyance I have just learned about is that tuple-encoding does not play nicely with graphsync. IPLD selectors operate over the encoded IPLD nodes, which in this case will be lists. So selectors for chain syncing need to be expressed with indices, rather than field names.

This is not the end of the world, of course. We can declare (or even reflect) a mapping of field name->index and use symbolic constants to construct queries.

anorth:
cc @hannahhoward @icorderi @whyrusleeping

Issue #615, opened 2019-10-31
Cleanup int, uint, uint64, Int, Uint, UInt, UVarint, Uint64, etc.


icorderi:
The spec has uppercase, lower case, newtype defs, encoding expectations, size bounds all used indiscriminately in the doc. 

We need to finalize expected size bound and encoding for each numerical type. 

We proposed the following sets of numbers:
- Special number types: 
    * `Varint64` (64-bit number encoded cheaply on low values), 
    * `UVarint64` (64-bit unsigned number encoded cheaply on low values), 
    * `BigInt` (arbitrary length integer, currently using leb128 encoding)  
    * _optional_ `[U]Varint{16-32}` to complete to the family (this doesn't affect the chain storage but puts memory size bounds and helps advise implementations on expected sizes, i.e. a `UVarint64` for a `MethodID` inside an actor seems a bit much). 
- Primitive number types: `uint{8-64}`, `int{8-64}` (this are NOT varlen encoded, LE/BE needs to be specified in the encoding section of the spec)

> **Note:** `Varint` and `UVarint` default to 64-bit, but it might be best to be consistent across the board and express numbers with their sizes.

zixuanzh:
@icorderi, yes, this was left as a TODO. Will put up a PR to clean up these types. Your PR will be welcomed too :heart:

Kubuxu:
> `BigInt` (arbitrary length integer, currently using leb128 encoding)

I'm not aware of `BigInt` being `leb128` encoded. `UVarint*` are `leb128` encoded.
Lotus currently encodes `BigInt` as big-endian bytearray.

Kubuxu:
Another thing is: we are using `CBOR`, we should follow CBOR encoding rules: https://tools.ietf.org/html/rfc7049#section-2.1 (Major type 0 and Major type 1).

hannahhoward:
^^^ +1 to @Kubuxu 's suggestion -- as I see it, CBOR is the lingau franca of Filecoin data exchange, and we should focus on staying in within rules of type encoding to CBOR, since we frequently then calculate CIDs that are dependent on that encoding.

jbenet:
Thanks for bringing this up. will take a look this week.

(some quick notes)
- Most numbers will be small in most cases but need to be able to grow (therefore varints)
- Yes i believe Varints should be LEB128 encoded
[...]

PR #779, opened 2020-01-01
Data formats and Serialization

sternhenri:
This PR establishes the data serialization format, and some data formats rules per https://github.com/filecoin-project/specs/issues/615 and https://github.com/filecoin-project/specs/issues/621.

@anorth feel free to piggyback off of this for the HAMT and Int work if useful to you. LMK if further work is required.

### DIFF
--- a/src/algorithms/crypto/randomness.md
+++ b/src/algorithms/crypto/randomness.md
@@ -13,7 +13,7 @@ We describe this formatting below.
 Any randomness derived from on-chain values uses the following encodings to represent these values as bytes:
 
 - **Bytes**: Bytes
-- **Ints**: Little-endian uint64 representation
+- **Ints**: Big-endian uint64 representation
 - **Strings**: ASCII
 - **Objects**: Their specified Serialization, currently CBOR-based serialization defined on algebraic datatypes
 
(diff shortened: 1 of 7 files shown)


src/algorithms/crypto/randomness.md
```diff
@@ -10,9 +10,10 @@ We describe this formatting below.
 
 ## Encoding On-chain data for randomness
 
-Any randomness derived from on-chain values uses the following encodings to represent these values as bytes
+Any randomness derived from on-chain values uses the following encodings to represent these values as bytes:
+
 - **Bytes**: Bytes
-- **Ints**: Little endian uint64 representation
+- **Ints**: Little-endian uint64 representation
```
anorth: As discussed [here](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1576707533036200), CBOR encodes ints big-endian, so specifiying little-endian here will require re-serialization which is pretty low value.

sternhenri: Happy to do big-endian, note though that this is encoding for the purposes of randomness generation, so it should not get CBOR serialized thereafter...

PR #783, opened 2020-01-07
EPoSt verif duplicate prevention

sternhenri:
**In this PR**:
- added check for duplicates in PoSt verification


sternhenri:
Hadn't realized that `challengeIndices` are unique across challenged sectors. Can you confirm this looks right?

PR #767, opened 2019-12-19
Mining Cycle cleanup

sternhenri:
**In this PR:**
- A lot of cleanup of stale files, links, code
- md touch up of `mining cycle` and `block producer`
- fleshed out all of the sms code working through all handlable TODOs and adding info on:
  - `createMiner`
  - `runMiningCycle`
  - surprisePoSt processing and submission flow
- some util code as pertains to all of the above

PR #784, opened 2020-01-07
fix addresses in msg sends

sternhenri:
Fix up to #767


src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
```diff
@@ -383,14 +385,18 @@ func (sms *StorageMiningSubsystem_I) _submitSurprisePoStMessage(state stateTree.
 	params := make([]util.Serialization, 1)
 	params[0] = sector.Serialize_OnChainPoStVerifyInfo(sPoSt)
 
-	sysActor, ok := state.GetActor(addr.SystemActorAddr)
+	workerAddr, err := addr.Address_Make_Key(node_base.NETWORK, addr.KeyHash(sms._keyStore().WorkerKey().VRFPublicKey()))
+	if err != nil {
+		return err
+	}
```
anorth: Alternatively it might be easier to read the worker address out of the MinerInfo on-chain structure and plumb it to here.

sternhenri: yeah I thought about it but thought it easier to avoid reading the on-chain structures in order to get info about your own node...

PR #788, opened 2020-01-07
Move all actor code to a top-level actors/ directory.

anorth:
This is in preparation for extraction to a new repository as an executable Go project.

Minimal change other than moving and fixing up imports, though I did package each actor to make any interdependencies more explicit. `git mv` should mean rebasing outstanding PRs against this is easy. 

There are still a bunch of dependencies from this subdir to other spec code. I'll attempt to resolve as many as I can before pulling the code out, so we see the diffs here. I'll work on top of this branch.


src/systems/filecoin_vm/interpreter/vm_registry.go
```diff
@@ -63,22 +65,22 @@ func init() {
...
 
-	cron := &sysactors.CronActorCode_I{}
+	cron := &cronact.CronActorCode_I{}
 
-	RegisterActor(InitActorCodeID, &sysactors.InitActorCode_I{})
+	RegisterActor(InitActorCodeID, &initact.InitActorCode_I{})
 	RegisterActor(CronActorCodeID, cron)
-	RegisterActor(AccountActorCodeID, &sysactors.AccountActorCode_I{})
-	RegisterActor(StoragePowerActorCodeID, &spc.StoragePowerActorCode_I{})
-	RegisterActor(StorageMarketActorCodeID, &market.StorageMarketActorCode_I{})
```
zixuanzh: Do we need to register BurntFund and Reward actors here too?

anorth: Yes we probably do, but I'm trying to avoid changes, including fixes, while doing this move. This piece in particular might change anyway as we hook up execution scaffolds.

PR #790, opened 2020-01-08
Break some dependencies from actors on other spec code.

anorth:
Much of this is noise from changing imports, but some significant changes:
- Moved some type aliases into the actors directory (more to follow).
- Removed the InvokeMethod calls and method dispatch, which was mostly just stubs.
  Dispatch belongs in the VM, not the actors themselves, and implementations have existing
  patterns. I do plan to add back exports lists defining the method numbers and types later.
- Changed MethodParams to a single serialized object and replaced use of codegen serialization
  placeholders with a global placeholder. This latter will be filled in with real CBOR
  serialization as part of making the actors executable.

This is based on #788, rebase before merging.

Imports remaining to break:
```
github.com/filecoin-project/specs/algorithms/crypto
github.com/filecoin-project/specs/libraries/filcrypto/filproofs
github.com/filecoin-project/specs/libraries/ipld
github.com/filecoin-project/specs/libraries/libp2p
github.com/filecoin-project/specs/systems/filecoin_blockchain/struct/block
github.com/filecoin-project/specs/systems/filecoin_markets/storage_market/storage_deal
github.com/filecoin-project/specs/systems/filecoin_mining/sector
github.com/filecoin-project/specs/systems/filecoin_nodes/node_base
github.com/filecoin-project/specs/systems/filecoin_vm/actor
github.com/filecoin-project/specs/systems/filecoin_vm/actor/address
github.com/filecoin-project/specs/systems/filecoin_vm/actor_interfaces
github.com/filecoin-project/specs/systems/filecoin_vm/indices
github.com/filecoin-project/specs/systems/filecoin_vm/runtime
github.com/filecoin-project/specs/util
```

cc @jzimmerman

### DIFF
--- a/src/actors/builtin/storage_miner/storage_miner_actor_code.go
+++ b/src/actors/builtin/storage_miner/storage_miner_actor_code.go
@@ -1,7 +1,11 @@
 package storage_miner
 
 import (
-	actor_util "github.com/filecoin-project/specs/actors/util"
+	"math/big"
+
+	actors "github.com/filecoin-project/specs/actors"
+	serde "github.com/filecoin-project/specs/actors/serde"
+	autil "github.com/filecoin-project/specs/actors/util"
 	filcrypto "github.com/filecoin-project/specs/algorithms/crypto"
 	filproofs "github.com/filecoin-project/specs/libraries/filcrypto/filproofs"
 	libp2p "github.com/filecoin-project/specs/libraries/libp2p"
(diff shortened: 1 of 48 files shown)


src/actors/builtin/storage_miner/storage_miner_actor_code.go
```diff
@@ -665,16 +676,16 @@ func (a *StorageMinerActorCode_I) _rtVerifySealOrAbort(rt Runtime, onChainInfo s
...
+			sectorSize,
+			onChainInfo.DealIDs(),
+		),
 	))
 	Assert(err == nil)
 
 	// Unless we enforce a minimum padding amount, this totalPieceSize calculation can be removed.
 	// Leaving for now until that decision is entirely finalized.
-	var totalPieceSize util.UInt
+	var totalPieceSize uint64
```
zixuanzh: We have removed the maximum padding amount enforcement, unclear what minimum padding here refers to (could be a typo). totalPieceSize is not used anywhere, we can remove this calculation.

src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -470,7 +471,8 @@ func _invokeMethodInternal(
 
 	rt._running = true
 	ret, exitCode = _catchRuntimeErrors(func() InvocOutput {
-		methodOutput := actorCode.InvokeMethod(rt, method, params)
+		IMPL_TODO("dispatch to actor code")
```
zixuanzh: Authorized user of IMPL_TODO 😄

src/actors/builtin/storage_miner/storage_miner_actor_code.go
anorth: Thanks, I'll save it for another PR.

PR #792, opened 2020-01-08
Improving Rational Calc on hash metrics

nicola:


nicola:
Adding this to the rational-calc-3 PR

PR #793, opened 2020-01-09
Move ChainEpoch, StoragePower, StorageWeight types into actors package.

anorth:
No semantic changes intended, a pure move.

### DIFF
--- a/src/algorithms/crypto/filcrypto.go
+++ b/src/algorithms/crypto/filcrypto.go
@@ -8,7 +8,7 @@ import (
 	util "github.com/filecoin-project/specs/util"
 )
 
-func RandomInt(randomness util.Randomness, nonce int, limit int) int {
+func RandomInt(randomness util.Bytes, nonce int, limit int) int {
 	nonceBytes := LittleEndianBytesFromInt(nonce)
 	input := randomness
 	input = append(input, nonceBytes...)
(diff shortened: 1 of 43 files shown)


src/algorithms/crypto/filcrypto.go
```diff
@@ -8,7 +8,7 @@ import (
 	util "github.com/filecoin-project/specs/util"
 )
 
-func RandomInt(randomness util.Randomness, nonce int, limit int) int {
+func RandomInt(randomness util.Bytes, nonce int, limit int) int {
```
zixuanzh: nit: should this be `actors.Randomness`?

src/systems/filecoin_blockchain/storage_power_consensus/storage_power_consensus_subsystem.go
```diff
@@ -65,18 +66,18 @@ func (spc *StoragePowerConsensusSubsystem_I) _getStoragePowerActorState(stateTre
 	return st
 }
 
-func (spc *StoragePowerConsensusSubsystem_I) GetTicketProductionRand(chain chain.Chain, epoch block.ChainEpoch) util.Randomness {
+func (spc *StoragePowerConsensusSubsystem_I) GetTicketProductionRand(chain chain.Chain, epoch actors.ChainEpoch) util.Randomness {
```
zixuanzh: nit: seeing util.Randomness in a few places, suggest deleting the type in util

anorth: There are a ton. I decided to leave util.Randomness behind because this change was all I needed to get the actors to not depend on it, and the name `actors` seemed out of place as an import. We could follow-up to replace it with abi.Randomness after #794. 

Or I would be just as happy deleting it and replacing with `[]byte`.

PR #794, opened 2020-01-09
Move types shared with actors to actors/abi package

anorth:
Moved shared types out of actors top-level to an `abi` (application binary interface) package. There are many more to come, including proofs and sector info used by more than one actor.

This was an automated refactor performed with find/replace tools. No semantic changes.

~Based on #793~

PR #795, opened 2020-01-09
Move the CodeID type to actors/abi and convert to a real CID

anorth:
~Based on #794~

### DIFF
--- a/src/actors/builtin/codes.go
+++ b/src/actors/builtin/codes.go
@@ -1,0 +1,42 @@
+package builtin
+
+import (
+	"github.com/filecoin-project/specs/actors/abi"
+	"github.com/ipfs/go-cid"
+	mh "github.com/multiformats/go-multihash"
+)
+
+// The built-in actor code IDs
+var SystemActorCodeID abi.ActorCodeID
+var InitActorCodeID abi.ActorCodeID
+var CronActorCodeID abi.ActorCodeID
+var AccountActorCodeID abi.ActorCodeID
+var StoragePowerActorCodeID abi.ActorCodeID
+var StorageMinerActorCodeID abi.ActorCodeID
+var StorageMarketActorCodeID abi.ActorCodeID
+var PaymentChannelActorCodeID abi.ActorCodeID
+var MultisigActorCodeID abi.ActorCodeID
+var RewardActorCodeID abi.ActorCodeID
+
+func init() {
+	builder := cid.V1Builder{Codec: cid.Raw, MhType: mh.IDENTITY}
+	makeBuiltin := func(s string) abi.ActorCodeID {
+		c, err := builder.Sum([]byte(s))
+		if err != nil {
+			panic(err)
+		}
+		id := abi.ActorCodeID(c)
+		return id
+	}
+
+	SystemActorCodeID = makeBuiltin("fil/1/system")
+	InitActorCodeID = makeBuiltin("fil/1/init")
+	CronActorCodeID = makeBuiltin("fil/1/cron")
+	AccountActorCodeID = makeBuiltin("fil/1/account")
+	StoragePowerActorCodeID = makeBuiltin("fil/1/storagepower")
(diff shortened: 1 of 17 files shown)


src/actors/builtin/codes.go
```diff
@@ -0,0 +1,49 @@
...
+var SystemActorCodeID abi.ActorCodeID
+var InitActorCodeID abi.ActorCodeID
+var CronActorCodeID abi.ActorCodeID
+var AccountActorCodeID abi.ActorCodeID
+var StoragePowerActorCodeID abi.ActorCodeID
+var StorageMinerActorCodeID abi.ActorCodeID
+var StorageMarketActorCodeID abi.ActorCodeID
+var PaymentChannelActorCodeID abi.ActorCodeID
+var MultisigActorCodeID abi.ActorCodeID
+
```
zixuanzh: do we need to include BurntFundActor and RewardActor?

src/actors/builtin/codes.go
```diff
@@ -0,0 +1,49 @@
...
+	SystemActorCodeID = makeBuiltin("fil/1/system")
+	InitActorCodeID = makeBuiltin("fil/1/init")
+	CronActorCodeID = makeBuiltin("fil/1/cron")
+	AccountActorCodeID = makeBuiltin("fil/1/account")
+	StoragePowerActorCodeID = makeBuiltin("fil/1/storagepower")
+	StorageMinerActorCodeID = makeBuiltin("fil/1/storageminer")
+	StorageMarketActorCodeID = makeBuiltin("fil/1/storagemarket")
+	PaymentChannelActorCodeID = makeBuiltin("fil/1/paymentchannel")
+	MultisigActorCodeID = makeBuiltin("fil/1/multisig")
```
zixuanzh: ditto

src/systems/filecoin_vm/interpreter/vm_registry.go
```diff
@@ -67,11 +56,11 @@ func _registerBuiltinActors() {
...
-	RegisterActor(InitActorCodeID, &initact.InitActorCode_I{})
-	RegisterActor(CronActorCodeID, cron)
-	RegisterActor(AccountActorCodeID, &accact.AccountActorCode_I{})
-	RegisterActor(StoragePowerActorCodeID, &spowact.StoragePowerActorCode_I{})
-	RegisterActor(StorageMarketActorCodeID, &smarkact.StorageMarketActorCode_I{})
+	RegisterActor(builtin.InitActorCodeID, &initact.InitActorCode_I{})
+	RegisterActor(builtin.CronActorCodeID, cron)
+	RegisterActor(builtin.AccountActorCodeID, &accact.AccountActorCode_I{})
+	RegisterActor(builtin.StoragePowerActorCodeID, &spowact.StoragePowerActorCode_I{})
+	RegisterActor(builtin.StorageMarketActorCodeID, &smarkact.StorageMarketActorCode_I{})
```
zixuanzh: ditto

src/actors/builtin/codes.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// Tests whether a CID is a builtin actor code ID.
+func IsBuiltinCodeID(c abi.ActorCodeID) bool {
```
zixuanzh: this function is never called and type abi.ActorCodeID can be any CID

src/actors/builtin/codes.go
anorth: BurntFunds is just an account. Will add Reward

src/systems/filecoin_vm/interpreter/vm_registry.go
anorth: There are many missing here, but I'm just trying to extract things, not fix them right now.

src/actors/builtin/codes.go
anorth: Thanks. I implemented it but then turns out all the call sites disappeared when it stopped being a union. I'll delete it.

PR #798, opened 2020-01-09
simplified surprise post randomness generation

sternhenri:
**In this PR:**
- Simplified SurprisePoSt randomness generation to not use VRF
- accordingly removed randomness from on-chain footprint for SurprisePoSt

the above necessitated splitting the `OnChainPoStVerifyInfo` for ePoSt and sPoSt. I'll note two things for @porcuquine to opine on:
- Do we need to include the round from which randomness was drawn for the surprise PoSt in the `OnChainSurprisePoStVerifyInfo`? It is already on chain in the actor code, but maybe it makes sense to have it there since it is needed for verif (so while we can get it elsewhere, maybe we shouldn't).
- It seems to me like the interface of `PoStVerifyInfo` potentially duplicates the `Randomness` field in the case of `OnChainElectionPoStVerifyInfo` (specifically randomness is both present in `Randomness` and `OnChain.Randomness`). I think that's fine. Just FYI.

### DIFF
--- a/src/actors/builtin/storage_miner/storage_miner_actor_code.go
+++ b/src/actors/builtin/storage_miner/storage_miner_actor_code.go
@@ -77,7 +77,7 @@ func (a *StorageMinerActorCode_I) OnSurprisePoStChallenge(rt Runtime) {
 }
 
 // Invoked by miner's worker address to submit a response to a pending SurprisePoSt challenge.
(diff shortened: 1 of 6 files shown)


sternhenri:
Noted. Happy to discuss with you and change when best for you!

PR #799, opened 2020-01-09
Fix/linux build

sternhenri:
Following up from https://github.com/filecoin-project/specs/pull/521.

I don't believe this fixed it @Stebalien. Can you test and close if not please?


Stebalien:
I wouldn't merge this. `make serve-website` doesn't work on linux but everything else does (including `make serve` which is the better target anyways). Some of these changes are still useful but some are definitely outdated.

sternhenri:
Ok. Then I'll close for now until I get a chance to spend time actually fixing this up for linux :). Thanks so much @Stebalien !

PR #800, opened 2020-01-09
Move the runtime interface (but not implementation) alongside actors

anorth:
The runtime interface on which actors depend must live alongside the actors so they can import it. Exitcode and indices come with it. The implementation remains behind, though.

Still TODO: 
- convert from `.id` to a Go interface. 

FYI @jzimmerman


src/actors/runtime/runtime_util.go
```diff
@@ -6,11 +6,11 @@ import (
 	filcrypto "github.com/filecoin-project/specs/algorithms/crypto"
 	addr "github.com/filecoin-project/specs/systems/filecoin_vm/actor/address"
 	ai "github.com/filecoin-project/specs/systems/filecoin_vm/actor_interfaces"
-	msg "github.com/filecoin-project/specs/systems/filecoin_vm/message"
-	exitcode "github.com/filecoin-project/specs/systems/filecoin_vm/runtime/exitcode"
 	util "github.com/filecoin-project/specs/util"
 )
 
+// TODO: most of this file doesn't need to be part of runtime, just generic actor shared code.
```
zixuanzh: move to actor_util at some point?

PR #801, opened 2020-01-10
Move all piece, sector, and proof types used on-chain into the actors abi.

anorth:
A bunch of types are used by multiple actors, so they moved to abi (I think with some refactoring some could move on to a single actor).

All the action is in the `actors/abi/*.go` files, and the various id files from which those types were yanked. The rest is mechanical update of imports and coping with conversion from `.id`.

### DIFF
--- a/src/actors/runtime/indices/indices.go
+++ b/src/actors/runtime/indices/indices.go
@@ -4,6 +4,7 @@ import (
 	abi "github.com/filecoin-project/specs/actors/abi"
 	actor_util "github.com/filecoin-project/specs/actors/util"
 	deal "github.com/filecoin-project/specs/systems/filecoin_markets/storage_market/storage_deal"
+	"math/big"
 )
 
 var PARAM_FINISH = actor_util.PARAM_FINISH
@@ -76,13 +77,13 @@ func (inds *Indices_I) PledgeCollateralReq(minerNominalPower abi.StoragePower) a
 	panic("")
 }
 
-func (inds *Indices_I) SectorWeightProportion(minerActiveSectorWeight abi.SectorWeight) abi.SectorWeight {
+func (inds *Indices_I) SectorWeightProportion(minerActiveSectorWeight abi.SectorWeight) big.Int {
 	// return proportion of SectorWeight for miner
 	PARAM_FINISH()
 	panic("")
 }
 
-func (inds *Indices_I) PledgeCollateralProportion(minerPledgeCollateral abi.TokenAmount) abi.TokenAmount {
+func (inds *Indices_I) PledgeCollateralProportion(minerPledgeCollateral abi.TokenAmount) big.Int {
 	// return proportion of Pledge Collateral for miner
 	PARAM_FINISH()
 	panic("")
@@ -101,7 +102,7 @@ func (inds *Indices_I) StoragePower(
 
 func (inds *Indices_I) StoragePowerProportion(
 	minerStoragePower abi.StoragePower,
-) abi.StoragePower {
+) big.Int {
 	PARAM_FINISH()
 	panic("")
 }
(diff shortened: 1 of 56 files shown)


src/systems/filecoin_mining/sector/sealing.go
```diff
@@ -19,18 +19,7 @@ var SealSeedHash = SHA256
 // 	return SealSeed(h)
 // }
 
-func (x PieceInfo_I) Ref() *PieceInfo_I {
-	return &x
-}
-
-func (svi *OnChainSealVerifyInfo_I) IsValidAtSealEpoch() bool {
```
anorth: Replaced with the runtime syscall for validating a seal.

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
```diff
@@ -289,42 +286,43 @@ func (sms *StorageMiningSubsystem_I) VerifyElectionPoSt(inds indices.Indices, he
 
 	postCfg := filproofs.ElectionPoStCfg(sectorSize)
 
-	pvInfo := sector.PoStVerifyInfo_I{
-		OnChain_:    onChainInfo,
-		Randomness_: onChainInfo.Randomness(),
+	pvInfo := abi.PoStVerifyInfo{
```
zixuanzh: some fields are missing here, perhaps optional

src/systems/filecoin_mining/storage_proving/sealer/sealer.go
```diff
@@ -77,18 +78,18 @@ func (s *SectorSealer_I) CreateSealProof(si CreateSealProofInputs) *SectorSealer
 	sdr := filproofs.WinSDRParams(cfg)
 	proof := sdr.CreateSealProof(randomSeed, auxTmp)
 
-	onChain := sector.OnChainSealVerifyInfo_I{
-		SealedCID_: auxTmp.CommR(),
+	onChain := abi.OnChainSealVerifyInfo{
```
zixuanzh: also looks like a lot of fields missing here

src/systems/filecoin_mining/storage_proving/sealer/sealer.go
```diff
@@ -77,18 +78,18 @@ func (s *SectorSealer_I) CreateSealProof(si CreateSealProofInputs) *SectorSealer
...
+		Proof: proof,
 	}
 
 	return SectorSealer_CreateSealProof_FunRet_Make_so(
 		SectorSealer_CreateSealProof_FunRet_so(
 			&CreateSealProofOutputs_I{
-				SealInfo_: &sector.SealVerifyInfo_I{
-					SectorID_: sid,
-					OnChain_:  &onChain,
+				SealInfo_: abi.SealVerifyInfo{
```
zixuanzh: ditto

src/actors/runtime/indices/indices.go
```diff
@@ -79,13 +76,13 @@ func (inds *Indices_I) PledgeCollateralReq(minerNominalPower abi.StoragePower) a
 	panic("")
 }
 
-func (inds *Indices_I) SectorWeightProportion(minerActiveSectorWeight abi.SectorWeight) util.BigInt {
+func (inds *Indices_I) SectorWeightProportion(minerActiveSectorWeight abi.SectorWeight) abi.SectorWeight {
```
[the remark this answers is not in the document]
anorth: Ok. I'll use `big.Int`. But I don't understand how an integer can represent a proportion (but if it's the numerator against the total SectorWeight denominator, then SectorWeight would be an appropriate type).

src/systems/filecoin_mining/storage_mining/storage_mining_subsystem.go
anorth: I'm not trying to fix anything here. There are many things missing all over the spec.

src/systems/filecoin_mining/storage_proving/sealer/sealer.go
anorth: Ditto

src/systems/filecoin_mining/storage_proving/sealer/sealer.go
anorth: Ditto

PR #803, opened 2020-01-11
refactor: replace ipld.CID with go-cid

frrist:
Update calling code to use semantics of go-cid


src/libraries/filcrypto/filproofs/algorithms.go
```diff
@@ -879,9 +880,14 @@ func zeroPadding(size UInt) PieceInfo {
 
 func joinPieceInfos(left PieceInfo, right PieceInfo) PieceInfo {
 	util.Assert(left.Size() == right.Size())
+
+	// FIXME: make this whole function generic?
+	sectorPieceCID, err := cid.Cast(BinaryHash_SHA256Hash(cid.Cid(left.PieceCID()).Bytes(), cid.Cid(right.PieceCID()).Bytes()))
```
anorth: This is slightly more complicated, see [this discussion](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1578629688082700)

We don't need to fully implement it here, but should make a note that, e.g. `cid.Bytes()` isn't actually the payload data that we want input to the binary hash function. Leaving a comment should suffice. FYI @porcuquine in case you want to specify what we leave here.

src/libraries/libp2p/libp2p.id
```diff
@@ -1,8 +1,8 @@
-import ipld "github.com/filecoin-project/specs/libraries/ipld"
 import mf "github.com/filecoin-project/specs/libraries/multiformats"
+import cid "github.com/ipfs/go-cid"
 
 // PeerID is the CID of the public key of this peer
-type PeerID ipld.CID
+type PeerID cid.Cid
```
anorth: I expect this will conflict with #804 but be easy to resolve.

src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -193,7 +194,7 @@ func (rt *VMContext) _createActor(codeID abi.ActorCodeID, address addr.Address)
 	// Create empty actor state.
 	actorState := &actor.ActorState_I{
 		CodeID_:     codeID,
-		State_:      actor.ActorSubstateCID(ipld.EmptyCID()),
+		State_:      actor.ActorSubstateCID(cid.Undef),
```
anorth: I think we want this to be the CID of an empty state. The lotus code has something for this - could you replicate that pattern here?

PR #806, opened 2020-01-13
Minor actor revisions

jzimmerman:
Fill in some missing functionality and address review comments from #774.

### DIFF
--- a/src/actors/builtin/storage_power/storage_power_actor_code.go
+++ b/src/actors/builtin/storage_power/storage_power_actor_code.go
@@ -190,6 +190,22 @@ func (a *StoragePowerActorCode_I) OnMinerSurprisePoStFailure(rt Runtime, numCons
 	}
 }
 
+func (a *StoragePowerActorCode_I) OnMinerEnrollCronEvent(rt Runtime, eventEpoch abi.ChainEpoch, sectorNumbers []sector.SectorNumber) {
+	rt.ValidateImmediateCallerAcceptAnyOfType(builtin.StorageMinerActorCodeID)
+	minerAddr := rt.ImmediateCaller()
+	minerEvent := &autil.MinerEvent_I{
+		MinerAddr_: minerAddr,
+		Sectors_:   sectorNumbers,
+	}
+
+	h, st := a.State(rt)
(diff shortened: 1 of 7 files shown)


src/actors/builtin/storage_power/storage_power_actor_code.go
```diff
@@ -189,6 +189,22 @@ func (a *StoragePowerActorCode_I) OnMinerSurprisePoStFailure(rt Runtime, numCons
...
+	rt.ValidateImmediateCallerAcceptAnyOfType(builtin.StorageMinerActorCodeID)
+	minerAddr := rt.ImmediateCaller()
+	minerEvent := &autil.MinerEvent_I{
+		MinerAddr_: minerAddr,
+		Sectors_:   sectorNumbers,
+	}
+
+	h, st := a.State(rt)
+	if _, found := st.Impl().CachedDeferredCronEvents_[eventEpoch]; !found {
+		st.Impl().CachedDeferredCronEvents_[eventEpoch] = autil.MinerEventSetHAMT_Empty()
```
anorth: An array is probably sufficient for this per-miner list, no?

sternhenri:
Could/Should this include a removal of explicit sector sampling in the actor code (actors should pass all their sectors to proofs libs). It would be great to get this in ASAP for @anorth's sake.

src/actors/builtin/storage_power/storage_power_actor_code.go
jzimmerman: Presumably we at least need some persistent data structure for this, since it will go on-chain. Do we have a persistent array implementation at the moment? If not, HAMT seems like the best choice.

jzimmerman:
> Could/Should this include a removal of explicit sector sampling in the actor code (actors should pass all their sectors to proofs libs). It would be great to get this in ASAP for @anorth's sake.

Yes, this is also on my to-do list, but doesn't need to be done in this particular PR I think (though let me know if you disagree).

src/actors/builtin/storage_power/storage_power_actor_code.go
anorth: I think I had misread the context, this isn't per-miner so yes a persistent structure is appropriate. I think the AMT is more suited to lists but a minor issue we can probably resolve later.

PR #807, opened 2020-01-14
add vesting back to multisig

zixuanzh:


src/actors/builtin/multisig/multisig_actor.go
```diff
@@ -162,7 +162,11 @@ func (a *MultiSigActorCode_I) _rtApproveTransactionOrAbort(
 	UpdateRelease_MultiSig(rt, h, st)
 
 	if thresholdMet {
-		// A sufficient number of approvals have arrived: relay the message and delete from pending queue.
+		if !st._hasAvailable(rt.CurrentBalance(), txn.Value(), rt.CurrEpoch()) {
+			rt.AbortArgMsg("insufficient funds unlocked")
```
jzimmerman: minor: should be AbortFundsMsg?

zixuanzh: ah not sure actually, I thought AbortFundsMsg is when user sends funds but not having sufficient funds. This is the other way around where user requests funds as a function argument.

PR #809, opened 2020-01-14
Proofs algo config

porcuquine:


anorth:
It would be really helpful to me if this landed ASAP, especially:
-  the removal of the `Cfg` objects from the `VerifyInfo` structures
- removal of ProofInstance from SealProof.

I am moving and translating the VerifyInfo structures to Go code to form part of the executable actor spec.

porcuquine:
Ack. @dignifiedquire Can you give this a quick review and see if it agrees with what we discussed yesterday? I will make a pass to see what plumbing may have been lost in my battle to get the first draft to type-check/compile. After those two things, I think we can merge as-is, since it's strictly more accurate than what it replaces and will help @anorth.

anorth:
The StorageMinerActorState has a SealInstanceCfg inside the MinerInfo struct, ultimately containing the WindowCount. Can this too be removed from miner state and hidden behind the registry?

src/systems/filecoin_mining/sector/posting.id
```diff
@@ -55,15 +55,18 @@ type OnChainPoStVerifyInfo interface {
...
+    Candidates        [PoStCandidate]
+    Proofs            [PoStProof]
+    Randomness        PoStRandomness
 }
 
 type OnChainSurprisePoStVerifyInfo struct {
-    Candidates  [PoStCandidate]
-    Proofs      [PoStProof]
-    CommT       ChallengeTicketsCommitment
+    RegisteredProof
```
anorth: Should this be an array like for ElectionPost?

src/systems/filecoin_mining/sector/posting.id
```diff
@@ -55,15 +55,18 @@ type OnChainPoStVerifyInfo interface {
 }
 
 type OnChainElectionPoStVerifyInfo struct {
-    Candidates  [PoStCandidate]
-    Proofs      [PoStProof]
-    Randomness  PoStRandomness
+    // There should be one RegisteredProof for each PoSt Candi
+    RegisteredProofs  [RegisteredProof]
```
anorth: For follow-up, talk to @zixuanzh about whether we should store the RegisteredProof for each sector on-chain, meaning we could omit it here.

PR #815, opened 2020-01-14
move actor code from filecoin_vm to new actor directory for extraction

frrist:


src/actors/actor.go
```diff
@@ -0,0 +1,13 @@
+package actors
+
+import (
+	cid "github.com/ipfs/go-cid"
+)
+
```
frrist: note: I have moved the methodID's previously defined here to methods.go, any objects?

frrist:
3950f4a merges the contents of `actor.id` into `actor.go` and replaces the name `ActorState_I` with `ActorState`. Is this is more change than we want here I can do this in a follow on.

src/actors/actor.go
```diff
@@ -1,8 +1,10 @@
-// This contains actor things that are _outside_ of VM exection.
-// The VM uses this to execute abi.
+package actors
 
-import abi "github.com/filecoin-project/specs/actors/abi"
-import cid "github.com/ipfs/go-cid"
+import (
+	abi "github.com/filecoin-project/specs/actors/abi"
+	util "github.com/filecoin-project/specs/util"
```
anorth: Avoid depending on `util`.

src/actors/actor.go
```diff
@@ -18,34 +20,59 @@ import cid "github.com/ipfs/go-cid"
 //   Yes, a user may submit N separate messages with increasing
 //   sequence number, causing them to execute in order.
 //
-type CallSeqNum UVarint
+type CallSeqNum util.UVarint
```
anorth: Make this an `int64`

src/actors/actor.go
```diff
@@ -18,34 +20,59 @@ import cid "github.com/ipfs/go-cid"
...
 
 // Code is a serialized object that contains the code for an Actor.
 // Until we accept external user-provided contracts, this is the
 // serialized code for the actor in the Filecoin Specification.
-type Code Bytes
+type Code abi.Bytes
```
anorth: I think this is unused: delete it if so

src/actors/actor.go
```diff
@@ -18,34 +20,59 @@ import cid "github.com/ipfs/go-cid"
...
+	CodeID_ abi.ActorCodeID
+	// CID of the root of optional actor-specific sub-state.
+	State_ ActorSubstateCID
+	// Balance of tokens held by this actor.
+	Balance_ abi.TokenAmount
+	// Expected sequence number of the next message sent by this actor.
+	// Initially zero, incremented when an account actor originates a top-level message.
+	// Always zero for other abi.
+	CallSeqNum_ CallSeqNum
+}
```
anorth: We don't need to move this structure, the actors can't see it! ActorSystemStateCID likewise.

It's probably best to restore this to the location you moved it from. It can remain an .id file in that case, at your option.

src/actors/actor.go
```diff
@@ -18,34 +20,59 @@ import cid "github.com/ipfs/go-cid"
...
+func (a *ActorState) CallSeqNum() CallSeqNum {
+	return a.CallSeqNum_
 }
 
-type ActorSystemStateCID cid.Cid
 type ActorSubstateCID cid.Cid
 
+func (x ActorSubstateCID) Ref() *ActorSubstateCID {
+	return &x
+}
```
anorth: Just change the callers of `Ref` to use `&`

src/actors/actor.go
```diff
@@ -18,34 +20,59 @@ import cid "github.com/ipfs/go-cid"
...
+
```
anorth: Delete the comment about ActorState below here.

src/actors/actor.go
frrist: done.

src/actors/actor.go
```diff
@@ -0,0 +1,27 @@
...
+//
+// Q&A
+// - > Does it have to be sequential?
+//   No, a random nonce could work against replay attacks, but
+//   making it sequential makes it much easier to verify.
+// - > Can it be used to order events?
+//   Yes, a user may submit N separate messages with increasing
+//   sequence number, causing them to execute in order.
+//
+type CallSeqNum int64
```
anorth: We might not have to move this either. Check before merging, and put it back if unreferenced from actors/

PR #817, opened 2020-01-15
Add runtime syscalls for interface to proofs. Remove filproofs direct access from storage miner actor.

anorth:
The actor code can't depend on filproofs (i.e. rust-fil-proofs) directly, but must go via runtime.

FYI @jzimmerman

### DIFF
--- a/src/actors/builtin/storage_miner/storage_miner_actor.id
+++ b/src/actors/builtin/storage_miner/storage_miner_actor.id
@@ -99,7 +99,6 @@ type MinerInfo struct {
 
     // Amount of space in each sector committed to the network by this miner.
     SectorSize              sector.SectorSize
-    RegisteredProof         sector.RegisteredProof
     SealPartitions          UVarint
     ElectionPoStPartitions  UVarint
     SurprisePoStPartitions  UVarint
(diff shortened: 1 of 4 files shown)


src/actors/builtin/storage_miner/storage_miner_actor.id
```diff
@@ -99,7 +99,6 @@ type MinerInfo struct {
 
     // Amount of space in each sector committed to the network by this miner.
     SectorSize              sector.SectorSize
-    RegisteredProof         sector.RegisteredProof
     SealPartitions          UVarint
```
porcuquine: I think if this is removed, then `RegisteredProof` needs to be added to the `OnChainSealVerifyInfo` constructed `ProveCommitSector` or somehow otherwise communicated to proofs at verification time. I might be missing the logic of how you are doing this, though. Can you help me understand how/that this is happening?

anorth: It's already in the OnChainSealVerifyInfo and OnChainElection*VerifyInfo structures which, as I understand it, means the caller will provide it.

anorth: But I see I may have misread

anorth: Added it to SectorProveCommitInfo so the caller provides it per-sector.

PR #818, opened 2020-01-15
refactor: replace account_actor.id with go code

frrist:


src/actors/builtin/account/account_actor.go
```diff
@@ -1,19 +1,72 @@
...
 type InvocOutput = vmr.InvocOutput
 
+type AccountActorCode interface {
+	Impl() *AccountActorCode_I
+}
+
+type AccountActorState interface {
+	Address() addr.Address
+	Impl() *AccountActorState_I
+}
```
anorth: As for #820, make the state a vanilla struct, remove the Code interface, and add an empty struct for the receiver type AccountActor.

src/actors/builtin/account/account_actor.go
```diff
@@ -1,19 +1,72 @@
...
+
+type AccountActorCode_I struct {
+	cached_cid []byte
```
anorth: We don't need this functionality, drop it.

PR #819, opened 2020-01-15
Proofs algo config

porcuquine:


anorth:
I think this is now redundant

PR #820, opened 2020-01-15
refactor: replace cron_actor.id with go code

frrist:


src/actors/builtin/cron/cron_actor.go
```diff
@@ -1,11 +1,28 @@
...
 
+type CronActorState interface {
+	Impl() *CronActorState_I
+}
+
+type CronTableEntry interface {
+	ToAddr() addr.Address
+	MethodNum() abi.MethodNum
+	Impl() *CronTableEntry_I
+}
```
anorth: This is too faithful a translation of the generated code :-) Please simplify these down to vanilla structs with exported fields and no methods.

src/actors/builtin/cron/cron_actor.go
```diff
@@ -1,11 +1,28 @@
...
+
+type CronActorCode interface {
+	Entries() []CronTableEntry
+	EpochTick(r vmr.Runtime)
+	Impl() *CronActorCode_I
+}
```
anorth: Delete this interface - we'll put in an exports mechanism later (with methodids etc).

In cron_actor.go, create an empty `type CronActor struct{}` as the method receiver (just as go-filecoin and lotus both do now).

src/actors/builtin/cron/cron_actor.go
```diff
@@ -1,26 +1,38 @@
...
 	abi "github.com/filecoin-project/specs/actors/abi"
 	builtin "github.com/filecoin-project/specs/actors/builtin"
 	vmr "github.com/filecoin-project/specs/actors/runtime"
 )
 
-func (a *CronActorCode_I) Constructor(rt vmr.Runtime) vmr.InvocOutput {
+type CronActorState struct{}
+
+type CronActor struct {
+	Entries []CronTableEntry
```
anorth: Interesting. I think this is a bug in the original. 

Please leave a TODO to move `Entries` into the `CronActorState` struct.

PR #822, opened 2020-01-15
refactor: replace init_actor.id with go code

frrist:


src/actors/builtin/init/init_actor.go
```diff
@@ -87,15 +96,27 @@ func (s *InitActorState_I) ResolveAddress(address addr.Address) addr.Address {
...
+}
+
+func Deserialize_InitActorState(Bytes) (InitActorState, error) {
+	panic("TODO")
+}
+
+func Deserialize_InitActorState_Assert(Bytes) InitActorState {
+	panic("TODO")
+}
+
```
frrist: required by `vm_interpreter.go`

src/actors/builtin/init/init_actor.go
```diff
@@ -87,15 +96,27 @@ func (s *InitActorState_I) ResolveAddress(address addr.Address) addr.Address {
...
+	s.AddressMap[address] = actorID
 	idAddr, err := addr.NewIDAddress(uint64(actorID))
 	autil.Assert(err == nil)
 	return idAddr
 }
 
+func (st *InitActorState) CID() cid.Cid {
+	panic("TODO")
+}
+
```
frrist: required to satisfy `ipld.Object` interface

src/actors/builtin/init/init_actor.go
anorth: Replace with use of the actors/serde module.

PR #825, opened 2020-01-15
refactor: replace multisig_actor.id with go code

frrist:


src/actors/builtin/multisig/multisig_actor.go
```diff
@@ -18,20 +18,98 @@ var IMPL_FINISH = autil.IMPL_FINISH
 var IMPL_TODO = autil.IMPL_TODO
 var TODO = autil.TODO
 
-////////////////////////////////////////////////////////////////////////////////
-// Actor methods
-////////////////////////////////////////////////////////////////////////////////
+type TxnID int
```
anorth: Use int64

src/actors/builtin/multisig/multisig_actor.go
```diff
@@ -18,20 +18,98 @@ var IMPL_FINISH = autil.IMPL_FINISH
...
+	if currBalance-amountToSpend < st.AmountLocked(currEpoch-st.StartEpoch) {
+		return false
+	}
+
+	return true
+}
+
+func (st *MultiSigActorState) CID() cid.Cid {
+	panic("TODO")
+}
```
anorth: Rather than move it up here, le'ts split this to a _state.go file like the other actors.

PR #830, opened 2020-01-16
Remove unnecessary message object from StorageDeal type.

anorth:
Add a Signature data type for actors (but retain the filcrypto one). This data type doesn't need
signing or verification code because that always uses a runtime syscall.


src/algorithms/crypto/signatures.id
```diff
@@ -9,8 +9,8 @@ type SigKeyPair struct {
 }
 
 type Signature struct {
-    Sig   SignatureBytes  @(internal)
     Type  SigType         @(internal)
+    Sig   SignatureBytes  @(internal)
```
anorth: Order matching Lotus

PR #828, opened 2020-01-15
refactor: replace storage_market_actor.id with go code 

frrist:

PR #831, opened 2020-01-16
refactor: replace sma_actor.id with go code

frrist:


frrist:
@anorth I'd prefer to merge this after #825 #828 and #829 as it will have some conflicts, please review last whenever you get to it.

src/actors/builtin/storage_miner/storage_miner_actor_state.go
```diff
@@ -1,25 +1,138 @@
...
+// that are not yet returned or burned.
+type StorageMinerActorState struct {
+	Sectors    SectorsAMT
+	PoStState  MinerPoStState
+	ProvingSet SectorNumberSetHAMT
+	Info       MinerInfo
+}
+
+type MinerPoStState struct {
+	// If > 0 miner is in an OK state.
```
anorth: I see some redundancy here. Let's say initially -1, then the epoch of the last successful post. Let's not nuke it when they enter another state. This means that  whether it's in an ok state is whether SurpriseChallengeEpoch is <0 and NumConsecutiveFailures == 0

src/actors/builtin/storage_miner/storage_miner_actor_state.go
```diff
@@ -64,64 +177,75 @@ func (st *StorageMinerActorState_I) _getStorageWeightDescsForSectors(sectorNumbe
 }
 
 func MinerPoStState_New_OK(lastSuccessfulPoSt abi.ChainEpoch) MinerPoStState {
```
anorth: Delete these MinePoStState_New methods and lets inline the updates to the state where these are called. I'll help you work out the right values.

src/actors/builtin/storage_miner/storage_miner_actor.go
```diff
@@ -100,20 +121,26 @@ func (a *StorageMinerActorCode_I) OnDeleteMiner(rt Runtime) {
...
+	updateSuccessEpoch := st.PoStState.Is_OK()
 	Release(rt, h, st)
 
 	// Advance the timestamp of the most recent PoSt success, provided the miner is currently
 	// in normal state. (Cannot do this if SurprisePoSt mechanism already underway.)
 	if updateSuccessEpoch {
-		a._rtUpdatePoStState(rt, MinerPoStState_New_OK(rt.CurrEpoch()))
+		newPostSt := MinerPoStState{
+			LastSuccessfulPoSt:     rt.CurrEpoch(),
+			SurpriseChallengeEpoch: st.PoStState.SurpriseChallengeEpoch, // expected to be undef because PoSt is OK
```
anorth:
```suggestion
			SurpriseChallengeEpoch: st.PoStState.SurpriseChallengeEpoch, // expected to be undef because PoStState is OK
```

PR #833, opened 2020-01-16
It compiles.

porcuquine:


src/actors/abi/sector.go
```diff
@@ -109,10 +109,11 @@ type OnChainSurprisePoStVerifyInfo struct {
 	RegisteredProof
 	Candidates []PoStCandidate
 	Proofs     []PoStProof
-	CommT      ChallengeTicketsCommitment
+	// CommT      ChallengeTicketsCommitment // Potential future optimization.
 }
 
 type PoStCandidate struct {
+	RegisteredProof
```
anorth: Either this one or the one in PrivatePoStCandidateProof is redundant. I'll land it like this, but lmk which one we should remove.

src/systems/filecoin_mining/sector/posting.id
```diff
@@ -1,31 +1,5 @@
...
-    ElectionPoSt
-    SurprisePoSt
-}
-
-type PoStWitness struct {
-    Candidates [abi.PoStCandidate]
-}
+// type PoStWitness struct {
+//     Candidates [abi.PoStCandidate]
+// }
```
anorth: You can remove this file in a follow-up

PR #836, opened 2020-01-16
refactor: extract actor_util.id to go code

frrist:
see TODO in `actor_util.go` about interface as map key before merging


src/actors/util/actor_util.go
```diff
@@ -1,10 +1,95 @@
...
+	return m.MinerAddress
+}
+
+func (m MinerEvent) Sectors() []abi.SectorNumber {
+	return m.MinerSectors
+}
+
+// TODO using an interface in a map is probably a bad idea since different implementations may provide incomparable
+// type, but for now lets do the easy thing
+type MinerEventSetHAMT map[MinerEventInterface]bool
```
frrist: Will this suffice until we complete the extraction?

anorth: No, we can't have any interfaces in these on-chain data types. Remove the interface entirely.

frrist: For now put this in a slice and deal with the slowness.

PR #837, opened 2020-01-16
refactor: replaces vm_exitcode.id with go code

frrist:


src/actors/runtime/exitcode/vm_exitcodes.go
```diff
@@ -65,27 +65,41 @@ const (
 	InsufficientPledgeCollateral
 )
 
+type ExitCode struct {
+	Success          struct{}
+	SystemError      SystemErrorCode
+	UserDefinedError UserDefinedErrorCode
+}
```
anorth: Thanks for this faithful port, but it's too complex. Can you replace all of this with a single `type ErrorCode int64` and some helper methods? I think we can remove the concept of UserDefinedErrorCode.

src/actors/runtime/exitcode/vm_exitcodes.go
```diff
@@ -65,56 +64,30 @@ const (
...
-	return x.Success == struct{}{} && x.SystemError == 0 && x.UserDefinedError == 0
+func (x ExitCode) IsError() bool {
+	return !x.IsSuccess()
 }
 
-func (x *ExitCode) IsError() bool {
-	return !x.IsSuccess()
+func (x ExitCode) Equals(e ExitCode) bool {
+	return x == e
 }
```
anorth: This whole function is probably unnecessary now

src/actors/runtime/exitcode/vm_exitcodes.go
```diff
@@ -65,56 +64,30 @@ const (
...
-func (x *ExitCode) AllowsStateUpdate() bool {
+func (x ExitCode) AllowsStateUpdate() bool {
 	return x.IsSuccess()
 }
 
-func (x *ExitCode) Equals(ExitCode) bool {
-	panic("")
+func OK() ExitCode {
+	return ExitCode(0)
 }
```
anorth: Replace this with a `const OK = ExitCode(0)`.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
```diff
@@ -136,8 +136,8 @@ func (vmi *VMInterpreter_I) ApplyMessage(inTree st.StateTree, chain chain.Chain,
 		retReceipt = vmri.MessageReceipt_Make(invocOutput, exitCode, vmiGasUsed)
 	}
 
-	_applyError := func(tree st.StateTree, errExitCode exitcode.SystemErrorCode, senderResolveSpec SenderResolveSpec) {
-		_applyReturn(tree, vmr.InvocOutput_Make(nil), exitcode.SystemError(errExitCode), senderResolveSpec)
+	_applyError := func(tree st.StateTree, errExitCode exitcode.ExitCode, senderResolveSpec SenderResolveSpec) {
```
anorth: As a follow-up, let's move this to a package with a less redundant name.

src/actors/runtime/exitcode/vm_exitcodes.go
frrist: True, but it's slightly more convenient than `if exitCode == Exitcode(foo) {`. I'll remove it if you'd like.

src/systems/filecoin_vm/interpreter/vm_interpreter.go
frrist: added a TODO for this.

PR #838, opened 2020-01-16
refactor: replace indices.id with go code

frrist:


src/actors/runtime/indices/indices.go
```diff
@@ -1,20 +1,35 @@
...
-	PARAM_FINISH()
-	return providerCollateral
-}
-
-func (inds *Indices_I) StorageDeal_DurationBounds(
+// Data in Indices are populated at instantiation with data from the state tree
+// Indices itself has no state tree or access to the runtime
+// it is a passive data structure that allows for convenience access to network indices
+// and pure functions in implementing economic policies given states
+type Indices struct {
```
anorth: Please define an interface with the name `Indices` and make that the return value from the runtime. It should probably have all the methods currently defined for this struct.

 Then rename this struct to implement the interface.

frrist: done.

PR #846, opened 2020-01-17
Replace build_go.mod specs-actors require directive with a replace

anorth:
Also exclude the actor code from being copied into the build dir.

After this, the build is performed directly with the code in the submodule, rather than fetching from GitHub. There's only one place to update when we want to pull in new code: the submodule ref.


anorth:
Had to update submodules in CI. Now done.

PR #847, opened 2020-01-20
Add Hello Protocol

ZenGround0:
This work is coming off of the effort to interop filecoin nodes.  Speccing the hello protocol is necessary for us to have an interoperable network.

Proposed diffs from current lotus hello Lotusers should be aware of:
1. added Height field to hello message (gfc wants to use this field).  
2. separated out Latency fields from the HelloMessage into their own message as they are only ever sent separately.

### DIFF
--- a/src/systems/filecoin_nodes/network/_index.md
+++ b/src/systems/filecoin_nodes/network/_index.md
@@ -31,6 +31,24 @@ Here is the list of libp2p protocols used by Filecoin.
 - DNSDiscovery: Design and spec needed before implementing
 - HTTPDiscovery: Design and spec needed before implementing
 - Hello:
-	- Hello protocol handles new connections to filecoin nodes.  It is an important part of the discovery process for ambient protocols (like KademliaDHT)
-	- Spec TODO.
-	- No Filecoin specific modifications to the protocol id.
+	- Hello protocol handles new connections to filecoin nodes to facilitate discovery
+	- the protocol string is `fil/hello/1.0.0`. 
+
+# Hello Spec 
+
+## Protocol Flow
+
+`fil/hello` is a filecoin specific protocol built on the libp2p stack.  It consists of two conceptual
+procedures: `hello_connect` and `hello_listen`.   
+
+`hello_listen`: `on new stream` -> `read peer hello msg from stream` -> `write latency message to stream` -> `close stream`
+
+`hello_connect`: `on connected` -> `open stream` -> `write own hello msg to stream` -> `read peer latency msg from stream`  -> `close stream`
+
+where stream and connection operations are all standard libp2p operations.  Nodes running the Hello Protocol should consume the incoming Hello Message and use it to help manage peers and sync the chain.
+
+## Messages
+{{< readfile file="hello.id" code="true" lang="go" >}}
+
+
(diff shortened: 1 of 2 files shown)


src/systems/filecoin_nodes/network/_index.md
```diff
@@ -31,6 +31,24 @@ Here is the list of libp2p protocols used by Filecoin.
 - DNSDiscovery: Design and spec needed before implementing
 - HTTPDiscovery: Design and spec needed before implementing
 - Hello:
-	- Hello protocol handles new connections to filecoin nodes.  It is an important part of the discovery process for ambient protocols (like KademliaDHT)
-	- Spec TODO.
-	- No Filecoin specific modifications to the protocol id.
+	- Hello protocol handles new connections to filecoin nodes to facilitate discovery
+	- the protocol string is `fil/hello/1.0.0`.
```
anorth: I think this protocol string should include the network name, as the pubsub topics do.

src/systems/filecoin_nodes/network/hello.id
```diff
@@ -0,0 +1,15 @@
...
+    HeaviestTipSet        [cid.Cid]
+    HeaviestTipSetWeight  BigInt
+    HeaviestTipSetHeight  Int
+    GenesisHash           cid.Cid
+}
+
+// LatencyMessage shares information about a peer's network latency
+type LatencyMessage struct {
+    TArrival  Int
+    TSent     Int
```
anorth: Document or name fields describing the units of these timestamps.

src/systems/filecoin_nodes/network/_index.md
ZenGround0: @whyrusleeping @anorth you two appear to disagree.  I can't merge this until I hear both sides of the story, please express your viewpoints in the comments below.  Thanks!

whyrusleeping: So, if we don't have the hello protocol on a common protocol name, nodes on different networks won't discover they are on different networks. Lotus triggers a disconnect if it connects to a node on a different genesis block

ZenGround0: This is a good point.  Rephrasing: as implemented by lotus, a node will leave open connections that do not respond to hellos but will close connections that respond to hellos with different genesis cids (and different networks always have different genesis cids).  If we separate out different networks at the protocol negotiation layer lotus would actually keep open the connections to nodes that are on a different network because they'd never even run hello.

anorth: Ok, so they _could_ discover it by inspecting the protocols, or just learning that a peer doesn't implement the expected protocol. I gather Lotus desires to keep open connections that don't implement hello at all?

It seems like all behaviours could be implemented in either case, so I don't mind. No network name is ok.

ZenGround0: > I gather Lotus desires to keep open connections that don't implement hello at all?

Yeah I think that is really the crux of this design decision if you pull on it long enough.  Its definitely something @whyrusleeping believes in strongly for interop across many different protocols.  Glad you are calling it out explicitly because it is not written down anywhere.

PR #854, opened 2020-01-23
add payment channel actor descr, links and high level interface

sternhenri:
This PR should be sufficient to implement payment channels as done in Lotus. To be merged alongside https://github.com/filecoin-project/specs-actors/pull/43


sternhenri:
Failure is due to how I update the src/actor subproj I think. If immediately clear to you @anorth would love help fixing, otherwise I'll dig in :)

src/systems/filecoin_token/payment_channels/payment_channel_actor.id
```diff
@@ -1,8 +1,48 @@
...
+type ModVerifyParams struct {
+    Actor   addr.Address
+    Method  uint64
+    Data    [byte]
+}
+
+type UpdateChannelStateParams struct {
+    Sv      SignedVoucher
+    Secret  [byte]
+    Proof   [byte]
```
anorth: Don't add these here, delete this file and express it just once in specs-actors.

anorth:
@sternhenri it is extremely likely that updating the actors submodule causes a bunch of build breakages due to things moving or being renamed. The submodule update should come in a PR of it's own that does only that.

I suggest you merge the prose here (after removing the code), and then follow-up to attempt the submodule update. It's not going to be pretty. I've been putting it off in the name of actual progress.

PR #864, opened 2020-01-28
fix layout thrashing in `fillCodeRefs`

### LINKED ISSUE #860: [web] `fillCodeRefs` slow on startup
*This might be an issue of my particular setup, if nobody else is experiencing something similar feel free to close.*

When loading the website https://filecoin-project.github.io/specs/ with Chrome (`Version 79.0.3945.88 (Official Build) (64-bit)`) in Ubuntu 18 it stalls for something like 20-30 seconds with very little responsiveness. The Chrome profiler pointed to

https://github.com/filecoin-project/specs/blob/7e4fde83c7699f8ec33a4c8735768e4bc0bbc6fd/hugo/layouts/shortcodes/js.html#L80-L84

(indicating that "Forced reflow is likely performance bottleneck").

This makes the basic workflow of opening the spec website to do a quick search of a term very slow (even for content that is not loaded through `fillSingleCodeRef`).

terorie:
Fixes #860 

Replaces HTMLElement.innerText queries with Node.textContent
because the former requires the layout to be rendered.
The layout is invalidated often though, leading to "forced synchronous layouts".

 - Google's performance advice https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing
 - `GetLayoutObject` call in WebKit's `Element::innerText()` https://chromium.googlesource.com/chromium/src/+/4380beb4e7b1f9339e86c5ad1e41b5ae87ba52c8/third_party/WebKit/Source/core/dom/Element.cpp#3342


hugo/layouts/shortcodes/js.html
```diff
@@ -54,7 +54,7 @@
   };
 
   fillCodeRefs = (xs, parent) => {
-    if (xs.length == 0 && parent.innerText.length != 0) {
+    if (xs.length == 0 && parent.textContent.length != 0) {
```
jimmylee: Nit: This will throw if `xs` or `parent` is `undefined` and does not have the length `property`, since I don't have the full context of the usage here I'm going to assume its okay.

terorie: I'm not familiar with the code either, but the only usage is this a few lines below
```js
let codeBlocks = document.getElementsByTagName("code");
for (let codeBlock of codeBlocks) {
    fillCodeRefs(codeBlock.children, codeBlock);
}
```
Seems unlikely that `parent` is undefined and all DOM Nodes have `textContent`

mishmosh:
This PR was closed & immediately reopened to restart CI (thanks for the troubleshooting help @travisperson).

Thanks again @terorie. Will be deployed tomorrow.

PR #872, opened 2020-01-31
clarify secp256k1 pubkey format

jleni:
Make explicit that uncompressed pubkeys must be used to generate addresses. 
Incorrectly using a compressed pubkey (33 bytes) can result in loss of funds as a different address will be generated and the corresponding pub/sec keys will not be accessible.


jleni:
any updates on this?

whyrusleeping:
@Kubuxu can you confirm that this is correct?

jleni:
yes, my suggestion is to make this explicit. 

If anyone in the community does this incorrectly (because is unclear), for a given private key if funds sent to `pubkey (compressed) => address`, they will be lost.

The other option is to add a workaround in the verification code that checks a match between the source address and the recovered pubkey in both compressed and uncompressed forms. 

Still, for now I would start by clarifying the current expectation in the specs.

Kubuxu:
Can you rebase the PR, thanks.

PR #876, opened 2020-02-09
ChainEpoch and CallSeqNum are int64

ZenGround0:
After a quick sync with lotus team I'm bringing up this change.  It is already what lotus has and is already way more bits than we need to represent block height without going full varint.

Because lotus and spec disagree here and it is an inescapable part of genesis block byte format this is a current interop blocker.  I'm moving forward with serializing block heights as cbor major type 0 as lotus does (instead of 2 with big int semantics on the byte string as spec wants) and submitting this change to the spec to align everyone.  This is time sensitive so please bring all dissenting opinions up ASAP.  

--edit-- 
we are in the same situation with CallSeqNum so I've added it here too

### DIFF
--- a/src/systems/filecoin_nodes/clock/clock_subsystem.id
+++ b/src/systems/filecoin_nodes/clock/clock_subsystem.id
@@ -7,7 +7,7 @@ type UTCClock struct {
 }
 
 // ChainEpoch represents a round of a blockchain protocol.
-type ChainEpoch UVarint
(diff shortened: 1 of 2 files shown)


mishmosh:
FYI there's still a commit message here referencing `ChainEpoch type is uint64`

anorth:
Both of these should now be imported from specs-actors, but don't take that on now.

Issue #868, opened 2020-01-29
Randomness bootstrap at genesis


sternhenri:
Per convo with Why, Dig, Porcu

Network bootstrap section, will need to know ticket values ahead of time and so on. —> predetermined first ticket, for preseal…
—> genesis ticket unknown, special case first preseal ticket, for the rest getRandomness for neg numbers hashes the ticket.
eg. getR(-499) = H(1)
      getR(-498) = H(2)
getR(negEpoch) = H(negEpoch + MaxSealTime) for that.
H(currentEpoch) for random epochs...

Issue #866, opened 2020-01-29
Update RT randomness 


sternhenri:
Clean up Randomness sampling to be simpler per discussion with @ZenGround0 
- just take in CBOR of needed entropy.
- use Digest of VRF output as input

PR #883, opened 2020-02-13
Bootstrapping, Randomness, weight md, etc

sternhenri:
In this PR
- start bootstrapping randomness note (per https://github.com/filecoin-project/specs/issues/868)
- other randomness changes (per https://github.com/filecoin-project/specs/issues/866 and with https://github.com/filecoin-project/specs-actors/pull/161)
- weight function update (replacing blocks by tickets in weight)

Code is NOT up to date. Current PR is only md. Code will be fixed as we realign with `specs-actors`


src/algorithms/crypto/randomness.md
```diff
@@ -10,54 +10,49 @@ We describe this formatting below.
...
 
 buffer = Bytes{}
 buffer.append(IntToBigEndianBytes(AppropriateDST))
-buffer.append(-1) // a flag to be used in cases where FIL might need longer randomness outputs. Currently unused
 buffer.append(ticket_randomness)
-buffer.append(other needed serialized inputs)
+buffer.append(s)
 
-randomness = SHA256(buffer)
+randomness = H(buffer)
```
ZenGround0: It is appropriate to specify that H is blake2b somewhere nearby.

sternhenri:
cc @anorth @icorderi

src/algorithms/crypto/randomness.md
```diff
@@ -10,54 +10,49 @@ We describe this formatting below.
...
 ```
 
 {{< readfile file="/docs/actors/actors/crypto/randomness.go" code="true" lang="go" >}}
 {{< readfile file="/docs/systems/filecoin_blockchain/struct/chain/chain.go" code="true" lang="go" >}}
+
+## Drawing randomness prior to genesis
```
ZenGround0: This section will be immensely clearer and more likely to be correct if you specify this special case in the already defined framework built up above.  You should specify that pre-genesis special cases all derive entropy bytes from a certain form.  The below section doesn't give enough detail for the implementer to understand how all types of pre-genesis drawing randomness work.  For example what if you want to draw with an epoch and miner address?  Do you cbor serialize curr - lookback in the epoch field of the entropy struct?  Do you serialize the epoch bytes of curr - lookback separately and fill the entropy struct with the epoch passed in?  By defining serialization separately here you also erase mention of domain separation tags which you probably don't want to do.

src/algorithms/crypto/randomness.md
ZenGround0: Overall this section is the right level of detail.

src/algorithms/crypto/randomness.md
```diff
@@ -10,54 +10,49 @@ We describe this formatting below.
...
+- `WindowedPoStChallengeSeed` -- uses ticket and epoch number
 ...
 
 In all cases, a ticket is used as the base of randomness (see {{<sref tickets>}}). In order to make randomness seed creation uniform, the protocol derives all such seeds in the same way, as follows (also see {{<sref tickets>}}):
+
+For a given randomness lookback `l`, and serialized entropy `s`:
+
 ```text
- For a given ticket's randomness ticket_randomness:
+ticket = draw_ticket_from_chain(l)
```
ZenGround0: where is this defined?  I think it means return the ticket `l` ancestors in the past but its not trivial what this does.

anorth:
Thanks, this makes sense to me now.

src/systems/filecoin_blockchain/struct/chain/chain.go
```diff
@@ -18,8 +18,17 @@ func (chain *Chain_I) TipsetAtEpoch(epoch abi.ChainEpoch) Tipset {
...
-	return ts.MinTicket().DrawRandomness(epoch)
+	if epoch < genesis {
+		genesisTS := chain.TipsetAtEpoch(genesis)
+		genesisTix := genesisTS.MinTicket().DrawRandomness(genesis)
+		buffer := []byte{}
+		buffer = append(buffer, genesisTix...)
+		buffer = append(buffer, BigEndianBytesFromInt(int64(epoch))...)
+		return blake2b.Sum256(buffer)
+	} else {
+		ts := chain.TipsetAtEpoch(epoch)
```
anorth: Please leave a big comment here describing that the epoch is intentionally not mixed in because some uses of randomness don't need it. It is at the caller's peril to mix in the chain epoch in order to reliably get different randomness for each epoch (given null blocks). But note that in cases of pre-genesis sampling this will be redundant. 

I think this should either always or never mix in the epoch number.

src/systems/filecoin_blockchain/struct/chain/chain.go
```diff
@@ -18,8 +18,17 @@ func (chain *Chain_I) TipsetAtEpoch(epoch abi.ChainEpoch) Tipset {
 
 // Draws randomness from the tipset at or immediately prior to `epoch`.
 func (chain *Chain_I) RandomnessAtEpoch(epoch abi.ChainEpoch) abi.RandomnessSeed {
-	ts := chain.TipsetAtEpoch(epoch)
-	return ts.MinTicket().DrawRandomness(epoch)
+	if epoch < genesis {
+		genesisTS := chain.TipsetAtEpoch(genesis)
+		genesisTix := genesisTS.MinTicket().DrawRandomness(genesis)
```
anorth: What's this Ticket.DrawRandomness? It looks like that also mixes in an epoch number, but in this call, the fixed epoch 0.

IMO it would be clearer to consider the ticket as strictly a byte array, and do any epoch mixing in this code, not the ticket code.

src/systems/filecoin_blockchain/struct/chain/chain.go
```diff
@@ -18,8 +18,17 @@ func (chain *Chain_I) TipsetAtEpoch(epoch abi.ChainEpoch) Tipset {
 
 // Draws randomness from the tipset at or immediately prior to `epoch`.
 func (chain *Chain_I) RandomnessAtEpoch(epoch abi.ChainEpoch) abi.RandomnessSeed {
-	ts := chain.TipsetAtEpoch(epoch)
-	return ts.MinTicket().DrawRandomness(epoch)
+	if epoch < genesis {
+		genesisTS := chain.TipsetAtEpoch(genesis)
```
anorth: When processing any messages in the genesis tipset (block) itself, there is no prior tipset to find here.

src/algorithms/crypto/randomness.go
```diff
@@ -0,0 +1,87 @@
...
+	autil "github.com/filecoin-project/specs-actors/actors/util"
+)
+
+// Specifies a domain for randomness generation.
+type DomainSeparationTag int
+
+const (
+	DomainSeparationTag_TicketProduction DomainSeparationTag = 1 + iota
+	DomainSeparationTag_ElectionPoStChallengeSeed
+	DomainSeparationTag_SurprisePoStChallengeSeed
```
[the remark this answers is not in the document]
sternhenri: it is. sorry about that! we'll fix that up when we match up specs-actors with spec

src/algorithms/crypto/randomness.go
```diff
@@ -0,0 +1,87 @@
...
+func DeriveRandWithMinerAddr(tag DomainSeparationTag, randSeed []byte, minerAddr addr.Address) abi.Randomness {
+	var addrBuf bytes.Buffer
+	err := minerAddr.MarshalCBOR(&addrBuf)
+	autil.AssertNoError(err)
+
+	return _deriveRandInternal(tag, randSeed, addrBuf.Bytes())
+}
+
+type AddressEpochEntropy struct {
+	MinerAddress addr.Address // Must be an ID-addr
```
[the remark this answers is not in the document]
sternhenri: No, this isn't the point imo. Your point is well taken with regards to the randomness, but what you're describing is a hash collision. We want to use a CBOR serialization in order to have simple interfaces. We can assume we are using CRHFs in which case the DSTs will suffice for what you say.

src/algorithms/crypto/randomness.md
sternhenri: Lmk if new version looks good. Thanks for this feedback. v useful!

src/systems/filecoin_blockchain/struct/chain/chain.go
sternhenri: Changed to always per our convo. Thank you

src/systems/filecoin_blockchain/struct/chain/chain.go
sternhenri: yes. fixed accordingly. Thank you!

src/systems/filecoin_blockchain/struct/chain/chain.go
sternhenri: By this I mean tipset made of the genesis block. Will comment accordingly

src/algorithms/crypto/randomness.md
```diff
@@ -10,54 +10,89 @@ We describe this formatting below.
...
 
 {{< readfile file="/docs/actors/actors/crypto/randomness.go" code="true" lang="go" >}}
 {{< readfile file="/docs/systems/filecoin_blockchain/struct/chain/chain.go" code="true" lang="go" >}}
+
+## Entropy to be used with randomness
+
+We currently distinguish the following entropy needs per use:
+
+- TicketProduction: requires MinerIDAddress
+- ElectionPoStChallengeSeed: requires current epoch and MinerIDAddress
```
anorth: This line misled me into thinking the epoch needed to be explicitly mixed in again, even though it's already mixed in  as described at line 49

anorth: Including ticket production and epost challenge seed here _at all_ misled me into thinking they could be calculated the same way. I suggest removing them entirely and then describing how they can be computed, given ticket-digest-plus-epoch described above (or, more importantly, if they diverge even lower down).

sternhenri: No but some of this is stale. Thank you!

Let me clean up.

PR #886, opened 2020-02-14
Link hamt spec

ZenGround0:
Figured we should link this as I couldn't find the link in the spec yet.  It's a draft but it appears to match go-hamt-ipld at a glance.  If they don't match either go-hamt-ipld or this spec should change so that they do.


vmx:
Please note that there is still small differences between that spec and the lotus implementation. e.g. the spec is more flexible (less hard-coded things), but there's also more to it. I think @rvagg knows more :)

PR #890, opened 2020-03-07
feat: change DHT protocol format

Stebalien:
Change the DHT protocol format from

    /fil/kad/<network-name>

to

    /fil/<network-name>/kad/1.0.0

The underlying format is `/CUSTOM_PREFIX/CANONICAL_DHT_PROTOCOL`.

This makes it possible for DHT implementations to add additional protocol versions without needing `{custom protocol -> canonical protocol}` map.

Context: The content routing team is currently adding a new DHT protocol and plans to add more in the future. Supporting a custom protocol map where the user specifies individual protocols for every DHT protocol variant is _possible_ but messy. We'd like to avoid that situation if at all possible.

Note: The current lotus testnet uses /lotus/kad/1.0.0 so upgrading the current live network should be seamless.


mishmosh:
Thanks for this @Stebalien! Tagging @ZenGround0 for review.

PR #893, opened 2020-03-16
move message field ordering back to what it was before

whyrusleeping:
Not sure when these got reordered, but we've sunk a significant amount of external work into the ordering that this PR restores.

### DIFF
--- a/src/systems/filecoin_vm/message/message.id
+++ b/src/systems/filecoin_vm/message/message.id
@@ -26,14 +26,14 @@ type UnsignedMessage struct {
     // Amount of value to transfer from sender's to receiver's balance.
     Value       abi.TokenAmount
 
+    // GasPrice is a Gas-to-FIL cost
+    GasPrice    abi.TokenAmount
+    GasLimit    GasAmount
+
     // Optional method to invoke on receiver, zero for a plain value send.
     Method      abi.MethodNum
     /// Serialized parameters to the method (if method is non-zero).
     Params      abi.MethodParams
-
-    // GasPrice is a Gas-to-FIL cost
-    GasPrice    abi.TokenAmount
-    GasLimit    GasAmount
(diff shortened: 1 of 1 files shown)


src/systems/filecoin_vm/message/message.id
```diff
@@ -26,14 +26,14 @@ type UnsignedMessage struct {
     // Amount of value to transfer from sender's to receiver's balance.
     Value       abi.TokenAmount
 
+    // GasPrice is a Gas-to-FIL cost
+    GasPrice    abi.TokenAmount
+    GasLimit    GasAmount
```
ZenGround0: No action required but FYI @anorth flagging that the definition of `GasAmount` found here wrapping a BigInt instead of an int64 implies we can't use `abi.GasUnits` here and stay up to spec.  Lotus uses a bigint based type and so this field is serialized as the cbor byte array major type rather than the signed int major type (which is how `abi.GasUnits` serializes).  For go-filecoin to use `abi.GasUnits` type instead of BigInt that type would need to switch to wrapping a big int.

I'm not sure if this is going to require annoying conversions between int64s and bigints but probably isn't too bad since lotus is doing it right now.

ZenGround0: @anorth just realized that's a gfc internal type so we're all good

anorth:
It didn't change. It's been this way [since October 2019](https://github.com/filecoin-project/specs/blob/04d5e1339b3bcad078e010871ddf6773576e40ce/src/systems/filecoin_vm/message/message.id) but was ignored, and now we've too many external dependencies locking it in.

whyrusleeping:
@anorth I consider a change in October 2019 a change, especially given that we copied our structs right out of the spec when lotus started, and the actual fields in the message struct havent changed, just their order.

Issue #894, opened 2020-03-17
Proposed changes to NewActorAddress


arajasek:
There are three proposed changes to how actor addresses are generated in the VM. They are:

1) Use the origin address, not the `_immediateCaller`. This seems necessary to avoid address collisions (for example, in power_actor's CreateMiner method).
2) Use the pubkey address, not the ID address, which is what the spec currently says we should do.
3) Use the top level message's nonce, and not the top level sender's actor's nonce (since that has already been incremented by this point).

I think we already have broad consensus on these proposed changes.

cc @frrist

whyrusleeping:
This SGTM, Thanks for writing this up @arajasek !

anorth:
I have added a fourth change to the issue description:

Replace internalCallSeqNum counting all internal message sends with a newActorAddressCount counter of calls to the method specifically, in the context of the originating message

PR #895, opened 2020-03-17
Re: #894: Implement the proposed changes to NewActorAddress

arajasek:
- Use the top level sender's pubkey address, not the immediate caller's ID address
- Use the top level message's CallSeqNum, not the top level sender's CallSeqNum

This ensures that address generation is deterministic and resistant to reorgs. The CallSeqNum change is just to make it a little more intuitive.

### DIFF
--- a/src/systems/filecoin_vm/runtime/impl/runtime.go
+++ b/src/systems/filecoin_vm/runtime/impl/runtime.go
@@ -119,8 +119,8 @@ type VMContext struct {
 	// Not necessarily the immediate caller.
 	_toplevelSender      addr.Address
 	_toplevelBlockWinner addr.Address
-	// Top-level call sequence number of the "From" actor in the initial on-chain message.
-	_toplevelSenderCallSeqNum actstate.CallSeqNum
+	// call sequence number of the top level message that began this execution sequence
+	_toplevelMsgCallSeqNum actstate.CallSeqNum
 	// Sequence number representing the total number of calls (to any actor, any method)
 	// during the current top-level message execution.
 	// Note: resets with every top-level message, and therefore not necessarily monotonic.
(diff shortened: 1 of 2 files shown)


src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -660,9 +662,10 @@ func (rt *VMContext) GetRandomness(epoch abi.ChainEpoch) abi.RandomnessSeed {
 func (rt *VMContext) NewActorAddress() addr.Address {
 	addrBuf := new(bytes.Buffer)
 
-	err := rt._immediateCaller.MarshalCBOR(addrBuf)
+	// note that this is a pubkey address
+	err := rt._toplevelMessage.From.MarshalCBOR(addrBuf)
```
anorth: No, there's no guarantee at all that `From` is a pubkey address. In fact, senders pay less gas if it's an ID address.

This needs to be resolved through the sender account actor state. That could be done inside this object, or before the ctx is constructed (pass it as a new ctor param).

src/systems/filecoin_vm/runtime/impl/runtime.go
```diff
@@ -121,6 +121,8 @@ type VMContext struct {
 	_toplevelBlockWinner addr.Address
 	// Top-level call sequence number of the "From" actor in the initial on-chain message.
 	_toplevelSenderCallSeqNum actstate.CallSeqNum
```
anorth: Can you remove this? Or perhaps rename and fix caller to use the message call seq num, instead of passing the whole message here.

arajasek:
All suggested changes implemented :)

anorth:
LGTM after fixing compilation

Issue #904, opened 2020-03-31
Change timestamp validation or remove timestamp from block header


anorth:
The spec constrains the timestamp to the range implied by the epoch. Lotus rejects any value except exactly the epoch start. Go-filecoin lied about the value anyway.

Either:
- change the spec to require the timestamp to be exactly the epoch start
- remove timestamp from the block header

cc @whyrusleeping @ZenGround0

sternhenri:
We have discussed the redundancy a few times in the past. The decision was to keep timestamps in spite of it for a few reasons: 
- convenience for products built atop filecoin, eg. block explorers, viz, etc that needn't do conversions themselves that the chain is doing.
- keeping the same header format even if the protocol changes making timestamps and epoch non-redundant.

I was not aware that lotus rejected all values but the epoch start, I knew it set it to epoch start but assumed that was simply a convenience, not a consensus rule. Happy to make the change but I don't understand why this should be a consensus rule. Namely, it is important that miners not accept blocks from the future, but I don't see an issue with "fast" miners releasing their blocks early (if appropriately timestamped -- making timestamp and epoch non-redundant). What is the issue there?

PR #908, opened 2020-04-07
correct mining cycle graph

sternhenri:
Fixing mining cycle to place cutoff after epoch boundary, and adding a slightly more succinct algorithm.


anorth:
I don't think the diagram matches the words. The orange Rx seems wrong since a node won't accept a block that arrives before the epoch starts. the orange should match the green exactly (and so is redundant).

sternhenri:
You're right. I'll fix it. I'm not sure why we force quantization of the timestamps? Ie why is a block with a non-quantized (but otherwise correct) timestamp invalid?

anorth:
Timestamps are redundant, a pure function of epoch. IMO we should remove
them. The spec is out of date here.

https://github.com/filecoin-project/specs/issues/904

sternhenri:
I'll address in your comment.

whyrusleeping:
(my approval was on commit 9f033cc)

PR #915, opened 2020-04-27
Add version field to messages, closes #913

### LINKED ISSUE #913: transaction/message versioning
At the moment, CBOR txs lack any kind of versioning support. Any changes to fields, or types may result in applications to incorrectly parse messages or transactions.
A very explicit example are airgapped devices or hardware wallets that are not frequently updated.

While unlikely, if the serialization format changes, it is possible that older implementations could still parse the data and show incorrect or unintended values.

I would recommend adding a few versioning bytes at the beginning of each tx / message.

whyrusleeping:


src/systems/filecoin_vm/message/message.id
```diff
@@ -16,6 +16,9 @@ type GasAmount struct {
 }
 
 type UnsignedMessage struct {
+    // Version of this message (0 until we have to have a breaking change)
+    Version int64
```
droark, line 20: Just checking if _int64_ is intentional vis-à-vis _uint64_. Granted, I don't anticipate Filecoin hitting `INT64_MAX`. :)

anorth:
I think message syntax validation should be changed to require the version to be zero, else someone could write any version in.

mishmosh:
@whyrusleeping This has 2 review comments to look at, but otherwise 👍.

Schwartz10:
Hey @whyrusleeping @mishmosh do either of you have a sense of a target completion date for this update to the spec and its implementation in Lotus?

whyrusleeping:
@Schwartz10 This has been implemented in lotus and go-filecoin

PR #939, opened 2020-05-16
Testnet Phase 2 Rework: Drand && Consensus

sternhenri:
**In This PR:**

- Updated Leader Election Section to fit chosen construction (DRand + WPoSt * 2)
- Updated ticket creation protocol to follow what Lotus does
- Updated SPC Section to remove ticket chain mentions and convert role of ticket chain to tie-breaks in FCR
- Added a Drand section
- Updated block header to purge electionPoSt and add beaconEntries

**Future Work:**
- Fill in details about the Drand network and grpc endpoints @nikkolasg 
- Rework all stale .id/.go files (in the interest of time I went for go-like pseudocode in the md)
- Pending WinningPoSt rework (@mishmosh)
  - remove ElectionPoSt section
  - rework blockHeader to include WinningPoSt
  - check that the randomness.md/go files still work with updated PoSt/PoRep 
- Pending all Proofs rework, remove all mentions of ticket chain or on-chain tickets as sources of randomness.
- You might consider (@mishmosh)
  - moving tickets section from spc to ec for clarity. I didn't do it since not sure whether they are used in PoRep or PoSt. If not, then they are only used in EC FCR tie-breaks and can be moved

**Asks for reviewers**:
- @ZenGround0 
  - basic protocol work and drand integration to make sure it fits current state of network, especially watching any edge cases I might have missed related to pre-genesis randomness
  - how does forksignaling work (in blockheader)
  - block validation spot check plz
- @zixuanzh 
  - In the EC section, making sure I am using the power appropriately
  - Help figuring out what the specs-actor spc endpoints for miner power and total network power are (I couldn't find them)
- @nicola 
  - Ensuring this reflects your design for LE
  - Deprecating the final construction doc if so (pending completion of wPoSt*2 spec work)
- @nikkolasg 
  - the drand sections

A couple questions for @whyrusleeping:
- Why does lotus even store the `Round` in their `BeaconEntry` I believe it suffices to just store the Signature given the time-to-round mapping. Seems like a waste of space on-chain.
- Is the `BeaconEntry` `Data` field the drand `Signature`? I assumed so but can't tell given the mockBeacon.


src/libraries/drand/_index.md
```diff
@@ -0,0 +1,120 @@
...
+  "previous_signature": "afc545efb57f591dbdf833c339b3369f569566a93e49578db46b6586299422483b7a2d595814046e2847494b401650a0050981e716e531b6f4b620909c2bf1476fd82cf788a110becbc77e55746a7cccd47fb171e8ae2eea2a22fcc6a512486d",
+  "randomness": "d7aed3686bf2be657e6d38c20999831308ee6244b68c8825676db580e7e3bec6"
+}
+```
+
+Specifically, we have:
+
+- `Signature`           -- the threshold BLS signature on the previous signature value `Previous` and the current round number `round`.
+- `PreviousSignature`   -- the threshold BLS signature from the previous drand round.
+- `Randomness`          --  the SHA256 hash of `Signature`, to be used as the random value for this round.
```
nikkolasg, line 34: It's a bit confusing for the user to know if it can use the randomness as is or not, that's stg on drand side. 
Nevertheless, here you mention sha256 but before you mention sha512. Pick one!

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,120 @@
...
+- `Round`               -- the index of Randomness in the sequence of all random values produced by this drand network.
+
+Specifically, the message signed is the concatenation of the round number treated as a uint64 and the previous signature. At the moment, drand uses BLS signatures on the BN256 curves and the signature is made over G1.
```
nikkolasg, line 37: drand uses the BLS12-381 curve with the latest v7 RFC of hash-to-curve - link to [spec](https://github.com/drand/drand/blob/master/docs/SPECS.md#cryptographic-specification)

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Polling the drand network
+
+Filecoin nodes can make use of [drand endpoints](https://github.com/drand/drand/blob/master/core/client_public.go) in working with a drand beacon.
```
nikkolasg, line 41: Replace with latest much more improved [client](https://github.com/drand/drand/blob/master/client/client.go#L12) over HTTP

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,120 @@
...
+
+To start, a Filecoin node must store a set of drand peer servers it will connect to to poll for values and the shared public key it expects them to have (TODO: explain how this list/key is obtained in below section). Polling is done using gRPCs (and protobufs).
```
nikkolasg, line 43: Now polling from drand will be from "relay nodes" that are NOT drand nodes, as to have a clear separation of concerns and better protection. 
There are two ways to fetch, either from HTTP and/or gossipsub - 
@davidias @Kubuxu @willscott for more info

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,120 @@
...
+
+On initialization, the Filecoin node can call the `Group` endpoint in order to obtain the beacon's [group file](https://github.com/drand/drand/blob/57a6056a24d4ebaa27a44852636807364624b9fc/key/group.go). Using this, the Filecoin node should:
```
nikkolasg, line 45: Note that a client should have the hash of the group file embedded to verify the group it receives corresponds to the one he expects. (or directly the full group if it has it already) 
Look https://github.com/drand/drand/blob/master/client/client.go#L147

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,120 @@
...
+While drand returns multiple values with every call to the beacon (see above), Filecoin blocks need only store a subset of these in order to track a full drand chain. This information can then be mixed with on-chain data for use in Filecoin. See {{<sref randomness>}} for more.
+
+#### Verifying an incoming drand value
+
+Upon receiving a new drand randomness value from a beacon, a Filecoin node should immediately verify its validity. That is, it should verify:
+
+- that the `Signature` field is verified by the beacon's `PublicKey` as the beacon's signature of `SHA256(PreviousSignature || Round)`.
+- that the `Randomness` field is `SHA256(Signature)`.
+
+See [drand](https://github.com/drand/drand/blob/master/core/client_public.go#L93) for an example.
```
nikkolasg, line 70: I would link to the more core library https://github.com/drand/drand/blob/master/beacon/beacon.go#L63

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,120 @@
...
+
+During the catch up period, Filecoin nodes will backdate their blocks in order to continue using the same time-to-round mapping to determine which drand round should be integrated according to the time. Miners can then choose to publish their null blocks for the outage period (including the appropriate drand entries throughout the blocks, per the time-to-round mapping), or (as is more likely) try to craft valid blocks that might have been created during the outage.
+
+Note that based on the level of decentralization of the Filecoin network, we expect to see varying levels of miner collaboration during this period. This is because there are two incentives at play: trying to mine valid blocks from during the outage to collect block rewards, not falling behind a heavier chain being mined by a majority of miners that may or may not have ignored a portion of these blocks.
+
+In any event, a heavier chain will emerge after the catch up period and mining can resume as normal.
+
+### Filecoin's drand network characteristics
```
nikkolasg, line 115: I would not call that `Filecoin's drand network` but `drand network specification` ;) 
And yes, once the main network is launched we can give info on how to connect - note the network is meant to be evolving so I wouldn't stick too much of what is the composition of the network nor the governance in a spec that is supposed to be fixed.

src/systems/filecoin_blockchain/struct/block/block.id
```diff
@@ -53,6 +56,13 @@ type Block struct {
     SECPMessages  [msg.SignedMessage]
 }
 
+type BeaconEntry struct {
+    // Drand Round for the given randomness
+    Round       uint64
+    // Drand Signature for the given Randomness
+    Data   []byte
```
nikkolasg, line 63: Again would call that signature for consistency.

src/algorithms/crypto/randomness.md
```diff
@@ -2,15 +2,28 @@
...
+
+Entropy from the drand beacon can be harvested into a more general data structure: a `BeaconEntry`, defined as follows:
 
-Entropy from the ticket-chain can be combined with other values to generate necessary randomness that can be
+```go
+type BeaconEntry struct {
+    // Drand Round for the given randomness
+    Round       uint64
+    // Drand Signature for the given Randomness
+    Data   []byte
```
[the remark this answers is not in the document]
sternhenri: This was a point from @whyrusleeping (as per lotus impl of the beacon), the idea being to adopt a more general name in case the beacon ever changes and no longer relies on a signature per se. 

I am very agnostic here. I think I tend to agree that the change would be large enough that leaving it as sig is simpler for now...

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,120 @@
+---
+menuTitle: drand
+title: drand - Distributed Randomness
+---
+
+DRand (Distributed Randomness) is a publicly verifiable random beacon protocol Filecoin relies on as a source of unbiasable entropy for leader election (see {{<sref leader_election>}}).
+
+At a high-level, the drand protocol runs a series of MPCs (Multi-Party Computations) in order to produce a series of deterministic, verifiable random values. Specifically, after a trusted setup, a known (to each other) group of n drand nodes sign a given message using t-of-n threshold BLS signatures in a series of successive rounds occuring at regular intervals (the drand round time).
+Any node that has gathered t of the signatures can reconstruct the full BLS signature. This signature is then hashed using SHA-512 in order to produce a collective random value which can be verified against the collective public key generated during the trusted setup.
```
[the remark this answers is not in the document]
sternhenri: ack. thank you. I missed this actually, since in any case Drand uses sha256 right now.

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,120 @@
...
+In the following sections we look in turn at how the Filecoin protocol makes use of drand randomness, and at some of the characteristics of the specific drand network Filecoin uses.
+
+### Drand randomness outputs
+
+By polling the appropriate endpoint (see below for specifics on the drand network Filecoin uses), a Filecoin node will get back a drand value formatted as follows (e.g.):
+
+```
+{
+  "round": 367,
+  "signature": "b62dd642e939191af1f9e15bef0f0b0e9562a5f570a12a231864afe468377e2a6424a92ccfc34ef1471cbd58c37c6b020cf75ce9446d2aa1252a090250b2b1441f8a2a0d22208dcc09332eaa0143c4a508be13de63978dbed273e3b9813130d5",
```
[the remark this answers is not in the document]
sternhenri: agree... see above

src/libraries/drand/_index.md
sternhenri: yeah should be 256. good catch. With that said, i am basing this on the drand docs... So whatever confusion stems from my reading them. What does drand output?

src/libraries/drand/_index.md
sternhenri: I see. Will fix. Current drand readme is stale then fyi

src/algorithms/expected_consensus/_index.md
```diff
@@ -2,119 +2,79 @@
...
-```text
-We elaborate on the above claim:
+At a high-level, leader election works as follows:
 
-We want the miner's expected wins over time to be equal to w = minerPower/networkPower * EC.ExpectedLeadersPerEpoch
+- A miner draws an appropriate random seed for this epoch
+- They generate an `ElectionProof` from this seed by using a VRF to generate a signature over the seed at a given epoch, as defined in {{<sref randomness>}}. The `ElectionProof` is the VRF Proof.
+  - If the `ElectionProof`'s normalized digest value (i.e. the VRF Digest) is below a miner-specific `ElectionTarget` determined by the miner's power in SPC, it is valid: the miner can craft a block (see {{<sref block_producer>}}). 
+    - The `ElectionProof` Digest must be normalized according to both the `ExpectedLeadersPerEpoch` in EC (a network parameter) and the maximum value of the digest (based on the hash used to produce it).
+    - The `ElectionTarget` is set as the proportion of the miner's `QualityAdjustedPower` over the total network power. Accordingly leader election in EC is proportional to miner power.
```
zixuanzh, line 45: `TotalQualityAdjPower` or total network quality adjusted power to be precise

src/algorithms/expected_consensus/_index.md
```diff
@@ -182,7 +142,7 @@ The probability that two Tipsets with different blocks would have all the same t
 
 {{<label finality>}}
 ## Finality in EC
-EC enforces a version of soft finality whereby all miners at round N will reject all blocks that fork off prior to round N-F. For illustrative purposes, we can take F to be 500. While strictly speaking EC is a probabilistically final protocol, choosing such an F simplifies miner implementations and enforces a macroeconomically-enforced finality at no cost to liveness in the chain.
+EC enforces a version of soft finality whereby all miners at round N will reject all blocks that fork off prior to round N-F. For illustrative purposes, we can take F to be 900. While strictly speaking EC is a probabilistically final protocol, choosing such an F simplifies miner implementations and enforces a macroeconomically-enforced finality at no cost to liveness in the chain.
```
zixuanzh, line 145: dont forget to update specs-actors for this too or just file an issue ❤️ 
https://github.com/filecoin-project/specs-actors/blob/eac38c1d153cb985fa147c97113b28718361c90b/actors/builtin/miner/policy.go#L50

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -5,6 +5,8 @@ entries:
 - storage_power_actor
 ---
 
+TODO: remove all stale .id, .go files referenced
+
 {{<label storage_power_consensus>}}
 The Storage Power Consensus subsystem is the main interface which enables Filecoin nodes to agree on the state of the system. SPC accounts for individual storage miners' effective power over consensus in given chains in its {{<sref power_table>}}. It also runs {{<sref expected_consensus>}} (the underlying consensus algorithm in use by Filecoin), enabling storage miners to run leader election and generate new blocks updating the state of the Filecoin system.
```
zixuanzh, line 11: Total network power states should live here https://github.com/filecoin-project/specs-actors/blob/eac38c1d153cb985fa147c97113b28718361c90b/actors/builtin/power/power_state.go#L17

src/algorithms/expected_consensus/_index.md
```diff
@@ -2,119 +2,79 @@
...
 
-Take P to be the average miner power over the N sectors in their ProvingSet.
-We have P = SUM_{i=0}^N P_i / N
-
-The miner's likelihood of winning an election in any epoch is, for C tickets randomly drawn
-W = C * target = C * P_i/networkPower * EC.ExpectedLeadersPerEpoch * N/C = N * P_i/networkPower * EC.ExpectedLeadersPerEpoch
-
-with N * P_i ~= minerPower on expectation over enough epochs. That is to say: W = minerPower/networkPower * EC.ExpectedLeadersPerEpoch as wanted.
-```
+Conceptually, EC yields winners proportionally to their power since it enables each miner to generate a single `ElectionProof` whose digest can be normalized to yield a uniformly-drawn random number between 0 and 1. It is then compared to the miner's power in proportion to total network power (i.e. also between 0 and 1). The more powerful the miner, the more frequently their `ElectionProof`s will be valid.
```
zixuanzh, line 48: ditto

src/libraries/drand/_index.md
sternhenri: when you say embedded, does that mean in requests? Or cached, to verify responses against?

src/algorithms/expected_consensus/_index.md
sternhenri: https://github.com/filecoin-project/specs-actors/pull/392 definitely would have forgotten. Thanks for the reminder, hawkeye

src/algorithms/expected_consensus/_index.md
sternhenri: sorry, what is ditto in reference to?

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: Thanks @zixuanzh. cc @mishmosh , this is probably a deeper question related to how the spec should acknowledge or integrate specs-actors

src/algorithms/expected_consensus/_index.md
zixuanzh: I believe it's total network power --> Total Quality Adjusted Power, ty!

src/algorithms/crypto/randomness.md
```diff
@@ -21,58 +34,25 @@ Further, we define Domain Separation Tags with which we prepend random inputs wh
...
 - the epoch number, ensuring
-    - liveness for leader election -- in the case of null rounds, the new epoch number will output new randomness for LE
-    - distinct values for randomness sought before genesis -- where the genesis ticket will be returned
-    - For instance, if in epoch `curr`, a miner wants randomness from `lookback` epochs back where `curr - lookback <= genesis`, the ticket randomness drawn would be based on `genesisTicket.digest` where the `genesisTicket` is the randomness included in the genesis block. Using the epoch as part of randomness composition ensures that randomness drawn at various epochs prior to genesis has different values.
-- other entropy,
-    - ensuring that randomness is modified as needed by other context-dependent entropy (e.g. a miner address if we want the randomness to be different for each miner).
+    - liveness for leader election -- in the case no one is elected in a round and no new drand entry has appeared, the new epoch number will output new randomness for LE
```
nicola, line 44: 1. "drand" -> "beacon".
2. I am not sure I agree with the fact that liveness is guaranteed in the absence of the beacon in the current version of the protocol.

src/algorithms/expected_consensus/_index.md
```diff
@@ -2,119 +2,79 @@
...
-
-We also use the VRF from {{<sref vrf>}} to draw randomness as part of crafting the Challenge Tickets in order to ensure leader election is secret (a miner cannot be known to win until they publish their ticket on the chain).
-
-It is important to note that every block contains three artifacts: one, a ticket derived from last block's ticket to extend the ticket-chain, two, a challenge ticket array PoSt process run in this epoch, and three a PoStProof to prove each ChallengeTicket was derived correctly.
-
 ### Election Validation
 
-In order to determine that the mined block was generated by an eligible miner, one must check its `ChallengeTickets`' validity and that they were generated appropriately by the PoSt generator. Thereafter, for each ticket the miner must check that it is a winning challenge ticket, per the above definition.
+In order to determine that the mined block was generated by an eligible miner, one must check its `ElectionProof`'s validity, that it was generated using the appropriate drand entry, and that it is valid according to the miner's `ElectionTarget`, per the above definition.
```
nicola, line 77: "drand entry" -> "beacon entry"

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,118 @@
...
+- `filEpochDuration`    -- the Filecoin network's epoch duration (between any two leader elections)
+- `filGenesisTime`      -- the Filecoin genesis timestamp
+- `filEpoch`            -- the current Filecoin epoch
+- `drandGenesisTime`    -- drand's genesis timestamp
+- `drandPeriod`         -- drand's epoch duration (between any two randomness creations)
+
+Using the above, a Filecoin node can determine the appropriate drand round value to be used for use in {{<sref leader_election>}} in an epoch using both networks' reliance on real time as follows:
+
+```go
+MaxBeaconRoundForEpoch(filEpoch) {
```
nicola, line 85: Why "Max"? this should just be BeaconRoundForEpoch, no?

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,54 +46,124 @@ More precisely,
...
+    for rand == nil {
+        seeds := GetRandomSeedsForEpoch(epoch)
+        if seeds == nil {
+            epoch -= 1
+        } else {
+            rand = seeds[len(seeds)-1].Randomness
+        }
+    }
+    return rand
+}
```
nicola, line 79: Although this may be what it's currently implemented, we already mentioned in different places that it's just a temporary resolution, since this leads to multiple randomness for the same height in absence of blocks.

We can decide to leave it like this and expect a PR that proposes a fix.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,54 +46,124 @@ More precisely,
...
+        prevEntry := entries[currIdx - 1]
+        err := drand.Verify(node.drandPubKey, prevEntry.Data, currEntry.Data, currEntry.Round)
+        if err != nil {
+            return err
+        }
+        currIdx -= 1
+    }
+
+    return nil
+}
```
nicola, line 141: this seems to verify the entries but not to check the validity with respect to filecoin (e.g. checking that we don't have more entries or less entries for this particular epoch)

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,54 +46,124 @@ More precisely,
...
-
-The miner runs the prior ticket through a Verifiable Random Function (VRF) to get a new unique ticket which can later be derived for randomness (as shown above). The prior ticket is prepended with the ticket domain separation tag and concatenated with the miner actor address (to ensure miners using the same worker keys get different randomness).
+The miner runs the beacon entry through a Verifiable Random Function (VRF) to get a new unique ticket. The beacon entry is prepended with the ticket domain separation tag and concatenated with the miner actor address (to ensure miners using the same worker keys get different tickets).
 
 To generate a ticket for a given epoch n:
 ```text
-LastTicket = MinTicketValueAtEpoch(n-1)
-newRandomness = VRF_miner(H(TicketProdDST || index || Serialization(pastTicket, minerActorAddress)))
+randSeed = GetRandomnessForEpoch(n)
+newRandomness = VRF_miner(H(TicketProdDST || index || Serialization(randSeed, minerActorAddress)))
```
nicola, line 163: We have different randomnesses in Filecoin. We should name them all individually. This should be ElectionProofRandomness (correct?)

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,54 +46,124 @@ More precisely,
...
+
+This randomness may be drawn from various Filecoin chain epochs by the respective protocols that use them according to their security requirements.
+
+It is important to note that a given Filecoin network and a given drand network need not have the same round time, i.e. blocks may be generated faster or slower by Filecoin than randomness is generated by drand.
+For instance, if the drand beacon is producing randomness twice as fast as Filecoin produces blocks, we might expect two random values to be produced in a Filecoin epoch, conversely if the Filecoin network is twice as fast as drand, we might expect a random value every other Filecoin epoch. Accordingly, depending on both networks' configurations, certain Filecoin blocks could contain multiple or no drand entries.
+
+In all cases, Filecoin blocks must include all drand beacon outputs generated since the last epoch in the `BeaconEntries` field of the block header. Any use of randomness from a given Filecoin epoch should use the last valid drand entry included in a Filecoin block, as follows:
+
+```go
+GetRandomnessForEpoch(epoch) []BeaconEntry {
```
nicola, line 68: Shouldn't this be called GetBeaconEntryForEpoch?

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,54 +46,124 @@ More precisely,
...
+        if seeds == nil {
+            epoch -= 1
+        } else {
+            rand = seeds[len(seeds)-1].Randomness
+        }
+    }
+    return rand
+}
+
+GetRandomSeedsForEpoch(epoch) []BeaconEntry {
```
nicola, line 81: Shouldn't this be called GetBeaconEntriesForEpoch?

src/algorithms/crypto/randomness.go
```diff
@@ -16,7 +16,7 @@ import (
 // random beacon at a given epoch and incorporating requisite entropy
 GetRandomness(chain personalization crypto.DomainSeparationTag, randEpoch abi.ChainEpoch, entropy []byte) abi.Randomness {
```
nicola, line 17: This should be renamed into GetBeaconEntry, no?

src/algorithms/crypto/randomness.md
sternhenri: where do you see that second claim made? The above statement only says that if the frequency of the beacon is less than Fil's, we still have liveness. I'll detail a bit to avoid confusion

src/libraries/drand/_index.md
sternhenri: I ripped this from lotus mostly, but the rationale is this will return the highest such epoch number. In the case the beacon is faster than Filecoin block production there might be multiple rounds for an epoch. 

You can list these out by looking at the gap in function output between two rounds (same as in your doc fwiw).

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: Make it a FIP. If we already have a solution that's great and I can update. If not, the spec should represent the state of the protocol... but it's worth noting indeed. (made a note of it).

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: That's a fair point... I'm not sure how you'd do that without overly constraining (nor why it's strictly necessary)...

@Kubuxu can you point to where this is done in Lotus?

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: not correct, this is for ticket production (for the vrf chain). You can tell from the DST. But overall randomness seeds are all taken using the same method (assuming you meant we should name the method individually), which seems to me to be the right level of abstraction. Thereafter, the randomness only changes as it's assembled (using the dst, serialization, etc)... If you meant name the var, indeed feel free to change it!

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: If you want, up to impl to name it however they want. This is half closer to what it was (ie just legacy), half the fact that I felt we can abstract away where randomness comes from (on chain vs beacon)

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: ibid

src/algorithms/crypto/randomness.go
sternhenri: This definitely not. It is no longer a beacon entry at this point. It is formatted randomness, specifically for Fil.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: but changed it

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: Changed

sternhenri:
@nicola 
- [x] use the word Beacon instead of Drand, 
- [ ] introduce the concept of Beacon Period (a period that uses the same Beacon Entry), 
- [ ] try to avoid the word randomness and instead name every randomness specifically.

For the latter two, I think the concept of a beacon period is more confusing than helpful, and while the definition made sense in your doc, I didn't really see it used in a way where it simplified any explanation. So not yet sure why it's needed and I'd rather hold off on nomenclature unless we need it (maybe this is wrt the verification of which beacon entries got to which blocks?).

For the word randomness, if you mean the method, it's important imo that it be left as a single factory method for producing all sorts of randomness (what you get is context dependent); if you mean var names... Sgtm, I think most are understandable given context in spec.

Last thing remaining imo is the beacon verification method and making sure it checks out!

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
nicola: yeah, this solution won't solve it (a block could return multiple randomnesses based on current and previous consecutive null blocks), I agree it will be an FIP.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
nicola: name of the var.

src/algorithms/crypto/randomness.go
nicola: What I mean to say by "Entry" is a randomness that could be the same randomness for multiple blocks.

If we say GetRandomness, which randomness is that? Is that from the VRF chain? is it the one we called in the design docs an "ElectionSeed"?

mishmosh:
@sa8 and @nicola are reviewing this week on security aspects, and are setting up a call with @sternhenri.

src/algorithms/crypto/randomness.go
nicola: This will always be from the beacon.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
nicola: SPC does not account for individual power

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,54 +46,126 @@ More precisely,
...
+    // at currIdx == 0, must fetch last Fil block's last BeaconEntry
+    entries := append(entries, prevEntries[len(prevEntries)-1])
+    entries := append(entries, currEntries...)
+
+    currIdx := len(entries) - 1
+    for currIdx >= 0 {
+        // walking back the entries to ensure they follow one another
+        currEntry := entries[currIdx]
+        prevEntry := entries[currIdx - 1]
+        err := drand.Verify(node.drandPubKey, prevEntry.Data, currEntry.Data, currEntry.Round)
```
nicola, line 135: It can't be in the future.

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,118 @@
...
+    // determine what the latest Filecoin timestamp was from the current epoch number
+    var latestTs
+    if filEpoch == 0 {
+        latestTs = filGenesisTime
+    } else {
+        latestTs = ((uint64(filEpoch) * filEpochDuration) + filGenesisTime) - filEpochDuration
+    }
+    // determine the drand round number corresponding to this timestamp
+    dround := (latestTs - drandGenesisTime) / uint64(drandPeriod)
```
nicola, line 94: This is incorrect and it is off by 1. @nikkolasg will verify this and check this in the code (Audit issue 2)

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,54 +46,126 @@ More precisely,
...
+
+GetBeaconEntriesForEpoch(epoch) []BeaconEntry {
+    maxDrandRound := MaxBeaconRoundForEpoch(epoch)
+
+    if epoch == 0 {
+        // special case genesis
+        entries := []
+        rand := drand.Public(curr)
+        return append(entries, rand)
+    }
```
nicola, line 91: 3. (medium) The spec has an incorrect definition of the randomness (`curr` is undefined) used at entry 0, in addition, it doesn't seem like that there should be an entry at the genesis block.
  - TODO: check how this is implemented in lotus, if this is incorrect change it.

src/libraries/drand/_index.md
nikkolasg: Below are the two functions that drand uses to determine the mapping round <-> timestamp
The main relevant lines is:
```go
fromGenesis := now - genesis
nextRound = uint64(math.Floor(float64(fromGenesis)/period.Seconds())) + 1
```
in short, drand is creating the round 1 at genesis time, like filecoin, so a +1 is needed in the line in the spec.
Full function:
```go
func CurrentRound(now int64, period time.Duration, genesis int64) uint64 {
	nextRound, _ := NextRound(now, period, genesis)
	if nextRound <= 1 {
		return nextRound
	}
	return nextRound - 1
}

// NextRound returns the next upcoming round and its UNIX time given the genesis
// time and the period.
// round at time genesis = round 1. Round 0 is fixed.
func NextRound(now int64, period time.Duration, genesis int64) (nextRound uint64, nextTime int64) {
	if now < genesis {
		return 1, genesis
	}
	fromGenesis := now - genesis
	// we take the time from genesis divided by the periods in seconds, that
	// gives us the number of periods since genesis. We add +1 since we want the
	// next round. We also add +1 because round 1 starts at genesis time.
	nextRound = uint64(math.Floor(float64(fromGenesis)/period.Seconds())) + 1
	nextTime = genesis + int64(nextRound*uint64(period.Seconds()))
	return nextRound + 1, nextTime
}

File: src/systems/filecoin_blockchain/storage_power_consensus/_index.md (hunk 25 above)
Comment (nicola) [reply src/systems/filecoin_blockchain/storage_power_consensus/_index.md:135]: Attack:
- The current block should have entries from X till Y, instead the miner adds entries from X until Y+1
- This function verifies all the entries X till Y+1 and this check will pass.

File: src/systems/filecoin_blockchain/storage_power_consensus/_index.md [hunk 28]
```diff
@@ -44,54 +46,126 @@ More precisely,
...
+// Note that the below function is a temporary solution and likely to be
+// updated in future. 
+GetBeaconEntryForEpoch(epoch) []BeaconEntry {
+    var rand Randomness
+    for rand == nil {
+        seeds := GetBeaconEntriesForEpoch(epoch)
+        if seeds == nil {
+            epoch -= 1
+        } else {
+            rand = seeds[len(seeds)-1].Randomness
```
nicola, line 77: This should be `GetBeaconEntriesForEpoch(epoch-1)`

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
nicola: Fix: check that currEntry.Round <= MaxBeaconRoundForEpoch(epoch)

sternhenri: @nicola check out https://github.com/filecoin-project/lotus/blob/master/chain/beacon/beacon.go#L30. They do a slightly diff check. First they check that if the last block includes the `maxBeaconEntry` for this block's epoch, then this block has no beacon entries (i.e. that entries cannot be replayed on the chain), second they check exactly that `currEntry.Round == MaxBeaconRoundForEpoch`, not `<=`. So their checks are more restrictive than ours. Lmk which approach you prefer, adapted to theirs for now.

src/libraries/drand/_index.md
sternhenri: @nikkolasg understood. That said, I don't understand in the above code why you add `+1` twice in `NextRound`. Ergh that 1-indexing is confusing. I would suggest renaming the `nextRound` variable to `currentRound`, which is what it is, since it only becomes the `nextRound` after you increment it again in the return statement.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: Check out https://github.com/filecoin-project/lotus/blob/master/chain/beacon/beacon.go#L54.

Updated spec given error you found. Nice catch. Should match lotus logic (checking on one uncertainty).

sternhenri:
`GetRandomnessFromVRFChain` in lotus:

from https://github.com/filecoin-project/storage-fsm/blob/fe71d5b42de3b048bb7d3d015d137a9332c07b0c/states_sealing.go#L79, uses https://github.com/filecoin-project/storage-fsm/blob/fe71d5b42de3b048bb7d3d015d137a9332c07b0c/states_sealing.go#L49 which in turn uses this https://github.com/filecoin-project/lotus/blob/fe4efa0e0ee045d38f63f4955d01cbf3e36bc41f/node/impl/full/chain.go#L59, our old `getRandomness` fn. 

Spec updated accordingly.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,70 +46,160 @@ More precisely,
...
+    }
+
+    entries := []
+    curr := maxDrandRound
+    for curr > prevMaxDrandRound {
+        rand := drand.Public(curr)
+        entries = append(entries, rand)
+        curr -= 1
+    }
+    return entries
```
Kubuxu, line 116: We fetch them in reverse order in lotus, but then we reverse the whole array. So entries always end up being in increasing order.

sternhenri: Thanks!

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,70 +46,162 @@ More precisely,
...
+```go
+// This need not be done for the genesis block
+// We assume that blockHeader and priorBlockHeader are two valid subsequent headers where block was mined atop priorBlock
+ValidateBeaconEntries(blockHeader, priorBlockHeader) error {
+    currEntries := blockHeader.BeaconEntries
+    prevEntries := priorBlockHeader.BeaconEntries
+
+    maxRoundForEntry := MaxBeaconRoundForEpoch(blockHeader.Epoch)
+    // ensure entries are not repeated in blocks
+    lastBlocksLastEntry := prevEntries[len(prevEntries)-1]
```
nikkolasg, line 138: What happens currently in Lotus is: 
* no drand value / information in the genesis block
* first epoch block contains one drand entry which is _not_ verifiable by itself since the entry does not contain the previous signature. We can't put a entry in the genesis block because we dont know it in advance (while the genesis block is quite predetermined I guess)
* At the second epoch, the "main chain" 2-nd epoch block contains a drand entry whose previous signature is the drand signature put in the epoch 1 block.

**consequence**: Miners will build on the first block without knowing if the randomness is valid or not. However, on the second block, miners will be able to verify if the randomness on the first round is correct. If not, that's a "re-org" as usual.

So here if `priorBlockHeader.Height == 0` we must skip that verification and return nil directly. This is what Lotus is doing [here](https://github.com/filecoin-project/lotus/blob/master/chain/beacon/drand/drand.go#L169).

That being said, there are other potential ways to handle genesis.  For example, in the epoch 1 header, include the full beacon entry with the previous signature as well.

sternhenri: right, I didn't carry through my change appropriately from `GetBeaconEntriesForEpoch`. Thanks. Should be fixed now.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
nicola: Following their way will work!

src/algorithms/crypto/randomness.md
```diff
@@ -86,6 +66,36 @@ GetRandomness(dst, l, s):
 {{< readfile file="/docs/actors/actors/crypto/randomness.go" code="true" lang="go" >}}
 {{< readfile file="/docs/systems/filecoin_blockchain/struct/chain/chain.go" code="true" lang="go" >}}
 
+## Drawing tickets from the VRF-chain for proof inclusion
+
+There are a few instances in which the protocol requires inclusion  to be drawn from the Filecoin blockchain's VRF-chain (which generates {{<sref tickets>}} with each new block), in order to tie certain proofs to a particular set of Filecoin blocks (i.e. a given chain or fork). This is notably the case for Seal Pre-commits (see {{<sref sealing>}}).
```
nikkolasg, line 71: Can we list out the instances where it is the case ? It's not obvous right now (to me and I believe @sa8 )

src/algorithms/expected_consensus/_index.md
```diff
@@ -2,119 +2,79 @@
...
-```text
-We elaborate on the above claim:
+At a high-level, leader election works as follows:
 
-We want the miner's expected wins over time to be equal to w = minerPower/networkPower * EC.ExpectedLeadersPerEpoch
+- A miner draws an appropriate random seed for this epoch
+- They generate an `ElectionProof` from this seed by using a VRF to generate a signature over the seed at a given epoch, as defined in {{<sref randomness>}}. The `ElectionProof` is the VRF Proof.
+  - If the `ElectionProof`'s normalized digest value (i.e. the VRF Digest) is below a miner-specific `ElectionTarget` determined by the miner's power in SPC, it is valid: the miner can craft a block (see {{<sref block_producer>}}). 
+    - The `ElectionProof` Digest must be normalized according to both the `ExpectedLeadersPerEpoch` in EC (a network parameter) and the maximum value of the digest (based on the hash used to produce it).
+    - The `ElectionTarget` is set as the proportion of the miner's `TotalQualityAdjPower` over the total network quality-adjusted power. Accordingly leader election in EC is proportional to miner power.
```
nicola, line 45: 5. (high) Leader election is not using power table lookback `ElectionPowerTableLookback`

This should be the total quality adjusted power at a previous epoch

src/libraries/drand/_index.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Edge cases and dealing with a drand outage
+
+It is important to note that any drand beacon outage will effectively halt Filecoin block production. Given that new randomness is not produced, Filecoin miners cannot generate new blocks.
+
+After a beacon downtime, drand nodes will work to quickly catch up to the current round, as defined by wall clock time. In this way, the above time-to-round mapping in drand (see above) used by Filecoin remains an invariant after this catch-up following downtime.
+
+So while Filecoin miners were not able to mine during the drand outage, they will quickly be able to run leader election thereafter, given a rapid production of drand values. We call this a "catch up" period.
+
+During the catch up period, Filecoin nodes will backdate their blocks in order to continue using the same time-to-round mapping to determine which drand round should be integrated according to the time. Miners can then choose to publish their null blocks for the outage period (including the appropriate drand entries throughout the blocks, per the time-to-round mapping), or (as is more likely) try to craft valid blocks that might have been created during the outage.
```
nikkolasg, line 108: Some comments:
1.
> Miners can then choose to publish their null blocks for the outage period 

Not clear - we can't publish a null block because a null block doesn't exist. 
2. Do we agree that one can "simply" append all the "catched-up" drand entries in the next block he's creating, right ? I believe the function below `ValidateBeaconEntries(blockHeader, priorBlockHeader) error {` does allow for that.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
```diff
@@ -44,70 +46,168 @@ More precisely,
...
+    prevMaxDrandRound := MaxBeaconRoundForEpoch(epoch - 1)
+    if (maxDrandRound == prevMaxDrandRound) {
+        // no new beacon randomness
+        return entries
+    }
+
+    entries := []
+    curr := maxDrandRound
+    for curr > prevMaxDrandRound {
+        rand := drand.Public(curr)
```
nikkolasg, line 112: We must make the assumption that this call is blocking: i.e. if the drand network is DOWN, then this call BLOCKS.

src/libraries/drand/_index.md
nikkolasg: >Miners can then choose to publish their null blocks for the outage period (including the appropriate drand entries throughout the blocks, per the time-to-round mapping)

Even though I don't quite understand this, this is not a behavior that Lotus codes at the moment (which is doing the "catch up" as well, mine as fast as possible).

src/algorithms/crypto/randomness.md
nicola: Tl;dr sealrandomness is VRF chain all the rest is drand. I believe this can come in a separate PR.

nicola:
In a new PR
- @nikkolasg will address the section "Polling the Drand network"
- @nikkolasg will address GetRandomness in the VM
- @sa8 will update the latest target function and will use correct lookbacks

src/algorithms/crypto/randomness.md
sternhenri: Per @nicola's point, I feel like current phrasing should suffice. Otherwise I can do this, but feels a bit restrictive:

```suggestion
There are a few instances in which the protocol requires inclusion  to be drawn from the Filecoin blockchain's VRF-chain (which generates {{<sref tickets>}} with each new block), in order to tie certain proofs to a particular set of Filecoin blocks (i.e. a given chain or fork). This is currently only the case for Seal Pre-commits (see {{<sref sealing>}}).
```

src/libraries/drand/_index.md
sternhenri: >  2. Do we agree that one can "simply" append all the "catched-up" drand entries in the next block he's creating, right ? I believe the function below ValidateBeaconEntries(blockHeader, priorBlockHeader) error { does allow for that.

I agree with that, on 1, what I mean is publish blocks, which will implicitly publish the null blocks. This is the key question, after an outage should we have a deterministic block/entry schedule (ie we know exactly how many entries should fit into the series of blocks and how they will be mapped onto those blocks), or not (ie there could be many valid drand/block mappings published on recovery).

I thought it was the latter @nicola , though you are right @nikkolasg that lotus implements one of those versions (ie catch up as fast as you can). IMO, that is not in conflict with the spec as written.

src/systems/filecoin_blockchain/storage_power_consensus/_index.md
sternhenri: added as a comment. thank you!

src/algorithms/expected_consensus/_index.md
sternhenri: seeing that @sa8 will address this in https://github.com/filecoin-project/specs/pull/967 right?

nicola:
Closing this issue for now and continuing on https://github.com/filecoin-project/specs/pull/967#issuecomment-654783785

PR #954, opened 2020-06-15
try to fix build.

raulk:
One of the compilation error messages has to do with a go version mismatch. Could that be the problem?


nonsense:
Now that the code compiles and renders locally, I think it is worth merging this as well, which probably will fix the `deploy` step on Github Actions, as currently the problem is:

```
Run sudo chown -R root:root /home/runner/work/specs
cd tools/codeGen && go build -o ../../bin/codeGen
# runtime/internal/atomic
compile: version "go1.14.4" does not match go tool version "go1.13.8"
# internal/cpu
compile: version "go1.14.4" does not match go tool version "go1.13.8"
# runtime/internal/sys
compile: version "go1.14.4" does not match go tool version "go1.13.8"
# internal/race
compile: version "go1.14.4" does not match go tool version "go1.13.8"
# sync/atomic
compile: version "go1.14.4" does not match go tool version "go1.13.8"
# unicode/utf8
compile: version "go1.14.4" does not match go tool version "go1.13.8"
# unicode
compile: version "go1.14.4" does not match go tool version "go1.13.8"
# math/bits
compile: version "go1.14.4" does not match go tool version "go1.13.8"
Makefile:186: recipe for target 'bin/codeGen' failed
make: *** [bin/codeGen] Error 2
##[error]Process completed with exit code 2.
```

Issue #949, opened 2020-05-29
Compilation errors


mishmosh:
Some problems have been introduced over recent sprints, causing compilation errors. This is pretty urgent as it affects readability for everyone & workflows for contributors:

- A CircleCI test is failing. Likely reasons:
  - syntax errors in go
  - some files that can’t be found (because renamed or moved to `specs-actors` repo).
- Failure to compile also means that auto-publish doesn’t work, so specs website is out of date.

mishmosh:
Reported by @hugodias:

```
algorithms/crypto/randomness.go:11:2: module github.com/filecoin-project/specs-actors@latest found (v0.5.6, replaced by ../../src/actors), but does not contain package github.com/filecoin-project/specs-actors/actors/abi/big
systems/filecoin_blockchain/struct/block/election.go:6:2: no matching versions for query "latest"
make: *** [build-code] Error 1
```

This is likely due to `.id` files. cc @zixuanzh suggests removing id files entirely.

hugodias:
Hey, @mishmosh I'm not related to this project. 

I'd say I never used it and also never reported such a problem.

mishmosh:
@hugodias apologies, wrong handle!

nonsense:
Fixed in https://github.com/filecoin-project/specs/pull/954 and https://github.com/filecoin-project/specs/pull/955

PR #955, opened 2020-06-15
fix compilation of filecoin specs

nonsense:
This PR is addessing https://github.com/filecoin-project/specs/issues/949 and hopefully fixing the compilation issues of the `filecoin-specs` website.

A bunch of problems were introduced to the Go code part of the `specs` in https://github.com/filecoin-project/specs/pull/883 - I have reverted the changes that don't compile, and have commented on every one of them explaining what went wrong, so that we see if the change is actually needed, but needs to be re-implemented.

---

I suggest we review the comments on the code I reverted, and make sure that we don't leave anything in terrible state, and create issues to update properly the codebase if the changes I am reverting are actually necessary, for example:
1. Introduce `GetRandomness` and actually use it where it should be used.
2. Remove `Output` and remove references to it.
3. Update function signature of `Verify`.
4. Define `genesisEpoch` correctly (currently it is defined as 0).
5. Remove `DrawRandomness` and all references to it.


src/algorithms/crypto/randomness.go
```diff
@@ -1,28 +0,0 @@
...
-
-import (
-	"bytes"
-	"encoding/binary"
-
-	addr "github.com/filecoin-project/go-address"
-	"github.com/minio/blake2b-simd"
-
-	abi "github.com/filecoin-project/specs-actors/actors/abi"
-	big "github.com/filecoin-project/specs-actors/actors/abi/big"
```
nonsense, line 11: Looks like `github.com/filecoin-project/specs-actors/actors` is replaced with vendored version at `src/actors` that doesn't have package `big`.

`github.com/filecoin-project/specs-actors` does have a package `abi/big`, so maybe this is what the author was referring to, but in this case we should probably remove `src/actors`.

src/algorithms/crypto/randomness.go
```diff
@@ -1,28 +0,0 @@
...
-	autil "github.com/filecoin-project/specs-actors/actors/util"
-)
-
-// Randomness returns a (pseudo)random byte array drawing from a
-// random beacon at a given epoch and incorporating requisite entropy
-GetRandomness(chain personalization crypto.DomainSeparationTag, randEpoch abi.ChainEpoch, entropy []byte) abi.Randomness {
```
nonsense, line 17: Missing `func`, so I don't think this can compile.

src/algorithms/crypto/vrf.go
```diff
@@ -20,9 +23,11 @@ func (self *VRFResult_I) MaxValue() util.Bytes {
 func (self *VRFKeyPair_I) Generate(input util.Bytes) VRFResult {
 	// sig := new(BLS).Sign(input, self.SecretKey)
 	var blsSig util.Bytes
+
+	digest := blake2b.Sum256(blsSig)
 	ret := &VRFResult_I{
-		Proof_: blsSig,
-		Digest_: blake2b.Sum256(blsSig)
```
nonsense, line 25: Missing `,` and different type. `blake2b.Sum256` returns `[32]byte`, and `Digest_` expects `[]byte`.

src/algorithms/crypto/vrf.id
```diff
@@ -13,6 +13,8 @@ type VRFKeyPair struct {
 }
 
 type VRFResult struct {
+    Output            Bytes  // @(internal)
```
nonsense, line 16: This was removed in the linked PR, however `Output` is still being referred to from other places, so I am reverting it back.

src/systems/filecoin_blockchain/struct/block/election.go
```diff
@@ -3,7 +3,7 @@ package block
 import (
 	addr "github.com/filecoin-project/go-address"
 	abi "github.com/filecoin-project/specs-actors/actors/abi"
-	crypto "github.com/filecoin-project/specs/algorithms/randomness"
```
nonsense, line 6: I am not sure what `randomness` package is referring to, because I can't find such package.

src/systems/filecoin_blockchain/struct/block/election.go
```diff
@@ -12,6 +12,16 @@ func (tix *Ticket_I) ValidateSyntax() bool {
...
 }
 
-func (tix *Ticket_I) Verify(proof util.Bytes, digest util.Bytes, pk filcrypto.VRFPublicKey) bool {
-	return tix.VRFResult_.Verify(proof, pk) && digest == blake2b.Sum256(proof)
+//func (tix *Ticket_I) Verify(proof util.Bytes, digest util.Bytes, pk filcrypto.VRFPublicKey) bool {
+//p := blake2b.Sum256(proof)
+//return tix.VRFResult_.Verify(proof, pk) && bytes.Compare(digest, p[:]) == 0
+//}
+
+func (tix *Ticket_I) Verify(randomness util.Bytes, pk filcrypto.VRFPublicKey, minerActorAddr addr.Address) bool {
```
nonsense, line 20: `Verify` func signature was being changed to the one commented out above, but there are a lot of places that actually refer to the previous signature, so I reverted it back (and left a commented out version of the new `Verify` at the top, in case we need to change this).

src/systems/filecoin_blockchain/struct/block/election.go
```diff
@@ -12,6 +12,16 @@ func (tix *Ticket_I) ValidateSyntax() bool {
...
+	input := acrypto.DeriveRandWithMinerAddr(acrypto.DomainSeparationTag_TicketProduction, randomness, minerActorAddr)
+	return tix.VRFResult_.Verify(input, pk)
+}
+
+func (tix *Ticket_I) DrawRandomness(epoch abi.ChainEpoch) util.Bytes {
```
nonsense, line 25: `DrawRandomness` was removed from the linked PR, but is referred to from other places.

src/systems/filecoin_blockchain/struct/chain/chain.go
```diff
@@ -10,8 +10,8 @@ import (
 // For negative epochs, it should return a tipset composed of the genesis block.
 func (chain *Chain_I) TipsetAtEpoch(epoch abi.ChainEpoch) Tipset {
 	current := chain.HeadTipset()
-
-	for current.Epoch() > epoch && epoch >= genesisEpoch{
+	genesisEpoch := abi.ChainEpoch(5)
```
nonsense, line 13: `genesisEpoch` did not exist, so we have to declare it. I realize that setting it to `abi.ChainEpoch` is wrong, but couldn't find a reference to `genesisEpoch` anywhere.

src/systems/filecoin_blockchain/struct/chain/chain.go
```diff
@@ -23,7 +23,8 @@ func (chain *Chain_I) TipsetAtEpoch(epoch abi.ChainEpoch) Tipset {
 func (chain *Chain_I) RandomnessSeedAtEpoch(epoch abi.ChainEpoch) abi.RandomnessSeed {
 
 	ts := chain.TipsetAtEpoch(epoch)
-	return ts.MinTicket().Digest()
+	//	return ts.MinTicket().Digest()
```
nonsense, line 26: `Digest()` doesn't exist, so I reverted this code to return `DrawRandomness(epoch)` as the version previous to the broken PR.

mishmosh:
Tagging @whyrusleeping for review (and cc @daviddias @momack2).

mishmosh:
Per @nicola and @zixuanzh recommendation, tagging @anorth for review as the most-familiar person with this area.

src/systems/filecoin_blockchain/struct/block/election.go
anorth: This change is problematic because the actors/crypto is implemented from the spec. This spec (or discussion/issues etc) is how we know what to write. It would really be much better if the spec duplicated what it needs from actors/crypto rather than import it. Otherwise, if we change actors, we'll have no way to know that we messed it up.

src/systems/filecoin_blockchain/struct/chain/chain.go
anorth: `0` would be a better value

src/systems/filecoin_blockchain/struct/chain/chain.go
anorth: Unless DrawRandomness does the digest (which I can't tell because it's now a reference to actors code), this isn't right and the digest must be taken afterwards.

anorth:
I am really not knowledgeable here. #883 told me what to do, and I don't know much more than having been able to implement it.

I have a vague memory of having been subsequently told that that PR was _also_ wrong, and the implementation changed again.

src/systems/filecoin_blockchain/struct/block/election.go
nonsense: I am not sure I understand how to solve this.

In this repo `github.com/filecoin-project/specs-actors/actors` is replaced with `replace github.com/filecoin-project/specs-actors => ../../src/actors`.

So `github.com/filecoin-project/spec-actors` is already duplicated/vendored (and actually differs, for example `DeriveRandWithMinerAddr` is present locally, and not part of `github.com/filecoin-project/spec-actors` AFAICT).

src/systems/filecoin_blockchain/struct/chain/chain.go
nonsense: `DrawRandomness` is referencing the local (replaced in go.mod) version of `spec-actors`, which if I understand what you mean should be the `source of truth` for what gets implemented / copied to `github.com/filecoin-project/specs-actors`

If it was referencing the latest version at `github.com/filecoin-project/specs-actors`, I think this won't compile, so I am not sure which is really the correct version we want to use.

nonsense:
Considering the comments from @anorth, I guess the `source-of-truth` version for Filecoin should be in this repo - either as markdown prose, or as both code + markdown prose, so that folks know what needs to be implemented on https://github.com/filecoin-project/specs-actors and other repos and we should not be referencing actual implementation from this repo.

I believe this is the case right now, as spec-actors is actually replaced with a local version.

However it appears that this repo has diverged from https://github.com/filecoin-project/specs-actors and I can't at all say which version is the latest, most correct one.

src/algorithms/crypto/randomness.go
nonsense: Having read what @anorth said about spec-actors being based on this repo, if `abi/big` is something needed for Filecoin (which I guess it is), we should probably copy it back to this repo as a `source-of-truth` - maybe in a follow up PR alongside the rest of the issues once identified.

anorth: So I do not think we should copy pedestrian code like a bigint wrapper into the spec (but IMO the spec shouldn't really have much code anyway).

I'm not sure if it's obvious, but the submodule actors repo is just the same specs-actors repo. It's not "vendored" or behind on purpose, but because updating the code here to keep up with it is painful busywork.

It would be better if this repo did not have code that depended on the actors. It's probably ok if pragmatically it does so just to import a bunch of types like the `abi` package and bigint wrappers.

src/systems/filecoin_blockchain/struct/chain/chain.go
anorth: No, the vendored actors repo is in no way a source of truth. It just points at an old commit of github.com/filecoin-project/specs-actors reflecting the point in time at which we liberated the actors code from the spec.

- Best would be that the relevant algorithm is expressed directly in this spec as pseudocode
- Slightly worse would be as code
- Worst of all is to reference an external repo like specs-actors

anorth:
> Considering the comments from @anorth, I guess the source-of-truth version for Filecoin should be in this repo - either as markdown prose, or as both code + markdown prose, so that folks know what needs to be implemented on filecoin-project/specs-actors and other repos and we should not be referencing actual implementation from this repo.

💯 

> I believe this is the case right now, as spec-actors is actually replaced with a local version.

It is not. The spec is behind and probably wrong. The locally-replaced spec-actors is merely an old commit from filecoin-project/specs-actors

To the extent that they differ, filecoin-project/specs-actors is definitely more correct. But on some tricky, shared things like randomness, this spec should directly contain the algorithm.

anorth:
I don't mean to block this change. Net, it's making the world a better place.

The work required to get the spec to point at up-to-date actors is huge and probably no-one's priority. IMO we should break the dependency, but I'm only a chicken in this bacon-and-eggs. If you wanna make a compromise for now and ask me to approve it (in lieu of any other owner), I will do so.

nonsense:
> I don't mean to block this change. Net, it's making the world a better place.

> The work required to get the spec to point at up-to-date actors is huge and probably no-one's priority. IMO we should break the dependency, but I'm only a chicken in this bacon-and-eggs. If you wanna make a compromise for now and ask me to approve it (in lieu of any other owner), I will do so.

This was going to be my next question! The idea behind this change was to make the code compile, so that we can render the latest version of the .md files and read the latest version of the spec.

Having said that, I was told multiple times over the last 1-2 days that these specs are also not up-to-date, behind and probably wrong.

---

It is not my decision to make, as I am really external to the team who is maintain these specs. I this it makes sense to merge this PR, so that specs compile and at least Markdown can be rendered.

It is really out of scope for this PR (or for me at this stage) to get the spec to be up-to-date.

PR #969, opened 2020-07-03
specs.next | beta.specs.filecoin.io

hugomrdias:
/cc @yiannisbot


next/content/math-mode.md
```diff
@@ -0,0 +1,432 @@
+---
+title: Math Mode
+bookhidden: true
+math-mode: true
+---
+
+# Introduction
+---
+{{<plain hidden>}}
```
vmx, line 9: That's not really needed. Without this shorttag it will put some dom there, but it will only display as an empty paragraph, which I think is good enough.

Without a shortcode it will be valid Markdown if you strip off the header.

next/.site/layouts/partials/docs/inject/body.html
```diff
@@ -1,12 +1,25 @@
...
-<script defer src="{{ $searchJS.RelPermalink }}" integrity="{{ $searchJS.Data.Integrity }}"></script>
-{{ end -}}
\ No newline at end of file
+  {{- $searchJSFile := printf "%s.search.js" .Language.Lang }}
+  {{- $searchJS := resources.Get "search.js" | resources.ExecuteAsTemplate $searchJSFile . | resources.Minify | resources.Fingerprint }}
+  <script defer src="{{ $searchJS.RelPermalink }}" integrity="{{ $searchJS.Data.Integrity }}"></script>
+{{ end -}}
+
+{{- if default false ($.Param "math-mode") -}}
+<script src="https://cdn.jsdelivr.net/npm/katex@0.11.1/dist/katex.min.js" integrity="sha384-y23I5Q6l+B6vatafAwxRu/0oK/79VlbSz7Q9aiSZUvyWYIYsd+qj+o24G5ZU2zJz" crossorigin="anonymous"></script>
```
vmx, line 11: Wouldn't it make sense to use the local version that it is part of the theme?

next/.site/layouts/partials/docs/inject/body.html
```diff
@@ -1,12 +1,25 @@
...
+<script src="https://cdn.jsdelivr.net/npm/katex@0.11.1/dist/contrib/auto-render.min.js" integrity="sha384-kWPLUVMOks5AQFrykwIup5lo0m3iMkkHrD0uJ4H5cjeGihAutqP0yW0J6dpFiVkI" crossorigin="anonymous"></script>
+<script>
+renderMathInElement(document.body, {
+    ignoredTags: ["script", "noscript", "style", "textarea"],
+    throwOnError: false,
+    delimiters: [
+        {left: "$$", right: "$$", display: true},
+        {left: "$", right: "$", display: false},
+        {left: "\\(", right: "\\)", display: false},
+        {left: "\\[", right: "\\]", display: true}
```
vmx, line 21: We don't need those, I would just keep the dollar ones.

next/content/math-mode.md
hugomrdias: strip off the header ?
this is needed so we dont get the flash of math code

next/.site/layouts/partials/docs/inject/body.html
hugomrdias: i will do that later

next/.site/layouts/partials/docs/inject/body.html
hugomrdias: we might in the future

next/content/math-mode.md
vmx: Stripping of the header is what I do if I render without Hugo (though I can see that it's not a priority and I can see that we don't really need it).

The flashing is a good point, other than that I don't see the need of using a special shortcode (but again I've fine with it). I just wanted to bring it up as less shortcodes also makes it easier for authors.

next/.site/layouts/partials/docs/inject/body.html
vmx: I prefer limiting things, so that you don't accidentally render things as math. With only dollar signs the rules are simple.

next/content/math-mode.md
hugomrdias: what about the `\_` ?
fyi in the future i will add a pipeline to make this static

vmx: You just put it into a code block, it will be hidden by KaTeX automatically as doesn't contain any data. The only artifact will be a bit of space as it renders it into a `<span>`.

vmx: Oh wait, I can even put it into the first actual code block, that would work as well. No special short codes would be needed. I can just do that an show it to you in my PR and then we can decide :)

hugomrdias: code blocks for hidden math dont work it renders an empty box

olizilla:
@jbenet work on the specs upgrade is going on in this PR. Latest rendered preview is here https://deploy-preview-969--fil-spec-staging.netlify.app/  (the deploy/netlify status check details will always have the most recent rendered preview)

daviddias:
Some nits on usability:

image

image

image

image

daviddias:
image

olizilla:
> can you make these always be expanded
@daviddias can you state the problem you are feeling? Can you frame your feedback in terms of what you need to be productive? Is it that you can't find sections without seeing an over view of all the headings on the left?

daviddias:
> @daviddias can you state the problem you are feeling? Can you frame your feedback in terms of what you need to be productive? Is it that you can't find sections without seeing an over view of all the headings on the left?

That's correct. And given that now each section is a separate page, I have to wait for the whole HTTP RTT to get the next page, which makes skimming and understanding where things are and how they are organized much harder.

The previous page solved this with a slider

image

davidad:
My feedback:
1. Using an animated icon as a symbolic glyph goes against UX guidelines, which point out that animation draws the user's attention, so it should be unobtrusive and brief ([source](https://www.nngroup.com/articles/animation-purpose-ux/)), or in response to user actions or state transitions ([source](https://material.io/design/iconography/animated-icons.html)).
2. It would be nice to be able to expand multiple parts of the sidebar at once; I'd also like an "expand all" button.
3. The search function doesn't show me where on the matching page the match was found, neither by highlighting or scrolling. I can do a Cmd-F, but now I have to search for the same thing twice. 
4. There's no "show as a single page" to enable Cmd-F alone to search the whole spec.

I see that all these points have basically already been covered by @daviddias but hope it's helpful to hear it from another user.

dignifiedquire:
Screenshot 2020-07-17 at 11 49 02

This tiny embedded frame of a PDF is not really useful, either this should be converted into spec content, or just an external link

dignifiedquire:
It would be great to have a light mode, I know there are a lot of fans of dark mode, but I feel it is quite hard to read for longer forms of text.

nicola:
Same here, light mode please, it is dark and the text is too gray for me :(

hugomrdias:
Thank for all the feedback keep it coming !! We will improve based on it, you can already see here https://github.com/filecoin-project/specs/projects/8 our planing.

hugomrdias:
This PR will be closed in favor of the beta branch which will be the default branch from now on. The previous setup still lives in the master branch until we finish the beta.

For more feedback please use this issue https://github.com/filecoin-project/specs/issues/992
Thank you.

PR #974, opened 2020-07-08
WIP: Some KaTeX example to play with

vmx:
Please see the notes in the source code that explains a few things.

The core issue here is mixing TeX and Markdown. Hugo isn't flexible
enough to introduce custom parsing of the dollar signs.

Ideas I have:

 - Use a Hugo fork with minimal changes that makes it possible to
   render KaTex on build time:
   https://graemephi.github.io/posts/server-side-katex-with-hugo-part-2/
   The patch is really minimal
   https://github.com/graemephi/kahugo
   and hence would be my preferred solution
 - Use a hacky way that doesn't need a fork, but lots of tooling:
   https://graemephi.github.io/posts/static-katex-with-hugo/
 - Insert the `katex` shortcodes in a pre-processing step around
   every `$`. This would make the source still easier to read/maintain,
   though it might be confusing where the content actually lives (except
   if we can make that somehow work so that the one editing the files
   won't notice that there's a special processing step involved

### DIFF
--- a/next/.site/layouts/shortcodes/katex.html
+++ b/next/.site/layouts/shortcodes/katex.html
@@ -1,0 +1,30 @@
+{{- if not (.Page.Scratch.Get "katex") -}}
+<!-- Include katext only first time -->
+<link rel="stylesheet" href="{{ "katex/katex.min.css" | relURL }}" />
+<script defer src="{{ "katex/katex.min.js" | relURL }}"></script>
+<script defer src="{{ "katex/auto-render.min.js" | relURL }}" onload='renderMathInElement(document.body, {
+  delimiters: [
+   {left: "$$", right: "$$", display: true},
+   {left: "$", right: "$", display: false},
+  ]
+});'></script>
+<style>
+.katex-display {
+  margin: 0;
+}
+.katex-display > .katex {
+  text-align: left;
+}
+</style>
+
+
+{{- .Page.Scratch.Set "katex" true -}}
+{{- end -}}
+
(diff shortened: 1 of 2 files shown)


vmx:
One more idea (which probably is not possible with Hugo, I mention it though): Having a shortcode that does proper escaping of the stuff between the dollar signs and then makrdownifying it.

porcuquine:
I think I previously had embedded LaTeX working fine in the spec using MathJax. Is there a reason that won't work or isn't the right answer for present purposes? I can believe there might be, and I don't know whether that support has been removed  from 'the spec' or is still lurking there somewhere.

I took a quick peek at the old Markdown, and it looks like the generated Markdown used `\(…\)` syntax rather than `$…$` syntax, but I'm not sure if that's a hard requirement.

Ignore all of this if it misses the point of what you're doing here.

vmx:
I'm closing this issue as @hugomrdias find a great way (with code blocks) to make things work. The new work is at https://github.com/filecoin-project/specs/pull/975.

Thanks @porcuquine for your comment, you're right that it used to work. Long story short, the whole system gets at revamp (which I totally agree with is a good idea) and things changes a lot from past iterations.

hugomrdias:
for the record and to answer @porcuquine 's question using mathjax its just too heavy has you can see if you open this document in hackmd it completely freezes the browser, not to mention mobile browsers.
With our current approach desktop browsers can easily render this HUGE document and mobile browser can do a good enough job also so win win.

porcuquine:
Thanks for the explanation.

Issue #581, opened 2019-10-21
Automatic draft publishing


jbenet:
- The spec uses `ipfs` to publish drafts
- websites like github.com/ipfs/website use some tooling to render + publish to ipfs per PR (see below)
- This would be useful for people to see the rendered output in a PR (and make sure it _can_ be published).

---

Example from: https://github.com/ipfs/website/pull/319 (click "show all checks" in the "All checks have passed" section).

![image](https://user-images.githubusercontent.com/138401/67234386-e1ab0a80-f3f9-11e9-9e62-688a8e2b0871.png)

andrew:
The code for that integration is here: https://github.com/ipfs-shipyard/ipfs-dns-deploy when using Circleci for running tests, built by @olizilla

olizilla:
I am working on adding this today. Adding it as a github action to integrate with the existing workflow.

Do we also want to migrate it from gh-pages to Fleek or our own dnslink update style publishing?

PR #976, opened 2020-07-13
feat: publish to IPFS on PR

olizilla:
- Adds step to pin the rendered website to IPFS, whenever a commit is pushed.
- Replaces hardcoded paths with the `$GITHUB_WORKSPACE` var. This makes the build work on forked repos, where the repo has to be renamed, and shields us from future issues if actions change their workspace dir.
- Updated the github action to pin to IPFS to work with the new style github actions and published as 1.0 https://github.com/ipfs-shipyard/ipfs-github-action

**IPFS ACTION IN ACTION**
Screenshot 2020-07-13 at 15 51 15

This will give us a github status entry on each commit that links to a rendered build, published to IPFS and pinned on cluster.ipfs.io

I've added the required `CLUSTER_USER` & `CLUSTER_PASSWORD` secrets to the repo.

Fixes: #581 

License: MIT

### DIFF
--- a/.github/workflows/add-to-ipfs.yml
+++ b/.github/workflows/add-to-ipfs.yml
@@ -1,21 +1,14 @@
-name: github pages deploy
+name: Add to IPFS 
 
-on:
-  push:
-    branches:
-      - master
+on: [push]
 
 jobs:
-  build-deploy:
+  add-to-ipfs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
         with:
           submodules: true
-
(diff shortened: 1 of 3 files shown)


olizilla:
@hugomrdias pointed out that this needs to build on PRs, but the exiting workflow is only configured for building on change to master, so am fixing that.

olizilla:
@hugomrdias it lives! IPFS preview on PR. The `Details` link takes you to the rendered site on IPFS.

Screenshot 2020-07-14 at 10 23 11

hugomrdias:
Nice thank you!! Can we make the github check link the actual http ipfs url ?

olizilla:
It already do!

PR #981, opened 2020-07-16
feat: simplified workflow to add next to ipfs

olizilla:
Adds a workflow to add the new site to IPFS. Updates the build process, for the new, much simpler flow. ❤️ 

our workflow has a lot of manual steps to ensure go etc is up to date... github provides a set of `setup-*` actions to ensure certain versions are installed, so lets use them!

License: MIT

### DIFF
--- a/.github/workflows/add-next-to-ipfs.yml
+++ b/.github/workflows/add-next-to-ipfs.yml
@@ -1,0 +1,31 @@
+name: Add to IPFS 
+
+on: [push]
+
+jobs:
+  add-next-to-ipfs:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/add-next-to-ipfs.yml
```diff
@@ -0,0 +1,31 @@
...
+
+on: [push]
+
+jobs:
+  add-next-to-ipfs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
```
hugomrdias, line 11:
```suggestion
```
We don't need this anymore it's all go modules.

PR #995, opened 2020-07-20
fix: link fixes in specs actors

yiannisbot:
This PR fixes some broken links to old specs-actors documentation in several files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,19 @@ You can find examples in the `config.toml`
     source = "actors"
     target = "content/modules/actors"
 ```
-This makes files from external repos available for Hugo rendering.  
+This makes files from external repos available for Hugo rendering and allows for linking to up-to-date sites that are directly pulled from other repositories.
+
+The configuration above gives the following information:
+
+- `path`: gives the repository you want to mount content from.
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -40,7 +40,19 @@ You can find examples in the `config.toml`
     source = "actors"
     target = "content/modules/actors"
 ```
-This makes files from external repos available for Hugo rendering.  
+This makes files from external repos available for Hugo rendering and allows for linking to up-to-date sites that are directly pulled from other repositories.
```
hugomrdias, line 43:
```suggestion
This makes files from external repos available for Hugo rendering and allows for linking to up-to-date files that are directly pulled from other repositories.
```

README.md
```diff
@@ -40,7 +40,19 @@ You can find examples in the `config.toml`
...
+
+The configuration above gives the following information:
+
+- `path`: gives the repository you want to mount content from.
+- `source`: the folder in the mounted repository on which you mount the the local Hugo site. This is the "root" seen from your Hugo site where you pull content from. In the above case, this means that the source on the mounted repository is `https://github.com/filecoin-project/specs-actors/actors/`.
```
hugomrdias, line 48:
```suggestion
- `source`: the folder from the repository referenced in the `path` that we want to mount into  our local Hugo filesystem. This is the "root" seen from your Hugo site where you pull content from. In the above case, this means that the source that will be mounted is `https://github.com/filecoin-project/specs-actors/actors/`.
```

PR #1000, opened 2020-07-21
fix: bring back download filecoin pdf link

### LINKED ISSUE #990: Add a download/external link for PDF
ref: https://github.com/filecoin-project/specs/pull/969#issuecomment-660004572

olizilla:
Fixes #990

License: MIT

### DIFF
--- a/content/algorithms/porep/_index.md
+++ b/content/algorithms/porep/_index.md
@@ -7,4 +7,6 @@ weight: 2
 # Proof-of-Replication
 ---
 
+See the [Filecoin Paper](https://filecoin.io/filecoin.pdf)
+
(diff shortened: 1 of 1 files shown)


olizilla:
@hugomrdias that fallback seems like the right thing for the shortcode to do. I'm more inclinded to change the shortcode to **always add a download link for an embeded pdf**.

Alternatively, we can just remove the embed... No one really wants to read a pdf embedded in the middle column of a web page.

hugomrdias:
Yeah you are right, I'm good with either.

olizilla:
@hugomrdias have replaced the pdf embed with a short intro taken from the abstract of https://filecoin.io/proof-of-replication.pdf

PR #1009, opened 2020-07-23
feat: use cid-in-subdomain IPFS preview urls

### LINKED ISSUE #993: Search isn't working through the gateway
check https://beta.spec.filecoin.io/ 
Screenshot 2020-07-20 at 12 40 29

### LINKED ISSUE #1007: search and service worker js not loaded from relative url
The service worker script sw.js and the flexsearch.js are both loaded dynamically, and the paths are stored in places that hugo cannot re-write to be relative urls...

Screenshot 2020-07-23 at 10 24 17

- sw.js uses a custom element and stores the path 
https://github.com/filecoin-project/specs/blob/71f37208a1f4f56b33ea307d7cbdb4b06996b115/layouts/partials/docs/inject/body.html#L2
- flexsearch.js is loaded dynamically by js
https://github.com/alex-shpak/hugo-book/blob/f416a71bf8bb77e7af9ade9eea7fe3b662962290/assets/search.js#L49-L50

the `relUrl` filter in hugo is not sufficient to get fully relative paths across the site. Hugos global `relativeUrls` parameter re-writes all URLs in standard src and href attributes after the site has been built. URLs stored in JS or custom elements aren't picked up.

The simple fix for this is to upate the preview URLs to use cid-in-subdomain, which is now the preferred mechanism. The service worker is not critical to have working in preview urls, and the flexsearch will be disabled in favour of #984

olizilla:
Upgrade to ipfs-github-action@v2 to get cid-in-subdomain preview urls

Fixes #1007 

License: MIT

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
 
       # Pin the built site to ipfs-cluster, output the cid as `steps.ipfs.outputs.cid`
       # see: https://github.com/ipfs-shipyard/ipfs-github-action
-      - uses: ipfs-shipyard/ipfs-github-action@use-cid-in-subdomain-preview-url
+      - uses: ipfs-shipyard/ipfs-github-action@v2.0.0
         id: ipfs
         with:
           path_to_add: public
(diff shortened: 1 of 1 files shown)


olizilla:
YES! IT BLENDS! I've updated the github action to use cid-in-subdomain preview urls. can comfirm that the service worker loaded and informed me that that page could be offlined, and the local search feature works.

Screenshot 2020-07-23 at 11 05 22

hugomrdias:
is this ready for review ?

olizilla:
@hugomrdias yep, the bulk of the work is over in ipfs-github-action https://github.com/ipfs-shipyard/ipfs-github-action/pull/3

PR #1016, opened 2020-07-25
fix(content): update filecoin piece content

yiannisbot:
This PR is updating the description of the Filecoin Piece and provides a description of Data Representations in the Filecoin Network. Although the section will likely go through some restructuring, with this PR I would like to make sure that the content in this subsection is correct and up-to-date.


content/systems/filecoin_files/piece/_index.md
```diff
@@ -8,20 +8,36 @@ dashboardAudit: 0
...
+Piece data structures.
 
 
 {{< figure src="pieces.png" title="Pieces, Proving Trees, and Piece Data Structures" zoom="true">}}
 
+## Data Representation in the Filecoin Network
+
+It is important to highlight that data submitted to the Filecoin network go through several transformations before they come to the format at which the `StorageProvider` stores it.
+
+1. When a piece of data, or file is submitted to Filecoin (in some raw system format) it is transformed into a _UnixFS DAG style data representation_ (in case it is not in this format already, e.g., from IPFS-based applications). The hash that represents the root of the IPLD DAG of the UnixFS file is the _Payload CID_, which is used in the Retrieval Market. The Payload CID is identical to an IPFS CID.
```
rvagg, line 32: I think this might be how Lotus does it when you import something via a CLI, but I don't think it's a strict system requirement that UnixFS get involved with this. For Filecoin Discover we are also importing some arbitrary IPLD blocks for some of the content that won't be interesting when viewed via UnixFS (i.e. the blockchain contents). From a user-perspective I guess mostly they'll see things as UnixFS but the system just needs data that can get bundled up into a CAR file.

content/systems/filecoin_files/piece/_index.md
```diff
@@ -8,20 +8,36 @@ dashboardAudit: 0
...
+2. In order to make a _Filecoin Piece_ the UnixFS IPLD DAG is serialised into a ["Content-Addressable aRchive" (.car)](https://github.com/ipld/specs/blob/master/block-layer/content-addressable-archives.md#summary) file, which is in raw bytes format.
+3. The resulting .car file is _padded_ with extra bits.
```
rvagg, line 34: Perhaps it's best to leave this vague and simple, but if it helps I tried to outline this (mainly in the context of how we get to CIDs and what those CIDs indicate) the other day @ https://github.com/ipfs-shipyard/ipld-explorer/issues/64#issuecomment-663401567

The main take-away for padding is that we need to get to a ^2 size so we can do a clean binary merkle. Before getting to that ^2, an extra 2 zero bits per 254 bits need to be added. So we take the original input, add zero bytes to the end of it until we reach exactly the size that we'll need in order to pad it out with those 2-zeros per 254 bits and end up with a ^2. Kind of tricky to explain when you have to work backward like that. Something like: zero-pad up to the `254/256` of the next ^2 value, then do the 2-bits-per-254 on top of that.

content/systems/filecoin_files/piece/_index.md
rvagg: "identical to an IPFS CID" isn't really correct. It's a CID in the way that IPFS has CIDs but it's very filecoin specific and now useful outside of filecoin. (another thing I outlined in https://github.com/ipfs-shipyard/ipld-explorer/issues/64#issuecomment-663401567 if it helps)

content/systems/filecoin_files/piece/_index.md
```diff
@@ -8,20 +8,36 @@ dashboardAudit: 0
...
+4. The next step is to calculate the Merkle root out of the hashes of the Piece. The resulting root of the Merkle tree is the **Piece CID**. This is also referred to as _CommP_ or _Piece Commitment_.
+5. At this point, the Piece is included in a Sector together with data from other deals. The `StorageProvider` then calculates Merkle root for all the Pieces inside the Sector. The root of this tree is _CommD_ (aka _Commitment of Data_ or `UnsealedSectorCID`).
+6. The `StorageProvider` is then sealing the sector and the root of the resulting Merkle root is the _CommR_ (or _Commitment of Replication_).
```
porcuquine, line 37: This is not really accurate according to how we prefer to present this.

There are two trees, each with a merkle root. One root is CommC. This is needed for PoRep. The other is CommRLast. This is needed for PoST. CommRLast is the merkle root of the replica.

CommR is a hash of CommC || CommRLast.

If you want to call CommR the merkle root of the sealed sector, then you have to accept that the tree of which it is the root is asymmetrical, has a weird shape, and includes all the intermediate data used in proof of replication but never afterward. We have chosen not to describe it in this way so far, and I think we should continue not to.

content/systems/filecoin_files/piece/_index.md
yiannisbot: @rvagg a couple of questions on your first point:

> the system just needs data that can get bundled up into a CAR file

What are the requirements for the data so that it can be bundled in a CAR file? Can it be any format?

What is the implication of the above on the Payload CID, which is generated before the data is "CAR'ed"?  Can it be of any format?

content/systems/filecoin_files/piece/_index.md
yiannisbot: Thanks @rvagg! I thought the extra explanation will be useful so I included a short description in the commit below.

content/systems/filecoin_files/piece/_index.md
```diff
@@ -31,7 +31,7 @@ It is important to highlight that data submitted to the Filecoin network go thro
...
 2. In order to make a _Filecoin Piece_ the UnixFS IPLD DAG is serialised into a ["Content-Addressable aRchive" (.car)](https://github.com/ipld/specs/blob/master/block-layer/content-addressable-archives.md#summary) file, which is in raw bytes format.
-3. The resulting .car file is _padded_ with extra bits.
+3. The resulting .car file is _padded_ with extra bits in order to get it to "power of 2" size. This is done in order for the file to make a binary Merkle tree. This means that two zero (0) bits need to be added to every 254 bits (to make the 256 bits). In case more padding is needed in order to reach the 254 bit threshold, then these bits are also filled with zeros.
 4. The next step is to calculate the Merkle root out of the hashes of the Piece. The resulting root of the Merkle tree is the **Piece CID**. This is also referred to as _CommP_ or _Piece Commitment_.
```
porcuquine, line 35: This description isn't clear to me. The bit padding (the two zero bits after every 254 bits) don't really have anything to do with the 'power of 2' padding. They're better thought of as two distinct, sequential steps. First the bit padding is added. Then the power-of-2 padding is added.

The description as written describes the motivation for bit padding as 'in order to get it to "power of 2" size' — but that's not the case. So the text above never really gets around to describing the actual power of two padding.

I don't understand what the last sentence of 3. is trying to communicate. It seems either to be a repetition of the previous sentence's content — or else meant to be a return to the promise of power of two padding. But if the latter, it doesn't actually describe that at all.

content/systems/filecoin_files/piece/_index.md
ribasushi: > What are the requirements for the data so that it can be bundled in a CAR file? 

(please do not copy this answer into the spec, I am just explaining mechanics)

- On the protocol level car files have no `must` type requirements (for several reasons, a notable one being difficulty of enforcement). Without the presence of some sort of index, a linearized DAG if indistinguishable from random blocks put in a line one after another (the current car spec does not define such an index: you "parse as you go").
In short `.car` files are a mechanism to efficiently transport a collection of blocks.
A **subset** of this capability is transport of well formed/complete DAGs ( multiple roots, all blocks present ).
A **sub-subset** of that is transport of single-root, deterministically processed DAGs ( you start from the root, and walk the dag left-hand-first, depth-first, skipping any already seen blocks ). You will see such car files being referred to as *"deterministic car"*.

A key implementation detail linking above and below bodies of text: on the **sector/proving** level the `.car` file is an opaque blob of data. Nothing is concerned with its internal structure. You can derive `commP` for a stream of random data, in the same manner you can derive an `commP` for the perfect `.car` file. What ties the two worlds is the presence of the `payloadCid` ( the single root within the `.car` file ) in the deal info recorded in the FIlecoin blockchain.

- In the context of filecoin itself things are slightly more complicated: you request a retrieval and start paying for it incrementally as you receive data, root-first. This makes it prudent to only accept a flow you can validate: the first block is **the** payload CID, the second block must be linked-to from the first one, the third from the second, etc. This is where the determinism comes back into focus, and becomes a strict requirement. Still: there are two submodes of operation:
   - In the default `import`/`retrieve` mode the `.car` for the deal is indeed formed by employing UnixFSv1, with [non-default parameters](https://github.com/filecoin-project/lotus/blob/v0.4.2/build/params_shared_vals.go#L16-L19) and is interpreted as such on retrieval.
   - By activating the `--car` option flag on both calls you can, however, operate over deterministically-ordered `.car` files, **containing a single-root DAG of any encoding and content**.

I believe this about sums things up.

content/systems/filecoin_files/piece/_index.md
yiannisbot: @porcuquine just to make sure we're on the same page before I rephrase:

> The bit padding (the two zero bits after every 254 bits) don't really have anything to do with the 'power of 2' padding.

AFAIU they (the 2 zero bits padding) are part of it - this is the second step to reach the "^2 padding".

> They're better thought of as two distinct, sequential steps. First the bit padding is added. Then the power-of-2 padding is added.

Can you briefly describe what you define as "bit-padding" and what you define as "^2 padding"? Do you agree with the definition/description that @rvagg gives [above](https://github.com/filecoin-project/specs/pull/1016#discussion_r460644727)?

rvagg: OK, so @porcuquine is almost certainly a better way of explaining it to communicate the concepts better! In code, the steps are:

1. find the next `^2 x 254/256` size above the input size
2. pad out the input to that size
3. pass through fr32 padding to add those extra 2 bits per 254 bytes

That first step is the most difficult to describe because it involves multiple parts and is just weird.

_BUT_ because we're dealing with zeros here, it can be flipped and the fr32 can be done beforehand, this code flow is probably just an artifact of the various streaming pieces or some legacy. The fr32 padding needs to be done across the entire piece, but since the byte padding is zero bytes anyway there's no good reason to require it be done as a last part of the step.

1. fr32 pad, adding 2 bits per 254 bits
2. figure out the next ^2 greater than the size after step 1
3. add zero bytes to pad input out to that length

content/systems/filecoin_files/piece/_index.md
rvagg: > have no must type requirements

well .... they _must_ be IPLD blocks, even if they are just `raw` bytes with an associated CID. That's important to communicate but maybe it's covered enough by linking to the CAR spec.

_(Warning: the following is my basic understanding, I'm **far** from expert in this and the world may have changed since I grasped even this tenuous understanding!)_ When making an online deal (as opposed to the less common offline ones we're focused on currently), I think graphsync gets involved in this process. Locally you do this process of building a CAR from your blocks to generate a local CommP for your deal, then you make the deal and use graphsync to transfer the IPLD blocks to the miner who then builds the CAR again from those blocks and gets to the same CommP.

In this flow, you'd get "deterministic" from the way that the CAR is built from the graphsync output just by default. I don't know how common these offline deals will be but we're being pretty liberal with our CAR layouts for it because there's no need for the CAR to be built the same on both sides, we're building them and then shipping them to the miner as they are.

There are implications for retrieval as @ribasushi said, I don't really understand those. I'm also not sure what the focus of this doc should be and whether it should cover both of these cases. I suspect online deals will be the most common. The client still needs to generate CommP from their data. I suppose a client like lotus will handle all of this process for you and it'll end up being responsible for building the graph of your data—and the most common will be file data which lotus will handle the UnixFSification for you.

wrt CAR files being an opaque blob—I don't know if you can avoid them entirely for anything but offline deals, but maybe that's worth noting too?

content/systems/filecoin_files/piece/_index.md
porcuquine: Yes, what @rvagg wrote above is correct. Sorry for the late response.

What I was describing corresponds to his second set of three steps. I think it's much simpler to understand.
That said, @rvagg is correct that the more complicated case is what is actually implemented, and  `rust-fil-proofs` does check the size before adding the bit-padding (by which I mean 2 zero bits after every 254 of raw input).

It's my belief that the spec only has to and only *should* describe the result — not the steps followed to produce it. So I would favor the simpler explanation, even though it doesn't correspond to the work flow imposed by the `rust-fil-proofs` API.

If you do want to use the other set of three steps (his original ones), then the description will need to find a way to articulate the hard-to-describe step 1.

Also missing entirely here is the minimum piece size, which is defined to be 127 bytes before bit padding (or 128 bytes after bit padding). For gory detail, see: https://github.com/filecoin-project/rust-fil-proofs/issues/1231

content/systems/filecoin_files/piece/_index.md
yiannisbot: I have updated the description to go with the suggestion. Thanks!

content/systems/filecoin_files/piece/_index.md
yiannisbot: Ok, I now got it! Thanks for the explanations! I agree that:

> the spec only has to and only should describe the result — not the steps followed to produce it

and therefore I'm keen to go back and just have that: _"The resulting .car file is _padded_ with extra bits in order to get it to "power of 2" size. This is done in order for the file to make a binary Merkle tree."_, which is the essence of it.

Are we all ok with that @porcuquine @rvagg?

If we want the process too, then this is the simplest explanation IMO:

> 1. fr32 pad, adding 2 bits per 254 bits
> 2. figure out the next ^2 greater than the size after step 1
> 3. add zero bytes to pad input out to that length

rvagg: I think you're a better judge of the audience here so should decide what level of detail to include. "in order for the file to make a binary Merkle tree" is only correct for one stage of padding, the fr32 is something else all-together. I'm not sure for this audience how much it matters that "your input padded internally and at the length such that it doesn't quite resemble your input before it is merklized". For now maybe it's fine as is, implementers and anyone else wanting to generate CommP separate from a full client (like we've had to do for these offline deals for Filecoin Discover data) will have to come to a more sophisticated understanding of the process, and this might not be the place for them to get that detail! Your call @yiannisbot, I think you have most of the details now.

porcuquine: >and therefore I'm keen to go back and just have that: "The resulting .car file is padded with extra bits in order to get it to "power of 2" size. This is done in order for the file to make a binary Merkle tree.", which is the essence of it.

I understand the impulse, but this description does not provide enough information to know *what* must be done regardless of the order of steps — so it's not equivalent.

> If we want the process too, then this is the simplest explanation IMO:

>> fr32 pad, adding 2 bits per 254 bits
>> figure out the next ^2 greater than the size after step 1
>> add zero bytes to pad input out to that length

I agree. I think you could use that, or a similar description, perhaps with some framing to make clear that the steps are descriptive of the result, not prescriptive of the algorithm/API which must be used/provided.

rvagg:
Hey, while we're doing this, if we end up with the longer description of _how to get to CommP_, how about we take the opportunity to insert a definition of "Fr32" into the specs because it's going to keep getting asked. @ribasushi was the latest one to ask just 2 days ago. Combining and rephrasing parts of @porcuquine's responses we could come up with something like:

> The term `Fr32` is derived from the name of a struct Filecoin uses to represent the elements of the arithmetic field of a pairing-friendly curve, specifically Bls12-381—which justifies use of 32 bytes. `F` stands for "Field", while `r` is simply a mathematic letter-as-variable substitution used to denote the modulus of this particular field.

content/systems/filecoin_files/piece/_index.md
```diff
@@ -3,24 +3,62 @@ title: Piece
...
+
+2. In order to make a _Filecoin Piece_, the IPLD DAG is serialised into a ["Content-Addressable aRchive" (.car)](https://github.com/ipld/specs/blob/master/block-layer/content-addressable-archives.md#summary) file, which is in raw bytes format. A CAR file is an opaque blob of data that packs together and transfers IPLD nodes. The _Payload CID_ is common between the CAR'ed and un-CAR'ed constructions. This helps later during data retrieval, when data is transferred between the storage  client and the storage provider as we discuss later.
+
+3. The resulting .car file is _padded_ with extra zero bits in order for the file to make a binary Merkle tree. To achieve a clean binary Merkle Tree the the .car file size has to be in some power of two (^2) size. Therefore, the padding process takes the input file and finds the size above the input size that makes for a power of two size. The gap between the input size and the next power of two size is padded with zeros. Finally, an extra padding process is applied, called `fr32 padding` which adds two (2) zero bits to every 254 bits (to make 256 bits in total).
```
rvagg, line 38: It looks like this para has gone through some iterations and lost the initial ordering in an attempt to make it easier to explain - which I understand, but it's now not quite accurate. The _final_ size needs to be a ^2 but the fr32 padding gets in there before it goes to ^2. So to calculate how much zero padding to add, you have to account for the 254/256 inflation that's going to occur.

I think in an earlier iteration it was suggested to switch the order of the steps and say that fr32 happens first and then inflation to ^2 after that. I think this is the ideal framing because it makes it easier to understand (even though in code it happens the other way around, it doesn't actually matter so it's worth doing what's best for descriptive purposes). So that would mean pulling the last sentence to be first, "A process called `fr32 padding` which adds ... is applied. Extra padding is applied to the end of the data, adding zero bytes until the resulting blob is a power of two sie.". Something like that.

content/systems/filecoin_files/piece/_index.md
```diff
@@ -3,24 +3,62 @@ title: Piece
...
 
+## Data Representation
+
+It is important to highlight that data submitted to the Filecoin network go through several transformations before they come to the format at which the `StorageProvider` stores it.
+
+Below is the process followed from the point a user starts preparing a file to store in Filecoin to the point that the provider produces all the identifiers of Pieces stored in a Sector.
+
+The first three steps take place on the client side.
+
+1. When a client wants to store a file in the Filecoin network, they start by producing the IPLD DAG of the file. The hash that represents root node of the DAG is an IPFS-style CID, called _Payload CID_.
```
rvagg, line 34: It might be worth synchronizing items 1 and 2 with this conversation https://github.com/filecoin-project/specs/pull/1151#discussion_r488746195 about how the whole IPLD+CAR frame isn't strictly a requirement, it's just how Lotus works currently and how the retrieval semantics work, but I don't believe any of this is strictly built into the piece requirements. Maybe that just means softening the language so it suggest the common path, not the required path.

content/systems/filecoin_files/piece/_index.md
yiannisbot: Good catch @rvagg! I've tried to explain it both ways, but indeed it becomes tricky, you're right. I've updated in commit https://github.com/filecoin-project/specs/pull/1016/commits/49a025b8da9309f3e30fad375d27f56659535a49 and included only the easier to grasp process.

content/systems/filecoin_files/piece/_index.md
yiannisbot: @rvagg that's a good point. However, I don't have a good way of phrasing this without being inaccurate. Instead, I have added the following note in order to clarify the point you're raising (in commit: https://github.com/filecoin-project/specs/pull/1016/commits/7db90517236d1ed4315981cb45b849a1cdb881fb), but will be more difficult to describe it in steps. Let me know if you think this works, or otherwise, could you provide some text include?

"- Steps 2 and 3 above are specific to the Lotus implementation. The same outcome can be achieved in different ways. However, any implementation has to make sure that the initial IPLD DAG is serialised and padded so that calculating the Merkle root out of the resulting blob of data gives the same **Piece CID**. As long as this is the case, implementations can deviate from the first three steps above."

content/systems/filecoin_files/piece/_index.md
yiannisbot: Forgot to mention that I've removed the definition of `fr32` from here and included it in the Glossary ([outstanding PR](https://github.com/filecoin-project/specs/pull/1151) atm).

rvagg (dismissed):
sgtm, at least the bits I've commented on, some of the rest are out of my scope

content/systems/filecoin_files/piece/_index.md
porcuquine: I'm just noticing this now. I'm commenting here without an entirely clear point, but so it doesn't get lost in the context shift:

`Fr32` is well-defined in the `rust-fil-proofs` implementation. As far as I can tell from the glossary, it's not given a substantive definition there. It's use here as an adjective was never particularly well-defined (from what I saw), although some true statements were made justifying the adjective.

Without making a specific suggestion for how to untangle this, let me offer the definition underlying the _type_ called `Fr32` in `rust-fil-proofs`. `Fr32` is a 32-bit representation of a field element (which, in our case, is now, is the arithmetic field of BLS12-381). To be well-formed, a value of type `Fr32` must _actually_ fit within that field, but this is not enforced by the type system. It is an invariant which must be perserved by correct usage.

In the case of so-called `Fr32` padding, two zero bits are inserted 'after' a number requiring at most 254 bits to represent. This guarantees that the result will be `Fr32`, regardless of the value of the initial 254 bits. This is a 'conservative' technique, since for some initial values, only one bit of zero-padding would actually be required.

It's possible that none of this needs to be spelled out explicitly here. It's not clear that an 'official' definition of the term makes much sense without at least some of this context, though. I've put this comment here rather than there because it's in the context of _this_ conversation that the added detail is obviously relevant.

porcuquine: I should note that `Fr32` is not obviously a concept which belongs in the spec, even though it seems to have made its way into the preferred name for what I historically called 'bit padding'. I don't mind if it does stay in the spec as a glossary term. I mention this just to clarify that its primary definition is as an implementation type with relatively elaborate characteristics.

In particular, it's important to understand that `Fr32`-padded data represents a smaller set of values than inhabit the `Fr32` type.

content/systems/filecoin_files/piece/_index.md
yiannisbot: Summary of above discussion has been integrated in latest commits.

content/systems/filecoin_files/piece/_index.md
yiannisbot: @porcuquine this is useful input. I have added your definition above as an extra "important note" in this section (see commit https://github.com/filecoin-project/specs/pull/1016/commits/972763d536936368db98668d3021c7afd4f48aca) and will also add this as a definition in the glossary too, if you don't mind.

> I should note that Fr32 is not obviously a concept which belongs in the spec

I generally tend to agree. However, if we don't mention anything specific to `Fr32`, then we should just mention what is in one of the notes: _"any implementation has to make sure that the initial IPLD DAG is serialised and padded so that it gives a clean binary tree, and therefore, calculating the Merkle root out of the resulting blob of data gives the same **Piece CID**"_, and leave it to that, without any mention of padding, .car etc.

I suggest leaving the text as is for now, given we mention that all this is Lotus-specific and the alternative would be slightly confusing to describe. Let me know if you're ok with this.

content/systems/filecoin_files/piece/_index.md
```diff
@@ -53,7 +54,9 @@ The following steps take place on the `StorageProvider` side (apart from step 4
...
+- `Fr32` is a 32-bit representation of a field element (which, in our case, is the arithmetic field of BLS12-381). To be well-formed, a value of type `Fr32` must _actually_ fit within that field, but this is not enforced by the type system. It is an invariant which must be perserved by correct usage.
+In the case of so-called `Fr32 padding`, two zero bits are inserted 'after' a number requiring at most 254 bits to represent. This guarantees that the result will be `Fr32`, regardless of the value of the initial 254 bits. This is a 'conservative' technique, since for some initial values, only one bit of zero-padding would actually be required.
+- Steps 2 and 3 above are specific to the Lotus implementation. The same outcome can be achieved in different ways, e.g., without using `Fr32` bit-padding. However, any implementation has to make sure that the initial IPLD DAG is serialised and padded so that it gives a clean binary tree, and therefore, calculating the Merkle root out of the resulting blob of data gives the same **Piece CID**. As long as this is the case, implementations can deviate from the first three steps above.
```
porcuquine, line 59: Two things:

1. I do think the padding method is prescribed. Otherwise there will not be a deterministic mapping from raw data to CommP. It's just that this padding method is not the only possible way to convert from raw data into a sequence of `Fr32` elements.

2. I don't think anything about `Fr32` is *Lotus*-specific: it's a `rust-fil-proofs` implementation detail.

I don't make these suggestions to complicate things. Probably incorporating them will lead to a simpler explanation. The current text is still not quite accurate.

yiannisbot: @porcuquine can you please provide the right phrasing for this, so that we finalise?

hugomrdias (approved):

PR #1034, opened 2020-07-30
feat: add algolia docsearch

### LINKED ISSUE #984: Add algolia search
Somewhat prescriptive this one, but we've had a good experience with the UX of Algolia search in the past, and it's not too much work to add, and could strike a balance of still being able to find everything quickly without needed to load MBs of HTML.

olizilla:
Integrate Algolia DocSearch into the site as per https://docsearch.algolia.com/docs/dropdown, along with UI fixes required to make it work... we have to pull the search bar out of the scrollable area, otherwise the search results are cropped / appear under the main body content. As such this pulls the header and search bar out of the nav, so it's just the TOC that scrolls.

Docsearch is like algolia premium. We dont have to set up algoila indexing at all, just add a js script to the page and they do all the scraping. Without that, we have to import our content to their servers and set up an action to re-import it when the main branch changes. With docsearch they scrape the deployed site for for us every 24 hours. For the curious, they maintain a public repo with the > 2k other docs sites that they index: https://github.com/algolia/docsearch-configs/tree/master/configs

**typo proof search results**
Screenshot 2020-07-30 at 16 23 05

TODO:
- [x] wire up docsearch
- [x] fix UI
- [x] get algolia indexing beta.spec.filecoin.io and update the index name and api key
- [ ] improve narrow screen layout

Fixes #984 

License: MIT


olizilla:
This has been updated to work with single page mode, but it will return results that link to the multipage site until algolia's next scrape which should be sometime today. (they scrape beta.spec.filecoin.io every 24hrs, and single-page mode got merged yesterday)

ianconsolata:
@olizilla this is more than 3 years old now -- is this PR still relevant, or can I close it?

PR #1044, opened 2020-08-05
Filecoin Message Pool Update

yiannisbot:
This PR includes updates on the Filecoin Message Pool.

### DIFF
--- a/content/systems/filecoin_blockchain/message_pool/_index.md
+++ b/content/systems/filecoin_blockchain/message_pool/_index.md
@@ -10,28 +10,16 @@ dashboardTests: 0
 
 # Message Pool
 ---
-The Message Pool, or `mpool` or `mempool` is a Memory Pool of Messages in the Filecoin blockchain system. It acts as the interface between Filecoin nodes (of any type) and the peer-to-peer network of other nodes used for off-chain message propagation. The message pool is used by nodes to maintain a set of messages to transmit to the Filecoin VM (for "on-chain" execution).
+The Message Pool, or `mpool` or `mempool` is a Memory Pool of _Transaction_ Messages in the Filecoin blockchain system. It acts as the interface between Filecoin nodes (of any type) and the peer-to-peer network of other nodes used for off-chain message propagation. The message pool is used by nodes to maintain a set of messages to transmit to the Filecoin VM (for "on-chain" execution).
 
-In order for a block message to end up in the blockchain it first has to be in the message pool. In reality, at least in the Lotus implementation of Filecoin, there is no central pool of messages stored somewhere. Instead, the message pool is an abstraction and is realised as a list of messages kept by every node in the network. Therefore, when a node puts a new message in the message pool, this message is propagated to the rest of the network using libp2p's pubsub protocol, GossipSub. Nodes need to subscribe to the corresponding pubsub topic in order to receive messages.
+In order for a transaction message to end up in the blockchain it first has to be in the message pool. In reality, at least in the Lotus implementation of Filecoin, there is no central pool of messages stored somewhere. Instead, the message pool is an abstraction and is realised as a list of messages kept by every node in the network. Therefore, when a node puts a new message in the message pool, this message is propagated to the rest of the network using libp2p's pubsub protocol, GossipSub. Nodes need to subscribe to the corresponding pubsub topic in order to receive messages.
 
 Message propagation using GossipSub does not happen immediately and therefore, there is some lag before message pools at different nodes can be in sync. In practice, and given continuous streams of messages being added to the message pool and the delay to propagate messages, the message pool is never synchronised across all nodes in the network. This is not a deficiency of the system, as the message pool does not _need_ to be synchronized across the network. 
 
 
 {{< embed src="message_pool_subsystem.id" lang="go" >}}
 
-The node types that use the message pool subsystem, as well as the specific reason they are using it for are as follows:
-
-- storage market provider and client nodes: for transmission of deals on chain,
-- storage miner nodes: for transmission of PoSts, sector commitments, deals, and other operations tracked on chain,
-- verifier nodes: for transmission of potential faults on chain,
-- relayer nodes: for forwarding and discarding messages appropriately.
-
 The message pool should have a maximum size defined to avoid DoS attacks, where nodes are spammed and run out of memory. The recommended size for the message pool is 5000 Transaction messages.
-
-The message pool subsystem is made of two components:
-
-- The [Message Syncer](message_syncer.md) -- which receives and propagates messages.
-- [Message Storage](message_storage.md) -- which caches messages according to a given policy.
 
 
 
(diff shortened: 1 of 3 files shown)


content/systems/filecoin_blockchain/message_pool/message_syncer.md
```diff
@@ -1,34 +1,32 @@
...
 dashboardAudit: 0
 dashboardTests: 0
 ---
 
-# Message Syncer
+# Message Propagation
 ---
 
-Transaction messages are sorted in the `mpool` of miners as they arrive according to cryptoeconomic rules followed by the miner.
+The message pool has to interface with the libp2p pubsub [GossipSub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) protocol. This is because transaction messages are propagated over [GossipSub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) _topics_. Every serialised `SignedMessage` is announced (see [Message](message)) in the corresponding topic by any node participating in the network.
```
schomatis, line 13: `Every serialised `SignedMessage` is announced (see [Message](message)) in the corresponding topic by any node participating in the network.`

There might be too much going on on this sentence. Maybe we can drop the serialization details and the struct name and focus more on the topic (maybe even naming it fully).

schomatis: (Also, minor detail, mentioning "topics" in plural might be confusing. I get that we're referring to a generic description, but it's a single topic, and using a generic unnamed plural forces later the use of the "corresponding" qualifier that doesn't add much more information because, for example, I don't know if there is a corresponding topic for different types of messages, which is not the case.)

content/systems/filecoin_blockchain/message_pool/message_syncer.md
```diff
@@ -1,34 +1,32 @@
...
 
-Block messages are ordered as they arrive at miner nodes in order to extend the blockchain.
-
-
-## Message Propagation
-
-Messages are propagated over libp2p pubsub topics, using the [GossipSub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) protocol. Every serialised `SignedMessage` is announced (see [Message](message)) in the corresponding topic by any node participating in the network.
-
-There are two pubsub topics related to the mempool: i) the `/fil/msgs/` topic that carries transaction messages and, ii) the `/fil/blocks/` topic that carries block messages. The process is as follows:
+There are two main pubsub topics related to transaction and block messages: i) the `/fil/msgs/` topic that carries transaction messages and, ii) the `/fil/blocks/` topic that carries block messages. The `/fil/msgs/` topic is linked to the `mpool`. The process is as follows:
```
schomatis, line 15:
```suggestion
There are two main pubsub topics related to transactions and blocks: i) the `/fil/msgs/` topic that carries transactions and, ii) the `/fil/blocks/` topic that carries blocks. The `/fil/msgs/` topic is linked to the `mpool`. The process is as follows:
```
I assume throughout we are talking about Filecoin (not libp2p) messages. Otherwise if we are referring to the two types of message in the same sentence without proper qualification it can get rather confusing. This suggested edit goes in the line of simplification by removing the mention of libp2p messages, we send Filecoin blocks and messages over the network (without the container "message" noun, how libp2p sees them is beyond this spec).

content/systems/filecoin_blockchain/message_pool/message_syncer.md
```diff
@@ -1,34 +1,32 @@
...
+Nodes must check that incoming transaction messages are valid, that is, that they have a valid signature. If the message is not valid it should be dropped and must not be forwarded.
 
 The updated, hardened version of the GossipSub protocol includes a number of attack mitigation strategies. For instance, when a node receives an invalid message it assigns a negative _score_ to the sender peer. Peer scores are not shared with other nodes, but are rather kept locally by every peer for all other peers it is interacting with. If a peer's score drops below a threshold it is excluded from the scoring peer's mesh. We discuss more details on these settings in the GossipSub section. The full details can be found in the [GossipSub Specification](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub).
 
+NOTES:
+- _Fund Checking:_ It is important to note that the `mpool` logic is not checking whether there are enough funds in the account of the transaction message issuer. This is checked by the miner before including a transaction message in a block.
```
schomatis, line 26: I'm not sure this is accurate, need to double check the code. We may do some basic funds checking in the pool.

content/systems/filecoin_blockchain/message_pool/message_syncer.md
```diff
@@ -1,34 +1,32 @@
...
+- _Message Syncing:_ Transaction messages are sorted in the `mpool` of miners as they arrive according to cryptoeconomic rules followed by the miner and in order for the miner to compose the next block.
```
schomatis, line 27: As discussed, I think we should drop the expression "message syncing" and reserve the _sync_ term only for blocks and in the context of consensus, not the pool.

content/systems/filecoin_blockchain/message_pool/message_syncer.md
yiannisbot: Hmm.. I'm fine with those edits, but I'm not sure I get why there's confusion. In both cases they're messages. Filecoin will use them as Filecoin messages and do whatever it needs to do and libp2p/gossipsub will again get them as messages that need to be propagated and do its own part of the magic.

content/systems/filecoin_blockchain/message_pool/message_syncer.md
yiannisbot: How about message sorting?

content/systems/filecoin_blockchain/message_pool/message_syncer.md
yiannisbot: Done! Good catch on the plural "topics" - corrected.

content/systems/filecoin_blockchain/message_pool/_index.md
```diff
@@ -10,28 +10,16 @@ dashboardTests: 0
 
 # Message Pool
 ---
-The Message Pool, or `mpool` or `mempool` is a Memory Pool of Messages in the Filecoin blockchain system. It acts as the interface between Filecoin nodes (of any type) and the peer-to-peer network of other nodes used for off-chain message propagation. The message pool is used by nodes to maintain a set of messages to transmit to the Filecoin VM (for "on-chain" execution).
+The Message Pool, or `mpool` or `mempool` is a Memory Pool of _Transaction_ Messages in the Filecoin blockchain system. It acts as the interface between Filecoin nodes (of any type) and the peer-to-peer network of other nodes used for off-chain message propagation. The message pool is used by nodes to maintain a set of messages to transmit to the Filecoin VM (for "on-chain" execution).
```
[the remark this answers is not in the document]
yiannisbot: Ok, I've excluded this sentence altogether and extended the following sentence.

content/systems/filecoin_blockchain/message_pool/message_syncer.md
schomatis: Sounds good.

PR #1046, opened 2020-08-06
GossipSub major spec update

yiannisbot:
This PR is updating the spec of the GossipSub protocol, which went through a major redesign recently. It adds details on how the mesh is formed, how gossip is emitted, as well as details on the hardening extensions (scoring function and mitigation strategies).

GossipSub has got its own complete and up-to-date specification, which is linked from the Filecoin Spec that this PR is updating, hence, the details on parameter values and settings have not been included here. Instead, the reader is referred to the main spec that lives in the libp2p repo.


content/algorithms/gossip_sub.md
```diff
@@ -1,21 +1,90 @@
...
+The topic parameters are implemented using counters maintained internally by the router whenever an event of interest occurs. The counters _decay_ periodically so that their values are not continuously increasing and ensure that a large positive or negative score isn't sticky for the lifetime of the peer.
+
+The decay interval is configurable by the application, with shorter intervals resulting in faster decay.
+
+Each decaying parameter can have its own decay _factor_, which is a configurable parameter that controls how much the parameter will decay during each decay period.
+
+For more details on the parameter calculation, please refer to the [GossipSub Specification](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md)
+
+
+### GossipSub Mitigation Strategies
```
vyzo, line 84: this is coming out of the blue a bit, maybe retitle this section and/or add an introductory paragraph.

yiannisbot: Good point @vyzo! I've added a brief description at the latest commit.

content/algorithms/gossip_sub.md
```diff
@@ -1,21 +1,92 @@
...
 
-Further more, every full node must implement and offer the bitswap protocol and provide all Cid Referenced objects, it knows of, through it. This allows any node to fetch missing pieces (e.g. `Message`) from any node it is connected to. However, the node should fan out these requests to multiple nodes and not bombard any single node with too many requests at a time. A node may implement throttling and DDoS protection to prevent such a bombardment.
+GossipSub is a gossip-based pubsub protocol that is utilising two types of links to propagate messages: i) _mesh links_ that carry full messages in an _eager-push_ manner and ii) _gossip-links_ that carry message identifiers only and realise a _lazy-pull_ propagation model.
```
daviddias, line 16: Everything below this line is copy & paste from the Gossipsub spec. I propose that we link to it rather than forcing us to maintain it in two places.

daviddias (changes requested):

content/algorithms/gossip_sub.md
vyzo: I totally agree!

yiannisbot: @daviddias @vyzo I'm totally fine hitting the "Delete" button - in fact I had the same thought for the score function and score parameters, which are the most vulnerable to change and I think should be removed from here.

The rest is not from the spec - most of it is actually from the paper :) Most of it provides a high-level view which adds to the spec IMO. A more useful exercise is to choose which parts are likely to change and exclude them, or put a clear message that this is a "summary, the GossipSub spec should be checked for details".

For example:
- if the mesh links and gossip links change, then we have a different protocol so need to retire this section anyway! :) 
- pretty much the same is for control messages.
- mesh maintenance is a simple and important part of the logic of the protocol
- peer discovery should probably go as it is likely to change 
- peer scoring should go.
- mitigation strategies can stay as a summary.

Happy to go either way.

vyzo: I don't think all this detail belongs in the spec; we have a (very) comprehensive spec and a paper, we shouldn't be repeating ourselves.
So yeah, delete and link to stuff.

yiannisbot: Okay! Done in latest commit. @daviddias @vyzo please have a look.

content/algorithms/gossip_sub.md
```diff
@@ -1,21 +1,18 @@
...
 
-## Bitswap
-
-Run bitswap to fetch and serve data (such as blockdata and messages) to and from other filecoin nodes. This is used to fill in missing bits during block propagation, and also to fetch data during sync.
-
-There is not yet an official spec for bitswap, but [the protobufs](https://github.com/ipfs/go-bitswap/blob/master/message/pb/message.proto) should help in the interim.
+GossipSub includes a number of security extensions and mitigation strategies that make the protocol robust against attacks. Please refer to the [protocol's specification](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md) for details on GossipSub's design and parameter settings.
```
vyzo, line 18: let's add a link to the paper as well.

content/algorithms/gossip_sub.md
```diff
@@ -1,21 +1,18 @@
...
 dashboardTests: 0
+auditURL: https://gateway.ipfs.io/ipfs/QmWR376YyuyLewZDzaTHXGZr7quL5LB13HRFnNdSJ3CyXu/Least%20Authority%20-%20Gossipsub%20v1.1%20Final%20Audit%20Report%20%28v2%29.pdf
 ---
 
 # GossipSub
 ---
 
-Messages and block headers along side the message references are propagated using the [gossipsub libp2p pubsub router](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub). Every full node must implement and run that protocol. All pubsub messages are authenticated and must be [syntactically validated](message#message-syntax-validation) before being propagated further.
+Transaction messages and block headers alongside the message references are propagated using the [libp2p GossipSub router](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub). **Both miners and full nodes must implement and run this protocol.** All pubsub messages are authenticated and must be [syntactically validated](message#message-syntax-validation) before being propagated further. The GossipSub Specification ++ provides all the implementation details and parameter settings. Here we provide an overview of the main characteristics and design choices of GossipSub.
```
vyzo, line 14: wait the miner doesn't actually run gossipsub, it defers to the full node.

content/algorithms/gossip_sub.md
yiannisbot: @vyzo @daviddias in the latest commit (https://github.com/filecoin-project/specs/pull/1046/commits/fb1fceba8dc7a5813c873d29819ddeaeeabd4ad1) I have extended the middle paragraph to give a tiny bit more detail on the protocol. Otherwise that paragraph is very thin and does not do a good job of giving an abstract view of the protocol. If we don't want to keep the extra text in the latest commit, then the whole of the paragraph should go IMO and just point to the spec and the technical report.

content/algorithms/gossip_sub.md
yiannisbot: done!

content/algorithms/gossip_sub.md
yiannisbot: Good catch. Rephrased in latest commit (https://github.com/filecoin-project/specs/pull/1046/commits/fb1fceba8dc7a5813c873d29819ddeaeeabd4ad1).

vyzo (commented):
This looks fine to me -- I'll let @david nitpick :)

daviddias (approved):

daviddias:
One last thing. Should the "dashboardAudit: done" be pointing to the LeastAuthority Audit?

daviddias:
@yiannisbot we need to have support for multiple audits. There is an audit coming for rust-libp2p gossipsub

PR #1054, opened 2020-08-10
Major update in EC and Secret Leader Election Specification

yiannisbot:
This PR is applying all the latest changes in the Expected Consensus, Leader Election and the rationale behind Poisson Sortition of `WinCount`. It removes all mention of Election PoSt which was very out-of-date.

### DIFF
--- a/content/algorithms/expected_consensus/_index.md
+++ b/content/algorithms/expected_consensus/_index.md
@@ -2,7 +2,7 @@
 title: "Expected Consensus"
 weight: 1
 dashboardWeight: 2
-dashboardState: incomplete
+dashboardState: stable
 dashboardAudit: 1
 dashboardTests: 0
 ---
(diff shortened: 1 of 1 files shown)


content/algorithms/expected_consensus/_index.md
```diff
@@ -28,115 +28,290 @@ The [Storage Power Consensus](storage_power_consensus) subsystem uses access to
...
+  - The probability of winning one block is `1-P[X=0]`, thus if `h_n` is less than `1-P[X=0]`, the miner wins at least one block.
+  - Similarly if `h_n` is less than `1-P[X=0]-P[X=1]` we have at least two blocks and so on.
+
+#### Explanations - Poisson Sortition
+
+Filecoin is building on the principle that a miner possessing X% of network power should win as many times as X miners with 1% of network power in the election algorithm.
+
+This is modelled via a Binomial distribution with parameter`p=ExpectedLeaderPerRound/TotalPower` and `n= MinerPower`.
+
+For ease of implementation, we use the the Poisson distribution with parameter `lambda = MinerPower*ExepectedLeader` to approximate this Binomial distribution.
```
davidad, line 149: I wouldn't say Poisson is an approximation for ease of implementation. I initially thought Binomial(ExpectedLeadersPerRound, MinerPower/TotalPower) was the "correct" or ideal distribution, but after much discussion with @jzimmerman, I came to realize this: since every miner is rolling uncorrelated/independent dice, and we want to be fully invariant to miner pooling (not just in expectation), Poisson is actually the ideal distribution.

The table below might still be worth presenting, but the framing should be something like "you might think binomial is better, but it isn't, but look, it doesn't matter much anyway" rather than "our approximation is good enough".

content/algorithms/expected_consensus/_index.md
```diff
@@ -28,115 +28,290 @@ The [Storage Power Consensus](storage_power_consensus) subsystem uses access to
...
+  - Get the power of the miner at block `ElectionPowerTableLookback`
+  - Get the total network power at block `ElectionPowerTableLookback`
+- Get randomness for the current epoch using `GetRandomness`.
+- Generate a VRF and compute its hash
+  - The storage miner's `workerKey` is given as input in the VRF process
+- Compute block quality: The smaller the hash is, the higher the block quality will be
+  - Compute the probability distribution of winning `k` blocks (i.e. Poisson distribution see below for details)
+  - Let `h_n` be the normalised VRF, i.e. `h_n = vrfout/2^256`
+  - The probability of winning one block is `1-P[X=0]`, thus if `h_n` is less than `1-P[X=0]`, the miner wins at least one block.
+  - Similarly if `h_n` is less than `1-P[X=0]-P[X=1]` we have at least two blocks and so on.
```
davidad, line 141: I might add: "While it is not permitted for a single miner to publish two distinct blocks, in this case, the miner produces a single block which earns two block rewards."

content/algorithms/expected_consensus/_index.md
```diff
@@ -28,115 +28,290 @@ The [Storage Power Consensus](storage_power_consensus) subsystem uses access to
...
+
+#### Explanations - Poisson Sortition
+
+Filecoin is building on the principle that a miner possessing X% of network power should win as many times as X miners with 1% of network power in the election algorithm.
+
+This is modelled via a Binomial distribution with parameter`p=ExpectedLeaderPerRound/TotalPower` and `n= MinerPower`.
```
davidad, line 147: If you want to pick an intuitive Binomial model to compare to, IMO the most intuitive is n=ExpectedLeadersPerRound and p=MinerPower/TotalPower

PR #1055, opened 2020-08-12
Adding Randomness Subsection in Spec

yiannisbot:
This PR is adding a new subsection in the Algorithms Section to discuss the details on Randomness and the interface with DRAND. There is also discussion around DRAND Outage and what happens in the catch up phase.

### DIFF
--- a/content/algorithms/randomness.md
+++ b/content/algorithms/randomness.md
@@ -1,0 +1,136 @@
+---
+title: "Randomness in FIL"
+weight: 5
+dashboardWeight: 1
+dashboardState: stable
+dashboardAudit: 0
+dashboardTests: 0
+---
+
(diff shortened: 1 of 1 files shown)


nikkolasg:
What's the scope of this PR ? It seems like it supersedes the previous PR of @nicola  ? 
And if that's the case (which is fine by me), then there are some things missing right ?
Namely, there are two ways to "search" for the drand values: either from the chain or from the network. Both are used: the former for chain validation for example, and the latter for lottery leader election.
Also the usage of the randomness in Filecoin.

ianconsolata:
@yiannisbot Is this change still relevant? Now that the CD is fixed, I'd like to start merging this backlog of PRs

PR #1059, opened 2020-08-14
FIL VM Message Validation Update

yiannisbot:
This PR is an update to the VM Message section. It does not include description of the gas fee calculation, which might need to go here - I'm not sure about this yet. A brief description will probably be included here anyway. This will be done in a subsequent commit or separate PR.

### DIFF
--- a/content/systems/filecoin_vm/message/_index.md
+++ b/content/systems/filecoin_vm/message/_index.md
@@ -116,3 +116,6 @@ The reason for this "no message semantic validation" policy is that the state th
 be applied to cannot be known before the message is executed _as part of a tipset_. A block producer
 does not know whether another block will precede it in the tipset, thus altering the state to
 which the block's messages will apply from the declared parent state.
+
+
+{{<embed src="/modules/lotus/chain/types/message.go"  lang="go">}}
--- a/content/systems/filecoin_vm/message/_index.md
+++ b/content/systems/filecoin_vm/message/_index.md
@@ -27,10 +27,10 @@ rewarded according to each message's gas price and consumption, forming a market
 ## Message syntax validation
 
 A syntactically invalid message must not be transmitted, retained in a message pool, or
-included in a block. Even if an invalid message ends up being sent (e.g., by a malicious player), the libp2p GossipSub propagation protocol will reject the message on the receiving end.
+included in a block. If an invalid message is received, it should be dropped and not propagated further.
 
 When transmitted individually (before inclusion in a block), a message is packaged as
-`SignedMessage`, regardless of signature scheme used.
+`SignedMessage`, regardless of signature scheme used. A valid signed message has a total serialized size no greater than `message.MessageMaxSize`.
 
 ```go
 type SignedMessage struct {
@@ -44,7 +44,7 @@ A syntactically valid `UnsignedMessage`:
 - has a well-formed, non-empty `To` address,
 - has a well-formed, non-empty `From` address, 
 - has `Value` no less than zero and no greater than the total token supply (`2e9 * 1e18`), and
-- has a non-negative `MethodNum`, or zero if the message is for a plain transfer of funds between two actors,
+- has a non-negative `MethodNum`,
 - has non-empty `Params` only if `MethodNum` is zero,
 - has non-negative `GasPrice`,
 - has `GasLimit` that is at least equal to the gas consumption associated with the message's serialized bytes,
@@ -78,70 +78,9 @@ type Message struct {
 }
 ```
 
-There should be several functions able to extract information from the `Message struct`, such as the sender and recipient addresses, the value to be transferred, the required funds to execute the message and others as appropriate.
+There should be several functions able to extract information from the `Message struct`, such as the sender and recipient addresses, the value to be transferred, the required funds to execute the message and the CID of the message.
 
-```go
-func (m *Message) Caller() address.Address {
-	return m.From
-}
-
-func (m *Message) Receiver() address.Address {
-	return m.To
-}
-
-func (m *Message) ValueReceived() abi.TokenAmount {
-	return m.Value
-}
-
-func (m *Message) RequiredFunds() BigInt {
-	return BigAdd(
-		m.Value,
-		BigMul(m.GasPrice, NewInt(uint64(m.GasLimit))),
-	)
-}
-```
-
-Given that Transaction Messages should eventually be included in a Block and added to the blockchain, the validity of the message should be checked.
-
-
-```go
-func (m *Message) ValidForBlockInclusion(minGas int64) error {
-	if m.Version != 0 {
-		return xerrors.New("'Version' unsupported")
-	}
-
-	if m.To == address.Undef {
-		return xerrors.New("'To' address cannot be empty")
-	}
-
-	if m.From == address.Undef {
-		return xerrors.New("'From' address cannot be empty")
-	}
-
-	if m.Value.LessThan(big.Zero()) {
-		return xerrors.New("'Value' field cannot be negative")
-	}
-
-	if m.Value.GreaterThan(TotalFilecoinInt) {
-		return xerrors.New("'Value' field cannot be greater than total filecoin supply")
-	}
-
-	if m.GasPrice.LessThan(big.Zero()) {
-		return xerrors.New("'GasPrice' field cannot be negative")
-	}
-
-	if m.GasLimit > build.BlockGasLimit {
-		return xerrors.New("'GasLimit' field cannot be greater than a block's gas limit")
-	}
-
-	// since prices might vary with time, this is technically semantic validation
-	if m.GasLimit < minGas {
-		return xerrors.New("'GasLimit' field cannot be less than the cost of storing a message on chain")
-	}
-
-	return nil
-}
-```
+Given that Transaction Messages should eventually be included in a Block and added to the blockchain, the validity of the message should be checked with regard to the sender and the receiver of the message, the value (which should be non-negative and always smaller than the circulating supply), the gas price (which again should be non-negative) and the `BlockGasLimit` which should not be greater than the block's gas limit.
 
 
 ## Message semantic validation
--- a/content/systems/filecoin_vm/message/_index.md
+++ b/content/systems/filecoin_vm/message/_index.md
@@ -178,7 +178,3 @@ The reason for this "no message semantic validation" policy is that the state th
 be applied to cannot be known before the message is executed _as part of a tipset_. A block producer
 does not know whether another block will precede it in the tipset, thus altering the state to
 which the block's messages will apply from the declared parent state.
-
-{{<embed src="message.id" lang="go" >}}
-
-{{<embed src="message.go" lang="go" >}}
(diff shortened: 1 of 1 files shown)


content/systems/filecoin_vm/message/_index.md
```diff
@@ -10,41 +10,139 @@ dashboardTests: 0
 # VM Message - Actor Method Invocation
 
 A message is the unit of communication between two actors, and thus the primitive cause of changes
-in state. A message combines:
+in state. When referring to a Message in the context of Filecoin and unless otherwise defined, we refer to a _Transaction Message_. A message combines:
```
anorth, line 13: What other kind of message is there?

content/systems/filecoin_vm/message/_index.md
```diff
@@ -10,41 +10,139 @@ dashboardTests: 0
...
+A message's _gas limit_ provides an upper bound on the computation required to process it. The sender of a message pays 
 for the gas units consumed by a message's execution (including all nested messages) at a
 gas price they determine. A block producer chooses which messages to include in a block and is
 rewarded according to each message's gas price and consumption, forming a market.
 
 ## Message syntax validation
 
 A syntactically invalid message must not be transmitted, retained in a message pool, or
-included in a block.
+included in a block. Even if an invalid message ends up being sent (e.g., by a malicious player), the libp2p GossipSub propagation protocol will reject the message on the receiving end.
```
anorth, line 30: will -> should? 

How about "if an invalid message is received, a node should not propagate it". An implementer needs to tell their gossipsub what a valid message looks like, we can't just say it'll magically be rejected.

content/systems/filecoin_vm/message/_index.md
```diff
@@ -10,41 +10,139 @@ dashboardTests: 0
...
+}
+```
 
 A syntactically valid `UnsignedMessage`:
 
 - has a well-formed, non-empty `To` address,
 - has a well-formed, non-empty `From` address, 
-- has a non-negative `CallSeqNum`,
 - has `Value` no less than zero and no greater than the total token supply (`2e9 * 1e18`), and
-- has a non-negative `MethodNum`,
```
anorth, line 38: zero is non-negative...

content/systems/filecoin_vm/message/_index.md
```diff
@@ -10,41 +10,139 @@ dashboardTests: 0
...
+- has a non-negative `MethodNum`, or zero if the message is for a plain transfer of funds between two actors,
 - has non-empty `Params` only if `MethodNum` is zero,
 - has non-negative `GasPrice`,
 - has `GasLimit` that is at least equal to the gas consumption associated with the message's serialized bytes,
 - has `GasLimit` that is no greater than the block gas limit network parameter.
 
-When transmitted individually (before inclusion in a block), a message is packaged as
-`SignedMessage`, regardless of signature scheme used. A valid signed message:
 
-- has a total serialized size no greater than `message.MessageMaxSize`.
```
anorth, line 47: This maximum size is a critical detail

content/systems/filecoin_vm/message/_index.md
```diff
@@ -10,41 +10,139 @@ dashboardTests: 0
...
 
 A syntactically valid `UnsignedMessage`:
 
 - has a well-formed, non-empty `To` address,
 - has a well-formed, non-empty `From` address, 
-- has a non-negative `CallSeqNum`,
 - has `Value` no less than zero and no greater than the total token supply (`2e9 * 1e18`), and
-- has a non-negative `MethodNum`,
+- has a non-negative `MethodNum`, or zero if the message is for a plain transfer of funds between two actors,
 - has non-empty `Params` only if `MethodNum` is zero,
```
anorth, line 48: We changed this. Messages with MethodNum=0 can have non-empty params. So just remove this point.

content/systems/filecoin_vm/message/_index.md
```diff
@@ -10,41 +10,139 @@ dashboardTests: 0
...
 - has `GasLimit` that is no greater than the block gas limit network parameter.
 
-When transmitted individually (before inclusion in a block), a message is packaged as
-`SignedMessage`, regardless of signature scheme used. A valid signed message:
 
-- has a total serialized size no greater than `message.MessageMaxSize`.
+```go
+type Message struct {
+	// Version of this message
+	Version int64
```
anorth, line 57: I'm surprised this isn't a `uint64` - could you check? If it's not, there should probably be a validation rule about it being non-negative.

content/systems/filecoin_vm/message/_index.md
```diff
@@ -10,41 +10,139 @@ dashboardTests: 0
...
+
+	// Address of the receiving actor.
+	To   address.Address
+	// Address of the sending actor.
+	From address.Address
+
+	Nonce uint64
```
anorth, line 64: This is called `CallSeqNum`.

content/systems/filecoin_vm/message/_index.md
```diff
@@ -10,41 +10,139 @@ dashboardTests: 0
...
+	Value BigInt
+
+	// GasPrice is a Gas-to-FIL cost
+	GasPrice BigInt
+	// Maximum Gas to be spent on the processing of this message
+	GasLimit int64
+
+	// Optional method to invoke on receiver, zero for a plain value transfer.
+	Method abi.MethodNum
+	//Serialized parameters to the method (if method is non-zero).
```
anorth, line 76:
```suggestion
	//Serialized parameters to the method.
```

content/systems/filecoin_vm/message/_index.md
```diff
@@ -10,41 +10,139 @@ dashboardTests: 0
...
+	return m.Value
+}
+
+func (m *Message) RequiredFunds() BigInt {
+	return BigAdd(
+		m.Value,
+		BigMul(m.GasPrice, NewInt(uint64(m.GasLimit))),
+	)
+}
+```
```
anorth, line 102: Implementation-specific details, not adding value here.

content/systems/filecoin_vm/message/_index.md
```diff
@@ -10,41 +10,139 @@ dashboardTests: 0
...
+	}
+
+	// since prices might vary with time, this is technically semantic validation
+	if m.GasLimit < minGas {
+		return xerrors.New("'GasLimit' field cannot be less than the cost of storing a message on chain")
+	}
+
+	return nil
+}
+```
```
anorth, line 144: Is this just duplicating the prose bullet list above? We should not do this. We should do one or the other (and I think the prose is superior).

content/systems/filecoin_vm/message/_index.md
yiannisbot: @anorth this is what I was based on: https://github.com/filecoin-project/lotus/blob/8c516df87b9f00c53c0767c607f955a7bab9d003/chain/types/message.go#L29

content/systems/filecoin_vm/message/_index.md
yiannisbot: Do you mean it's not called `Nonce` anymore, or you mean that it's pointing to `CallSeqNum`?

content/systems/filecoin_vm/message/_index.md
yiannisbot: The term message is rather liberally used throughout (e.g., for blocks, exit codes, etc.) and I felt that it's useful to clarify. Do you think it's not needed?

content/systems/filecoin_vm/message/_index.md
yiannisbot: Removed. Does this mean that a `MethodNum` needs to have a positive value?

content/systems/filecoin_vm/message/_index.md
yiannisbot: I've excluded the code and extended the description.

content/systems/filecoin_vm/message/_index.md
yiannisbot: I've excluded the code and extended the description.

content/systems/filecoin_vm/message/_index.md
yiannisbot: I've brought it back. Thanks

content/systems/filecoin_vm/message/_index.md
yiannisbot: I've integrated the suggested text in the subsequent commit.

content/systems/filecoin_vm/message/_index.md
anorth: Hmm. The message in blocks is very nearly the same concept as message passed between actors - close enough that I think it's ok to use the same term. And an exit code is associated with this same message concept - the internal and top-level ones both have exit codes.

My concern is that this sentence introduces a new term, "transaction", that isn't defined or used anywhere else in the spec. (The Runtime state accessor uses the term, but it means something different there).

content/systems/filecoin_vm/message/_index.md
anorth: No, it can be zero and that message can still have params.

content/systems/filecoin_vm/message/_index.md
anorth: Ok, I've asked what's up with this [in Slack](https://filecoinproject.slack.com/archives/CP50PPW2X/p1597812900247300).

content/systems/filecoin_vm/message/_index.md
anorth: I mean the spec has always called it `CallSeqNum`, and since the rest of the spec uses that term we should match that here.

It kinda doesn't matter what any implementation calls it internally, since the name is not part of the serialized format.

content/systems/filecoin_vm/message/_index.md
yiannisbot: Done

content/systems/filecoin_vm/message/_index.md
yiannisbot: Ok, you convinced me! :) I've removed the extra clarification.

content/systems/filecoin_vm/message/_index.md
yiannisbot: I've rephrased to `CallSeqNum` to be consistent with the spec. I was also wondering why `CallSeqNum` is not here, so good that we clarified.

content/systems/filecoin_vm/message/_index.md
anorth: There has been some confusion. Please remove the line 

> has non-empty `Params` only if `MethodNum` is zero,

(`MethodNum` is a `uint64` so it's correct to have removed that line also.)

content/systems/filecoin_vm/message/_index.md
anorth: FYI I'm gunna change this to uint64 in lotus

anorth (approved):

content/systems/filecoin_vm/message/_index.md
yiannisbot: Changed it in the latest commit, so that it's up-to-date.

content/systems/filecoin_vm/message/_index.md
yiannisbot: Done in latest commit.

PR #1061, opened 2020-08-14
fix(content): FIL VM System Actors Update

yiannisbot:
This PR is updating the description of the system actors that interact with the VM (init, cron, account, reward).

There is more to the reward actor than is said in this and the previous version of the spec, so I would welcome opinions on whether we should add more, e.g., related to the calculation of the reward, the penalties and so on. On the other hand, some of this is text for the cryptoecon part, so an option is to include it there.


content/systems/filecoin_vm/sysactors/cron_actor.md
```diff
@@ -10,4 +10,25 @@ dashboardTests: 0
 # CronActor
 ---
 
-{{<embed src="/modules/actors/builtin/cron/cron_actor.go"  lang="go">}}
+The `CronActor` is responsible for sending messages to other registered actors at the end of every epoch. It interfaces with the Storage Power Actor on `OnEpochTickEnd` to check:
+- `CronEventPreCommitExpiry`: if the miner has submitted their `ProveCommit` proofs,
+- `CronEventProvingPeriod`: the proving period identified,
+- `CronEventWorkerKeyChange`: if a miner has submitted a worker key address change.
```
anorth, line 16: IMO, the cron actor should not be defined in terms of the other actors that use it, but the service that it provides.

content/systems/filecoin_vm/sysactors/_index.md
```diff
@@ -11,93 +11,22 @@ dashboardTests: 0
...
+- `MultisigActor` (or Multi-Signature Wallet Actor): responsible for dealing with operations involving the Filecoin wallet [[Multisig Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/multisig/multisig_actor.go)]
+- `PaychActor` (or Payment Channel Actor): responsible for setting up and settling funds related to payment channels [[Paych Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/paych/paych_actor.go)]
+-  `PowerActor` (or Storage Power Consensus Actor): responsible for keeping track of the storage power allocated at each storage miner [[Storage Power Actor](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/power/power_actor.go)]
+- `verifregActor` (or Verified Registry Actor): responsible for managing verified clients [[Verifreg Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/verifreg/verified_registry_actor.go)]
```
anorth, line 31: `VerifiedRegistryActor`

content/systems/filecoin_vm/sysactors/_index.md
```diff
@@ -11,93 +11,22 @@ dashboardTests: 0
...
   - the [RewardActor](reward_actor.md) for block reward and token vesting (singleton).
+
+
+The remaining seven (7) builtin System Actors that do not interact directly with the VM are the following:
+
+- `MarketActor`: responsible for managing storage and retrieval deals [[Market Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_actor.go)]
+- `MinerActor` (or Storage Miner Actor): actor responsible to deal with storage mining operations and collect proofs [[Storage Miner Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go)]
+- `MultisigActor` (or Multi-Signature Wallet Actor): responsible for dealing with operations involving the Filecoin wallet [[Multisig Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/multisig/multisig_actor.go)]
+- `PaychActor` (or Payment Channel Actor): responsible for setting up and settling funds related to payment channels [[Paych Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/paych/paych_actor.go)]
```
anorth, line 29: `PaymentChannelActor`

content/systems/filecoin_vm/sysactors/_index.md
```diff
@@ -11,93 +11,22 @@ dashboardTests: 0
...
   - the [InitActor](init_actor.md), which initializes new actors and records the network name, and
   - the [CronActor](cron_actor.md), a scheduler actor that runs critical functions at every epoch.
 There are another two actors that interact with the VM:
   - the [AccountActor](account_actor.md) responsible for user accounts (non-singleton), and
   - the [RewardActor](reward_actor.md) for block reward and token vesting (singleton).
+
+
+The remaining seven (7) builtin System Actors that do not interact directly with the VM are the following:
+
+- `MarketActor`: responsible for managing storage and retrieval deals [[Market Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/market/market_actor.go)]
```
anorth, line 26: `StorageMarketActor`

content/systems/filecoin_vm/sysactors/_index.md
```diff
@@ -11,93 +11,22 @@ dashboardTests: 0
...
+- `MinerActor` (or Storage Miner Actor): actor responsible to deal with storage mining operations and collect proofs [[Storage Miner Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go)]
```
anorth, line 27: `StorageMinerActor`

content/systems/filecoin_vm/sysactors/_index.md
```diff
@@ -11,93 +11,22 @@ dashboardTests: 0
...
+- `MultisigActor` (or Multi-Signature Wallet Actor): responsible for dealing with operations involving the Filecoin wallet [[Multisig Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/multisig/multisig_actor.go)]
+- `PaychActor` (or Payment Channel Actor): responsible for setting up and settling funds related to payment channels [[Paych Actor Repo](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/paych/paych_actor.go)]
+-  `PowerActor` (or Storage Power Consensus Actor): responsible for keeping track of the storage power allocated at each storage miner [[Storage Power Actor](https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/power/power_actor.go)]
```
anorth, line 30: `StoragePowerActor`

content/systems/filecoin_vm/sysactors/cron_actor.md
```diff
@@ -2,33 +2,18 @@
...
 dashboardTests: 0
 ---
 
 # CronActor
 ---
 
-The `CronActor` is responsible for sending messages to other registered actors at the end of every epoch. It interfaces with the Storage Power Actor on `OnEpochTickEnd` to check:
-- `CronEventPreCommitExpiry`: if the miner has submitted their `ProveCommit` proofs,
+The `CronActor` is responsible for scheduling jobs for several operations that need to take place between actors. This is primarily realised in the form of sending messages to other registered actors at the end of every epoch. It interfaces with the Storage Power Actor on `OnEpochTickEnd` to check:
+- `CronEventPreCommitExpiry`: if the miner has submitted their `ProveCommit` proofs (given enough time for the miner to complete the proofs of replication and finalize the commit phase). If the miner has completed their commit phase on time, they can claim the `PreCommiDeposit` back. If not, the `PreCommitDeposit` is lost and no power is added for this miner to the power table.
```
anorth, line 14: These events are all really internal details of the power and miner actor. The cron actor itself has its own registry of calls, and it contains two items: a call to the power and a call to the market actor.

Note that names of things like methods and cron event tags are liable to change, which will lead to confusion here (the names aren't part of the "spec").

content/systems/filecoin_vm/sysactors/reward_actor.md
```diff
@@ -12,38 +12,11 @@ dashboardTests: 0
...
 
-```go
-type AwardBlockRewardParams struct {
-	Miner     address.Address
-	Penalty   abi.TokenAmount // penalty for including bad messages in a block, >= 0
-	GasReward abi.TokenAmount // gas reward from all gas fees in a block, >= 0
-	WinCount  int64           // number of reward units won, > 0
-}
-```
+A `Reward` struct contains a `StartEpoch` that keeps track of when this `Reward` is created, a `Value` that represents the total number of tokens rewarded, and an `EndEpoch` which is when the reward will be fully vested. The `AmountWithdrawn` records how many tokens have been withdrawn from a `Reward struct` so far. Owner addresses can call `WithdrawReward` which will withdraw all vested tokens that the investor address has from the RewardMap so far. When `AmountWithdrawn` equals `Value` in a `Reward struct`, the `Reward struct` will be removed from the `RewardMap`.
```
anorth, line 17: This is all really old and not at all how it works now. I suggest avoiding description at this low a level.

content/systems/filecoin_vm/sysactors/cron_actor.md
yiannisbot: > These events are all really internal details of the power and miner actor. 

I included those to provide some links between the different sections and description of actors. I thought it's going to be useful for the reader. But I'm fine removing, which I have done in the latest commit.

> The cron actor itself has its own registry of calls, and it contains two items: a call to the power and a call to the market actor.

The latest description is more general to say what the `CronActor` _could_ do. I think this is what we want from the spec? Should I just mention these two?

> Note that names of things like methods and cron event tags are liable to change, which will lead to confusion here (the names aren't part of the "spec").

Agreed, but I think these are descriptive-enough, so even if the actual names in the implementation change, this still provides the concept. Do you want me to come up with more generic ones?

content/systems/filecoin_vm/sysactors/cron_actor.md
yiannisbot: I've made an effort to revise the text towards this direction in the latest commit. Not sure it's enough. Let me know if some of it should be more or less specific.

content/systems/filecoin_vm/sysactors/reward_actor.md
yiannisbot: Excluded this para in latest commit. What are the important points about the `RewardActor` that need to be in spec and are not misleading?

yiannisbot: Also marked it as "stable" - let me know if it's not the case and "incomplete" is more appropriate :)

content/systems/filecoin_vm/sysactors/cron_actor.md
anorth: I would just say that, built in to the genesis state, the CronActor's dispatch table invokes the StoragePowerActor and StorageMarket actor for them to maintain internal state and process deferred events. It could in principle invoke other actors after a network upgrade.

There's no " `OnEpochTickEnd` call the `InitActor`"

content/systems/filecoin_vm/sysactors/init_actor.md
```diff
@@ -2,12 +2,16 @@
...
+dashboardState: stable
 dashboardAudit: 0
 dashboardTests: 0
 ---
 
 # InitActor
 ---
 
-{{<embed src="/modules/actors/builtin/init/init_actor.go" lang="go">}}
+The `InitActor` has the power to create new actors, e.g., those that enter the system. It maintains a table resolving a public key and temporary actor addresses to their canonical ID-addresses. Invalid CIDs should not get committed to the state tree.
```
anorth, line 13:
```suggestion
The `InitActor` has the power to create new actors, e.g., those that enter the system. It maintains a table resolving a public key or temporary actor addresses to their canonical ID-addresses. Invalid CIDs should not get committed to the state tree.
```

content/systems/filecoin_vm/sysactors/init_actor.md
```diff
@@ -2,12 +2,16 @@
...
+
+The actor should be able to construct a canonical ID address for the actor that will persist even in case of a chain re-organisation.
```
anorth, line 15: The canonical ID address does _not_ persist even in case of chain re-org. The address types are distinguished by this:
- the ID address is canonical but does not survive reorg
- the actor or pubkey address survives reorgs

content/systems/filecoin_vm/sysactors/reward_actor.md
```diff
@@ -1,18 +1,20 @@
...
+dashboardState: stable
 dashboardAudit: 0
 dashboardTests: 0
 ---
 
 # RewardActor
 ---
 
-RewardActor is where unminted Filecoin tokens are kept. RewardActor contains a `RewardMap` which is a mapping from owner addresses to `Reward` structs. 
+The `RewardActor` is where unminted Filecoin tokens are kept. The `RewardActor` contains a `RewardMap` which is a mapping from owner addresses to `Reward` structs.
```
anorth, line 13: This `RewardMap` stuff is out of date. The actor distributes rewards directly to miner actors, where they are locked for vesting.

The paragraph below is wrong too.

content/systems/filecoin_vm/sysactors/reward_actor.md
```diff
@@ -1,18 +1,20 @@
...
+A `Reward struct` is created to preserve the flexibility of introducing block reward vesting into the protocol. `MintReward` creates a new `Reward struct` and adds it to the `RewardMap`.
 
-A `Reward` struct contains a `StartEpoch` that keeps track of when this `Reward` is created, `Value` that represents the total number of tokens rewarded, and `EndEpoch` which is when the reward will be fully vested. `VestingFunction` is currently an enum to represent the flexibility of different vesting functions. `AmountWithdrawn` records how many tokens have been withdrawn from a `Reward` struct so far. Owner addresses can call `WithdrawReward` which will withdraw all vested tokens that the investor address has from the RewardMap so far. When `AmountWithdrawn` equals `Value` in a `Reward` struct, the `Reward` struct will be removed from the `RewardMap`.
+The award value used for the current epoch is updated at the end of an epoch through a cron tick. In the case previous epochs were null blocks this is the reward value as calculated at the last non-null epoch.
```
anorth, line 17: The "null block" case isn't relevant any more, we fixed it.

anorth (approved):
It depends what you want to mean by "stable". I would not say "will not change before mainnet", that is way too strong.

These bits are all ~80% right. We've done the 20% work for the 80% benefit. Getting them complete and correct 100% will be much more effort, especially from the implementers who know all the gnarly details.

content/systems/filecoin_vm/sysactors/reward_actor.md
yiannisbot: Marking all of the VM sections as "reliable" for now.

content/systems/filecoin_vm/sysactors/init_actor.md
yiannisbot: Noted and changed.

content/systems/filecoin_vm/sysactors/reward_actor.md
yiannisbot: Updated.

hugomrdias (approved):

PR #1071, opened 2020-08-19
fix: dashboard numbering Only shows the first section for each file because we only have values per file.

hugomrdias:


yiannisbot (commented):
It all looks good to me @hugomrdias. But it would be good to have an issue that explains the changes, what was the problem before, how this PR fixes it and what (if anything) we're missing by going this way.

olizilla:
Nice work @hugomrdias 👌

PR #1072, opened 2020-08-20
added the last three requirements for leader election 

sa8:
I would like to add a link to the "cryptoeco" part of the specs (where we talk about initial pledge etc) but I don't know where that is.

### DIFF
--- a/content/algorithms/expected_consensus/_index.md
+++ b/content/algorithms/expected_consensus/_index.md
@@ -412,7 +412,7 @@ Specifically, this can be proven with a "witness" block, that is by submitting b
 
 ### Penalization for faults
 A single consensus fault results into:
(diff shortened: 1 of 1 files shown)


content/algorithms/expected_consensus/_index.md
```diff
@@ -400,11 +412,13 @@ Specifically, this can be proven with a "witness" block, that is by submitting b
 
 ### Penalization for faults
 A single consensus fault results into:
-- miner termination and removal of power from the power table,
-- loss of all pledge collateral (which includes the initial pledge and blocks rewards yet to be vested)
+- miner suspension for 24 hours (the miner is not allowed to create block during that time)
```
nicola, line 415: we currently agreed on finality, not 24h

PR #1077, opened 2020-08-24
fix: layout jump and more

### LINKED ISSUE #1065: Code blocks in spec
We need to make a decision on the best way to include code in the spec.

The requirements for this are:
1. the lotus version of the code should be present in the spec (this will be the Go code as agreed in https://github.com/filecoin-project/specs/issues/998 - at least for now)
2. the code presented in the specs site should be up to date.

The options we have for the first point are:
- include blocks of code in the corresponding sections. IMO, this breaks the flow and disrupts attention as code blocks are very often very long. An alternative is to do some css magic (suggested by @olizilla) and include code in "scrollable" windows/frames in each section. In this case, we would have up to "includeHeight" lines of code directly visible in the window/frame and the rest would be "scrollable". This has the advantage of both having the code in place, but also avoid having endless blocks of code within the section.
- include the code blocks in a the "Listings" section and link to each listing from the corresponding section in the main spec. This again provides easy access to the code block without being disruptive to the text of the specification.

The only option I see for the second point, which will meet our requirement without incurring extra manual effort is to pull the corresponding .go files from the main repositories and rebuild the site every some amount of time (say every 1h/1d). Another less optimal option is to manually copy the .go files on "code freeze" to the specs repo. In the future, this means that any changes to the main code will have to be updated manually on the specs repo, or otherwise the specs version of the code will run out of date.

Until we start working towards the vision of "compilable" spec site, these options look reasonable to me. It would be great if we can reach a decision asap.

cc: @whyrusleeping @anorth @schomatis @daviddias

hugomrdias:
fix: fix hugo modules mounts for retrieval market

fix: fix scrollbars and add fixed height to code blocks

   closes #1065

fix: fix dashboard resize

   anchor links wouldnt scroll to the proper section because dashboard
   creation would make the layout jump in size at runtime

fix: remove logs


README.md
```diff
@@ -119,7 +144,7 @@ stringit, frustra Saturnius uteroque inter! Oculis non ritibus Telethusa
 ```
 
 ## Page Header
-The first heading should be # Head with `---` like below and should refer to the overall title of the document. The right nav **only** starts on the second level of headings. 
+The first heading should be # Head with `---` like below and should refer to the overall title of the document.
```
olizilla, line 147: i really really want to fix this. The first heading should just be a `#` and the horizontal line should be added in with css not a hr element.

olizilla: given that the "first" header will be re-written to be a different heading level depending on its depth in the tree, it doesn't even make sense to include explict hr elements, as they will be inconsistently applied to either `h1`s `h2`s or `h3`s depending on what level this doc starts at

README.md
```diff
@@ -146,18 +171,18 @@ Description for all the available frontmatter properties
...
 <!-- This will hidden the page from the navigation -->
 bookhidden: true
 <!-- This is used in the dashboard to describe the importance of the page content -->
 dashboardWeight: 2
-<!-- This is used in the dashboard to describe the state of the page content options are "incorrect", "wip", "incomplete" and "stable" -->
+<!-- This is used in the dashboard to describe the state of the page content options are "missing", "incorrect", "wip", "reliable", "stable" or "n/a" -->
 dashboardState: stable
-<!-- This is used in the dashboard to describe if the theory of the page has been audited, options are 1 or 0 -->
-dashboardAudit: 1
+<!-- This is used in the dashboard to describe if the theory of the page has been audited, options are "missing", "wip", "stable" or "n/a" -->
```
olizilla, line 184: `stable` doesn't make sense here. Ideally it would be the date of the last audit. if we can't have that then `done` or `exists` would make sense.

README.md
hugomrdias: yes yes!! https://github.com/filecoin-project/specs/issues/1081

olizilla (approved):
This is all good. I would swap the colors for reliable and stable... like green is reliable, good enough to use, and filecoin blue is stable as in "this is filecoin! it (probably) wont change".

that and the dashboardAudit values need tweaking.

olizilla:
I think the table needs to be rendereded in it's entirety... can you set the height to be the height of all the rows for now until we find a better fix?

olizilla:
... or it needs a link that takes you to a full page view of the overview table.

hugomrdias:
> I think the table needs to be rendereded in it's entirety... can you set the height to be the height of all the rows for now until we find a better fix?

You think ? that's harder to do in a way that doesn't break the layout if we change something. Being the size of the viewport isn't enough ?

olizilla:
Can tackle it in another PR... I just think that folks will want the full table, not inside an inner scroll container.

hugomrdias:
So 
- swap reliable/stable colors
- use `done` for audit `stable`
- remove scroll from table without breaking the anchor links
@olizilla can you handle these i another PR and i will merge this ?

olizilla:
@hugomrdias I am fine with that. Pls assign me an issue for any bits you want me to do.

PR #1099, opened 2020-08-26
feat: add instructions to enter multi-page mode

vmx:
This commit also adds a multi-page landing page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ brew install go bzr
 yarn start
 # open http://localhost:1313/ in the browser
 ```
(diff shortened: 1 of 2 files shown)


vmx:
Rebased on top of current beta branch.

olizilla:
We have a feature request to support extra documents - single page docs that dont have the ToC of the rest of the site... What i'd like to do to support the use-case of "let me focus on just this page while i'm developing" is to tweak things so that in dev-mode, there is a link in each page heading that would take you to the single page for that doc. I'll demo it in a seperate PR.

Conanza00 (approved):

PR #1104, opened 2020-08-26
chore: tidy up the npm scripts

olizilla:
A proposal for @hugomrdias's eyes.

- Pull out each job into it's own target so we can use `concurently`'s support for globs over job names.
- Create a clean target for all of the things.
- make `serve` an alias of `start` so folks dont have to think.

arent they lovely

https://github.com/filecoin-project/specs/blob/efce26a63c2b29f105596397e5ec0df152b14db8/package.json#L6-L19

License: MIT


package.json
```diff
@@ -4,12 +4,18 @@
...
+    "serve": "npm start",
+    "build": "concurrently npm:build-*",
+    "clean": "concurrently npm:clean-*",
+    "watch-hugo": "hugo server --bind=0.0.0.0 --disableFastRender",
+    "watch-diagrams": "tools/diagrams.js --watch",
+    "build-hugo": "hugo --gc --minify",
+    "build-diagrams": "tools/diagrams.js --all",
+    "clean-hugo": "rm -rf ./public",
+    "clean-hugo-modules": "hugo mod clean --all",
+    "clean-diagrams": "rm -rf ./_gen"
```
hugomrdias, line 18:
```suggestion
    "clean": "premove public resources static/_gen && hugo mod clean --all && hugo mod tidy"
```

yarn add premove --dev

hugomrdias (changes requested):

package.json
hugomrdias: i was doing the same thing lol :)

hugomrdias:
something when wrong with this PR i think you started from the pipelines branch but the target was beta 
Superseed by #1105

PR #1107, opened 2020-08-26
fix(content): revise language & recommendations around timekeeping

jsoares:
Revises the clock section to:
* Correct misuse of the terms "offset", "skew", and "drift" in the context of timekeeping
* Correct inconsistent use of "client" wrt the rest of the spec
* Provide appropriate recommendations for system-level timekeeping. Just getting a new time every hour is far from ideal: in particular, it can cause time to move back in a running system, which is sometimes problematic. Taking 3 samples and dropping outliers is also not effective. The proper way to keep time is to run a daemon performing continuous synchronisation against multiple servers and skew correction. 
* Provide appropriate recommendations for facility-level timekeeping. There is a whole range of alternatives between a "computer-grade" oscillator and a caesium clock (e.g. TXCOs). The latter certainly isn't necessary for this level of precision (none of them are, really, unless you're going for sub-ms). Also, the best thing you can do for timekeeping is just getting a GPS antenna.
* Remove misleading reference (and broken link) to Ethereum offset correction. What the example shows is just go-ethereum doing an NTP check and alerting the user to excessive offsets -- it does not synchronise the clock. Wall clock synchronisation really should be handled at the system level and not by the Filecoin (or Ethereum) binary.


content/systems/filecoin_nodes/clock/_index.md
```diff
@@ -27,34 +27,31 @@ The Filecoin system clock is used:
...
-  - We further recommend making 3 measurements in order to drop by using the network to drop outliers
-  - See how [go-ethereum does this](https://github.com/ethereum/go-ethereum/blob/master/p2p/discv5/ntp.go) for inspiration
-- clients MAY consider using cesium clocks instead for accurate synchrony within larger mining operations
+- Larger mining operations MAY consider using local NTP/PTP servers with GPS references and/or frequency-stable external clocks for improved timing.
 
-Mining operations have a strong incentive to prevent their clock from drifting ahead more than one epoch to keep their block submissions from being rejected.  Likewise they have an incentive to prevent their clocks from drifting behind more than one epoch to avoid partitioning themselves off from the synchronized nodes in the network.
+Mining operations have a strong incentive to prevent their clock skewing ahead more than one epoch to keep their block submissions from being rejected.  Likewise they have an incentive to prevent their clocks skewing behind more than one epoch to avoid partitioning themselves off from the synchronized nodes in the network.
 
 ## Future work
```
jsoares, line 51: The adoption of drand introduces a strong dependency on a common external clock; it may make sense to reword (or delete) this section. Since I am not fully aware of the implications and constraints on FIL's side, I'm not the person to do it.

content/systems/filecoin_nodes/clock/_index.md
```diff
@@ -27,34 +27,31 @@ The Filecoin system clock is used:
...
 
-If either of the above metrics show significant network skew over time, future versions of Filecoin may include potential timestamp/epoch correction periods at regular intervals.
+If continued network timing aberration is seen over time, future versions of Filecoin may include potential timestamp/epoch correction periods at regular intervals.
 
 When recoverying from exceptional chain halting outages (for example all implementations panic on a given block) the network can potentially opt for per-outage "dead zone" rules banning the authoring of blocks during the outage epochs to prevent attack vectors related to unmined epochs during chain restart.
 
 Future versions of the Filecoin protocol may use Verifiable Delay Functions (VDFs) to strongly enforce block time and fulfill this leader election requirement; we choose to explicitly assume clock synchrony until hardware VDF security has been proven more extensively.
```
jsoares, line 57: The adoption of drand introduces a strong dependency on a common external clock; it may make sense to reword (or delete) this section. Since I am not fully aware of the implications and constraints on FIL's side, I'm not the person to do it.

olizilla:
Thanks @jsoares this looks good to me. Keep 'em coming.

For PRs that update terminology, please help your friendly neighbourhood reviewers who may not be as well versed in the topic by linking to sources in the desctiption. I've learnt a lot from spelunking here...

>   To begin, we define a clock's **"offset"** at a particular moment as the
>   difference between the time reported by the clock and the "true" time
>   as defined by UTC.  If the clock reports a time Tc and the true time
>   is Tt, then the clock's offset is Tc - Tt.

>   We will refer to a clock as "accurate" at a particular moment if the
>   clock's offset is zero, and more generally a clock's "accuracy" is
>   how close the absolute value of the offset is to zero.  For NTP,
>   accuracy also includes a notion of the frequency of the clock; for
>   our purposes, we instead incorporate this notion into that of **"skew"**,
>   because we define accuracy in terms of a single moment in time rather
>   than over an interval of time.
>
>   A clock's "skew" at a particular moment is the frequency difference
>   (first derivative of its offset with respect to true time) between
>   the clock and true time.
>
>   As noted in RFC 1305, real clocks exhibit some variation in skew.
>   That is, the second derivative of the clock's offset with respect to
>   true time is generally non-zero.  In keeping with RFC 1305, we define
>   this quantity as the clock's **"drift"**.

from: https://tools.ietf.org/html/rfc2330#section-10
found via: https://www.iol.unh.edu/sites/default/files/knowledgebase/1588/clock_synchronization_terminology.pdf

olizilla:
@yiannisbot can you take a moment to review this pls.

jsoares:
@olizilla Will do, thanks for the feedback!

jsoares:
@yiannisbot Updated to current master, as requested. The diff looks a little less clear than it used to but the changes should be the same.

yiannisbot (approved):
Very good set of changes.

jsoares:
Thanks, @yiannisbot! I don't have merge rights so someone else will need to handle it.

hugomrdias (approved):

PR #1111, opened 2020-08-27
Updating Data Transfer Protocol section status to "Stable"

yiannisbot:
Due to parallel commits and PR merges the status of this section was accidentally downgraded to "wip" from "stable". This PR is setting the status back to the correct one.


olizilla (approved):
👍

Issue #992, opened 2020-07-20
Beta version feedback


hugomrdias:
This issue will be used to get feedback from the team.

We started the first round of feedback in this [PR](https://github.com/filecoin-project/specs/pull/969)  

You can see our current planning [here](https://github.com/filecoin-project/specs/projects/8) and the latest deploy here https://beta.spec.filecoin.io

alanshaw:
I can't read the specs website on my iphone it crashes repeatedly:

![IMG_B77BE30BEE0E-1](https://user-images.githubusercontent.com/152863/91560299-89a23600-e931-11ea-99f8-29a54d324c87.jpeg)

alanshaw:
Suggestion: add a link back to the github repositiory so that the community can fix typos etc.

alanshaw:
Does the default view _have_ to be a single page? Why not offer this as an alternative like the nodejs docs do:

<img width="1212" alt="Screenshot 2020-08-28 at 13 32 33" src="https://user-images.githubusercontent.com/152863/91561235-26b19e80-e933-11ea-924f-6ca28c14cd44.png">

olizilla:
The single page view is somewhat hostile to the mobile view. We have a requirement to make the single page view be the canonical view. I would like to reverse that decision, but I dont have time to fight that cause right now.

yiannisbot:
The site has shipped. Thanks for all the feedback!

PR #1118, opened 2020-08-28
wip: multi-page mode for dev

olizilla:
aims to provide clean multi page layout and a dev-mode only link to the isolated page for a given section

RE: #1099 

License: MIT

### DIFF
--- a/assets/_icons.scss
+++ b/assets/_icons.scss
@@ -41,3 +41,30 @@
     border-top: 2px solid;
     top: -4px
 }
+.gg-file {
+    box-sizing: border-box;
+    position: relative;
+    display: block;
+    transform: scale(var(--ggs,1));
+    width: 14px;
+    height: 16px;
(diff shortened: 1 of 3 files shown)


vmx:
I like that idea, but wouldn't something like the multi-page hack at https://github.com/filecoin-project/specs/pull/1099 also help with https://github.com/filecoin-project/specs/issues/992#issuecomment-682500687? It would be quiet easy to tell people hey, this is not officially supported, but if you want a multi-page version, just go to spec.filecoin.io/multipage.

Anyway, you are the one who makes UX/DX decisions, so I can live with either (or both :P)

olizilla:
The minor conflict is that the multi-page jump off point requires the multi-page ToC to work (scroll is broken currently) and it would need to be toggleable... additonal pages that a suplemental to the spec do not want to include the spec ToC.

This is not insurmountable, but I'm gonna park it for today and review on next week.

PR #1122, opened 2020-08-28
feat: build toc and dashboard in hugo

### LINKED ISSUE #1125: Issues building ToC and Dashboard on the client
We have multiple reports of the page crashing on safari on iOS, which we blieve to be due to the parsing the entire dom into a model that is then used to generate the ToC and the spec dashboard.

The bulding the dashboard on the client also causes the page height to change after the page has loaded, which makes in-page anchors seem broken, as the page is scrolled to where the anchor was, then the page gets longer, and the viewport no longer shows the anchor that you scrolled to. The current workaround for that is to fix the height of the dashboard in css, but  that introduces a chore to maintain it.

A fix for this is underway in #1122 where we parse the html at build time, after hugo has generated the index.html, then write the mode to the hugo data dir, which triggers a rebuild that then includes the ToC and Dashboard. The end result is the final html includes everything, allowing us to remove the client side dom building entirely. The thing to verify is if the dev / author experience remains a joy. If it causes any problems, we could use the client side dom building in dev, and then post-process the html once for a production deploy, but I am keen to avoid having to maintain two pathways.

olizilla:
Avoid building the ToC and the dashboard on the client side by parse data from dom and building the toc from data. 

Generates a toc.json file in the hugo data dir, which is then used by the toc partial and the dashboard-spec shortcode.

**example toc.json**
```json
[
  {
    "id": "intro",
    "number": "1",
    "tagName": "H1",
    "text": "Introduction",
    "page": true,
    "dashboardWeight": "0.2",
    "dashboardAudit": "n/a",
    "dashboardAuditURL": "",
    "dashboardAuditDate": "",
    "dashboardState": "reliable",
    "children": [
      {
        "id": "intro__spec-status",
        "number": "1.1",
        "tagName": "H2",
        "text": "Spec Status",
        "page": false,
        "dashboardWeight": "0.2",
        "dashboardAudit": "n/a",
        "dashboardAuditURL": "",
        "dashboardAuditDate": "",
        "dashboardState": "reliable",
        "children": [
```

TODO
- [x] ensure livereload picks up edits.
- [x] add watch mode for dom parser
- [x] tidy up
- [x] build dashboard from data

fixes #1125 

License: MIT


olizilla:
@hugomrdias local testing shows we can have hugo render to disk in dev mode, parse the generated dom, write the toc model as JSON to the hugo data dir, and have hugo pick up the change, and rebuild the ToC 🎉  

Next up i'm going to move the spec dashboard back to hugo but this time built from the toc model.

Would be great to get your eyes on it in the mean time to double check this is not too outrageous.

hugomrdias:
LGTM can you make it work with the diagrams watcher ? we should use only one watcher.
maybe rename diagrams.js to pipelines and pack everything into a single watcher.

hugomrdias:
hidden toc sections are only actually hidden at runtime with js, can we build the toc already hidden at build time ?

olizilla:
> hidden toc sections are only actually hidden at runtime with js, can we build the toc already hidden at build time ?

Done!

> GTM can you make it work with the diagrams watcher ? we should use only one watcher. maybe rename diagrams.js to pipelines and pack everything into a single watcher.

Will do in separate PR

hugomrdias (changes requested):
the build is not minifying the output html

olizilla:
@hugomrdias the npm run script `build` is minifying the html output with the `--minify`  flag to hugo, there is no change there. I've compared the output and the only diffs are the inclusion of the ToC & Dashboard, and the cache busting values on all images.

The dev build probably doesn't minify the output, but I dont see a problem there.

hugomrdias:
I know but some the preview url doesnt load the minified html

hugomrdias (approved):

PR #1123, opened 2020-08-29
fix(content): chainsync section update

yiannisbot:
This PR is updating the ChainSync section to match more closely the latest developments. Although nothing was terribly wrong in the previous version of the section, the text looked more like a rather theoretical specification. Furthermore, lots has changed in the way pubsub (i.e., GossipSub) works, which made the chainsync text look out of date.

We will try to keep a version of the previous text and link to it for those that will want to refer back to it.


content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,42 +1,52 @@
...
 
 `ChainSync` is the protocol Filecoin uses to synchronize its blockchain. It is
 specific to Filecoin's choices in state representation and consensus rules,
 but is general enough that it can serve other blockchains. `ChainSync` is a
 group of smaller protocols, which handle different parts of the sync process.
 
+Chain synchronisation generally applies to three main processes:
+1. Chain synchronization when a node first joins the network and needs to get to the current state before validating or extending the chain.
+2. Chain synchronisation when a node has fell out of sync, e.g., due to a brief disconnection.
+3. Continuous chain synchronisation to keep up with the latest messages and blocks.
```
schomatis, line 28: These 3 are the same process. The differentiation is useful as examples of how the sync behaves in different circumstances but let's not call each of them processes.

content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,42 +1,52 @@
...
+
+There are three main protocols used to achieve synchronisation for these three processes.
+- `GossipSub` is the libp2p pubsub protocol used to propagate messages and blocks. It is mainly used in the third process above when a node needs to stay in sync with new blocks being produced and propagated.
+- `BlockSync` is used to synchronise specific parts of the chain, that is from and to a specific height.
```
schomatis, line 32: FYI, this may change on the near future (will keep you updated): https://github.com/filecoin-project/lotus/issues/2612

content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,42 +1,52 @@
...
 specific to Filecoin's choices in state representation and consensus rules,
 but is general enough that it can serve other blockchains. `ChainSync` is a
 group of smaller protocols, which handle different parts of the sync process.
 
+Chain synchronisation generally applies to three main processes:
+1. Chain synchronization when a node first joins the network and needs to get to the current state before validating or extending the chain.
+2. Chain synchronisation when a node has fell out of sync, e.g., due to a brief disconnection.
+3. Continuous chain synchronisation to keep up with the latest messages and blocks.
+
+There are three main protocols used to achieve synchronisation for these three processes.
```
schomatis, line 30: I would also add the `hello` protocol, when two peers meet they exchange their chain heads (from there it goes through the same program flow as an incoming block from gossipsub).

content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,42 +1,52 @@
...
+- `GossipSub` is the libp2p pubsub protocol used to propagate messages and blocks. It is mainly used in the third process above when a node needs to stay in sync with new blocks being produced and propagated.
+- `BlockSync` is used to synchronise specific parts of the chain, that is from and to a specific height.
+- `Bitswap` is used to request and receive blocks, when a node is synchonized but GossipSub has failed to deliver some blocks to a node.
+- `GraphSync` can be used to fetch parts of the blockchain as a more efficient version of `Bitswap`.
```
schomatis, line 34: These I think are secondary in the sync process, we may be able to remove them or list them separately to keep the main list concise and easier to grasp. They don't fetch block headers (they might be used to fetch messages or complete chain information in some other way).

content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -1,42 +1,52 @@
...
+
 ## Terms and Concepts
 
 - `LastCheckpoint` the last hard social-consensus oriented checkpoint that `ChainSync` is aware of.
   This consensus checkpoint defines the minimum finality, and a minimum of history to build on.
   `ChainSync` takes `LastCheckpoint` on faith, and builds on it, never switching away from its history.
 - `TargetHeads` a list of `BlockCIDs` that represent blocks at the fringe of block production.
   These are the newest and best blocks `ChainSync` knows about. They are "target" heads because
-  `ChainSync` will try to sync to them. This list is sorted by "likelihood of being the best chain" (eg for now, simply `ChainWeight`)
+  `ChainSync` will try to sync to them. This list is sorted by "likelihood of being the best chain". At this point this is simply realized through `ChainWeight`.
```
schomatis, line 45: `ChainSync` (a term I insist doesn't exist in the code) seems to refer to a process while `ChainWeight` is a parameter of our current chain. They may be related but saying one is used to achieve the other might be a bit of a stretch that may require more explanatory text adapting the two terms together.

content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -45,12 +55,11 @@ At a high level, `ChainSync` does the following:
   - Resource expensive verification MAY be skipped at nodes' own risk
 - **Part 2: Bootstrap to the network (`BOOTSTRAP`)**
   - Step 1. Bootstrap to the network, and acquire a "secure enough" set of peers (more details below)
-  - Step 2. Bootstrap to the `BlockPubsub` channels
-  - Step 3. Listen and serve on Graphsync
+  - Step 2. Bootstrap to the `GossipSub` channels
 - **Part 3: Synchronize trusted checkpoint state (`SYNC_CHECKPOINT`)**
-  - Step 1. Start with a `TrustedCheckpoint` (defaults to `GenesisCheckpoint`).
+  - Step 1. Start with a `TrustedCheckpoint` (defaults to `GenesisCheckpoint`). The `TrustedCheckpoint` SHOULD NOT be verified in software, it SHOULD be verified by operators.
```
schomatis, line 60: ?

content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -178,177 +80,24 @@ of the altered protocol.
...
-      fetch ancestors connecting back to `ChainSync.FinalityTipset` within a reasonable amount of time).
-    - All blocks between `ChainSync.FinalityTipset ... ChainSync.TargetHeads` have been validated
-      This means all blocks _before_ the best heads.
-  - Not under a temporary network partition
-- **transitions out:**
-  - once gaps between `ChainSync.FinalityTipset ... ChainSync.TargetHeads` are closed: move to `CHAIN_FOLLOW`
-  - (Perhaps moving to `CHAIN_FOLLOW` when 1-2 blocks back in validation may be ok.
-    - we dont know we have the right head until we validate it, so if other heads of similar height are right/better, we won't know until then.)
+Peer discovery is a critical part of the overall architecture. Taking this wrong can have severe consequences for the operation of the protocol. This is especially important for security because network "Bootstrap" is a difficult problem in peer-to-peer networks. The set of peers a new node initially connects to may completely dominate the node's awareness of other peers, and therefore the view of the state of the network that the node has.
```
schomatis, line 85: > Taking this wrong can have severe consequences for the operation of the protocol.
What does this mean?

Other than bootstrap nodes (hardcoded in the implementation) there is no wrong or right peer, we just can't assume anything and verify everything.

content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -178,177 +80,24 @@ of the altered protocol.
...
-  - Encounter gaps of `>MaxBlockGap` or `>MaxEpochGap` between Validated set and a new `ChainSync.BestTargetHead`
-- **transitions out:**
-  - if a temporary network partition is detected: move to `CHAIN_CATCHUP`
-  - if `BlockGap > MaxBlockGap`: move to `CHAIN_CATCHUP`
-  - if `EpochGap > MaxEpochGap`: move to `CHAIN_CATCHUP`
-  - if node is shut down: move to `INIT`
+### Peer Exchange
 
-## Block Fetching, Validation, and Propagation
+This process is supported through either bootstrap nodes or other normal peers. Bootstrap nodes must be maintained by system operators. They have to be stable and operate independently of protocol constructions, such as the GossipSub mesh construction, that is, bootstrap nodes do not maintain connections to the mesh.
```
schomatis, line 92: I think this basically replaces the previous paragraph I commented on.

content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -178,177 +80,24 @@ of the altered protocol.
...
 
-- `TargetHeads` is changing, as `ChainSync` must be aware of the best heads at any time. reorgs happen, and our first set of peers could've been bad, we keep discovering others.
-  - Hello protocol is good, but it's polling. Unless node is constantly polllng, won't see all the heads.
-  - `BlockPubsub` gives us the realtime view into what's actually going on.
-  - Weight can also be close between 2+ possible chains (long-forked), and `ChainSync` must select the right one (which, we may not be able to distinguish until validating all the way)
-- fetching + validation are strictly faster per round on average than blocks produced/block time (if they're not, will always fall behind), so we definitely catch up eventually (and even quickly). The last couple rounds can be close ("almost got it, almost got it, there").
+### Explicit Peering Agreements
 
-### General notes on fetching Blocks
+With explicit peering agreements, the operators must specify a list of peers which nodes should connect to when joining. The protocol must have options available for these to be specified. For every explicit peer, the router must establish and maintain a bidirectional (reciprocal) connection.
```
schomatis, line 98: I don't follow this but if it's part of the gossipsub spec ignore this comment.

content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -178,177 +80,24 @@ of the altered protocol.
...
-- `ChainSync` makes requests for chains of `BlockHeaders` to close gaps between  `TargetHeads`
-- `ChainSync` forms partial unvalidated chains of `BlockHeaders`, from those received via `BlockPubsub`, and those requested via `Graphsync`.
-- `ChainSync` attempts to form fully connected chains of `BlockHeaders`, parting from `StateTree`, toward observed `Heads`
-- `ChainSync` minimizes resource expenditures to fetch and validate blocks, to protect against DOS attack vectors.
-  `ChainSync` employs **Progressive Block Validation**, validating different facets at different stages of syncing.
-- `ChainSync` delays syncing `Messages` until they are needed. Much of the structure of the partial chains can
-  be checked and used to make syncing decisions without fetching the `Messages`.
-
-### Progressive Block Validation
+## Progressive Block Validation
```
schomatis, line 100: We should merge https://github.com/filecoin-project/lotus/blob/master/documentation/en/block-validation.md here if possible, or at least extract the checks required by the spec and leave there just implementation specific ones.

content/systems/filecoin_blockchain/chainsync/_index.md
```diff
@@ -367,138 +116,6 @@ of the altered protocol.
...
-
-### PartialGraph
-
-The `PartialGraph` of blocks.
-
-> Is a graph necessarily connected, or is this just a bag of blocks, with each disconnected subgraph being reported in heads/tails?
-
-The latter.  The partial graph is a DAG fragment-- including disconnected components.
-here's a visual example, 4 example PartialGraphs, with Heads and Tails. (note they aren't tipsets)
+## Summary
```
schomatis, line 119: This is a great summary and should be on top.

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: This says that the genesis block should be checked by a human. Is it not the case?

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: This is putting emphasis on the fact that bootstrap nodes should be configured correctly. Do you think this is not needed?

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: The previous paragraph is generally about peer discovery, which includes both peer exchange and explicit peering agreements. This one is focusing on bootstrap nodes more in specific. But I understand it sounds like repetition. I'm fine with the repetition due to the importance of this part, but I'm also fine to remove one of the two paragraphs if you think it's disrupting the flow.

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: I was also thinking the same. Done!

content/systems/filecoin_blockchain/chainsync/_index.md
schomatis: Agree that it's important, let's make it a bit more explicit if possible. Instead of taking about 'peer discovery' in general and mentioning the bootstrap nodes as a particular case down the road (when it's actually the only case where it's important) I would just say something more direct as you just did now in your response.

content/systems/filecoin_blockchain/chainsync/_index.md
schomatis: I would mention bootstrap nodes in a single paragraph to make sure it has enough emphasis (fine if the end result is two paragraphs after that refactor).

schomatis:
> > this may seem like a slaughter
> 
> It's a very helpful review! Addressed part of the comments and will follow up with the rest tomorrow to get this done.

Thanks, I was actually referring to the fact we're killing 400 lines in this PR (which makes sense to me).

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: Correct! Rushed editing. Fixed.

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: This is a good point, I agree. I have separated the two groups of protocols.

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: I get your point, but I don't really find anything wrong with the phrasing. Someone might choose to implement chainsync as a separate process, which wouldn't be a problem, I believe? Can you please suggest a better phrasing, or thoughts on what explanatory text should roughly say?

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: I do agree that a big part (if not all) of the content in the link you shared is spec material. However, I think it should go under "Blockchain"->"Blocks" (current section 2.4.1: https://beta.spec.filecoin.io/#systems__filecoin_blockchain__struct) and be linked from there, rather than have it in ChainSync (and link from 2.4.1). The summary in the current section is fine IMO. What do you think?

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: I did that and mention bootstrap nodes only in this subsection (`## Peer Exchange`). Let me know if that's what you had in mind.

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: I have now excluded reference of "Bootstrap" in this leading paragraph, although there wasn't mention of bootstrap _nodes_ - just bootstrapping as a process. But I've cleaned it up a bit and made it clearer.

schomatis (approved):
Talked this over with @yiannisbot in the sync, his comments SGTM.

hugomrdias (approved):

content/systems/filecoin_blockchain/chainsync/_index.md
yiannisbot: Block validation will be explained in more detail in the Blockchain section.

PR #1133, opened 2020-09-01
chore: refactor to a single watcher

olizilla:
- add watch.js to configure chokidar.
- pass watcher to each tool via an exported `configureWatcher` method.

If run directly, the tool assumes it should process all things it cares about. If required as module, it does nothing, and allows the calling code to use it to configure the watcher.

License: MIT


olizilla:
This is based off of the tic-toc branch for #1122 as re-working the watcher is a significant change that is unrelated to the goal of that PR.

hugomrdias (approved):

PR #1145, opened 2020-09-03
feat: improve mobile experience

hugomrdias:


assets/_custom.scss
```diff
@@ -110,15 +116,8 @@ html {
...
 .container {
-    max-width: 89rem;
-}
-.book-menu {
-    flex: 0 0 20rem;
-}
-.book-page {
-    padding-top: 0rem;
-    padding-left: 2rem;
-    max-width: 52rem;
```
olizilla, line 121: Pls bring back the `max-width` for the content. 

> “Anything from 45 to 75 characters is widely regarded as a satisfactory length of line for a single-column page… the 66-character line (counting both letters and spaces) is widely regarded as ideal. For multiple-column work, a better average is 40-50 characters.” 
> — Robert Bringhurst, The Elements of Typographic Style

https://tachyons.io/docs/typography/measure/

olizilla:
Tested this on Chrome on a Pixel 4, and the sidebar is significantly more responsive with these changes 👍

PR #1147, opened 2020-09-08
fix: fix scrollbars in firefox

### LINKED ISSUE #1144: No scrollbar handles in FF
On Firefox 76.0.1 on Debian I don't have handles on the scrollbars.

image

In Chromium you can see that small handle on the right edge and a bigger one at the the second code block.

image

hugomrdias:
closes #1144


vmx (approved):
Works for me, I can see the scrollbar handles now.

PR #1151, opened 2020-09-10
feat(content): major update to the glossary content

yiannisbot:
This is a major update to the terms and definitions of the Glossary section.


content/glossary/_index.md
```diff
@@ -2,311 +2,411 @@
...
+
+The Filecoin network expects constant availability of stored data. Failing to submit WindowPoSt for a sector will result in a [fault](glossary#fault), and the storage miner supplying the sector will be [slashed](glossary#slash).
+
+## Winning Proof-of-Spacetime (WinningPoSt)
+
+_Winning Proof-of-Spacetime_ (WinningPoSt) is the mechanism by which [storage miners](glossary#storage-miner-actor) are rewarded for their contributions to the Filecoin network. At the beginning of each [epoch](glossary#epoch), a small number of storage miners are [elected](glossary#election) to each mine a new [block](glossary#block). As a requirement for doing so, each miner is tasked with submitting a compressed [Proof-of-Spacetime](glossary#proof-of-spacetime-post) for a specified [sector](glossary#sector). Each elected miner who successfully creates a block is granted [FIL](glossary#fil), as well as the opportunity to charge other Filecoin participants fees to include [messages](glossary#message) in the block.
+
+Storage miners who fail to do this in the necessary window will forfeit their opportunity to mine a block, but will not otherwise incur penalties for their failure to do so.
+
+## zkSNARK
```
porcuquine, line 398: zk-SNARK is more common and what I think we should use.

Similarly, for capitalization of the expanded term, let's consistently use: 'Zero-Knowledge Succinct Non-Interactive Argument of Knowledge'

content/glossary/_index.md
```diff
@@ -2,311 +2,411 @@
...
+
+A _zero-knowledge_ argument of knowledge has the requirement that the verifier should not need access to the knowledge the prover has access to in order to verify the prover's claim.
+
+A _succinct_ argument of knowledge is one that can be "quickly" verified, and which is "small", for appropriate definitions of both of those terms.
+
+A zero-knowledge, succinct non-interactive argument of knowledge (zk-SNARK) embodies all of these properties. Filecoin utilizes these constructions to enable its distributed network to efficiently verify that [storage miners](glossary#storage-miner-actor) are storing files they pledged to store, without requiring the verifiers to maintain copies of these files themselves.
 
-Zero Knowledge Succinct ARguments of Knowledge. A way of producing a small 'proof' that convinces a 'verifier' that some computation was done correctly.
+In summary, a zkSNARK is a way of producing a small 'proof' that convinces a 'verifier' that some computation on a stored file was done correctly, without the verifier needing to have access to the stored file itself.
```
porcuquine, line 412: I would say something like 'Filecoin uses zk-SNARKs to produce…' to make clear we are not attributing a domain-related usage to SNARKs in general.

content/glossary/_index.md
yiannisbot: Good catch! Done!

content/glossary/_index.md
```diff
@@ -2,311 +2,411 @@
...
+## Payment Channel
 
-A payment channel is set up between actors in the Filecoin system to enable off-chain payments with on-chain guarantees, making settlement more efficient.
+A payment channel is set up between actors in the Filecoin system to enable off-chain payments with on-chain guarantees, making settlement more efficient. Payment channels are managed by the Payment Channel Actor, who is responsible for setting up and settling funds related to payment channels.
 
-### Piece
+## Piece
 
-A piece is a portion of a file that gets fitted into a sector.
+The main unit of account and negotiation for the data that a user wants to store on filecoin. A piece is a [CAR file](https://github.com/ipld/specs/blob/master/block-layer/content-addressable-archives.md#summary) produced by an IPLD DAG with its own *payload CID* and *piece CID*. A piece is not a unit of storage, that is, it can be of any size up to the size of a [sector](glossary#sector). If a piece is larger than a sector, then it has to be split in two (or more) pieces.
```
hannahhoward, line 207: This is not exactly accurate.

Two things:
- a piece is not gauranteed to be a CAR file - Lotus software assumes this to be the case, but there is nothing preventing one from making a storage deal for non-CAR data as long as it matches the PieceCID.
- a piece cannot be "of any size" -- it must be a size that slots into a sector -- i.e. a power of 2 when bit-padded. (so without bit padding it must be a (power of 2) * 127/128)
- so technically, even Lotus's pieces are not simply CAR data -- they are CAR data + a bunch of zeros to make it the right size

content/glossary/_index.md
```diff
@@ -2,311 +2,411 @@
...
 
-A `TipSet` is a set of blocks that have the same parent set and same number of `tickets`, which implies they will have been mined at the same `height`. A `TipSet` can contain multiple blocks if more than one miner successfully mines a block at the same `height` as another miner.
+## Verified client
 
+To further incentivize the storage of "useful" data over simple [capacity commitments](glossary#capacity-commitment), [storage miners](glossary#storage-miner-actor) have the additional opportunity to compete for special [deals](glossary#deal) offered by verified clients. Such clients are certified with respect to their intent to offer deals involving the storage of meaningful data, and the power a storage miner earns for these deals is augmented by a multiplier.
```
hannahhoward, line 352: Do we have a section explaining verified client? Asking cause the Verified Client concept came about after the initial writing of the spec

hannahhoward (commented):
Comments left. I dunno how blocking any of them are, other than the Piece description has some inaccuracies.

content/glossary/_index.md
```diff
@@ -2,311 +2,411 @@
...
+## CID
 
-Derived from a `Partial Ticket` as part of ElectionPoSt and used to run leader election.
+CID is short for Content Identifier, a self describing content address used throughout the IPFS ecosystem. CIDs are used in Filecoin to identify files submitted to the decentralized storage network. For more detailed information, see [the github documentation for it](https://github.com/ipld/cid).
 
-### Cid
+## Client
 
-CID is short for Content Identifier, a self describing content address used throughout the IPFS ecosystem. For more detailed information, see [the github documentation for it](https://github.com/ipld/cid).
+A client is any user with an account who wishes to store data with a storage miner. A client's account is used to pay for the storage, and helps to prove the client's ability to pay.
```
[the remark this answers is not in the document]
yiannisbot: Good point. I rephrased slightly to make it more straightforward: "There are two types of clients in Filecoin, the storage client and the retrieval client, both of which can be implemented as part of the same physical host. All clients have an account, which is used to pay for the storage or retrieval of data."

content/glossary/_index.md
```diff
@@ -2,311 +2,411 @@
...
 
-### Generation Attack Threshold
+When a [storage miner](glossary#storage-miner-actor) fails to complete [Window Proof-of-Spacetime](glossary#window-proof-of-spacetime-windowpost) for a given sector, the Filecoin network registers a _fault_ for that sector, and the miner is [_slashed_](glossary#slashing). If a storage miner does not resolve the fault quickly, the network assumes they have abandoned their commitment.
 
-Security parameter. Number of rounds within which a new Proof-of-Storage must be submitted in order for a miner to retain power in the network (and avoid getting slashed). This number must be be smaller than the minimum time it takes for an adversarial miner to generate a replica of the data (thereby not storing it undetectably for some period of time).
+## File
+
+Files are what clients bring to the filecoin system to store. A file is split up into [pieces](glossary#piece). A piece is the basic unit of account in the storage network and is what is actually stored by the Filecoin network.
```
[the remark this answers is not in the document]
yiannisbot: Oooops! Good catch! Rephrased

content/glossary/_index.md
yiannisbot: Thanks for the clarification @hannahhoward. I wanted to leave it a bit more general in the glossary and point to the actual section for more details, but added some more details here too - let me know if you agree:

"The main unit of account and negotiation for the data that a user wants to store on Filecoin. In the Lotus implementation, a piece is a [CAR file](https://github.com/ipld/specs/blob/master/block-layer/content-addressable-archives.md#summary) produced by an IPLD DAG with its own *payload CID* and *piece CID*. However, a piece can be produced in different ways as long as the outcome matches the _piece CID_. A piece is not a unit of storage and therefore, it can be of any size up to the size of a [sector](glossary#sector). If a piece is larger than a sector (currently set to 32GB or 64GB and chosen by the miner), then it has to be split in two (or more) pieces. For more details on the exact sizing of a pieces as well as how it can be produced, see the [Piece section](piece)."

content/glossary/_index.md
yiannisbot: We have some description scattered here and there, which I've added recently, but I intend to add a separate subsection dedicated to verified clients as it's become a central part.

sa8 (approved):

hugomrdias (approved):

PR #1158, opened 2020-09-17
fix(content): update CryptoEcon Section content

yiannisbot:
This PR is adding new description to the CryptoEconomics section of the spec. The description includes a table with the basic parameters, as well as justification for the design principles of Filecoin's cryptoeconomic model.

TODOs:
- [x] Link to the minting model and token allocation when the corresponding PRs get merged.
- [ ] There will soon be a separate section that will summarise the parameters of the protocol. Once that section exists, the table included here will likely need to be excluded from here and replace with a pointer to that section, in order to have a single source of truth.


zixuanzh (approved):
LGTM

PR #1161, opened 2020-09-17
fix(content): remove sharded IPLD array file

yiannisbot:
The Sharded IPLD Arrays are not used in Filecoin anymore, hence, this PR removes the subsection from the Appendix.


mikeal:
@rvagg should we pull any of this into the IPLD docs or specs site? We still have implementations of this, they just aren’t used in Filecoin anymore.

rvagg:
Our vector spec covers this well enough, I think it's already 1:1 (although slightly more configurable?) with sharray so I don't think we need to capture anything here. This will be a good one to get to with ipld-prime now that they're poking at ADLs but we have it covered with the spec and the sharray code.

rvagg (approved):

hugomrdias (approved):

PR #1170, opened 2020-09-22
fix: fix some details in CC Sectors

yiannisbot:
This small PR clarifies a few details regarding CC sectors and the fact that by default they store randomness. It also makes clearer the fact that resealing is needed when upgrading to a deal or verified deal and that miners are free to reject deals and instead use their storage in the form of committed capacity.


zixuanzh (approved):
LGTM

hugomrdias (approved):

PR #1179, opened 2020-09-25
fix(content): on-chain storage market update

yiannisbot:
This PR updates the storage market parts that take place on-chain and relate to the storage market actor and the storage deal flow.


zixuanzh (dismissed):
Updated storage deal flow but this does not include improvements introduced by [FIP-0002](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0002.md).

content/systems/filecoin_markets/onchain_storage_market/storage_market_actor.md
```diff
@@ -29,5 +37,5 @@ This collateral is returned to the storage provider when all deals in the sector
 
 
 {{<katex>}}
-$MinimumProviderDealCollateral = \\[0.2cm] \ \ \ \ \ \ \ \ 5\% \times FILCirculatingSupply \times \frac{DealRawByte}{max(NetworkBaseline, NetworkRawBytePower)}$
+$MinimumProviderDealCollateral = \\[0.2cm] \ \ \ \ \ \ \ \ 1\% \times FILCirculatingSupply \times \frac{DealRawByte}{max(NetworkBaseline, NetworkRawBytePower)}$
```
hugomrdias, line 40: This is not inline! Please read https://github.com/filecoin-project/specs#katex

hugomrdias (changes requested):

content/systems/filecoin_markets/onchain_storage_market/storage_market_actor.md
yiannisbot: @hugomrdias can you clarify the problem? It's not inline, but is a single formula. Does it break something as is?

hugomrdias: inline means inline with other content, there's no content around this formula. If you open the devtools you will see errors because [0.2cm] as no meaning when rendered in inline mode.

yiannisbot: @hugomrdias I've done it with math mode now. Can you check if it's ok now?

hugomrdias: you still using inline mode (`$math$`) when you should be using display mode (`$$math$$`)

yiannisbot: I'm not using katex in this latest commit. I'm using `math-mode: true` and put the formula in code-fences. So according to the [README](https://github.com/filecoin-project/specs#wrap-math-blocks-with-code-fences) it should be in single `$`, right?

hugomrdias: either way theres always inline and display mode, in this case where its a single formula with nothing around it, you want to highlight so it should be display mode `$$math$$`

hugomrdias (approved):

content/systems/filecoin_markets/onchain_storage_market/storage_market_actor.md
yiannisbot: okay! done!

PR #1181, opened 2020-09-26
fix(content): chain manager and block producer updates

yiannisbot:
This PR updates the chain manager and block producer sections, which were mostly in a good state. @schomatis, as discussed, please edit directly to update the text where needed.

Two notes:
- I would probably exclude completely current section 2.4.1.3 "Chain" as I don't see what it adds. @schomatis let me know what you think.
- In section 2.4.1.5.2 on Block Rewards, I have excluded all text and have instead pointed to the correct sections.

@irenegia I would welcome a quick opinion on whether the Election PoSt text in section 2.4.1.5 Block Producer is accurate or needs update.


schomatis (dismissed):
I'm not finding much to review here apart from some minor prose edits (I may be missing something though).

> I would probably exclude completely current section 2.4.1.3 "Chain" as I don't see what it adds. @schomatis let me know what you think.

Sure.


> In section 2.4.1.5.2 on Block Rewards, I have excluded all text and have instead pointed to the correct sections.

I can't find that section index.

irenegia:
@yiannisbot 

> @irenegia I would welcome a quick opinion on whether the Election PoSt text in section 2.4.1.5 Block Producer is accurate or needs update

I can not find any text about Election PoSt in the "Block Producer" section..

irenegia:
> @irenegia it's this text in line 20:
> 
> "Producing a block for epoch `H` requires computing a tipset for epoch `H-1` (or possibly a prior epoch, if no blocks were received for that epoch). Using the state produced by this tipset, a miner can scratch winning ElectionPoSt ticket(s). Armed with the requisite `ElectionPoStOutput`, as well as a new randomness ticket generated in this epoch, a miner can produce a new block."

I see, this part should be removed. 
Instead we should write something like:

"Producing a block for epoch `H` requires waiting for the beacon entry for that epoch and using it to run GenerateElectionProof. If WinCount ≥ 1 (the miner is elected), the same beacon entry is used to run WinningPoSt. Armed by the ElectionProof ticket (output of GenerateElectionProof) and the WinningPoSt proof, the minere can produce an new block."

irenegia (approved):

content/systems/filecoin_blockchain/struct/chain_manager/_index.md
```diff
@@ -13,32 +13,28 @@ The _Chain Manager_ is a central component in the blockchain system. It tracks a
...
-The chain manager interfaces and functions are included here, but we expand on important details below for clarity.
 
-
-## Chain Expansion
+## Chain Extension
 
 ### Incoming block reception
 
-Once a block has been received and passes syntactic and semantic validation it must be added to the local datastore, regardless whether it is understood as the best tip at this point. Future blocks from other miners may be mined on top of it and in that case we will want to have it around to avoid refetching.
+Once a block has been received and passes syntactic and semantic validation (as explained above) it must be added to the local datastore, regardless of whether it is understood as the best tip at this point. Future blocks from other miners may be mined on top of it and in that case we will want to have it around to avoid refetching.
```
schomatis, line 21:
```suggestion

```

This entire line sounds like implementation details. We can rework it if we think it's sound advise and worth staying here but the term *must* should be used with care in a spec I think.

content/systems/filecoin_blockchain/struct/chain_manager/_index.md
```diff
@@ -13,32 +13,28 @@ The _Chain Manager_ is a central component in the blockchain system. It tracks a
...
 
-Chain selection is a crucial component of how the Filecoin blockchain works. Every chain has an associated weight accounting for the number of blocks mined on it and so the power (storage) they track. It is always preferable to mine atop a heavier Tipset rather than a lighter one. While a miner may be foregoing block rewards earned in the past, this lighter chain is likely to be abandoned by other miners forfeiting any block reward earned as miners converge on a final chain. For more on this, see [chain selection](expected_consensus#chain-selection) in the Expected Consensus spec.
+Chain selection is a crucial component of how the Filecoin blockchain works. Every chain has an associated weight accounting for the number of blocks mined on it and so the power (storage) they track. It is always preferable to mine atop a heavier Tipset rather than a lighter one. While a miner may be foregoing block rewards earned in the past, this lighter chain is likely to be abandoned by other miners forfeiting any block reward earned as miners converge on a final chain. For more on this, see [chain selection](expected_consensus#chain-selection) in the Expected Consensus section.
```
schomatis, line 25: Also sounds a little superficial for this section. The advise of staying in consensus is fundamental for any protocol (and if any we should emphasize it there, not here).
```suggestion
```

schomatis (changes requested):
So, after another pass I think we should drop the entire Chain Manager, it doesn't add much value from the spec perspective. At most we can keep the Go API.

content/systems/filecoin_blockchain/struct/chain_manager/_index.md
yiannisbot: I think the following is an important detail:

> regardless of whether it is understood as the best tip at this point. Future blocks from other miners may be mined on top of it and in that case we will want to have it around to avoid refetching.

If we want to rephrase from `must` to `should` we can do that instead of deleting.

content/systems/filecoin_blockchain/struct/chain_manager/_index.md
yiannisbot: That's a fair point. I would keep only the following as I think it is needed to link to the paragraph after.

> Chain selection is a crucial component of how the Filecoin blockchain works. Every chain has an associated weight accounting for the number of blocks mined on it and so the power (storage) they track. The full details of how selection works are provided in the [Chain Selection](expected_consensus#chain-selection) section.

schomatis:
I'll leave that up to you. When I started the review I marked those two for delete for sure, and after looking at what was left I didn't see much more remaining so I (confusingly) summed up my review as 'just remove it all', but I'm fine with anything in between those two extremes.

content/systems/filecoin_blockchain/struct/chain_manager/_index.md
schomatis: I'm fine with just changing the word if you think it's useful. It still seems to me too much of an implementation detail, should we really store those tipsets preemptively? Is it really worth the disk space to save some bandwidth? Maybe yes, maybe no, but you see we are already talking about details that don't reflect how the protocol works. They sound to me more of an engineering issue than anything else.

yiannisbot: Yes, it is a detail, I agree. I also think it will not hurt as a recommendation, which is what I have added it as. It will be interesting to see what is cheaper to do in terms of resources. I have a feeling that as the network grows, it will be better if everyone kept a copy temporarily, instead of having to fetch thousands of copies of the same thing to thousands of nodes.

hugomrdias (approved):

PR #1187, opened 2020-09-30
fix(tooling): fix math in multipage layout

### LINKED ISSUE #1182: Fix math mode in multi-page mode
The multi-page mode is useful if you're authoring new stuff, especially when using math. You don't want to see the full spec, but only the part you're currently editing. You can get into the multi-page mode when using slashed URLs, e.g. https://spec.filecoin.io/algorithms/sdr/

Currently the math mode doesn't render there. It would be great to have that fixed.

hugomrdias:
closes #1182


vmx (dismissed):
Thanks. It works in the deployed version and the code also looks simpler.

layouts/partials/body.html
```diff
@@ -1,7 +1,7 @@
-{{ $opts := dict "targetPath" "main.js" "minify" "true" }}
+{{ $opts := dict "targetPath" "main.js" "minify" "false" }}
```
hugomrdias, line 1:
```suggestion
{{ $opts := dict "targetPath" "main.js" "minify" "true" }}
```

PR #1203, opened 2020-10-12
fix: rename index.md to _index.md

### LINKED ISSUE #1196: Wrong placement of _index.md file
The _index.md file in this path/folder: https://github.com/filecoin-project/specs/blob/master/content/systems/filecoin_mining/storage_mining/mining_cycle/ gets the wrong position in the ToC and text. It becomes a subsubsection of the previous subsection, instead of becoming a subsection itself. All the frontmatter is the same as in all other cases, so I don't know how to make it become section 2.6.2.4, instead of 2.6.2.3.3.- or perhaps I'm missing something?

Screenshot 2020-10-06 at 21 55 54

hugomrdias:
closes #1196


yiannisbot (approved):
Good catch!

PR #1211, opened 2020-10-14
feat(tooling): section menu, links, blockquotes and content boxes

### LINKED ISSUE #1157: Status labels in H2 sections - Glossary & Libraries sections
I notice that there are status labels on every H2 subsection in the Glossary section of the spec. I was under the impression that in the current setup, only subsections or subsubsections that live in their own file take labels (and a place in the dashboard), therefore, this behaviour looks unusual to me. All other H2 sections of the spec do have their own file and are supposed to have a label, so this might be an edge case we had not seen before.

Screenshot 2020-09-16 at 23 17 44

The same seems to be happening with H3 subsusections in the Libraries section, which is more unusual, as there are many H3 subsections in the spec that don't have a label.

This is not an urgent issue, but would be good to solve for consistency throughout.

### LINKED ISSUE #1173: Normalise examples and figure boxes

### LINKED ISSUE #1185: Symbol shortcode is pulling code comments too
When pulling code blocks using the symbol shortcode, text or comments that exist just above the symbol in the code are also pulled and included in the spec. These comments appear as normal text that is obviously not possible to edit. See the bulletpoint list in the screenshot below.

@hugomrdias it would be great to: i) get a confirmation that this will not change, i.e., these comments will continue to be pulled and appear in the spec, and ii) know whether there will be any change to make these comments appear as code comments, inside a box or something similar.

Knowing this will allow me to make decisions on whether extra text is needed in the actual spec to cover those comments.

Screenshot 2020-09-30 at 08 54 37

### LINKED ISSUE #1202: Build section info box
Each section should have a dropdown type of thing with at least github edit url and unique identifier for that section with copy to clipboard (this ID can be used to support integration like oni tests and FIPS).

/cc @yiannisbot

hugomrdias:
This PR improves a couples of things 

# Depth slider css
Screenshot 2020-10-14 at 16 16 53
Screenshot 2020-10-14 at 16 17 32

# Blockquotes now have proper background
Screenshot 2020-10-14 at 16 19 12

# Figures and embeds now have permalinks
Screenshot 2020-10-14 at 16 19 57

# Figures and embeds look the same 
Screenshot 2020-10-14 at 16 21 42

# Embeds integrate all the elements inside same box as Figures
Screenshot 2020-10-14 at 16 23 48

# Section Menu 
Now each section has a menu to aggregate all the info like state, audit, github edit link and unique identifier and more in the future like tests etc.  

menu gif

Check https://bafybeiexx7iex3upkzf76svkvvwhlg6xh6sxh5tq4k2t5hdvcjtsq63iuy.on.fleek.co/test/ to play with these new improved components.

# Even faster CI without Go and fully on Fleek

Github action now only need node js and just run build and test scripts.

# Other small stuff
- Remove multilang files
- CSS tweaks
- New system ui fonts
- Better permalinks
- All new svg icons, using svg symbols refs for perf 

closes #1173
closes #1202
closes #1157
closes #1185


yiannisbot (approved):
Great set of improvements! Makes it look much better!

olizilla (changes requested):
I've not had time to review the implementation, but the changes look beautiful!

The section heading menu is a neat idea. The icons could use a little more space and there is an overflow issue (in FF at least)

Screenshot 2020-10-14 at 21 42 19

PR #1216, opened 2020-10-15
fix(tooling): remove display mode katex

### LINKED ISSUE #1193: katex display mode
The katex environment is expected to work both in inline mode (with `$`) and in display mode (with `$$`). However, the display mode does not seem to render correctly the formulas, which seems to be a bug. Inline more works fine.

hugomrdias:
closes #1193


yiannisbot (approved):
Great! Good catch the katex issue.

FeliciaAKelley (approved):

PR #1218, opened 2020-10-16
restructuring audits, adding theory audits

dkkapur:


daviddias:
`Done`  is not good because Security is never done (recall to Security Philosophy "Achieving “100% fully secure” is an untenable goal.")

content/appendix/audit_reports.md
```diff
@@ -0,0 +1,47 @@
...
+
+### 2020-06-03: GossipSub Design and Implementation
+
+Audit report: [GossipSub v1.1 Protocol Design + Implementation](https://gateway.ipfs.io/ipfs/QmWR376YyuyLewZDzaTHXGZr7quL5LB13HRFnNdSJ3CyXu/Least%20Authority%20-%20Gossipsub%20v1.1%20Final%20Audit%20Report%20%28v2%29.pdf)
+
+This audit focused specifically on GossipSub as a pubsub protocol built on libp2p, looking at v1.1 which includes a peer scoring layer to mitigate certain types of attacks that could compromise a network. The audit covered the [spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub), [go-libp2p-pubsub](https://github.com/libp2p/go-libp2p-pubsub) and [gossipsub-hardening](https://github.com/libp2p/gossipsub-hardening/). The report found 4 issues, primarily in the Peer Scoring that was introduced in v1.1, and includes additional suggestions. All the issues raised in the report have been resolved, and additional details are available in the report linked above.
+
+### 2020-04-18: GossipSub Evaluation
+
+Audit report: [GossipSub-v1.1 Evaluation Report](https://gateway.ipfs.io/ipfs/QmRAFP5DBnvNjdYSbWhEhVRJJDFCLpPyvew5GwCCB4VxM4)
```
yiannisbot, line 23: This is not really an audit report, it's an evaluation report, so I would call it such. It gives further evidence on the correctness of the protocol. Also, since we're including this, how about linking to the technical report/paper that came out of it, that: i) details the whole design, ii) the rationale behind design decisions, iii) related work in the area and iv) includes more test scenarios. The link is here: https://arxiv.org/abs/2007.0275.

content/appendix/audit_reports.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Audit Reports
+
+Security is a critical component in ensuring Filecoin can fulfill its mission to be the storage network for humanity. In addition to robust secure development processes, trainings, theory audits, and investing in external security research, the Filecoin project has engaged reputable third party auditing specialists to ensure that the theory behind the protocol and its implementation delivers the intended value, enabling Filecoin to be a safe and secure network. This section covers a selection of audit reports that have been published on Filecoin's theory and implementation.
+
+## GossipSub
+
+### 2020-06-03: GossipSub Design and Implementation
+
+Audit report: [GossipSub v1.1 Protocol Design + Implementation](https://gateway.ipfs.io/ipfs/QmWR376YyuyLewZDzaTHXGZr7quL5LB13HRFnNdSJ3CyXu/Least%20Authority%20-%20Gossipsub%20v1.1%20Final%20Audit%20Report%20%28v2%29.pdf)
```
yiannisbot, line 17: With the exception of the GossipSub 'Evaluation Report', the rest of the reports here are basically _External Audit Reports_. I would be in favour of explicitly marking them as such, especially if in the future we plan to include also internal audit reports.

content/algorithms/gossip_sub.md
```diff
@@ -5,8 +5,8 @@ dashboardWeight: 2.0
 dashboardState: stable
 dashboardTests: 0
 dashboardAudit: done
-dashboardAuditDate: '2020-06-03'
 dashboardAuditURL: https://gateway.ipfs.io/ipfs/QmWR376YyuyLewZDzaTHXGZr7quL5LB13HRFnNdSJ3CyXu/Least%20Authority%20-%20Gossipsub%20v1.1%20Final%20Audit%20Report%20%28v2%29.pdf
+dashboardAuditDate: 'Done'
```
yiannisbot, line 9: To cover @daviddias' comment on the "Done" label, I think that the date in a green cell does the job. Alternatively, we could have "Done: date (in MM/YY format)". This would obviously not scale for multiple dates, but we don't need to have multiple dates in the dashboard, just the latest.

yiannisbot (commented):
Looks good! Thanks @dkkapur for tidying this up! Just a couple of suggestions.

content/appendix/audit_reports.md
daviddias: I believe @dkkapur wanted to link to Least Authority Security Audit instead.

content/appendix/audit_reports.md
daviddias: The section should state the partner that ran the audit, yes.

content/algorithms/gossip_sub.md
daviddias: @yiannisbot remember how we discussed that there had to be support for listing multiple audits in each cell? Right now, the direction is leaning towards "done if an audit has been done". Let's roll this back. 

Each cell should have the dates of the multiple audits and links to them.

In the future, when the spec gets versioning and there is a major upgrade. The color should change to signal that the latest version hasn't been audited.

content/appendix/audit_reports.md
yiannisbot: The LA report seems to be the one just above this one 😉

content/algorithms/gossip_sub.md
yiannisbot: > remember how we discussed that there had to be support for listing multiple audits in each cell?

Sure - I thought this decision changed after the discussion between you and @dkkapur. I'm fine to go back to the previous format.

The only problem with the date-based display is that we can't fit many dates, probably two maximum, unless we squeeze the format to something like MMYY only. IMO, having only the latest date will do it, especially now that we have a separate section with more details where we can link all the reports.

dkkapur: Fair - what does this mean for all the other `Done`s that are in the dashboard that don't have an audit report? Do we want to set a goal for having the last theory audit date shared there at some point as well? cc @daviddias

content/appendix/audit_reports.md
dkkapur: Yeah the LA report is included as well. I think I will change the "audit report" to just "report" and since this has evaluation in its title, hopefully should be OK?

content/appendix/audit_reports.md
dkkapur: Will add this as a field.

content/algorithms/gossip_sub.md
yiannisbot: > People will typically want to review the latest audit so I think it is OK for now while we figure out the right long term design.

I agree with that and I'm fine going this way.

yiannisbot: @dkkapur regarding your question above, as agreed off-band, let's mark those `Done` which don't have a report as "Report coming soon" in a yellow cell background.

yiannisbot: Let's do this last one ^^ in a separate PR. I'll take care of it.

yiannisbot (approved):
Great addition! Thanks!

PR #1226, opened 2020-10-19
fix(tooling): fix audit relative urls

### LINKED ISSUE #1215: Audit relative urls opening in a new tab
https://github.com/filecoin-project/specs/pull/1214 

The URLs in the dashboard for the audit reports, introduced through linked PR, result in a new tab opening up, would be great if it could just scroll on the same tab.

hugomrdias:
closes #1215


yiannisbot (dismissed):
Thanks!

PR #1227, opened 2020-10-19
fix(content): rename go-filecoin to venus

hugomrdias:


yiannisbot (approved):
Great! Let's merge it! It was failing CI when I did it.

PR #1230, opened 2020-10-20
Update audit_reports.md

daviddias:


yiannisbot (approved):
Nice tidy up! How about linking to the repositories, given that they do not all live at the same place? We've got links elsewhere, but I think wouldn't hurt adding here too. Just a suggestion.

daviddias:
> How about linking to the repositories, given that they do not all live at the same place?
Let's handle that in a separate issue 👍🏽

PR #1241, opened 2020-11-03
feat(tooling): add releases

hugomrdias:
Screenshot 2020-11-03 at 11 58 00
Screenshot 2020-11-03 at 11 58 33


yiannisbot (approved):
Great addition!

PR #1242, opened 2020-11-07
fix(content): minting model math update

yiannisbot:
This PR adds details on the math specification of minting. In particular touches on both simple and baseline minting, but also includes the block reward issuance.


zixuanzh:
LGTM, we also need to update the parameter in the [minting model section](https://bafybeicmehplalchzzxvl3ov6mprswksccfbxprym4takfap6d4pakugay.on.fleek.co/#section-systems.filecoin_token.minting_model) and changed it to "the network baseline will start from **2.5EiB, or 2.88888888EB,** (which is less than 0.01% of the worldʼs storage today) and grow at an annual rate of **100%** (higher than the usual world storage annual growth rate at 40%)"

PR #1276, opened 2021-10-07
Update poseidon.md

jbaylina:
Original reference: https://github.com/filecoin-project/neptune/issues/113
@DrPeterVanNostrand

The documentation says that w is the first column of m_hat or second column of m without the first row.. 

Actually it is the first column of m   without the first row.

The implementation is ok: https://github.com/filecoin-project/neptune/blob/bafd77a5014e3b6a6b40359097835c3eb1dd533f/src/mds.rs#L197


ianconsolata:
closing as duplicate

PR #1286, opened 2021-11-19
add details of consensus fault slashing

hunjixin:


hunjixin:
https://github.com/filecoin-project/venus/pull/4588   https://github.com/filecoin-project/venus/pull/4586

hunjixin:
@jennijuju It may be better to put these contents in the spec.

PR #1333, opened 2024-10-31
Remove generated diagrams from git state

ianconsolata:


ianconsolata:
Ok, it looks like Fleek can't actually generate the diagrams, which is why I think I added them to git state in the first place

Issue #1335, opened 2024-11-01
add Plausible.io tracking to replace GA. 


smagdali:
Better to use the privacy respecting Plausible. Also I don't think the legacy GA tag was even working anymore since Google went to V4.

smagdali:
https://plausible.io/spec.filecoin.io

PR #1336, opened 2024-11-01
add plausible analytics fixing #1335

smagdali:
also removed dead GA tag.


smagdali:
works - https://plausible.io/spec.filecoin.io

PR #1339, opened 2025-05-15
chore: update Forest entry

LesnyRumcajs:
Updated the entry to make it reflect the reality + fixed the docs link.

### DIFF
--- a/content/implementations/forest.md
+++ b/content/implementations/forest.md
@@ -10,8 +10,18 @@ implRepos:
 
 # Forest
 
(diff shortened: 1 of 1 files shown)


content/implementations/forest.md
```diff
@@ -10,8 +10,20 @@ implRepos:
...
 
-You can find the Forest codebase [here](https://github.com/ChainSafe/forest) and the documentation site [here](https://chainsafe.github.io/forest/).
+Forest does not support all features of the reference implementation but is a good fit for specific applications, such as:
 
-The Forest implementation of Filecoin is supported by [ChainSafe](https://chainsafe.io/).
+- generating Filecoin snapshots,
+- running a bootstrap node,
+- running an RPC node.
+
+Forest does not support mining.
```
lanzafame, line 21: I would reframe this as Forest is only a syncing implementation or Forest does not provide storage provider functionality. Just like we don't expect the Lotus node to do mining, that is the job lotus-miner etc.

lanzafame (changes requested):
Other than the mining support comment, looks good.

content/implementations/forest.md
LesnyRumcajs: Rephrased.

lanzafame (approved):

PR #1342, opened 2025-07-10
FIP-0003 to FIP-0085 fips update

lanzafame:


dannyob:
This is a *lot* of updates! Apart from Claude and @lanzafame , do we have a way to spot-check these changes to make sure they're accurate; and maybe also drip-feed it a bit more, rather all in one batch.

I worry that a) a plausible-sounding but wrong spec is harder to work from than a known out-of-date one, and b) that multiple small flaws will compound quickly if we don't fix the early ones soon.

